PR #22, opened 2018-08-03
Interstitial ads black screen issue

Simba-Mupfunya:
When implementing full page or interstitial ads such as google admob a black screen appears on any iOS device after the ad is loaded. This is a known problem with OpenTK-1.1 and the solution is to override WillMoveToWindow in iPhoneOSGameView so as to not suspend the window which causes the problem.

FYI, I had this patch done for me back in version 1.10.2 beta by Virgile and i hoped it would propagate into the other versions but my guess is no one was/is using full page ads in their game(s) yet

### DIFF
--- a/sources/engine/Xenko.Engine/Starter/iOSXenkoView.cs
+++ b/sources/engine/Xenko.Engine/Starter/iOSXenkoView.cs
@@ -101,6 +101,12 @@ namespace Xenko.Starter
             isRunning = false;
         }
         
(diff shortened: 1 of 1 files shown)


xen2:
Agree with @Kryptos-FR, just add a small comment so that we remember why it's there, and good to go.

Simba-Mupfunya:
@Kryptos-FR @xen2 Just added a small comment.

PR #45, opened 2018-08-05
[Editor] Rename sskk namespace prefix to xk in XAML files.

Kryptos-FR:
We were still using `sskk` which stands for SiliconStudio. Replaced it with `xk` which is a shortcut to mean Xenko.


xen2:
Thanks, good catch!
LGTM

Issue #42, opened 2018-08-04
GameStudio: when creating a new game/samples, detection code for platforms is broken


xen2:
When creating a new game/samples or updating platforms, the wizard says the required components are not installed for iOS, Android and UWP, even if the appropriate programs/modules are installed.
This is because the detection code is pretty outdated and needs to be updated.

XenkoConfig.cs contains the code: 
https://github.com/xenko3d/xenko/blob/0669deabfe65a04f80041433420f0ace0ad5ca9a/sources/engine/Xenko.Assets/XenkoConfig.cs#L97

xen2:
Additional note: best way now might be to check packages installed with Visual Studio.
We already have a small abstraction for that: `VisualStudioVersions.AvailableVisualStudioInstances` which contains `PackageVersions`. This lists the options selected by users during VS install.

AirAKose:
@xen2  Should it only check your default IDE for the packages, or all available?

xen2:
Unfortunately the Default IDE is defined in `Xenko.Core.Assets.Editor` so it can't be accessed from `XenkoConfig.cs`. So for now, I would say it's enough to detect if any IDE match.

Then we could later adjust the editor side to also follow same logic, by either:
- automatically use different IDE than default one if there is one better matching requirement?
- or at least displaying a warning saying there is a better IDE to choose (maybe better to avoid confusing user by launching a different instance than expected and he knows what to fix/reinstall)

xen2:
Fixed by #48

PR #48, opened 2018-08-05
Updated Package Support Checks

AirAKose:
Checking for Android, iOS, and UWP packages should now be up-to-date for VS2017 and still work with VS2015 for issue #42 

I tried to make it somewhat expandable via Version-to-path dictionaries with a default. They shouldn't need to be updated unless a package path changes.

I... Couldn't find a better iOS package to check for. That was the only one to mention iOS, unless that's rolled into Xamarin.VisualStudio.Apple.Sdk? Any suggestions for better packages would be much appreciated!

### DIFF
--- a/sources/engine/Xenko.Assets/XenkoConfig.cs
+++ b/sources/engine/Xenko.Assets/XenkoConfig.cs
@@ -20,7 +20,7 @@ namespace Xenko.Assets
         private static readonly string ProgramFilesX86 = Environment.GetEnvironmentVariable(Environment.Is64BitOperatingSystem ? "ProgramFiles(x86)" : "ProgramFiles");
 
         private static readonly Version VS2015Version = new Version(14, 0);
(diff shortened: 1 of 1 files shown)


xen2:
Thanks!

Took a look at https://docs.microsoft.com/en-us/visualstudio/install/workload-component-id-vs-community
Would `Component.Xamarin` work for both iOS and Android?
It seems safer to rely on documented package since smaller ones might changes (merge/split, name changes, etc...).

Also, it might be a good opportunity to rename Packages to Components since that's how they are called (it would avoid confusion with Xenko packages and NuGet packages). Sorry that I used the wrong nomenclature in #42.

AirAKose:
That's entirely fair! Thanks for the link, I tried to find that list and was definitely searching for the wrong terms. n_n' I completely agree that `Component.Xamarin` would be best.

No problem on the nomenclature; terminology gets jumbled pretty often. I'm happy to not make it any more jumbled!

Those changes should be in now

xen2:
Thanks!

PR #52, opened 2018-08-05
update Microsoft.Build.* packages

jhorv:
update Microsoft.Build.* packages, fixing "Could not find a compatible version of MSBuild." errors

### DIFF
--- a/sources/assets/Xenko.Core.Assets.CompilerClient/Xenko.Core.Assets.CompilerClient.csproj
+++ b/sources/assets/Xenko.Core.Assets.CompilerClient/Xenko.Core.Assets.CompilerClient.csproj
@@ -28,9 +28,6 @@
     </Compile>
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.Build" Version="15.7.179" ExcludeAssets="runtime" />
-  </ItemGroup>
-  <ItemGroup>
     <ProjectReference Include="..\..\tools\Xenko.ExecServer\Xenko.ExecServer.csproj">
       <Private>False</Private>
     </ProjectReference>
--- a/sources/assets/Xenko.Core.Assets.CompilerClient/Xenko.Core.Assets.CompilerClient.csproj
+++ b/sources/assets/Xenko.Core.Assets.CompilerClient/Xenko.Core.Assets.CompilerClient.csproj
@@ -29,7 +29,6 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.Build" Version="15.7.179" ExcludeAssets="runtime" />
-    <PackageReference Include="Microsoft.Build.Locator" Version="1.0.18" ExcludeAssets="runtime" />
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\..\tools\Xenko.ExecServer\Xenko.ExecServer.csproj">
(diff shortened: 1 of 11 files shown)


sources/assets/Xenko.Core.Assets.CompilerClient/Xenko.Core.Assets.CompilerClient.csproj
```diff
@@ -27,11 +27,15 @@
       <Link>Package.Constants.cs</Link>
     </Compile>
   </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Build" Version="15.7.179" ExcludeAssets="runtime" />
+    <PackageReference Include="Microsoft.Build.Locator" Version="1.0.18" ExcludeAssets="runtime" />
```
Kryptos-FR: I don't think the locator should be excluded at runtime. Or it means it is not needed here.

sources/assets/Xenko.Core.Assets/Xenko.Core.Assets.csproj
```diff
@@ -12,21 +12,11 @@
...
-    <PackageReference Include="Microsoft.Build.Tasks.Core" Version="15.6.82">
-      <ExcludeAssets>runtime</ExcludeAssets>
-    </PackageReference>
-    <PackageReference Include="Microsoft.Build.Utilities.Core" Version="15.6.82">
-      <ExcludeAssets>runtime</ExcludeAssets>
-    </PackageReference>
     <Reference Include="Microsoft.CSharp" />
     <Reference Include="System.ServiceModel" />
+    <PackageReference Include="Microsoft.Build" Version="15.7.179" ExcludeAssets="runtime" />
+    <PackageReference Include="Microsoft.Build.Locator" Version="1.0.18" ExcludeAssets="runtime" />
```
Kryptos-FR: Same here: locator excluded...

jhorv:
In cases where the Build.Framework reference is omitted, I did so because Microsoft.Build references it implicitly.

Re: your other comments, I'll test whether the Locator reference is needed there.

jhorv:
I went back and made adjustments:

- Xenko.Core.Assets references Locator without ExcludeAssets="runtime"
- Xenko.Core.Assets.CompilerClient doesn't need a ref to Locator at all

As for why this works - my guess/hope is simply that the latest version of MSBuild API does a better job of finding MSBuild.

Also, it didn't work for me at first either - I had to clean and manually kill a bunch of lingering MSBuild.exe processes before it built without errors.

Kryptos-FR:
> As for why this works - my guess/hope is simply that the latest version of MSBuild API does a better job of finding MSBuild.

Let's hope so.

Aside from that, LGTM.

xen2:
Thanks!
Probably the Locator assembly (we were still using an early beta version).

2 small details:
* Seems like you missed one reference: `sources/core/Xenko.Core.AssemblyProcessor/Xenko.Core.AssemblyProcessor.csproj:    <PackageReference Include="Microsoft.Build.Utilities.Core" Version="15.6.82">`.
* CompilerClient shouldn't need `Microsoft.Build` either.

jhorv:
> * Seems like you missed one reference: sources/core/Xenko.Core.AssemblyProcessor/Xenko.Core.AssemblyProcessor.csproj: <PackageReference Include="Microsoft.Build.Utilities.Core" Version="15.6.82">.

Interesting!  I missed AssemblyProcessor because it isn't part of the Xenko.sln solution, and it is only referenced as an Assembly reference by Xenko.GameStudio.  Despite that, it gets built somewhere along the way (I didn't investigate but I assume you know how).

> * CompilerClient shouldn't need Microsoft.Build either.

Removed.

PR #62, opened 2018-08-06
[Math] Add Single2/3/4, Double2/3/4, make Vectors etc use real_t

aaronfranke:
Xenko is an engine designed for the future. You know what is a great feature to have for future games? Double precision float support! We're a long way off from that, but this PR is the first stepping stone.

Why double precision? Here's how Unigine brands it: [Unbounded worlds](https://unigine.com/en/products/engine/unbounded-world). I highly recommend reading it.

Some games using double precision already include Minecraft, Star Citizen, and Space Engineers. Currently most games using double-precision use a custom engine. Unigine is the only one I know of with full double precision support, while CryEngine and Godot have half-finished support.

This pull request does two things. 

1. It adds Single2/3/4 and Double2/3/4 classes, for when you explicitly need to work with either Single or Double precision. These have conversions between each other, Vectors, and Halfs. I added these since the standard in Xenko seems to be parallel implementations, judging by the existing Int2/3/4, Size2/3/4, Half2/3/4, etc, classes.

2. This PR changes "float" in several math types to "real_t". The easiest way to support doubles, at least for transitioning, is to support Single and Double precision simultaneously. This can be done in C# with "using" and preprocessor directives. This commit abstracts away the "float" variables in a few classes to a new variable called "real_t", which for now can only represent "float".

The keyword "real_t" is chosen because it's a type used for "real" numbers, and because it's an abstract type, which are typically denoted by the suffix `_t` (for example, Unix `time_t`).

### DIFF
--- a/sources/core/Xenko.Core.Mathematics/Double2.cs
+++ b/sources/core/Xenko.Core.Mathematics/Double2.cs
@@ -41,167 +41,106 @@ using real_t = System.Single;
 namespace Xenko.Core.Mathematics
 {
     /// <summary>
-    /// Represents a four dimensional mathematical vector.
+    /// Represents a two dimensional mathematical vector.
     /// </summary>
-    [DataContract("real_t4")]
+    [DataContract("double2")]
     [DataStyle(DataStyle.Compact)]
     [StructLayout(LayoutKind.Sequential, Pack = 4)]
-    public struct Vector4 : IEquatable<Vector4>, IFormattable
+    public struct Double2 : IEquatable<Double2>, IFormattable
     {
         /// <summary>
-        /// The size of the <see cref="Xenko.Core.Mathematics.Vector4"/> type, in bytes.
+        /// The size of the <see cref="Xenko.Core.Mathematics.Double2"/> type, in bytes.
         /// </summary>
-        public static readonly int SizeInBytes = Utilities.SizeOf<Vector4>();
+        public static readonly int SizeInBytes = Utilities.SizeOf<Double2>();
 
         /// <summary>
-        /// A <see cref="Xenko.Core.Mathematics.Vector4"/> with all of its components set to zero.
+        /// A <see cref="Xenko.Core.Mathematics.Double2"/> with all of its components set to zero.
         /// </summary>
-        public static readonly Vector4 Zero = new Vector4();
+        public static readonly Double2 Zero = new Double2();
 
         /// <summary>
-        /// The X unit <see cref="Xenko.Core.Mathematics.Vector4"/> (1, 0, 0, 0).
+        /// The X unit <see cref="Xenko.Core.Mathematics.Double2"/> (1, 0).
         /// </summary>
-        public static readonly Vector4 UnitX = new Vector4(1.0f, 0.0f, 0.0f, 0.0f);
+        public static readonly Double2 UnitX = new Double2(1.0, 0.0);
 
         /// <summary>
-        /// The Y unit <see cref="Xenko.Core.Mathematics.Vector4"/> (0, 1, 0, 0).
+        /// The Y unit <see cref="Xenko.Core.Mathematics.Double2"/> (0, 1).
         /// </summary>
-        public static readonly Vector4 UnitY = new Vector4(0.0f, 1.0f, 0.0f, 0.0f);
+        public static readonly Double2 UnitY = new Double2(0.0, 1.0);
 
         /// <summary>
-        /// The Z unit <see cref="Xenko.Core.Mathematics.Vector4"/> (0, 0, 1, 0).
+        /// A <see cref="Xenko.Core.Mathematics.Double2"/> with all of its components set to one.
         /// </summary>
-        public static readonly Vector4 UnitZ = new Vector4(0.0f, 0.0f, 1.0f, 0.0f);
-
-        /// <summary>
-        /// The W unit <see cref="Xenko.Core.Mathematics.Vector4"/> (0, 0, 0, 1).
-        /// </summary>
-        public static readonly Vector4 UnitW = new Vector4(0.0f, 0.0f, 0.0f, 1.0f);
-
-        /// <summary>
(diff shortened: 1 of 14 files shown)


xen2:
While I am fine with the concept to offer double precision to solve large world issue, isn't it usually better solved by moving your center of the world next to the player?
Of course, double/float switch is much more easy to use (single setting) than recentering, so I agree it could be good to have this option as well.

If we allow it, my main practical problem is: if using `#ifdef`, it means we have to ship different runtime of the engine? or do you expect people to recompile engine? The problem with the second approach will be with external plugins, which won't be compiled for both versions as they will likely be distributed as binaries.
Ideally I would prefer not to multiply binaries because it will easily end up with combinatorial explosion in terms of number or runtime versions to compile (just think of a few other features like this and we will quickly end up with 32 or 64+ versions to compile).
Would be good to find a solution for how to deal with that early on.

aaronfranke:
In all cases, you have to store double-precision coordinates *somewhere*. 

Yes, you can move things around so that the player is closer to the world origin, and most large-scale games would want to do this at least at a low-level for rendering, since GPUs mostly only accept single-precision. You do still need to store the absolute positions of things using double-precision with a large-scale game. For example, Kerbal Space Program is made in Unity, which uses single-precision, but KSP has its own double-precision Vector3 and Quaternion etc for internal math, I highly recommend the [video on how they implemented it](https://www.youtube.com/watch?v=mXTxQko-JH0&t=4m15s). But if you've ever played KSP you would know that there are *still* plenty of precision-caused bugs that can tear your spacecraft apart and distort rendering.

The best option is for the engine to have first-class support for double-precision. Having Double2/3/4 classes built-in helps for sure, but ideally in the far future we'd want the engine to be using double-precision math everywhere. Now, in terms of implementation:

> a few other features like this and we will quickly end up with 32 or 64+ versions to compile

There isn't really any downside to this PR as-is, single-precision works fine and compiling with double is not even an option yet. At worst you could make it an unsupported option.

In the far future you could remove support for single-precision floats. Modern FPUs and instruction sets included with 64-bit CPUs can process double-precision floats at basically the same speed. This isn't the case for old 32-bit devices, but since Xenko is an engine targeted at the future, this isn't a problem. The only real downside to removing single-precision for Xenko would be breaking existing projects.

> it means we have to ship different runtime of the engine?

Yes, or users self-compile. Until polished the latter would likely be the only option. As stated if code complexity is an issue you could remove single-precision in the far future.

xen2:
Idea for a fun R&D project: would be quite interested to see how much the overall codebase would differ between float/double, to see if there's a pattern/code construct (attribute?) we could use to be able to automatically patch those during game build time (depending on the float/double setting) to keep a single runtime.

mafiesto4:
I would stick to using single precision floats (#performance). To support large worlds engine should implement world origin shifting like PhysX lib does (https://docs.nvidia.com/gameworks/content/gameworkslibrary/physx/guide/Manual/OriginShift.html#shifting-the-scene-origin).

aaronfranke:
As for performance, FPU performance on 64-bit CPUs is nearly identical, most convert to 80-bit internally ("extended double"). SIMD optimization benefits can be lessened for devices with SIMD widths of less than 256 bits. I don't even know if this code is designed to take advantage of SIMD instructions. The obvious and unavoidable one is storage size, but that's not too significant of an increase (an extra 50 bytes per object transformation?)

Still, unless you decide to remove single precision support, you could always just compile with singles. And we could do performance tests/comparisons in the future to decide that.

HeadClot:
I would like to see the PR merged into the repo if possible. This solves one of my problems for a Game project that I would like to make. :)

Kryptos-FR:
@HeadClot No. Definitely not.

Like @xen2 explained it has a lot of impact that haven't been completely evaluated yet. And we probably don't want to have to maintain it while we are still lacking resources for other more urgent tasks.

See with @aaronfranke if you can grab it and merge it into your own fork.

aaronfranke:
@HeadClot While this PR does add Double2/3/4 and Single2/3/4 which are usable now, double-precision support for the rest of the engine is very very incomplete and will not work yet. Still, you are welcome to add my changes into your own fork, as suggested.

Take your time, Xenko team. I definitely think that double-precision is the way to go but I encourage you to explore options to get a better understanding of what's best. 

By the way, I'm a C# programmer, not an existing Xenko user. The decision y'all make should be one that benefits all Xenko users, and only Xenko users know what's best for Xenko users.

xen2:
@aaronfranke Thanks for your understanding!
All options are still open, I just don't want to close doors too early by locking us into a specific path until we have some time to think it though and make a plan that will cover all use cases.

Hope you're fine compiling and using a fork in the meantime.

sources/core/Xenko.Core.Mathematics/Double2.cs
```diff
@@ -0,0 +1,1506 @@
+// Copyright (c) Xenko contributors (https://xenko.com) and Silicon Studio Corp. (https://www.siliconstudio.co.jp)
+// Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
```
Kryptos-FR: Just a thought.
Should the copyright of new files still include "SiliconStudio"? I would assume their copyright does not extend past existing files. But I'm not a lawyer...

aaronfranke: Good question, but also, these probably count as existing files. I literally did a copy/paste Vector2/3/4 and find/replace "float" with "double" to create Double2/3/4. Then a few other minor things like fixing errors and converting between Vectors and Singles.

EmptyKeys:
As one of the lead developers of Space Engineers i would suggest not to put that to master. It would be cool as feature tho, but not as replacement of current floats. Some kind of addon ... why not.

xen2:
Due to lack of decision/consensus, I will close this PR for now.

Feel free to open an issue if you want to track decisions/progress on large world support. You could also link your repo fork in that issue so that anybody interested could find and use it in the meantime).

PR #71, opened 2018-08-06
fix 600+ xunit usage warnings

jhorv:
Fixes many warnings related to xunit usage, reducing total # of warnings on my machine from 1223 to 616.  Ultimate goal is to fix all warnings so they are useful again.

Prior to this I had 33 tests fail.  After this commit I have 30 tests fail.

I haven't yet dissected which ones differ because the VS test runner isn't so great at managing this many tests.  Is there any automation/script for running tests via the console runner?


xen2:
Great work, thanks!
We did the switch from NUnit to xUnit recently (https://github.com/xenko3d/xenko/commit/085bce68167d589e067ec4d1539e51a3bf7b115b), so there was still some code using the obsolete NUnit-like API.

Currently running unit tests on teamcity to make sure everything is OK.

gitfool:
Nice. I was doing exactly the same thing! 😉 

I was also going to point out that switching to FluentAssertions would be much easier to use...

xen2:
Thanks! (no worries for failed checks, the D3D tests were already broken and number of failed tests didn't change)

PR #75, opened 2018-08-07
[GameStudio] Don't block the UI thread while loading the Backers file.

Kryptos-FR:
Using `Task.Result` blocks the calling thread (here main UI thread) while the task is not completed.

Replaced with a proper async method and a `DependencyProperty`.


xen2:
FYI, clicking links like Privacy Policy and 3rd party were already blocking so I didn't bother...
Thanks for fixing!

Kryptos-FR:
True. Should also be done in a worker thread I think. I will fix that as well.

PR #76, opened 2018-08-07
[GameStudio] Don't block the UI thread when clicking on the hyperlinks.

Kryptos-FR:
Following #75, load the markdown document inside a task when clicking on the hyperlink.


xen2:
Thanks!

PR #77, opened 2018-08-07
[GitHub] changed abstract 'vvvv' to vvvv.org with link

tebjan:
vvvv is pretty abstract for people who don't know it, so i changed it to vvvv.org


xen2:
No problem, thanks for your support!

PR #84, opened 2018-08-09
Fix null reference exception thrown when TextBlock text is null.

jayrulez:
Prevent TextBlock from throwing null reference exception when WrapText is set to true and the value of text is null. Fixes issue#82


Kryptos-FR:
Thanks. LGTM

PR #85, opened 2018-08-09
Add Windows Mixed Reality support

Aminator9000:
In addition to Oculus and OpenVR I added Windows Mixed Reality support for UWP. This brings not only support for WMR headsets (which were already supported thanks to SteamVR) but also HoloLens, although testing for the latter would be needed. As the WMR motion controllers have a thumbstick and a touchpad, I made it available to differentiate between the two while they still map to the same on the other platforms.

To use Windows Mixed Reality you just add the API to the list of required APIs and preferably set it at the top of the list so that it won't use OpenVR which would result in an exception. Next you need to start the game from a CoreWindow context and set the optional parameter "isWindowsMixedReality" to true or else it uses the swap chain of the CoreWindow instead of the holographic space. Be sure to not activate the CoreWindow as this must happen after creating the holographic space which is done internally.

Current issues:

- No post-processing
- No MSAA
- Controller movement is still a bit finicky especially when throwing things.
- Just-In-Time debugging breaks when the headset looses tracking (just turn it off).

I hope this addition will lead to more titles produced with WMR support and as always please test it if you have a WMR device (or emulate it with the Windows Mixed Reality Portal app).

### DIFF
--- a/sources/core/Xenko.Core.Mathematics/Matrix.cs
+++ b/sources/core/Xenko.Core.Mathematics/Matrix.cs
@@ -3379,40 +3379,6 @@ namespace Xenko.Core.Mathematics
             return Equals((Matrix)value);
         }
 
-#if XENKO_PLATFORM_UWP
-        /// <summary>
-        /// Performs an implicit conversion from <see cref="Xenko.Core.Mathematics.Matrix"/> to <see cref="System.Numerics.Matrix4x4"/>.
-        /// </summary>
-        /// <param name="value">The value.</param>
-        /// <returns>The result of the conversion.</returns>
-        public static implicit operator System.Numerics.Matrix4x4(Matrix value)
-        {
-            return new System.Numerics.Matrix4x4()
-            {
-                M11 = value.M11, M12 = value.M12, M13 = value.M13, M14 = value.M14,
-                M21 = value.M21, M22 = value.M22, M23 = value.M23, M24 = value.M24,
-                M31 = value.M31, M32 = value.M32, M33 = value.M33, M34 = value.M34,
-                M41 = value.M41, M42 = value.M42, M43 = value.M43, M44 = value.M44
-            };
-        }
-
-        /// <summary>
-        /// Performs an implicit conversion from <see cref="System.Numerics.Matrix4x4"/> to <see cref="Xenko.Core.Mathematics.Matrix"/>.
-        /// </summary>
-        /// <param name="value">The value.</param>
-        /// <returns>The result of the conversion.</returns>
-        public static implicit operator Matrix(System.Numerics.Matrix4x4 value)
-        {
-            return new Matrix()
-            {
-                M11 = value.M11, M12 = value.M12, M13 = value.M13, M14 = value.M14,
-                M21 = value.M21, M22 = value.M22, M23 = value.M23, M24 = value.M24,
-                M31 = value.M31, M32 = value.M32, M33 = value.M33, M34 = value.M34,
-                M41 = value.M41, M42 = value.M42, M43 = value.M43, M44 = value.M44
-            };
-        }
-#endif
-
 #if SlimDX1xInterop
         /// <summary>
         /// Performs an implicit conversion from <see cref="Xenko.Core.Mathematics.Matrix"/> to <see cref="SlimDX.Matrix"/>.
--- a/sources/core/Xenko.Core.Mathematics/Matrix.cs
+++ b/sources/core/Xenko.Core.Mathematics/Matrix.cs
@@ -3379,6 +3379,40 @@ namespace Xenko.Core.Mathematics
             return Equals((Matrix)value);
         }
 
+#if XENKO_PLATFORM_UWP
+        /// <summary>
+        /// Performs an implicit conversion from <see cref="Xenko.Core.Mathematics.Matrix"/> to <see cref="System.Numerics.Matrix4x4"/>.
+        /// </summary>
+        /// <param name="value">The value.</param>
+        /// <returns>The result of the conversion.</returns>
+        public static implicit operator System.Numerics.Matrix4x4(Matrix value)
+        {
+            return new System.Numerics.Matrix4x4()
+            {
+                M11 = value.M11, M12 = value.M12, M13 = value.M13, M14 = value.M14,
+                M21 = value.M21, M22 = value.M22, M23 = value.M23, M24 = value.M24,
+                M31 = value.M31, M32 = value.M32, M33 = value.M33, M34 = value.M34,
+                M41 = value.M41, M42 = value.M42, M43 = value.M43, M44 = value.M44
+            };
+        }
+
+        /// <summary>
+        /// Performs an implicit conversion from <see cref="System.Numerics.Matrix4x4"/> to <see cref="Xenko.Core.Mathematics.Matrix"/>.
+        /// </summary>
+        /// <param name="value">The value.</param>
+        /// <returns>The result of the conversion.</returns>
+        public static implicit operator Matrix(System.Numerics.Matrix4x4 value)
+        {
+            return new Matrix()
+            {
+                M11 = value.M11, M12 = value.M12, M13 = value.M13, M14 = value.M14,
+                M21 = value.M21, M22 = value.M22, M23 = value.M23, M24 = value.M24,
+                M31 = value.M31, M32 = value.M32, M33 = value.M33, M34 = value.M34,
+                M41 = value.M41, M42 = value.M42, M43 = value.M43, M44 = value.M44
+            };
+        }
+#endif
+
 #if SlimDX1xInterop
         /// <summary>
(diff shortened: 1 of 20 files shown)


sources/core/Xenko.Core.Mathematics/Quaternion.cs
```diff
@@ -1358,6 +1358,28 @@ public override bool Equals(object value)
...
+        {
+            return new System.Numerics.Quaternion(value.X, value.Y, value.Z, value.W);
+        }
+
+        /// <summary>
+        /// Performs an implicit conversion from <see cref="System.Numerics.Quaternion"/> to <see cref="Xenko.Core.Mathematics.Quaternion"/>.
+        /// </summary>
+        /// <param name="value">The value.</param>
+        /// <returns>The result of the conversion.</returns>
+        public static implicit operator Quaternion(System.Numerics.Quaternion value)
```
Kryptos-FR: Same remark about both-way implicit conversion.

sources/core/Xenko.Core.Mathematics/Vector3.cs
```diff
@@ -1688,6 +1688,27 @@ public override bool Equals(object value)
...
+        {
+            return new System.Numerics.Vector3(value.X, value.Y, value.Z);
+        }
+
+        /// <summary>
+        /// Performs an implicit conversion from <see cref="System.Numerics.Vector3"/> to <see cref="Xenko.Core.Mathematics.Vector3"/>.
+        /// </summary>
+        /// <param name="value">The value.</param>
+        /// <returns>The result of the conversion.</returns>
+        public static implicit operator Vector3(System.Numerics.Vector3 value)
```
Kryptos-FR: Same remark about both-way implicit conversion.

sources/engine/Xenko.Engine/Rendering/Compositing/ForwardRenderer.cs
```diff
@@ -116,7 +116,7 @@ public partial class ForwardRenderer : SceneRendererBase, ISharedRenderer
         [DefaultValue(true)]
         public bool BindDepthAsResourceDuringTransparentRendering { get; set; } = true;
 
-        private Logger logger { get; } = GlobalLogger.GetLogger(nameof(ForwardRenderer));
+        private Logger Logger { get; } = GlobalLogger.GetLogger(nameof(ForwardRenderer));
```
Kryptos-FR: I was about to say that `private` fields are always lowercase. But then I realized it is actually a `private` property. 

I don't see the point of having a `private` property here. I think it should be converted to a `private readonly` field.

sources/engine/Xenko.Graphics/WindowsMixedReality/WindowsMixedRealityGraphicsPresenter.cs
```diff
@@ -0,0 +1,151 @@
...
+using System;
+using System.Runtime.InteropServices;
+using SharpDX.Direct3D11;
+using Windows.Graphics.DirectX.Direct3D11;
+using Windows.Graphics.Holographic;
+using Windows.UI.Core;
+
+namespace Xenko.Graphics
+{
+    public class WindowsMixedRealityGraphicsPresenter : GraphicsPresenter
```
Kryptos-FR: Having that class in `Xenko.Graphics` forces to have dependency on `Windows.Graphics.Holographic` for all UWP projects (even those that are not targeting such device. I think we should come up with a better design.

sources/engine/Xenko.Games/GameContext.UWP.cs
```diff
@@ -44,11 +43,14 @@ public GameContextUWPXaml(SwapChainPanel control, int requestedWidth = 0, int re
...
         /// <inheritDoc/>
-        public GameContextUWPCoreWindow(CoreWindow control, int requestedWidth = 0, int requestedHeight = 0)
+        public GameContextUWPCoreWindow(CoreWindow control, int requestedWidth = 0, int requestedHeight = 0, bool isWindowMixedReality = false)
             : base(control ?? CoreWindow.GetForCurrentThread(), requestedWidth, requestedHeight)
         {
             ContextType = AppContextType.UWPCoreWindow;
+            IsWindowMixedReality = isWindowMixedReality;
         }
+
+        public bool IsWindowMixedReality { get; }
```
Kryptos-FR: Abstraction leak?

sources/engine/Xenko.Graphics/WindowsMixedReality/WindowsMixedRealityGraphicsPresenter.cs
xen2: Yes and ideally it should all be doable as a plugin, but there is unfortunately quite a bunch of non-trivial extensions point to add, could be a lot of work (it affects renderer code and game creation).

HeadClot:
Will start testing this. Good work :)

sources/engine/Xenko.Graphics/WindowsMixedReality/WindowsMixedRealityGraphicsPresenter.cs
```diff
@@ -0,0 +1,155 @@
...
+
+                // This can change every frame as the system moves to the next buffer in the
+                // swap chain. This mode of operation will occur when certain rendering modes
+                // are activated.
+                Texture2D d3DBackBuffer = new Texture2D(resource);
+
+                backBuffer = new Texture(GraphicsDevice).InitializeFromImpl(d3DBackBuffer, false);
+
+                LeftEyeBuffer = backBuffer.ToTextureView(new TextureViewDescription() { ArraySlice = 0, Type = ViewType.Single });
+                RightEyeBuffer = backBuffer.ToTextureView(new TextureViewDescription() { ArraySlice = 1, Type = ViewType.Single });
```
xen2: Those two buffers should be disposed if they exist before being recreated.

sources/core/Xenko.Core.Mathematics/Matrix.cs
```diff
@@ -3379,6 +3379,40 @@ public override bool Equals(object value)
...
+                M41 = value.M41, M42 = value.M42, M43 = value.M43, M44 = value.M44
+            };
+        }
+
+        /// <summary>
+        /// Performs an implicit conversion from <see cref="System.Numerics.Matrix4x4"/> to <see cref="Xenko.Core.Mathematics.Matrix"/>.
+        /// </summary>
+        /// <param name="value">The value.</param>
+        /// <returns>The result of the conversion.</returns>
+        public static implicit operator Matrix(System.Numerics.Matrix4x4 value)
```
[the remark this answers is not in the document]
Aminator9000: @Kryptos-FR As Windows.Perception.Spatial uses System.Numerics types I needed to add these implicit conversions to reduce bloat. There should be nothing wrong with making the conversion implicit both ways as both use floats. I mean, even the existing conversions are implicit both ways if both types use the same underlying primitive types.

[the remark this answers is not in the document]
xen2: @Kryptos-FR It seems like `System.Numerics` is already an implicit dependency (quite surprised, not sure where it comes from, seems automatic for Windows and UWP platforms!)
@Aminator9000 Not a big fan of having things halfway (conversion only on UWP).
So either we:
- Add `System.Numerics` and this conversion for all platforms. It implies you are able to tests those platforms.
- You move this `System.Numerics` implicit conversion temporarily where you need it as an explicit method (and we can always add it back to Mathematics later for everybody)

sources/engine/Xenko.VirtualReality/WindowsMixedReality/WindowsMixedRealityHmd.cs
```diff
@@ -0,0 +1,202 @@
+// Copyright (c) Xenko contributors (https://xenko.com) and Silicon Studio Corp. (https://www.siliconstudio.co.jp)
```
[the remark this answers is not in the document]
Kryptos-FR: @xen2 Let's add something in the doc/readme or somewhere else (wiki?) to precise a template for new files.
I think we could also port the "coding guidelines" we had in confluence and ad it to the wiki, to ensure a uniform coding style and naming conventions.

xen2:
@Aminator9000 Please let me know when this is ready for final review/merge (if not already).

Aminator9000:
@xen2 It is ready now although the above mentioned issues are still there. Those could be fixed later as I am currently not sure how to fix them.

xen2:
Sorry for the late review, LGTM!
Thanks!

yar3333:
Please, help me to create a minimal WMR application. I do all what neccessary after reading all about VR in docs. I create UWP app, remove XAML Page component and edit `App.OnLaunched` code to this:
```c#
protected override void OnLaunched(LaunchActivatedEventArgs e)
{
	Game = new Game();
	Game.UnhandledException += Game_UnhandledException;
	Game.Run(new GameContextUWPCoreWindow(null, 0, 0, true));
}
```
WMR portal started, VR controllers visible in simulator, but all around is black. What I do wrong?

yar3333:
I found solution - *.t4 in Xenko sources contain code to create `FrameworkView`. No action needed :)

Issue #88, opened 2018-08-09
Physics error "Pair not present" - TopDownRPG and Physics Sample


aurodev:
TopDownRPG crashes at runtime after upgrading in GameStudio v3.0.0.3 Dev build
Also occurs with a new 3.0.0.3 dev build generated sample.

Physics Sample displays the same "Pair not present" error 
when switching screens after "Shoot the cubes"

**System**
**OS/device including version:**
Windows 8.1 64bit, 8GB AMD FX 6350 6 core

**Graphics Chipset**
AMD Radeon HD 7700 Series
Memory Size 1024 MB
Radeon Software Version 17.4.4

________________________________

VS 2017 Debugger
_______________
System.Exception
  HResult=0x80131500
  Message=**Pair not present**.
  Source=Xenko.Physics
  StackTrace:
   at **Xenko.Physics.Simulation.ContactRemoval**(ContactPoint contact, PhysicsComponent component0, PhysicsComponent component1) in C:\Users\Greg\Documents\Xenko\sources\engine\Xenko.Physics\Simulation.cs:line 1077
   at Xenko.Physics.Simulation.EndContactTesting() in C:\Users\Greg\Documents\Xenko\sources\engine\Xenko.Physics\Simulation.cs:line 1214
   at Xenko.Physics.Bullet2PhysicsSystem.Update(GameTime gameTime) in C:\Users\Greg\Documents\Xenko\sources\engine\Xenko.Physics\Bullet2PhysicsSystem.cs:line 105
   at Xenko.Games.GameSystemCollection.Update(GameTime gameTime) in C:\Users\Greg\Documents\Xenko\sources\engine\Xenko.Games\GameSystemCollection.cs:line 98
   at Xenko.Games.GameBase.Update(GameTime gameTime) in C:\Users\Greg\Documents\Xenko\sources\engine\Xenko.Games\GameBase.cs:line 918
   at Xenko.Games.GameBase.UpdateAndProfile(GameTime gameTime) in C:\Users\Greg\Documents\Xenko\sources\engine\Xenko.Games\GameBase.cs:line 926
   at Xenko.Games.GameBase.TickInternal() in C:\Users\Greg\Documents\Xenko\sources\engine\Xenko.Games\GameBase.cs:line 677
   at Xenko.Games.GameBase.Tick() in C:\Users\Greg\Documents\Xenko\sources\engine\Xenko.Games\GameBase.cs:line 510
[...]

PR #99, opened 2018-08-11
change physics exception to log a warning instead

jhorv:
Workaround for [issue #88](https://github.com/xenko3d/xenko/issues/88).  Logs a Warning instead of throwing an exception.  Easily to reproduce via:

* In TopDownRPG, punch a cube, pick up coins, issue is triggered.
* In PhysicsSample, change scenes to trigger.

While playing with PhysicsSample, I started to get the feeling that it was related to objects disappearing from the scene, but not being fully removed from the physics simulation?


xen2:
Thanks!

Issue #12, opened 2018-08-03
Improve Build System Epic


xen2:
Right now we build Xenko as a single package.
Ideally we should move on to do something more in line with what NuGet expects:
* One NuGet package per dll
* Use "runtimes" for platforms and graphics API
* Games would move from referencing huge `Xenko` to referencing `Xenko.Engine` and `Xenko.Physics` for example.

Other area of improvements:
* Compiling with PackageReference against a local development version (we currently use this "redirect" concept but it's not great). This might need some improvements/discussion with NuGet team.
* Simplify target files
* Review up to date check
* Possibly cross compile multiple platforms from a single solution (`TargetFrameworks`)

Use cases:
* Create solution from VS & add Xenko from NuGet package manager, then open it with Game Studio
* Use only specific part of Xenko from any Visual Studio project
* Create nuget packages from your code/assets and publish them on nuget so that anybody can use them easily
* People can customize small part of Xenko and only publish the modified part in their own streams.
* Much faster to update (i.e. no need for everybody to redownload sample if we update only a specific part of the engine)

DW01:
A good number of large projects on NuGet (i.e. Discord.NET, Newtonsoft.JSON, etc..) also tend to utilise metapackages to make referencing all components or specific subsets of components easier on the developer. Given how many libraries Xenko creates, this could be an idea.

We could, for example, do this:

- Xenko (overall metapackage)
--- Xenko.Graphics (compartmentalised render backends)
--- Xenko.Editor (Game Studio / "native" UI for editor and extensions)
--- Xenko.Tests (unit testing)
--- etc

as well as individual library packages for the adventurous.

weekendkoder:
will there be a webGL export ?

tebjan:
@weekendkoder this is discussed in: #677

weekendkoder:
@tebjan .. okay. Thx.

PR #107, opened 2018-08-14
[Math] Add Double2/3/4 only, fix grammar and inconsistencies

aaronfranke:
As I stated in [my previous PR](https://github.com/xenko3d/xenko/pull/62), if that didn't get merged I would create a PR that just adds Double2/3/4 types (and also I made a few grammar tweaks and fixed inconsistencies). 

Unlike my previous PR, this does NOT change anything to `real_t` and does NOT add Single2/3/4. It does not change any existing code. I don't think anyone would object to adding Double2/3/4 math types built-in for users who wish to do their own math in double-precision without having to re-implement built-in types.

Plus, some of the grammar and wording was incorrect and inconsistent, I've fixed some of that.

### DIFF
--- a/sources/core/Xenko.Core.Mathematics/Double2.cs
+++ b/sources/core/Xenko.Core.Mathematics/Double2.cs
@@ -1,0 +1,1485 @@
+// Copyright (c) Xenko contributors. (https://xenko.com)
+// Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using System;
+using System.ComponentModel;
+using System.Globalization;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+namespace Xenko.Core.Mathematics
+{
+    /// <summary>
+    /// Represents a two dimensional mathematical vector with double-precision floats.
+    /// </summary>
+    [DataContract("double2")]
+    [DataStyle(DataStyle.Compact)]
+    [StructLayout(LayoutKind.Sequential, Pack = 4)]
+    public struct Double2 : IEquatable<Double2>, IFormattable
+    {
+        /// <summary>
+        /// The size of the <see cref="Xenko.Core.Mathematics.Double2"/> type, in bytes.
+        /// </summary>
+        public static readonly int SizeInBytes = Utilities.SizeOf<Double2>();
+
+        /// <summary>
+        /// A <see cref="Xenko.Core.Mathematics.Double2"/> with all of its components set to zero.
+        /// </summary>
+        public static readonly Double2 Zero = new Double2();
+
+        /// <summary>
+        /// The X unit <see cref="Xenko.Core.Mathematics.Double2"/> (1, 0).
+        /// </summary>
+        public static readonly Double2 UnitX = new Double2(1.0, 0.0);
+
+        /// <summary>
+        /// The Y unit <see cref="Xenko.Core.Mathematics.Double2"/> (0, 1).
+        /// </summary>
+        public static readonly Double2 UnitY = new Double2(0.0, 1.0);
+
+        /// <summary>
+        /// A <see cref="Xenko.Core.Mathematics.Double2"/> with all of its components set to one.
+        /// </summary>
+        public static readonly Double2 One = new Double2(1.0, 1.0);
+
+        /// <summary>
+        /// The X component of the vector.
+        /// </summary>
+        [DataMember(0)]
+        public double X;
+
+        /// <summary>
+        /// The Y component of the vector.
+        /// </summary>
+        [DataMember(1)]
+        public double Y;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Xenko.Core.Mathematics.Double2"/> struct.
+        /// </summary>
+        /// <param name="value">The value that will be assigned to all components.</param>
+        public Double2(double value)
+        {
+            X = value;
+            Y = value;
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Xenko.Core.Mathematics.Double2"/> struct.
+        /// </summary>
+        /// <param name="x">Initial value for the X component of the vector.</param>
+        /// <param name="y">Initial value for the Y component of the vector.</param>
+        public Double2(double x, double y)
+        {
+            X = x;
+            Y = y;
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Xenko.Core.Mathematics.Double2"/> struct.
+        /// </summary>
+        /// <param name="values">The values to assign to the X and Y components of the vector. This must be an array with two elements.</param>
+        /// <exception cref="ArgumentNullException">Thrown when <paramref name="values"/> is <c>null</c>.</exception>
+        /// <exception cref="ArgumentOutOfRangeException">Thrown when <paramref name="values"/> contains more or less than two elements.</exception>
+        public Double2(double[] values)
+        {
+            if (values == null)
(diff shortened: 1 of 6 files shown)


tebjan:
Wouldn't this also need a double precision matrix type? The translation component of a matrix should also be able to express all positions of the extended space.

Kryptos-FR:
@tebjan We are not supporting extended space with that PR, just adding a few types. See #62.

sources/core/Xenko.Core.Mathematics/Double2.cs
```diff
@@ -0,0 +1,1480 @@
+// Copyright (c) Xenko contributors (https://xenko.com) and Silicon Studio Corp. (https://www.siliconstudio.co.jp)
```
Kryptos-FR: New files should not include "Silicon Studio" in the copyright header.

sources/core/Xenko.Core.Mathematics/Double2.cs
```diff
@@ -0,0 +1,1480 @@
...
+// Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
+//
+// -----------------------------------------------------------------------------
+// Original code from SlimMath project. http://code.google.com/p/slimmath/
+// Greetings to SlimDX Group. Original code published with the following license:
+// -----------------------------------------------------------------------------
+/*
+* Copyright (c) 2007-2011 SlimDX Group
+*
```
Kryptos-FR: That class is not part of SlimDX, so that copyright is misleading an potentially illegal.

sources/core/Xenko.Core.Mathematics/Vector3.cs
```diff
@@ -144,6 +144,17 @@ public Vector3(float[] values)
             Z = values[2];
         }
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Xenko.Core.Mathematics.Vector3"/> struct.
+        /// </summary>
+        /// <param name="v">The Double3 to construct the Vector3 from.</param>
+        public Vector3(Double3 v)
```
Kryptos-FR: I'm not a fan of constructors that might cause runtime exception because of the downcast from `double`to `float`.
I would prefer to have an explicit operator:
```csharp
public static explicit operator Vector3(Double3 value)
{
    return new Vector3((float)value.X, (float)value.Y, (float)value.Z);
}
```

tebjan:
The question was rather mathematically, what is the use for double types if you can't do proper transformations with them? Multiplying with a single precision matrix will lose information... So what would be the use cases for these new double precision types?

sources/core/Xenko.Core.Mathematics/Vector3.cs
aaronfranke: I wasn't aware that it was possible for that to cause a runtime exception? Are there any particular bad values you have in mind which would cause this?

aaronfranke:
@tebjan It's only useful for manual math, as like you said, it doesn't make sense to use the engine's built-in single-precision types with it. One of many possible use cases (mostly pseudocode):

    Double3 playerPos = new Double3(1000000, 0, 0);
    Double3 enemyPos = new Double3(1000010.1234, 0, 0);
    Double3 floatingOrigin = -playerPos;
    
    // If above math was done in single-precision then the enemy would jitter
    Vector3 enemyWorldPos = (Vector3)(enemyPos + floatingOrigin);

Also, in Kerbal Space Program, they had to implement their own "Vector3d" type, and they calculate positions of orbits etc with double-precision and then make a single-precision local version.

sources/core/Xenko.Core.Mathematics/Vector3.cs
Kryptos-FR: My bad. I thought it would throw an `OverflowException` if you had something like `float f = (float)(((double)float.MaxValue)*2)`. But apparently, that's only the case for signed/unsigned pair for integral types.

But as @xen2 said, when losing precision explicit cast is preferred anyway.

aaronfranke: This, and the licensing stuff, have now been changed.

xen2:
Two questions/remarks:

1. Wondering about the naming. We have this inconsistency in Xenko: we have `Half4`, `Int4` and `Vector4`. Wondering if `Vector4` should be `Float4`? (and what about `Matrix`, should it be `Float4x4`?)
Another approach is `Vector4d` for double, but not so great...

2. Since nothing in the engine depends on those types, I was wondering if they should be part of `Mathematics` or a new optional assembly (that could even be maintained by @aaronfranke out of band so that it can be updated/improved anytime.
Note that this will make much more sense as soon as we switch to proper NuGet packages (#12) and I am fine to make it part of `Mathematics` for now and split it later once we have NuGet packages if we deem necessary.

aaronfranke:
Good questions!

1. That's why I added Single2/3/4 in my last PR. I figured it was best to have "Vector" just be "the number type used by the engine" and "Single" be "float" forever. I can still add these if you wish but it's A) it's redundant if Vector is always float and B) I would NOT recommend renaming Vector since it's a very common name.

2. I'm not familiar with NuGet and I'm not a Xenko user, I only planned to add these to make Xenko more suitable for my uses so that I have more options in the future. So I don't think I should be maintaining a separate package (especially if it just has 3 files).

sources/core/Xenko.Core.Mathematics/Half2.cs
```diff
@@ -175,13 +175,33 @@ public override bool Equals(object obj)
...
-        /// Performs an explicit conversion from <see cref="Vector3"/> to <see cref="Half3"/>.
+        /// Performs an explicit conversion from <see cref="Half2"/> to <see cref="Vector2"/>.
         /// </summary>
         /// <param name="value">The value.</param>
         /// <returns>The result of the conversion.</returns>
         public static explicit operator Vector2(Half2 value)
         {
             return new Vector2(value.X, value.Y);
         }
+
```
xen2: Let's move over all those new implicit operators from `HalfX`/`VectorX` files to `DoubleX` file so that `DoubleX` files are self contained.

sources/core/Xenko.Core.Mathematics/Half2.cs
```diff
@@ -175,13 +175,33 @@ public override bool Equals(object obj)
...
+        /// <summary>
+        /// Performs an explicit conversion from <see cref="Half2"/> to <see cref="Double2"/>.
+        /// </summary>
+        /// <param name="value">The value.</param>
+        /// <returns>The result of the conversion.</returns>
+        public static explicit operator Double2(Half2 value)
+        {
+            return new Double2(value.X, value.Y);
+        }
```
xen2: Let's move over all those new implicit operators from `HalfX`/`VectorX` files to `DoubleX` files so that `DoubleX` files are self contained.

aaronfranke: Alright, so now, `DoubleX.cs` depends on `VectorX` & `HalfX`, but not the other way around.

tebjan:
I would also add a double precision matrix as well, just because you will need it pretty quickly if you mess around with double precision vectors.

Hyrtwol:
my 2 cent would use

Float2/3/4

to mimic hlsl

aaronfranke:
@tebjan Honestly that could be placed in another PR if it's desired.

xen2:
@Hyrtwol Thanks for the feedback. Still not sure what will be done on that front yet, but if something happens it can be done as part of a separate discussion/PR.

xen2:
Merged, thanks!

PR #114, opened 2018-08-16
New public key token for Irony and selecting that specific Irony.dll in Xenko.Core.Shaders

### LINKED ISSUE #112: Conflict with official Irony.dll
Apparently xenko compiles a [custom Irony.dll](https://github.com/xenko3d/xenko/tree/master/sources/shaders/Irony) that is used in the shader compiler. This creates an conflict that cannot be solved if a project that loads the [official Irony.dll from the nuget](https://www.nuget.org/packages/Irony/). Our project uses Irony as well to parse generic type names.

The reason for the conflict is that the custom Irony.dll from xenko uses the same public token as the official one and has no version set, it's 0.0.0.0.
so the custom Irony project should use a different public key token for the strong name in order to load it with the official one side-by-side in the same process.

tebjan:
fixes #112


xen2:
@tebjan Out of curiosity, would it work if the key was removed?
We needed those when we were signing assemblies but it's not the case anymore.

tebjan:
from my understanding, loading two assemblies with the same name only works if the assemblies have specific identifiers apart from the name. correct versioning might also work, but strongly named assemblies are more error prone. more info on that:
https://docs.microsoft.com/en-us/dotnet/framework/deployment/side-by-side-execution

PR #123, opened 2018-08-20
Added InitialCounterOffsetValue property to Buffer

### LINKED ISSUE #122: Add initial counter offset value of compute shader
The initial counter value should be exposed and have a default of -1:

see for example:
https://github.com/sharpdx/SharpDX/blob/master/Source/SharpDX.Direct3D11/DeviceContext.ComputeShaderStage.cs#L65

for reference the issue from the old repo: https://github.com/SiliconStudio/xenko/issues/600

currently working on a pull request.

tebjan:
first draft of adding the reset counter value to compute shaders. the parameter is needed in SetUnorderedAccessView which influences a few signatures. not sure if it is the best way, i am open for suggestions. i also thought about adding it as a property to the parameter updater...

i've also removed "Offset" and "Size" from DescriptorSetEntry since it was never used. It got replaced with the UAVInitialOffset, hope that is fine.

this PR fixes #122

### DIFF
--- a/sources/engine/Xenko.Graphics/DescriptorSet.cs
+++ b/sources/engine/Xenko.Graphics/DescriptorSet.cs
@@ -79,7 +79,7 @@ namespace Xenko.Graphics
             if(unorderedAccessView is Buffer buffer)
             {
                 HeapObjects[DescriptorStartOffset + slot].Value = buffer;
-                HeapObjects[DescriptorStartOffset + slot].Offset = buffer.InitialCounterOffsetValue;
+                HeapObjects[DescriptorStartOffset + slot].Offset = buffer.InitialCounterOffset;
             }
             else
             {
--- a/sources/engine/Xenko.Graphics/DescriptorSet.cs
+++ b/sources/engine/Xenko.Graphics/DescriptorSet.cs
@@ -66,7 +66,7 @@ namespace Xenko.Graphics
         /// <param name="size">The constant buffer view size.</param>
         public void SetConstantBuffer(int slot, Buffer buffer, int offset, int size)
         {
-            HeapObjects[DescriptorStartOffset + slot] = new DescriptorSetEntry(buffer);
+            HeapObjects[DescriptorStartOffset + slot] = new DescriptorSetEntry(buffer, offset, size);
         }
 
         /// <summary>
--- a/sources/engine/Xenko.Graphics/DescriptorSet.cs
+++ b/sources/engine/Xenko.Graphics/DescriptorSet.cs
@@ -74,10 +74,14 @@ namespace Xenko.Graphics
         /// </summary>
         /// <param name="slot">The slot.</param>
         /// <param name="unorderedAccessView">The unordered access view.</param>
+        /// <param name="uavInitialOffset">The Append/Consume buffer offset. A value of -1 indicates the current offset
+        /// should be kept. Any other values set the hidden counter for that Appendable/Consumable
+        /// UAV. uavInitialCount is only relevant for UAVs which have the 'Append' or 'Counter' buffer
+        /// flag, otherwise the argument is ignored.</param>
         public void SetUnorderedAccessView(int slot, GraphicsResource unorderedAccessView, int uavInitialOffset)
         {
             HeapObjects[DescriptorStartOffset + slot].Value = unorderedAccessView;
-            HeapObjects[DescriptorStartOffset + slot].UAVInitialOffset = uavInitialOffset;
+            HeapObjects[DescriptorStartOffset + slot].Offset = uavInitialOffset;
         }
 #endif
     }
--- a/sources/engine/Xenko.Graphics/DescriptorSet.cs
+++ b/sources/engine/Xenko.Graphics/DescriptorSet.cs
@@ -74,14 +74,18 @@ namespace Xenko.Graphics
         /// </summary>
         /// <param name="slot">The slot.</param>
         /// <param name="unorderedAccessView">The unordered access view.</param>
-        /// <param name="uavInitialOffset">The Append/Consume buffer offset. A value of -1 indicates the current offset
-        /// should be kept. Any other values set the hidden counter for that Appendable/Consumable
-        /// UAV. uavInitialCount is only relevant for UAVs which have the 'Append' or 'Counter' buffer
-        /// flag, otherwise the argument is ignored.</param>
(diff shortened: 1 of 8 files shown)


xen2:
Thanks.

`Offset` and `Size` are used on some platforms (D3D12 & Vulkan). The behavior is enabled when `BufferPool.useBufferOffsets` is set.
(note: you couldn't see it easily from D3D11 solution -- we've got to find a way to deal with that)

Since UAV don't need `Offset`, please reuse this field to hold the initial offset value in case of UAV (please add a comment on `Offset` so that people know about its double usage).
Also, will need to implement the UAV offset on all platforms (`DescriptorSet.Direct3D12.cs` and `DescriptorSet.Vulkan.cs`); let me know if you need help in that area.

tebjan:
Ah, i was wondering why i couldn't find usages of the Offset parameter. re-using it is a very good idea.
sure, will add it to DX12 and Vulkan too.

sources/engine/Xenko.Graphics/DescriptorSet.cs
```diff
@@ -66,17 +66,22 @@ public void SetSamplerState(int slot, SamplerState samplerState)
         /// <param name="size">The constant buffer view size.</param>
         public void SetConstantBuffer(int slot, Buffer buffer, int offset, int size)
         {
-            HeapObjects[DescriptorStartOffset + slot] = new DescriptorSetEntry(buffer, offset, size);
+            HeapObjects[DescriptorStartOffset + slot] = new DescriptorSetEntry(buffer);
```
xen2: Please revert this, might be used later (D3D11.2 or OpenGL supports cbuffer offsets/size as well)

tebjan: yes, just noticed myself that i forgot to revert it.

sources/engine/Xenko.Graphics/Rendering/EffectInstance.cs
```diff
@@ -92,11 +92,11 @@ protected virtual void ChooseEffect(GraphicsDevice graphicsDevice)
         {
         }
 
-        public void Apply(GraphicsContext graphicsContext)
+        public void Apply(GraphicsContext graphicsContext, int uavInitialOffset = -1)
```
xen2: The offset is per UAV so it shouldn't end up being global to the effect instance call.
It sounds like it should fit as a `(GraphicsResource, int)` in the `ParameterCollection`.
The issue is that `ParameterCollection` is good to hold either blittable struct (int, float, matrix, etc...) or objects (i.e. GraphicsResource), but not managed struct containing both of them...
We could temporarily wrap this pair of parameter into a class but this is not great for locality/perf (we can deal with that later though).

xen2: You can still set them one by one, so they should probably be stored as `(GraphicsResource, int)` in the `ParameterCollection` (instead of `GraphicsResource` currently), so that you have all the data to call `descriptorSet.SetUnorderedAccessView` in `EffectParameterUpdater.Update`.

tebjan: had some time to look at this again and it seems that adding the counter value to the `ParameterCollection` is a bit tricky because then the shader key generator (visual studio plugin) would need to generate keys with different types for counter or append buffers. the easiest solution seems to be to add a public `int` field to the `Buffer` type and init it with -1. the only drawback i see would be that the parameter setter would have a try cast to `Buffer` to retrieve the counter value, but that would only be in `SetUnorderedAccessView`;
```csharp
public void SetUnorderedAccessView(int slot, GraphicsResource unorderedAccessView, int uavInitialOffset)
{
    if(unorderedAccessView is Buffer buffer)
    {
        HeapObjects[DescriptorStartOffset + slot].Value = buffer;
        HeapObjects[DescriptorStartOffset + slot].Offset = buffer.UAVInitialOffset;
    }
    else
    {
        HeapObjects[DescriptorStartOffset + slot].Value = unorderedAccessView;
        HeapObjects[DescriptorStartOffset + slot].Offset = -1;
    }
}
```
and to set the buffer:
```csharp
myBuffer.UAVInitialOffset = 0;
myEffect.Parameters.Set(myEffectKeys.MyBuffer, myBuffer);
```

tebjan:
i've updated the PR according to [my comment](https://github.com/xenko3d/xenko/pull/123#discussion_r212470810).

sources/engine/Xenko.Graphics/Rendering/EffectInstance.cs
xen2: Two questions then:
1. Is there some cases where we might want to bind the same buffer with different counts within the same frame?
2. If yes, would this count be kind of static (i.e. -1, 0 or 12 in which case we could allow creating static views) or dynamic (can change any frame and needs to be computed and be updated every draw call within same frame -- in which case your approach would be problematic).

I will also try to look more advanced examples (simple ones seems to always be 0 or -1) or try to think of few scenarios on my side to better understand use cases.

tebjan: yes, the count can change every frame. for example we have a use case where we made something like a RingBuffer data structure with a counter buffer of 10.000 elements divided into 10 sections of 1.000 elements (Dispatch call with max. 1.000 threads) and in each frame we write into one of the sections, then the counter cycles thru values of 0, 1000, 2000 etc. during the frames. the next shader in the pipeline that reads the counter buffer doesn't need to set the counter value, it reads the sections by index.

so is it slow to call `myEffect.Parameters.Set(myEffectKeys.MyBuffer, myBuffer);` each frame? 

but yes this is an advanced case, in mostly it is just -1 or 0.

tebjan: i've edited the comment, the question is: would it be enough to only do `myBuffer.UAVInitialOffset = value;` every frame and the `myEffect.Apply` will take care of the updated value? when the value is a property of the buffer.

xen2:
Merged, thanks!

xen2:
@tebjan Sorry, you will have to reopen it.
I had to revert the merge because
* OpenGL/Vulkan/D3D12 the parameter was not added (doesn't compile)
* Your commit changes CRLF into LF. Please check your git configuration.

tebjan:
no problem. yes, i wasn't finished with the other platforms. odd with the line endings... will check.

PR #129, opened 2018-08-22
[Build] Enable StyleCop for various runtime libraries, and remove lot of warnings

xen2:


Kryptos-FR:
1000+ changes files. OK I'm not reviewing it :sweat:

Let's assume you did everything right and let's merge it quickly before it creates conflicts with other branches.

gitfool:
@xen2 Did you use Resharper Code Cleanup? 😉

xen2:
@gitfool no, mostly roslyn fixes (took me a while)
@Kryptos-FR yes I know... (idea is to clean the codebase little bit before more people start joining to the project)

gitfool:
Another good cleanup is using `var` everywhere.

xen2:
@gitfool yes a mix of resharper + roslyn fixes (StyleCop being a roslyn analyzer, it provides the codefixes for its warnings).
Like `var` too (but not everybody does). I know it's part of SonarAnalyzer rules (I intend to enable some of its rules later too).

gitfool:
I added a comment inline, but I can't see it now, so just in case... what do you think about removing the double indenting? e.g. [sources/core/Xenko.Core.Design/PackageVersionRange.cs#L187](https://github.com/xenko3d/xenko/blob/87b14e3cf426e0c5f8263181282ba2bf2b6c3446/sources/core/Xenko.Core.Design/PackageVersionRange.cs#L187) could change from:
```
result = new PackageVersionRange
    {
        MinVersion = version,
        IsMinInclusive = true,
    };
```
to:
```
result = new PackageVersionRange
{
    MinVersion = version,
    IsMinInclusive = true,
};
```

xen2:
@gitfool I personally prefer the second version.
I remember vaguely Resharper was by default doing the first version which was annoying, but not sure.

Kryptos-FR:
@xen2 This commit broke all samples that have code dealing with render feature (e.g SpaceEscape): `SubRenderFeature.RootRenderFeature` was changed to `SubRenderFeature.rootRenderFeature` but no samples template were upgraded to reflect the change.

We should enable the building of all samples in TC and make sure to run it at least once before each release (optionally on every pull-request).

xen2:
@Kryptos-FR Good point. We had this kind of sample building in Silicon Studio teamcity, I will need to check how much of it I can reuse/salvage.

xen2:
Independently, I will revert `rootRenderFeature`, it should be accessible.

Kryptos-FR:
In the meantime I'm fixing the samples...

xen2:
I remember having this conflict on stylecop wanting lower case for protected fields (and upper case for protected properties -- or something along those lines).
However I was quite unsure about it... should protected fields be allowed at all or force protected to be property only? and if we allow protected properties, do we really want them lower case? (note: internal or public fields starts with upper case)

Kryptos-FR:
@xen2 having fields other than private is usually considered a code smell. The only exception is if those fields should be accessed by `ref` for performance.

So yes, properties are always Uppercase. And if we change that field to be a property then there is no compilation error for SpaceEscape ;)

xen2:
@Kryptos-FR Ideally it should be a property but I think it was a field because used in a lot of performance critical tight-loop of rendering code.
So either turn it into a property (probably fine if properly optimized by JIT), or a field with uppercase (for perf) to pretend it's like a protected property that can be used as such in the inheriting classes without much risks of breaking changes.

xen2:
Fixed `rootRenderFeature` with ce3f3f9

xen2:
@Kryptos-FR Reactivated sample unit testing with 2fd02363a
https://teamcity.xenko.com/viewLog.html?buildId=2237&buildTypeId=Engine_Tests_WindowsSamplesD3D&tab=buildResultsDiv
I did it as pure xunit tests (no need to run various manual commands from `XenkoSamples.build`) so it should be much easier for everybody to run those tests.

Issue #110, opened 2018-08-15
Editor crashes when editing script.


jayrulez:
The editor seems to crash constantly when editing scripts.

I start by typing public UIPage...
as soon as I start to type the property name the editor crashes.

```
Application: GameStudio
UserEmail: 
UserMessage: 
XenkoVersion: 3.0.0.3
GameStudioVersion: 4
ThreadName: Main thread
DefaultGraphicProfile: Level_10_0
OpenedAssets: dd73cc2c-f632-4387-9b37-98dac320df05:MainScene (Scene)
4d7fbe74-f6dc-f306-0646-8c39b84fdac3:BattleScript (Script Source Code)
44a4cf64-3848-4e33-8f25-d2ff27542aa9:BattleUI (UI page)

LastActions: * +[Update property Appearance.Clip To Bounds]
    * {AssetContentValueChangeOperation: ClipToBounds = False}
* +[Update property Appearance.Visibility]
    * {AssetContentValueChangeOperation: Visibility(∅ -> *) = Hidden}
* +[Update property Appearance.Visibility]
    * {AssetContentValueChangeOperation: Visibility = Collapsed}
* +[Update property Appearance.Visibility]
    * {AssetContentValueChangeOperation: Visibility = Visible}
* +[Reload game assemblies]
    * {ContentValueChangeOperation: EntityComponentCollection[1] --[1] = }
    * {ContentValueChangeOperation: EntityComponentCollection[2] --[2] = }
    * {ReloadAssembliesOperation}
    * {AnonymousDirtyingOperation}
    * {ContentValueChangeOperation: EntityComponentCollection[2] ++[2] = TurnBasedBattleSim.BasicCameraController}
    * {ContentValueChangeOperation: EntityComponentCollection[1] ++[1] = TurnBasedBattleSim.Game.BattleScript}
    * {AnonymousDirtyingOperation}

CurrentDirectory: C:\Program Files\Xenko
CommandArgs: /LauncherWindowHandle 11274966 C:\Users\Robert Campbell\Desktop\GameDev\Xenko\TurnBasedBattleSim\TurnBasedBattleSim.sln
OsVersion: Microsoft Windows 10 Pro 10.0.17134 x64
ProcessorCount: 4
Exception: TaskCanceledException: A task was canceled.
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
[...]

PR #135, opened 2018-08-24
Fix GameStudio crashing when editing files with the simple code editor

Eideren:
Fixes issue #110. 
Once "CaretOnPositionChanged" is called, cancels the previous working call through its cancellation token. The exception generated through that token has to be caught somewhere, this pull catches it where it should be most relevant.

### DIFF
--- a/sources/editor/Xenko.Assets.Presentation/AssetEditors/ScriptEditor/SimpleCodeTextEditor.cs
+++ b/sources/editor/Xenko.Assets.Presentation/AssetEditors/ScriptEditor/SimpleCodeTextEditor.cs
@@ -181,17 +181,17 @@ namespace Xenko.Assets.Presentation.AssetEditors.ScriptEditor
             braceMatchingCts = cts;
 
             var document = workspace.Host.GetDocument(documentId);
-            var text = await document.GetTextAsync().ConfigureAwait(false);
-            var caretOffset = CaretOffset;
-            if (caretOffset <= text.Length)
+            try
             {
-                try
+                var text = await document.GetTextAsync(token).ConfigureAwait(false);
+                var caretOffset = CaretOffset;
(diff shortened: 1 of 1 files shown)


sources/editor/Xenko.Assets.Presentation/AssetEditors/ScriptEditor/SimpleCodeTextEditor.cs
```diff
@@ -185,8 +185,12 @@ private async void CaretOnPositionChanged(object sender, EventArgs eventArgs)
             var caretOffset = CaretOffset;
             if (caretOffset <= text.Length)
             {
-                var result = await braceMatchingService.GetAllMatchingBracesAsync(document, caretOffset, token).ConfigureAwait(true);
-                braceMatcherHighlighter.SetHighlight(result.leftOfPosition, result.rightOfPosition);
+                try
```
Kryptos-FR: In your experience, can the token be already cancelled at that point?

If so, checking `if (token.IsCancellationRequested)` might save some computation (no await state-machine, no exception). Of course, the `try...catch` is still necessary.

Kryptos-FR: Looking at the whole code more closely, it could happen in a large document since there is an async call to `document.GetTextAsync()` a few lines above.

So while we are at it, let's give the `token` to the `document.GetTextAsync()` call as well and put a `try...catch` around all of it. In which case, testing the token is not necessary anymore.

Eideren:
Done. I'm fairly certain that the UI is single threaded, as such, concurrency before the first await will never occur, so checking for cancellation at that time is not necessary but should I still include that for cases where this method would be used elsewhere in the future ?

PR #143, opened 2018-08-25
[Editor] Minor UI tweaks.

Eideren:
Moved the render mode dropdown next to the editor camera settings, leaving it next to the translation mode doesn't make a lot of sense.

On the project selection window, the default tab selected points to recent projects instead of new when the user already has one.

The other changes can be seen in the two screenshots below, they are described right under.

old
new

In the asset view pane, reduced the font size, weight and opacity of the asset's type, the name of the asset should be the most visually important element, the icon already serves the same purpose as that label does and users will more often find their assets by visually scanning through asset names first.

Added a separator between the asset view option and filter to avoid grouping them visually since they aren't related.
I'm not sure about the size fix for it though, it sets itself by default to {16,16} since it's not a bitmap, I set it to fallback to auto instead, no idea if this was intended.

I tried to fix a bunch of other stuff but couldn't get the WPF preview to work so I'll leave it at that.

### DIFF
--- a/sources/editor/Xenko.Assets.Presentation/AssetEditors/EntityHierarchyEditor/Views/EntityHierarchyEditorView.xaml
+++ b/sources/editor/Xenko.Assets.Presentation/AssetEditors/EntityHierarchyEditor/Views/EntityHierarchyEditorView.xaml
@@ -673,6 +673,24 @@
                   </StackPanel>
                 </Grid>
               </Popup>
+              <ComboBox ItemsSource="{Binding Rendering.AvailableRenderModes}" SelectedItem="{Binding Rendering.RenderMode}" DisplayMemberPath="Name" Width="80">
+                <ComboBox.ItemContainerStyle>
+                  <Style TargetType="{x:Type ComboBoxItem}" BasedOn="{StaticResource {x:Type ComboBoxItem}}">
+                    <Style.Triggers>
+                      <DataTrigger Binding="{Binding Name}" Value="-">
+                        <Setter Property="IsEnabled" Value="False"/>
+                        <Setter Property="Template">
+                          <Setter.Value>
+                            <ControlTemplate TargetType="{x:Type ComboBoxItem}">
+                              <Separator Style="{StaticResource ComboBoxSeparator}"/>
+                            </ControlTemplate>
+                          </Setter.Value>
+                        </Setter>
+                      </DataTrigger>
+                    </Style.Triggers>
+                  </Style>
+                </ComboBox.ItemContainerStyle>
+              </ComboBox>
             </WrapPanel>
             <WrapPanel>
               <ListBox Style="{StaticResource EnumListBox}" SelectedItem="{Binding Transform.ActiveTransformation, Mode=TwoWay}"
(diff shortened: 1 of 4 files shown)


tebjan:
nice, i like the font weight switch for assets. if i might have a wish, isn't there a nicer flat icon for folders? this mountain/rose things looks a bit alien besides for example the game settings or scene icon...

Eideren:
I'll see if I can create a bunch of those tomorrow, 'not sure where/if I should store the vector files and resources in the repo though.

tebjan:
how about this one? 
folder_64

its pretty minimal: 
image

Kryptos-FR:
> this mountain/rose things looks a bit alien

Yeah, it does. In previous versions of Xenko, we had a better icon but we couldn't use it for the MIT version because of license issues.

xen2:
@tebjan @Kryptos-FR Yes, I remember not being able to quickly find a nice-looking large size flat folder icon with proper license. I agree this one is quite... ugly and out of place.
You're very welcome to replace it with another one, thanks!

tebjan:
is creative commons license ok? the one i made is derived from one that only needs author attribution. i can make a PR that changes the file and adds the original author info to the Thirdparty.md.

xen2:
@Kryptos-FR @tebjan @SebastianStehle @Eideren @0xaryan  Wanted some opinions, do you think monochrome is better or something with color (i.e. similar to VS solution explorer icons) so that it is easier to recognize?
I would personally vote for colored one (esp, so that it matches Solution Explorer treeview, and to be able to use color as faster visual cue). Switching to a .svg one should make it looks good even if big.

Kryptos-FR:
Yeah I'd also prefer a colored icon for the folder. Monochrome blends too much into the background, but folders should stand out in my opinion.

Eideren:
I would rather use vectors since we already have a bunch of other vector icons.
Here's a bunch I just did:
folderpreview

The left one is windows-like flat, middle one is generic and right one is based on the current icon.
I could also redo the one @tebjan gave us in vectors if you guys would rather have that one.

Eideren:
'just saw the last couple of messages, I'll use VS's folder color then once you choose the design you prefer.

0xaryan:
I'm not exactly sure that which icon is suitable in that case, but I prefer the colored one. As I remember that folder icon is not located in a `ResourceDictionary`, but in a `resx` file. I think some additional changes needed in code-behind if you want to use a vector icon.

Eideren:
We already have a couple of examples for vector images in ImageDictionary.xaml, we just have to convert vector files to wpf's format.

0xaryan:
@Eideren I know, but those are `DrawingImage` resources, code-behind only accepts `BitmapImage` as thumbnail. Not sure, but I think you should change that Thumbnail loader method to pass vectors to the view.

tebjan:
@Eideren the original file i used is a .svg you can find it here, its under creative commons: https://thenounproject.com/search/?q=folder

0xaryan:
@Eideren Take a look at project `Xenko.Core.Assets.Editor`. folder icon is located in `Resources/Images.resx`. 
In `ViewModel/DirectoryViewModel.cs` there is a method named `GetFolderThumbnail` which access to that resource.

```
private static ThumbnailData GetFolderThumbnail(IDispatcherService dispatcher)
{
    using (var ds = new DigestStream(new MemoryStream()))
    {
        var bytes = Resources.Images.folder;
        ds.Write(bytes, 0, bytes.Length);
        var objectId = ds.CurrentHash;
        var data = new ThumbnailData(objectId, new MemoryStream(bytes));
        data.PrepareForPresentation(dispatcher).Forget();
        return data;
    }
}
```
It passes a `MemoryStream` of image to `Xenko.Core.Assets.Editor.Services.ThumbnailData` class. Then that stream will be converted to a `BitmapImage` with `BuildBitmapImage` method. That's why I say you can't simply pass a `DrawingImage` to the view without changing the mechanism, although the View only wants an `ImageSource` which both `DrawinImage` and `BitmapImage` are accepted.

Toriadus:
So it looks really better, thanks for the work you've done!

tebjan:
how about something like that?

image

Toriadus:
I`m also  supporter of yellowish folders..

tebjan:
folder icon PR continued in #155

xen2:
Merged, thanks!

PR #151, opened 2018-08-26
[Editor] Prevent editing the prefab root or the scene root (fixes #150)

### LINKED ISSUE #150: Renaming "prefab root" in prefab editor causes a crash.
Open the prefab editor. Click on prefab root. Push f2. Rename it or don't, as soon as it's deselected, Xenko will crash.

Kryptos-FR:
The root node in the prefab or scene editors should not be editable.


xen2:
Thanks!

PR #155, opened 2018-08-26
[Editor] Flat folder icon

tebjan:
Changes the image for folders in the asset view. double resolution as the image before. color as discussed in #143:

image

### DIFF
--- a/THIRD PARTY.md
+++ b/THIRD PARTY.md
@@ -199,3 +199,7 @@ Contact
 - Facebook
 - Twitter
 - etc...
+
+### Folders 2 Bold
+The folder icon in the asset view is licensed under creative commons and derived from:
+- Alfredo Hernandez: https://www.alfredocreates.com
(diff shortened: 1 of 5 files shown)


Toriadus:
Looks like pretty good, but in my opinion there isn't enough chamfers on the edges of tiles. Also, I previously talked about a too dark (almost black) color scheme (my opinion)...

makotech222:
Agreed on the too black color scheme. Needs just a bit more lightness in my opinion. Otherwise, new folder icon looks good.

tebjan:
yes, i also think that the current dark theme is a bit too dark. it would be quite nice to have it in VS2017 dark colors... ;) the blue selection already looks quite similar. and as you often switch between visual studio and game studio it might be quite natural to have the same color scheme. maybe @0xaryan can help us out here and brighten it up a bit.

image

Toriadus:
Thank you for understanding @tebjan, I hope @0xaryan will understand and support our thought (after all,  we're talking about  a little bit lighter, nothing radical)

xen2:
Merged, thanks!

PR #171, opened 2018-08-31
[Graphics] Add buffer initial counter offset

### LINKED ISSUE #122: Add initial counter offset value of compute shader

tebjan:
update of my previous pull request #123 which fixes #122 

@xen2 i fixed the signature for OpenGL and added a comment for DX12. most platforms don't seem to have proper uav support. solutions build properly here. tell me if something is missing.

### DIFF
--- a/sources/engine/Xenko.Graphics/Buffer.cs
+++ b/sources/engine/Xenko.Graphics/Buffer.cs
@@ -127,6 +127,14 @@ namespace Xenko.Graphics
         public PixelFormat ViewFormat { get; private set; }
 
         /// <summary>
+        /// The initial Append/Consume buffer counter offset. A value of -1 indicates the current offset
+        /// should be kept. Any other values set the hidden counter for that Appendable/Consumable
+        /// Buffer. This value is only relevant for Buffers which have the 'Append' or 'Counter'
+        /// flag, otherwise it is ignored. The value get's initialized to -1.
+        /// </summary>
+        public int InitialCounterOffset { get; set; } = -1;
+
+        /// <summary>
         /// Return an equivalent staging texture CPU read-writable from this instance.
         /// </summary>
         /// <returns>A new instance of this buffer as a staging resource</returns>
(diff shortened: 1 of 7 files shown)


xen2:
Thanks!
I think there's still some CRLF issues (some lines were changed from CRLF to LF only).

Kryptos-FR:
@xen2 Are you sure it is not your local git config that is incorrect? The default behavior on Windows has always been to "checkout windows-style, commit Unix-style", i.e. `core.autocrlf` set to `true` in the global `.git/config`. 
So I'm not sure how you could have a conflict on that. File on GitHub would always have LF line ending. Maybe I'm missing something...

tebjan:
hm, that is weird, my git is set to "checkout as-is, commit as-is" so my git will never change any files. also my visual studio is set to CRLF. i hope the files are CRLF in the repo... or not?

Kryptos-FR:
@tebjan No, as is usually the convention, files in the repo are LF. So "checkout as-is, commit as-is" is dangerous because it might keep mixed CRLF and LF line endings in the same file.

xen2:
@Kryptos-FR I was also looking for a github status checker for CRLF but couldn't find anything. Could be handy because hard to notice!

xen2:
@tebjan @Kryptos-FR OK what I have on my computer:
* `Buffer.Direct3D12.cs` (and any other file) on master: all CRLF (probably all LF on repo and autocrlf fixes it for me to CRLF)
* `Buffer.Direct3D12.cs` on your branch: all LF except the part you changed which are CRLF. I assume that's exactly how it is on the repo. It just come to me unmodified by autocrlf since it has a mix of CRLF and LF which is unexpected and git might think it's on purpose and important to keep.

tebjan:
ok, makes perfect sense now. i'll change the settings for my local xenko get repo then. thanks for the input!

xen2:
@tebjan are you going to fix the commit or should I do it on my side?

tebjan:
yes, i'll fix that when i am back at the office. sorry for the delay, had a few days on vacation.

xen2:
No worries, I don't mind doing it.

tebjan:
@xen2 i did a reset on the branch and re-commited the files with git set to "checkout windows-style, commit Unix-style". hope the line endings are correct now.

xen2:
LGTM, launching some tests and will merge.

xen2:
Merged, thanks!

PR #179, opened 2018-09-02
[Physics] Make CharacterComponent.Gravity a Vector3 instead of a float

### LINKED ISSUE #170: Avoid assuming negative-Z gravity
**Release Type**: Both

**Version**: any future

**Describe the feature**
Not so much a bug report as a feature request, but the engine should avoid explicitly assuming a negative Z vector for gravity. Currently, the engine assumes negative-Z gravity (see Additional context for example).

**To Reproduce**
Not applicable.

**Expected behavior**
User-configurable gravity vector, possibly also localisable to volume triggers. Useful for, among other things, proper vehicle behavior through structures such as loops.

**Additional context**
CharacterComponent.cs (part of Xenko.Physics) assumes negative-Z as gravity, but uses a full Vector3 internally:

        /// <summary>
        /// Gets or sets if this character gravity
        /// </summary>
        /// <value>
        /// true, false
        /// </value>
        /// <userdoc>
        /// The gravity force applied to this character
        /// </userdoc>
        [DataMember(95)]
        public float Gravity
        {
            get
            {
                return gravity;
            }
            set
            {
                gravity = value;

                if (KinematicCharacter != null)
                {
                    KinematicCharacter.Gravity = new Vector3(0.0f, value, 0.0f);
                }
            }
        }

Perhaps make all three components of the Vector3 configurable?

DW01:
(Closes #170 )

This allows the user to configure all three dimensions in `CharacterComponent.Gravity` rather than assuming gravity is always on the Z plane.

Also some miscellaneous comment updates to clear up some underexplained properties in the documentation.

(ignore all these merges, I'm bad at Git haha)


sources/engine/Xenko.Assets/Entities/EntityHierarchyAssetBase.Upgraders.cs
```diff
@@ -56,5 +56,49 @@ protected override void UpgradeAsset(AssetMigrationContext context, PackageVersi
...
+
+                // Loop through the YAML file.
+                foreach (dynamic entityDesign in entities)
+                {
+                    // Get the entity.
+                    var entity = entityDesign.Entity;
+
+                    // Save override and remove old Gravity element.
+                    var gravityOverride = entity.GetOverride("Gravity");
+                    entity.RemoveChild("Gravity");
```
Kryptos-FR: That won't work. The current `Gravity` property is not on the `Entity` itself.

sources/engine/Xenko.Assets/Entities/EntityHierarchyAssetBase.Upgraders.cs
```diff
@@ -56,5 +56,49 @@ protected override void UpgradeAsset(AssetMigrationContext context, PackageVersi
...
+                    foreach (var component in entity.Components)
+                    {
+                        try
+                        {
+                            var componentTag = component.Value.Node.Tag;
+
+                            // Is this a character component?
+                            if (componentTag == "!CharacterComponent")
+                            {
+                                component.Value.SetOverride("Gravity", gravityOverride);
```
Kryptos-FR: Here is where the upgrade will happen:
* get the current float value for Gravity
* create a "Vector3" value (YAML version) and assign the previous value to its Y component.

DW01: I assume I can get the "old value" from YAML by assigning `component.Value.PropertyName` (`PropertyName` is `Gravity` in this case) to a new `var`? This seems to be how it's done within `MoveRenderGroupToEntityUpgrader`, so I hope I'm correct in assuming this.

As for writing to YAML, I've not really worked with YAML serialisation before (although JSON is also ugly and I've worked with it), so would I initialise the `Vector3` representation as a string containing `oldGravity` as the Y component?

xen2: Code should look like:
```
if (component.Value.ContainsChild("Gravity"))
{
    component.Value.Gravity = ...
}
```
For the actual float=>Vector3 transform code, take a look at `ProceduralModelAsset.Upgrader`, it contains an upgrader that does exactly that.

xen2: @DW01 https://github.com/xenko3d/xenko/blob/releases/3.0.0.4/sources/engine/Xenko.Assets.Models/ProceduralModelAsset.cs#L70

DW01:
Given the test case @Kryptos-FR gave me in Gitter passed with flying colors, is this good to merge?

Kryptos-FR:
@xen2 when we change assets (breaking change) what is our policy regarding the version number? Should we increase the minor or major one? In that case we might want to wait for more changes or create a "future" branch to differentiate bug fixes and new features.

xen2:
@Kryptos-FR good question.
Contrary to before, we should now **technically** be fine to have asset changes for minor upgrades since we upgrade project dependencies to Xenko even for minor releases.
Still, I am not decided on whether to allow such changes in minor release (bump to 3.0.1.x? or bump to 3.1.x same as before?)

xen2:
@DW01 yes, PR looks good to me. Just need to figure out the last point @Kryptos-FR brought up, but that's on my side.

DW01:
@xen2 @Kryptos-FR No problem, it's best that gets figured out for future changes much larger than this anyway.

aaronfranke:
> (ignore all these merges, I'm bad at Git haha)

You should squash all of these commits together into one. Clean history is important for Git workflow.

Via GitKraken: Ctrl-Click commits, then right-click, and click "Squash",

Via Git in the terminal: `git rebase -i HEAD~10` and change "pick" to "fixup" or just "f". Replace `10` with the number of commits you're trying to squash.

xen2:
@aaronfranke not necessary, github has an option to squash before merging PR

xen2:
Did a few minor tweaks in your branch.
Merged, thanks!

Issue #159, opened 2018-08-27
Input.IsKeyPressed does not behave as intended.


justStand:
It seems to rely on the Windows keyboard "Repeat rate" setting.

If you hold a key down; whenever Windows will output the key, the input.iskeypressed will return true.

Kryptos-FR:
From what I can read in the code (https://github.com/xenko3d/xenko/blob/master/sources/engine/Xenko.Input/Windows/KeyboardWinforms.cs) it seems to rely on low-level Win32 API.

That said it is possible that the "Repeat rate" setting simulates up and down events which will explain what you described (I don't know much about that feature). Were that the case, I don't think there is much we can do (since it means it is affecting low-level Win32 API).

We need to investigate more...

Eideren:
@Kryptos-FR You were right; changing windows' keyboard repeat rate changes the rate at which it is repeated in-engine. 

Based on [this page](https://docs.microsoft.com/en-us/windows/desktop/inputdev/wm-syskeydown) we can see that the second to last bit on the LParam provides us a filter which would help us to ignore repeating messages. I added logging code to [InputSourceWinforms](https://github.com/xenko3d/xenko/blob/master/sources/engine/Xenko.Input/Windows/InputSourceWinforms.cs)'s HandleKeyDown / HandleKeyUp and here's the content of the lParam IntPtr casted to an int32 and converted to bit representation in the console : 
![inputissue](https://user-images.githubusercontent.com/5742236/44920307-de5d4700-ad3f-11e8-83df-4d204160667f.jpg)
As you can see on line 'A', the first down event for the key starts with a zeroed second bit, which, on the following event while holding the key, turns to one.

On line 'B' and 'C', you can see that the previous key was released and another one pressed but no up event occurred. We are actively sampling the API, It isn't interrupting with messages right ? If that's the case then this is just an editor side sample issue.
You can also see a variant of that issue crop up on line 'D' were the first down event wasn't caught but the following repeating events were.


I could push a fix for this since I already had a couple of other minor fix in other parts of the input system but just to be sure, can I safely filter out those repeated inputs based on what I found or am I forgetting something ? And is the issue shown on line B, C and D expected behavior ?

xen2:
I agree with your analysis, it shouldn't trigger `IsKeyPressed` multiple time if it's a single long keypress.
Would gladly accept a PR fix, thanks!

justStand:
[...]

Issue #94, opened 2018-08-11
Editor camera stutters/does not process inputs


Eideren:
Keys held are sometimes ignored or processed half a second later when controlling the editor camera, it can show up on every other input or won't for multiple minutes.
Sampling inputs through 'Input.IsKeyDown' in a game build works as expected.

Mouse inputs seems to suffer as well, seemingly missing frames or inputs when rotating the camera.
I fiddled with the editor camera's input-handling code (removing cases where it would exit out before hitting 'UpdateCameraAsPerspective'), I forced the camera to rotate automatically by increasing it's yaw inside the function mentioned before and capturing a video of it next to my moving cursor, there were indeed frames (sometimes up to 6 in a row) captured where the camera wouldn't rotate but the windows cursor would move. The video captured was at 60fps, I couldn't find the editor view's framerate since RivaTuner flickers for some reason( perhaps related to the view implementation inside wpf? ) but if it is vsynced it would probably be at 144hz to follow my screen's refresh rate ?

While capturing the video through OBS, my gpu and cpu where under minimal load, rivatuner wasn't running and my web browser was idle.
When running a single camera build of the animation sample RivaTuner reported ~900 fps so it's unlikely that performances are a concern.

Is something wrong with how the camera 'MicroThread' created through 'Game.Script.AddTask' is scheduled in editor ?

Is this issue on my end only ?

phr00t:
I notice this problem too. When I move the camera forward with the WASD keys, it seems to continue moving even after releasing the keys for random periods < 1sec. Just overall laggy camera operation in the editor.

xen2:
Fixed by #359

Eideren:
Re-investigating this as #713 was only a workaround and relied upon rawinput which isn't part of the editor anymore.
Setting ``WS_CHILD`` flag to false instead of true here ``https://github.com/stride3d/stride/blob/master/sources/presentation/Stride.Core.Presentation/Controls/GameEngineHost.cs#L131`` resolves the issue but causes others. Not exactly sure why it kind of works yet though.

xen2:
Temporarily reverted #1274 so that we can release 4.1

jusvit:
Any update on this? It leaves a sour taste on my tongue when I use the Editor.

Eideren:
I'll look into it again later today

Eideren:
Opened a workaround for this issue, see https://github.com/stride3d/stride/pull/1274 @jusvit

PR #181, opened 2018-09-02
[Input] Windows input fixes

Eideren:
Fixing issue #159 
While this first push addresses the main issue mentioned there it makes the scene view hard to use because of the missing events mentioned lower in that same issue.
I'm investigating it, if you have any clue feel free to share !

n.b.:Those `Console.WriteLine` will be gone once I'm done of course.

### DIFF
--- a/sources/engine/Xenko.Input/InputManager.State.cs
+++ b/sources/engine/Xenko.Input/InputManager.State.cs
@@ -110,7 +110,7 @@ namespace Xenko.Input
         {
             get
             {
-                if (!HasMouse) return NoKeys;
+                if (!HasKeyboard) return NoKeys;
                 return Keyboard.PressedKeys;
             }
         }
@@ -122,7 +122,7 @@ namespace Xenko.Input
         {
             get
             {
-                if (!HasMouse) return NoKeys;
+                if (!HasKeyboard) return NoKeys;
                 return Keyboard.ReleasedKeys;
             }
         }
@@ -134,7 +134,7 @@ namespace Xenko.Input
         {
             get
             {
-                if (!HasMouse) return NoKeys;
+                if (!HasKeyboard) return NoKeys;
                 return Keyboard.DownKeys;
             }
         }
--- a/sources/engine/Xenko.Input/Windows/InputSourceWinforms.cs
+++ b/sources/engine/Xenko.Input/Windows/InputSourceWinforms.cs
@@ -104,6 +104,10 @@ namespace Xenko.Input
 
         internal void HandleKeyDown(IntPtr wParam, IntPtr lParam)
         {
+            if (MessageIsDownAutoRepeat(lParam))
+                return;
+            Console.WriteLine("DOWN");
+
             var virtualKey = (WinFormsKeys)wParam.ToInt64();
             virtualKey = GetCorrectExtendedKey(virtualKey, lParam.ToInt64());
             keyboard?.HandleKeyDown(virtualKey);
@@ -112,6 +116,8 @@ namespace Xenko.Input
 
         internal void HandleKeyUp(IntPtr wParam, IntPtr lParam)
         {
+            Console.WriteLine("UP");
+
             var virtualKey = (WinFormsKeys)wParam.ToInt64();
             virtualKey = GetCorrectExtendedKey(virtualKey, lParam.ToInt64());
             heldKeys.Remove(virtualKey);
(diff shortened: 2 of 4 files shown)


sources/engine/Xenko.Input/InputManager.State.cs
```diff
@@ -110,7 +110,7 @@ public IReadOnlySet<Keys> PressedKeys
         {
             get
             {
-                if (!HasMouse) return NoKeys;
+                if (!HasKeyboard) return NoKeys;
```
Kryptos-FR: :open_mouth: good catch!

sources/engine/Xenko.Input/Windows/InputSourceWinforms.cs
```diff
@@ -165,6 +171,20 @@ private static WinFormsKeys GetCorrectExtendedKey(WinFormsKeys virtualKey, long
...
+        /// <br/>This function is used to find out if the received message is a repeated KEYDOWN.
+        /// </summary>
+        /// <param name="lParam">lParam of the KEYDOWN message</param>
+        /// <returns><c>True</c> if this message is a repeated KeyDown, <c>false</c> if it's an actual keydown</returns>
+        private static bool MessageIsDownAutoRepeat(IntPtr lParam)
+        {
+            // According to the microsoft docs on WM_KEYDOWN
+            // (https://docs.microsoft.com/en-us/windows/desktop/inputdev/wm-keydown)
+            // The second to last bit is 0 when the last keyboard message was up and 1 if it was already down
+            return ((int)lParam & (1 << 30)) != 0;
```
Kryptos-FR: @xen2 should that be enclosed by `unchecked`? `IntPtr` can be 32-bis or 64-bits depending on the platform if I'm not mistaken.

Eideren: You're right, I'll just use a long since it's already used in the other function.

Eideren:
So, the input system is fine, native WndProc just fails to send us all of the user's inputs. 
It's most likely related to the UI element bound to that function since that's really the only other variable in this equation and it also fails to get some inputs.

We could use direct or raw input and call it a day but I'm pretty sure that this issue is larger than just input related and could solve a bunch of those on #94 .

What's going on with that UI element ?

Kryptos-FR:
That's out of my expertise :sweat_smile: let's ask @xen2

xen2:
@Eideren Interesting find! Unfortunately I have no idea why it fixes the issue. Pinging @xoofx :)

xoofx:
Can't tell why (don't remember the codebase of the winform in Xenko), but before hacking with RawInput here, I would verify the messages at SPY++ level on the form to make sure that it is not WinForms early filtering something badly. In that case, it might be possible to find a solution within it.

Eideren:
Spy++ does report that we receive all of our keyup/down messages.

Now just to triple verify our filtering, I took a look at how the forms are used and found out that we don't actually use the common windows forms way of initializing interfaces, we pull and push messages to the UI through [this function](https://github.com/xenko3d/xenko/blob/1cc4d2761db038ebe89f04c1787b572b9ebc6dc0/sources/engine/Xenko.Games/Desktop/WindowsMessageLoop.cs#L126) instead ; so I removed all of the overridden key functions on the form and turned `UseApplicationDoEvents` to true to use the standard `Application.DoEvents()` in place of our custom version and ... nothing changed.

'Implemented a [PreProcessMessage](https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.control.preprocessmessage?view=netframework-4.7.2) method on the form and nope, same results there.

I then tried to launch our scene view through Application.Run or form.Show, I'm really grasping at straws here, but both of them couldn't act properly.

There doesn't seem to be any kind of concrete source/explanation as to what RegisterRawInputDevices does either, besides the fact that it [enables raw input reception for applications](https://docs.microsoft.com/en-us/windows/desktop/api/winuser/nf-winuser-registerrawinputdevices).

I can't think of anything else, if you have any idea at this point, I'm all ears.

Eideren:
Alright, since this is getting nowhere and we still have to at least provide a work around for this issue I pushed it anyway.
Would you like me to create a new issue to replace the work around or does the 'TODO' inside the summary for the function suffice ?

xen2:
Yes I am good with using `RegisterRawInputDevices` for now considering how much improved the situation would be.
TODO with a quick explanation and link to this conversation should be enough for now.
Thanks!

xen2:
Thanks!

Issue #185, opened 2018-09-05
NullReferenceException in DebugTextSystem


jonstelly:
**Release Type**: GitHub

**Version**: master

**Platform(s)**: Windows/(All I think?)

**Describe the bug**
`DebugText.Print("something", new Int2(20, 20))` from a script causes a Null Reference Exception in DebugTextSystem line 88.  It looks like `DebugTextSystem.Content` is initialized to null since the DebugTextSystem is created and registered in the `Game` constructor, while the `ContentManager` is created and registered later in `GameBase.PrepareContext()`.

**To Reproduce**
1. Create a basic Game
2. In the BasicCamerController.Update method, add this line: `DebugText.Print("Hello, Crash", new Int2(20, 20));`
3. Get the below exception/crash:

**Expected behavior**
Debug Text

**Screenshots**
If applicable, add screenshots to help explain your problem (use GitHub drag & drop)

**Log and callstacks**
```
[Game]: Error: Unexpected exception. System.NullReferenceException: Object reference not set to an instance of an object.
   at Xenko.Profiling.DebugTextSystem.Draw(GameTime gameTime)
   at Xenko.Games.GameSystemCollection.Draw(GameTime gameTime)
   at Xenko.Games.GameBase.Draw(GameTime gameTime)
   at Xenko.Games.GameBase.DrawFrame()
   at Xenko.Games.GameBase.TickInternal()
System.NullReferenceException: Object reference not set to an instance of an object.
   at Xenko.Profiling.DebugTextSystem.Draw(GameTime gameTime)
   at Xenko.Games.GameSystemCollection.Draw(GameTime gameTime)
   at Xenko.Games.GameBase.Draw(GameTime gameTime)
   at Xenko.Games.GameBase.DrawFrame()
   at Xenko.Games.GameBase.TickInternal()

Unhandled Exception: System.NullReferenceException: Object reference not set to an instance of an object.
   at Xenko.Profiling.DebugTextSystem.Draw(GameTime gameTime)
   at Xenko.Games.GameSystemCollection.Draw(GameTime gameTime)
   at Xenko.Games.GameBase.Draw(GameTime gameTime)
   at Xenko.Games.GameBase.DrawFrame()
   at Xenko.Games.GameBase.TickInternal()
   at Xenko.Games.GameBase.Tick()
   at Xenko.Games.GamePlatform.Tick()
   at Xenko.Games.GamePlatform.OnRunCallback()
[...]

PR #186, opened 2018-09-05
Workaround for DebugTextSystem NRE

jonstelly:
- DebugTextSystem initialized before ContentManager, so Content property is always null, throwing NRE

Workaround for: #185


xen2:
I took a look and decided to go straight for the proper fix rather than a workaround for only this specific case. This is now in master with commit b887069aacce
Thanks for your help!

Issue #183, opened 2018-09-03
CPU/GPU is working abnormally when the window is minimized


EssenRoc:
**Release Type**: Both

**Version**: 3.0.0.4

**Platform(s)**: Windows

**Describe the bug**
CPU/GPU is working abnormally when the window is minimized.

**To Reproduce**
1. Create a project
2. Run project
3. Minimize window

**Expected behavior**
I think CPU/GPU should not be so exaggerated.

**Screenshots**

![gif](https://user-images.githubusercontent.com/12685126/44970306-90388580-af83-11e8-9f4d-dc6e397f1df9.gif)

domske:
Confirmed. I tested the same with similar results.

Eideren:
Xenko doesn't follow `GraphicsDevice.Presenter.PresentInterval` value while minimized : the game loop is ran at a rate higher than what the display syncs to even though it still runs GraphicsDevice.Presenter.Present(). 

Perhaps related to [our use of SharpDX's SwapChain.TryPresent](https://github.com/xenko3d/xenko/blob/3e89094a076aa5d64344af5aa64a13847f32d910/sources/engine/Xenko.Graphics/Direct3D/SwapChainGraphicsPresenter.Direct3D.cs#L154) ? Not sure how presenting a frame acts in general for minimized windows.

Weirdly enough, for me, it's also hitting the GPU far harder than `Immediate` does. 
On `Immediate` the process' game-loop runs at ~2ms/loop, while minimized it runs at ~0.05ms/loop, so the GPU is probably not actually rendering, we're just prodding it and SharpDX's TryPresent might be returning that the graphics device is disconnected ?

Here's the presenter callstack:
```
Xenko.Games.GameBase.Run(GameContext)
Xenko.Games.GameWindowWinforms.Run()
Xenko.Games.WindowsMessageLoop.Run(Control, RenderCallback, Boolean)
Xenko.Games.GamePlatform.OnRunCallback()
Xenko.Games.GamePlatform.Tick()
Xenko.Games.GameBase.Tick()
Xenko.Games.GameBase.TickInternal()
Xenko.Engine.Game.EndDraw(Boolean)
Xenko.Games.GameBase.EndDraw(Boolean)
Xenko.Games.GraphicsDeviceManager.EndDraw(Boolean)   (from Xenko.Games.IGraphicsDeviceManager)
Xenko.Graphics.SwapChainGraphicsPresenter.Present()
[...]

PR #188, opened 2018-09-08
Fix CPU/GPU is working abnormally...

EssenRoc:
Try to fix [#183](https://github.com/xenko3d/xenko/issues/183)

### DIFF
--- a/sources/engine/Xenko.Graphics/Direct3D/SwapChainGraphicsPresenter.Direct3D.cs
+++ b/sources/engine/Xenko.Graphics/Direct3D/SwapChainGraphicsPresenter.Direct3D.cs
@@ -24,6 +24,7 @@
 #if XENKO_GRAPHICS_API_DIRECT3D
 using System;
 using System.Reflection;
+using System.Threading;
 using SharpDX;
 using SharpDX.DXGI;
 using SharpDX.Mathematics.Interop;
@@ -46,6 +47,8 @@ namespace Xenko.Graphics
         private readonly Texture backBuffer;
 
         private int bufferCount;
+
+        private bool isOccluded;
 
 #if XENKO_GRAPHICS_API_DIRECT3D12
         private int bufferSwapIndex;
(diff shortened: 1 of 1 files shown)


sources/engine/Xenko.Graphics/Direct3D/SwapChainGraphicsPresenter.Direct3D.cs
```diff
@@ -151,7 +154,26 @@ public override void Present()
...
+                if (isOccluded)
+                {
+                    var result = swapChain.Present((int)PresentInterval, PresentFlags.Test);
+                    if (result == Result.Ok)
+                    {
+                        isOccluded = false;
+                    }
+                    else
+                    {
+                        Thread.Sleep(1);
```
Kryptos-FR: What is the reason for sleeping here?

I'm worried `Thread.Sleep(1)` might cause some issues. There is no guarantee a thread will sleep for only 1 ms. In fact, it is more likely to sleep for much more than that (OS interrupt, context switching, etc.). If a sleep can't be avoided, better use `Thread.Yield()` or `Thread.Sleep(0)`.

EssenRoc: Use Thread.Yield() or Thread.Sleep(0). The CPU is still very exaggerated. Use Thread.Sleep(1) shouldn't have much impact, even if it sleeps 15 milliseconds. look this [https://docs.microsoft.com/en-us/windows/desktop/direct3ddxgi/dxgi-status (DXGI_STATUS_OCCLUDED)](https://docs.microsoft.com/en-us/windows/desktop/direct3ddxgi/dxgi-status)

EssenRoc: If Utilities.Sleep is more accurate, it can also be used.

Eideren: With this change, the game loop while minimized runs at a rate higher than refresh rate which doesn't seem expected ?
This is probably because rendering is running asynchronously to logic ?

Eideren: No ? As Kryptos said: sleep is unreliable, using it as is to try to produce constant frame times is unwise.
If you aren't too sure how to properly implement it, verify that you understand how the game and render loop are setup and find a couple of articles on the subjet, try searching for "programming constant framerate" or similar on google.

EssenRoc: When the window is minimized, the game screen cannot be rendered and the frame rate cannot be maintained. At this time only PressFlags.Test (not rendering) is waiting for recovery.
My understanding is that the game screen is now sleeping and waiting for recovery.

Eideren: I'm not sure I understood what you're saying, we're talking about the game loop here, we can't stop it while the window is minimized, right ?
We could implement a toggle to reduce the game's update rate or stop it entirely, we can safely ignore rendering but I'm pretty sure that by default *even while minimized* the expected behavior is that the game loop keeps running in the background.

Issue #189, opened 2018-09-09
Layout are not resizing when hides all panels on the right side


xwellingtonx:
**Release Type**: Both

**Version**:  All

**Platform(s)**: Windows

**Describe the bug**
When you hide all the panels on the right side the layout are not resizing properly.

**To Reproduce**
1. Open a project in Game Studio
2. Hide all panels on the right side

**Expected behavior**
The main scene need to resize.

**Screenshots**
![bug](https://user-images.githubusercontent.com/10274446/45265494-fe68c680-b421-11e8-8a94-d84cc4bce230.gif)

**Log and callstacks**
Not applicable

**Additional context**
Not applicable

**Please assign to me.

xwellingtonx:
Will be necessary to change default editor layout of **DockingManager** so do i need know if has any away to update the config file for xenko old installations?

xwellingtonx:
It's resolved with PR #192.

Issue #133, opened 2018-08-23
Build fails with private nuget sources


xwellingtonx:
I have found one bug with private nuget sources. Xenko makes use of Nuget.exe that does not have permission to access private repositories such as the VSTS Package Management.. 

If you have configured any private source in your Visual Studio the build will fail without throwing exception.

![sem titulo](https://user-images.githubusercontent.com/10274446/44542686-ef74db00-a6e3-11e8-89fa-10e9d67b09bd.jpg)

The unique form that's I found it to fix it is to disable the private sources in your Visual Studio but it's necessary to update the code of InstallPackage method in NugetStore for throw a message about this error.

I'm studying the better form to update it because i don't know much yet about Xenko.

xen2:
This is a duplicate of #1 (fixed, will be released with next launcher update).

xen2:
Note: need to double check if the fix done in launcher also fixes it for Xenko itself.

xen2:
Note: need to double check if it's the same issue as launcher.

xwellingtonx:
I think it's the same, but really needs to check. When you pull the update I will test again in my machine.

Issue #190, opened 2018-09-09
Unpinned panels are not shown with a proper height/width


xwellingtonx:
**Release Type**: Both

**Version**: All

**Platform(s)**: Windows

**Describe the bug**
When panels is auto hide and you mouse over it the height/width are not the same when visible.

**To Reproduce**
1. Open a project in Game Studio
2. Auto hide any painel
3. Mouse over it

**Expected behavior**
The height/width must be the same size as when the panel is visible.

**Screenshots**
![bug2](https://user-images.githubusercontent.com/10274446/45265800-56ee9280-b427-11e8-9752-6fe167cb9cd1.gif)

**Log and callstacks**
Not applicable

**Additional context**
Not applicable

**Please assign to me.

xwellingtonx:
It's resolved with the PR #192.

PR #192, opened 2018-09-09
[Editor] Fix editor resizing and to insert auto hide minimum size

xwellingtonx:
This PR is to resolve the issues #189 and #190.

Before Updated:
before

After Updated:
now

**See my question about Xenko old installations in issue #189. 
I will update if necessary, because for now it's necessary you reset manually the layout of Game Studio for the update works.**


sources/editor/Xenko.GameStudio/GameStudioWindow.xaml
```diff
@@ -433,51 +433,51 @@
...
             <xcad:LayoutPanel Orientation="Vertical">
               <xcad:LayoutPanel Orientation="Horizontal">
-                <xcad:LayoutAnchorablePane DockWidth="240">
+                <xcad:LayoutAnchorablePane DockWidth="300">
                   <!-- SOLUTION EXPLORER -->
-                  <xcad:LayoutAnchorable ContentId="SolutionExplorer" Title="{Binding Source={xk:Localize Solution explorer, Context=View}}"
-                                         gs:AvalonDockHelper.IsVisible="{Binding Panels.SessionExplorerPanelVisible, Mode=TwoWay,
-                                         Source={x:Static gs:GameStudioViewModel.GameStudio}, FallbackValue={xk:True}}">
+                  <xcad:LayoutAnchorable AutoHideMinWidth="300" ContentId="SolutionExplorer" Title="{Binding Source={xk:Localize Solution explorer, Context=View}}"
+                                           gs:AvalonDockHelper.IsVisible="{Binding Panels.SessionExplorerPanelVisible, Mode=TwoWay,
```
Kryptos-FR: Nit: unnecessary whitespace change (+10 more occurrences below).

xwellingtonx: I will fix it, i think was my VS :(

xen2:
@Kryptos-FR what would happen to already saved layouts?

sources/editor/Xenko.GameStudio/GameStudioWindow.xaml
xwellingtonx: Done!

Kryptos-FR:
@xen2 Good question. The basic layout itself has not changed (only one `LayoutPanel` has been replaced by a `LayoutAnchorablePaneGroup`. I'll have to do a few tests but my guess is the old layout will be loaded and behave like before. I don't think that it will fail to load, if that was the question.

So as @xwellingtonx mentioned, user will have to reset the layout. Let's do some tests.

xwellingtonx:
@xen2  It won't fail as @Kryptos-FR said it. The GameStudioInternal.config file in a machine that's has built master branch or installed the release version of Xenko will be loading the old XAML of this file without any problem.

Follow steps to test it:
1. Build master branch
2. Open a project
3. Confirm the bug
4. Close it and build my update branch
5. Confirm that's bug is yet happening
6. Reset interface manually on edit options
7. See the update work

xen2:
@xwellingtonx OK, that's what I was curious to know: does keeping existing config remove future functionality.
Since it seems to be the case, shouldn't we just force a layout reset to make sure everybody has a fully working layout system?

xwellingtonx:
@xen2 I believe that, however, this will cause the personal preferences of each user to be lost.

xen2:
@xwellingtonx 
We have three options:
1. Reset layout to new one (but user loose previous settings)
2. Upgrade (little bit more work)
3. Keep layout (could lead to complex situation with people dragging on old config/bugs that will be difficult to repro)

Do you think an upgrader (i.e. using Regex) would be simple to write for this case? (if it takes more than 20 min I wouldn't bother).
IMHO I would rather avoid surprises and stick to 1 rather than 3, but I'll leave that one up to you, I am fine to go with any of those 3 options.

xwellingtonx:
I think the best option really is the first one, because if we upgrade with Regex it can cause problems with custom layout by users. The XAML is very different from the default one.

My choice is the first option but how will it be done? Because we only need to force the restart of the layout in the next version, after that will be not more necessary.

xen2:
@xwellingtonx that's where we also need to store a version number and discard loaded value if loaded version number is lower than expected version number.
Some idea where/how to store the version number, not sure which one will work best/easier:
- Separate entry (i.e. `MRUAdditionalData.DockingLayoutVersion`)
- Turn `MRUAdditionalData.DockingLayout` into a Version+Value struct
- Inside the xml (maybe a root attribute, like `<?xml version="1.0" layout-version="3">`
- Bump entry name (i.e. `MRUAdditionalData.DockingLayout2`)

The version checking/reset logic should go in `DockingLayoutManager.LoadLayoutFromSettings`.

xwellingtonx:
@xen2 I need a time to do it because i do not know much how MRUAdditionalData works or do you intend to do it?

xen2:
@xwellingtonx Sure, I can take care of it

xen2:
@xwellingtonx Added a commit there: https://github.com/xenko3d/xenko/commits/fix_editor_resizing (not sure why but I couldn't push directly in your branch. Please cherry pick it and check if it is all working as intended.

xwellingtonx:
@xen2 Done! ;)

xwellingtonx:
@xen2 Thanks, I tested here and it's all work very well.
I have only seen one point for the future that is about the old versions still being saved in the file, if we have many updates in the layout in the future the file may end up staying a lot of dirt.

sources/editor/Xenko.GameStudio/DockingLayoutManager.cs
```diff
@@ -149,6 +149,10 @@ private void LoadLayoutFromSettings(MRUAdditionalData data, bool resetLayoutIfFa
             BindableSelectedItemsControl.DisableBindings = true;
             try
             {
+                // This exception is normal and will trigger a reset of the layout, since no layout can be loaded from the settings file.
+                if (GameStudioInternalSettings.CurrentLayoutVersion > data.DockingLayoutVersion)
```
xen2: Can you adjust this check to `!=` rather than `>` please?

sources/editor/Xenko.GameStudio/DockingLayoutManager.cs
```diff
@@ -161,7 +165,11 @@ private void LoadLayoutFromSettings(MRUAdditionalData data, bool resetLayoutIfFa
                     return;
 
                 // Erase saved layout if we're unable to load it.
-                if (isInEditorLayout)
+                if (GameStudioInternalSettings.CurrentLayoutVersion > data.DockingLayoutVersion)
```
xen2: Can you adjust this check to `!=` rather than `>` please?

xen2:
I was thinking about the fact we update layouts one by one (old versions staying in file), but since this file is shared and some projects might still open with previous Xenko versions, we can't upgrade until each specific project is open with a given version of Xenko.

Also I added two last comments and I will merge it.

sources/editor/Xenko.GameStudio/DockingLayoutManager.cs
xwellingtonx: Yes, sure!

xwellingtonx:
I has add you as a collaborator, could you try to push again?

xen2:
@xwellingtonx 
I thought it was automatic with this:
image

Anyway, it will need further work, older versions of Xenko don't like to read the file back with this new variable.

xwellingtonx:
@xen2, I thought that deserializing the file would discard any property that is not used.

This option is enabled in this PR but it does not seem to work :/
image

xen2:
@xwellingtonx This mode where we discard unknown property is optional and not used in this case.
Thinking about it, it wouldn't be good if discarded because it would be saved back without it for all projects. This is not what we want.
I will need to figure out another way to support upgrades. A simple way would be to version using the setting key instead (i.e. `Internal/MostRecentlyUsedSessionsData2`).

xen2:
I separated branch `master` and `master-3.0` so I will finish and merge this one soon.
Note: it will require a launcher update/release too (but it was overdue to fix #133).

xwellingtonx:
> I separated branch `master` and `master-3.0` so I will finish and merge this one soon.
> Note: it will require a launcher update/release too (but it was overdue to fix #133).

Thanks, @xen2!

xen2:
Merged with 7d2e0f91bd7
Thanks!

PR #202, opened 2018-09-14
[Game] Update-rate throttling

Eideren:
Related to #183 & #188
First pass on general frame-rate control features, this first commit implements a generic maximum frame rate controller (here, while testing, set to 40/second).
Next commit will implement a feature to reduce frame rate automatically while the game isn't in focus and lastly deal with VSync while the device is lost/disabled.

I would like to have your input on a couple of things:
- Where should we provide access to the variables ?
Move it to the presenter and extend our PresentInterval enum to a class to set it at either a specific value, vsync1, vsync2 or unlocked ?
Keep it inside of GameBase and initialize it to a fairly large rate ?
Force user to initialize it through GameBase's constructor ?
- Should I move the function out of the Profiler sampling for GameEndDraw ?
I'm pretty sure that if we set our interval to VSync it's included inside that sample.

Now if we really want to be more precise we could sleep (the amount - average sleep latency) and spinwait for the rest of the duration, but it's already pretty precise as it is, I don't really see the point of wasting those cpu cycles.


xen2:
Thanks for starting working on this.

* I would say those variables should stay where they are being used. `PresentInterval` is used by `GraphicsPresenter`, but the new variables you introduced are used only by `GameBase` so I suppose that's where they should be declared.
* Good question.. Maybe it can be a separate profiler to identify throttling? (you can make sure to skip the profiler if no throttling but putting it around `Utilities.Sleep` only)

> Next commit will implement a feature to reduce frame rate automatically while the game isn't in focus and lastly deal with VSync while the device is lost/disabled.
Maybe you could cover the case of #188 by @EssenRoc to treat Present=occluded same way as game isn't in focus?

EssenRoc:
This is a question of limiting the frame rate . For example: normal game frame rate, lost focus frame rate, vertical sync frame rate...

EssenRoc:
Present=occluded doesn't need to be processed, just wait for recovery. So we should reduce the frame rate and reduce the performance consumption.

EssenRoc:
[#183](https://github.com/xenko3d/xenko/issues/183) When vertical sync, Present=occluded has no frame rate limit

xen2:
@EssenRoc Yes that's what I was thinking too, doing the Occluded test mode at a slower frame rate (to keep update & CPU low) until Present TEST is not occluded again. This could be covered by using almost the same code (with a different framerate).

Eideren:
I ended up making it a 'Utilities' function since it could be used for other kinds of loops like network threads and the likes.

We can't guarantee occlusion detection for every platform at this moment, right ? So should I really implement a platform specific (dx11&12) path ? Users don't really expect this kind of behaviors and keeping it as minimized and unfocused covers all of our most obvious cases.

Note: mobile solutions keeps crashing for me, the code I wrote to handle focus on those platforms should work but I couldn't test it.

tebjan:
@Eideren thanks for working on this. your approach looks very good to me. i've also did a lot of work on mainloops and can provide some knowledge if you need. from my testings `Sleep(1)` usually is in the range of 1-4ms. occasionally even 16ms, but that is very rare. so if you want to create a very precise mainloop you need to handle the error range of sleep at the end of the wait period in a different way, e.g. do a busy wait phase. see the busy wait section here in my blog post on the timers and main loop:
https://vvvv.org/blog/vl-patch-your-own-mainloops

but maybe that is not the goal to be super precise, but only to throttle frames and save CPU time.

Eideren:
@tebjan thanks, yes, the current solution is very precise if we're only looking for constant FPS but falls short in terms of precise frame-times, I alluded to the same solution on the last paragraph of the first post.
Users would probably expect locking FPS to first reduce CPU consumption before increasing image fluidity so I'll implement it as a feature/toggle.

tebjan:
@Eideren totally agree. toggle sounds great, also a variable to adjust when to enter the busy wait.

sources/engine/Xenko.Games/GameBase.cs
```diff
@@ -97,6 +102,12 @@ protected GameBase()
             TargetElapsedTime = TimeSpan.FromTicks(10000000 / 60); // target elapsed time is by default 60Hz
```
xen2: Might be good to use TimeSpan.TicksPerSecond rather than 10000000.

sources/core/Xenko.Core/Utilities.cs
```diff
@@ -803,5 +876,17 @@ public static IEnumerable<T> IterateTree<T>(T root, Func<T, IEnumerable<T>> chil
...
+        
+
+        /// <summary>
+        /// Converts a <see cref="Stopwatch" /> raw time to a <see cref="TimeSpan" />.
+        /// </summary>
+        /// <param name="delta">The delta.</param>
+        /// <returns>The <see cref="TimeSpan" />.</returns>
+        public static TimeSpan ConvertRawToTimestamp(long delta)
+        {
+            return new TimeSpan((delta * 10000000) / Stopwatch.Frequency);
```
xen2: Might be good to use TimeSpan.TicksPerSecond rather than 10000000.

xen2:
LGTM, thanks!

Eideren:
@tebjan I ended up making an automatically scaled window instead of user-side variable, 'seems to be working pretty well from my tests.

Eideren:
Oh, and, PR ready for review / merging of course.

tebjan:
that's a smart way to figure the time out. however it should be optional. the setting might be influenced by the timer precision and that can change at runtime. also as a professional user i want to decide on my own how much performance i want to sacrifice for precision. so its important that this can be set from outside. but the auto mode can be the default. for example by a flag or if you set wait time to -1 or so...

sources/engine/Xenko.Games/GameBase.cs
```diff
@@ -676,6 +723,11 @@ private void TickInternal()
             }
         }
 
+
+
+
```
xen2: Remove empty lines (should be only 1 in between functions)

sources/engine/Xenko.Games/OpenTK/GameWindowOpenTK.cs
```diff
@@ -257,6 +257,14 @@ public override bool IsMinimized
             }
         }
 
+        public override bool Focused
```
xen2: This one would be nicer as `public override bool Focused => gameForm.Focused;`

xen2: Never mind, rest of file is not... we can do that in a separate major style PR later.

xen2:
@Eideren LGTM and ready for merge. What about latest @tebjan comment, something that can be done after or should I wait?

Eideren:
Implementing @tebjan 's request made a mess of the GameBase's structure so I pushed all of this to a new class/file instead.

xen2:
Did a few adjustments and merged it, thanks!

xen2:
@Eideren Looks like you forgot `GameWindowUWP`: https://teamcity.xenko.com/viewLog.html?buildId=2200&buildTypeId=Engine_BuildWindowsUWP&tab=buildLog&_focus=1288

sources/core/Xenko.Core/ThreadThrottler.cs
```diff
@@ -0,0 +1,217 @@
...
+
+            // We skipped the loop, probably because the system is/was extremely busy, 
+            // reduce delay slowly to force loop re-entry and re-evaluation of the current averageSleepDelay
+            if (looped == false && Type == ThrottlerType.PreciseAuto)
+                spinwaitWindow = spinwaitWindow == 0 ? 0 : spinwaitWindow - spinwaitWindow / 10;
+
+            // prefer one tick too soon than two ticks too late
+            var epsilon = TimeSpan.FromTicks(1);
+            // 'spinwait' for the rest of the duration, that duration should take close to threadSleepDurationMs
+            while (MinimumElapsedTime - callWatch.Elapsed > epsilon) { }
```
Kryptos-FR: Wouldn't it be better to use `System.Threading.SpinWait` instead of manual aggressive empty loop?

Eideren: The class defaults to the other mode, this one's point is to be as precise as possible even to the point of wasting cycles.
Spinwait will yield in certain cases so we won't be able to guarantee users that our timing is precise.

Kryptos-FR: But it is very important to yield after some time, otherwise the system might become unstable (CPU starvation), hence why this struct exist in the .Net framework. In the above code, there is no check to ensure that `MinimumElapsedTime - callWatch.Elapsed` is not a very big number, which I find a bit dangerous.

Eideren: We do yield, Utilities.Sleep yields for 1+ms and that function is called multiple times within this function until the time left to sleep is just slightly larger than one ms. 
if `MinimumElapsedTime - callWatch.Elapsed` is large, that's expected behavior, Utilities.Sleep(Thread.Sleep) will just run a bunch more ; what might be worrying is the `spinwaitWindow` scaling up and ignoring the sleep loop but it will always auto-correct itself back the next couple of calls.

We didn't yield before this PR and we still won't yield if we're running slowly, if yielding is necessary it shouldn't lie in this function's implementation since this one's about capping the maximum execution rate not managing load balancing.

sources/engine/Xenko.Games/GameBase.cs
```diff
@@ -94,9 +94,13 @@ protected GameBase()
             timer = new TimerTick();
             IsFixedTimeStep = false;
             maximumElapsedTime = TimeSpan.FromMilliseconds(500.0);
-            TargetElapsedTime = TimeSpan.FromTicks(10000000 / 60); // target elapsed time is by default 60Hz
+            TargetElapsedTime = TimeSpan.FromTicks(TimeSpan.TicksPerSecond / 60); // target elapsed time is by default 60Hz
             lastUpdateCount = new int[4];
             nextLastUpdateCountIndex = 0;
+            
+            TreatNotFocusedLikeMinimized = true;
```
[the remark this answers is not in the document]
Eideren: I'l take a look tomorrow, probably just need to include TextInput focus into the detection.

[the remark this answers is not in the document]
Eideren: See #229

Issue #195, opened 2018-09-10
Most recent projects are showing deleted items


xwellingtonx:
**Release Type**: Both

**Version**: All

**Platform(s)**: Windows

**Describe the bug**
When you delete a xenko project the Launcher and Game Studio don't remove it from recently files list.

**To Reproduce**
Launcher:
1. Delete a project 
2. Open Launcher

Game Studio:
1. Delete a project 
2. Open Game Studio or go to new project window

**Expected behavior**
If the project was deleted is expected that the project doesn't show more in recent files list.

**Screenshots**
![launcher](https://user-images.githubusercontent.com/10274446/45313690-f5512580-b505-11e8-93bf-e5e2f4aa61ee.jpg)

![gamestudio](https://user-images.githubusercontent.com/10274446/45313700-f97d4300-b505-11e8-8474-1afe6820e18b.PNG)

**Log and callstacks**
Not applicable

**Additional context**
Not applicable

**Please, assign to me.

xen2:
FYI some software (i.e. Visual Studio) don't delete until you actually try to open and it asks you if you want to delete it.
One guess why they do this is that your project might be an a USB stick or network drive that you might have forgot to mount before start starting VS.
Simply wanted to remind you of this use case, not saying we have to do the same.

![image](https://user-images.githubusercontent.com/527565/45328942-00a16100-b599-11e8-8c00-cbaff4dfbcd4.png)

xwellingtonx:
This way is interesting I'll test it! I realized now that we'll need an option to remove a project from the list too, just like visual studio does.

PR #203, opened 2018-09-14
[Game Studio] Recent files property update notification fixed

xwellingtonx:
This PR will be fixed the issue #195.

My first commit it's about the problem with the UI update on editor when you clean the list of recent projects, apparently it's not possible to use an ObservableCollection with value converter `Take`, I don't know why this, because to use a converter is the best solution as informed here [Creating a WPF ItemSource Length Filter Converter](https://weblog.west-wind.com/posts/2018/Jun/23/Creating-a-WPF-ItemSource-Length-Filter-Converter). I had tried many different solutions using it, but none works.


xen2:
Thanks! True that Converter often have lot of limitations...
Let's keep it simple the way you did (esp. since this MRU list is rarely updated) rather than using obscure WPF tricks, looks good to me!

PR #211, opened 2018-09-20
Fix for `Normalize` methods of Int2 and Int3

nemerle:
The previous methods would always produce `inv` == 0, and thus normalization would always zero-out the vector components.

### DIFF
--- a/sources/core/Xenko.Core.Mathematics/Int2.cs
+++ b/sources/core/Xenko.Core.Mathematics/Int2.cs
@@ -130,7 +130,7 @@ namespace Xenko.Core.Mathematics
         /// </summary>
         public bool IsNormalized
         {
-            get { return Math.Abs((X * X) + (Y * Y) - 1f) < MathUtil.ZeroTolerance; }
+            get { return (X * X) + (Y * Y) == 1; }
         }
 
         /// <summary>
--- a/sources/core/Xenko.Core.Mathematics/Int2.cs
+++ b/sources/core/Xenko.Core.Mathematics/Int2.cs
@@ -195,7 +195,7 @@ namespace Xenko.Core.Mathematics
         /// </summary>
         public void Normalize()
         {
-            if (X!=0 && Y!=0)
+            if (X!=0 || Y!=0)
             {
                 double inv = 1 / Math.Sqrt((X * X) + (Y * Y));
                 X = (int)(X*inv);
--- a/sources/core/Xenko.Core.Mathematics/Int2.cs
+++ b/sources/core/Xenko.Core.Mathematics/Int2.cs
@@ -126,14 +126,6 @@ namespace Xenko.Core.Mathematics
         }
 
         /// <summary>
-        /// Gets a value indicting whether this instance is normalized.
-        /// </summary>
-        public bool IsNormalized
-        {
-            get { return (X * X) + (Y * Y) == 1; }
-        }
-
-        /// <summary>
         /// Gets or sets the component at the specified index.
         /// </summary>
         /// <value>The value of the X or Y component, depending on the index.</value>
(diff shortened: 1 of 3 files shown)


sources/core/Xenko.Core.Mathematics/Int2.cs
```diff
@@ -195,12 +195,11 @@ public int LengthSquared()
         /// </summary>
         public void Normalize()
         {
-            int length = Length();
-            if (length > MathUtil.ZeroTolerance)
+            if (X!=0 && Y!=0)
```
Kryptos-FR: That's incorrect. A vector with one component set to zero still has a normalized version. In that case it is simply 1 in the other component.

sources/core/Xenko.Core.Mathematics/Int3.cs
```diff
@@ -214,13 +214,12 @@ public int LengthSquared()
         /// </summary>
         public void Normalize()
         {
-            int length = Length();
-            if (length > MathUtil.ZeroTolerance)
+            if (X != 0 && Y != 0 && Z != 0)
```
Kryptos-FR: Incorrect, see previous comment.

sources/core/Xenko.Core.Mathematics/Int2.cs
tebjan: yes, just replace '&&' with "||"

nemerle:
Gah.. that's what I get for doing 3 things at once :)

Kryptos-FR:
@nemerle That would be more consistent indeed.

tebjan:
now that i think about it again, a vector with integer components can only have length 1 if all components are 0 except one. so basically the normalization can be simplified by finding the biggest absolute component and return a vector that has set this component to 1 or -1 depending on its sign.

xen2:
@tebjan agree, those methods don't make much sense
they don't exist in HLSL either:
https://docs.microsoft.com/en-us/windows/desktop/direct3dhlsl/dx-graphics-hlsl-normalize
maybe we should just get rid of them for `IntX` types?

sources/core/Xenko.Core.Mathematics/Int4.cs
```diff
@@ -186,6 +186,31 @@ public Int4(int[] values)
...
+        /// Calculates the length of the vector.
+        /// </summary>
+        /// <returns>The length of the vector.</returns>
+        /// <remarks>
+        /// <see cref="Int4.LengthSquared"/> may be preferred when only the relative length is needed
+        /// and speed is of the essence.
+        /// </remarks>
+        public int Length()
+        {
+            return (int)Math.Sqrt((X * X) + (Y * Y) + (Z * Z)+ (W * W));
```
xen2: Very small detail, but there's a missing whitespace there

sources/core/Xenko.Core.Mathematics/Int4.cs
```diff
@@ -186,6 +186,31 @@ public Int4(int[] values)
...
+        /// Calculates the squared length of the vector.
+        /// </summary>
+        /// <returns>The squared length of the vector.</returns>
+        /// <remarks>
+        /// This method may be preferred to <see cref="Int4.Length"/> when only a relative length is needed
+        /// and speed is of the essence.
+        /// </remarks>
+        public int LengthSquared()
+        {
+            return (X * X) + (Y * Y) + (Z * Z)+ (W * W);
```
xen2: Very small detail, but there's a missing whitespace there

sources/core/Xenko.Core.Mathematics/Int4.cs
xen2: Nevermind, could fix it in your PR myself

sources/core/Xenko.Core.Mathematics/Int4.cs
xen2: Nevermind, could fix it in your PR myself

xen2:
Merged, thanks!

PR #213, opened 2018-09-24
Vr tracker support

joreg:
- added a list of TrackedItems to Xenko.VirtualReality.VRDevice
- for now only supported by OpenVR, the other APIs return new TrackedItem[0]
- it returns all devices, including HMD, Controllers, GenericTrackers, LightHouses,..
- added DeviceClass enum to file DeviceEnums.cs (previously named DeviceState.cs). the enum is taken from OpenVR
- a TrackedItem also returns its SerialNumber so we can distinguish several devices of the same class
- a TrackedItem also returns its BatteryPercentage
- SetTrackingSpace on VRDevice now allows to choose between Seated, Standing and Raw


sources/engine/Xenko.VirtualReality/TrackedDevice.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        public abstract DeviceClass Class { get; }
+
+        public abstract string SerialNumber { get; }
+
+        public virtual void Update(GameTime time)
+        {           
+        }
+
+        public virtual void Dispose()
```
Kryptos-FR: If it doesn't do anything then either implementing `IDisposable` is not required or it should follow the proper pattern:
```csharp
public void Dispose()
{
   Dispose(true);
}

protected abstract void Dispose(bool disposing)
{
}
```

sources/engine/Xenko.VirtualReality/DeviceEnums.cs
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Xenko contributors (https://xenko.com) and Silicon Studio Corp. (https://www.siliconstudio.co.jp)
+// Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
+namespace Xenko.VirtualReality
+{
+    public enum DeviceState
```
Kryptos-FR: Why moving the enum into a different file?

sources/engine/Xenko.VirtualReality/DummyDevice.cs
```diff
@@ -43,6 +43,8 @@ public class DummyDevice : VRDevice
 
         public override TouchController RightHand => null;
 
+        public override TrackedDevice[] TrackedDevices => null;
```
Kryptos-FR: I personally prefer collection properties to return empty rather than null.

sources/engine/Xenko.VirtualReality/OpenVR/OpenVR.cs
```diff
@@ -150,6 +150,27 @@ public void Update()
             }
         }
 
+        public class TrackedDevice
```
Kryptos-FR: Not a fan of having classes with the same name in different namespaces. Can cause confusion and makes it harder to find/navigate the code.

sources/engine/Xenko.VirtualReality/TrackedDevice.cs
xen2: It's not being done on any other input classes, so for consistency let's keep it that way for now.

sources/engine/Xenko.VirtualReality/DeviceEnums.cs
joreg: before, it was one enum in a file with the same name as the enum (DeviceState). since i added a second enum concerning the Device i thought, instead of adding a second file, i'd put them both in the same file and renaming the file to a more generic name, ie. DeviceEnums.cs.

please let me know if you prefer one file per enum and i'll change it accordingly.

sources/engine/Xenko.VirtualReality/OpenVR/OpenVR.cs
joreg: renamed the Xenko.VirtualReality one to TrackedItem so the one in OpenVR can keep the name of its original api.

sources/engine/Xenko.VirtualReality/DummyDevice.cs
joreg: done. now returns new ..[0]

xen2:
I assume this is ready for review? I will take a look and merge it soon.

joreg:
@tebjan i put the SerialNumber and DeviceClass getters in Update instead of the constructor because i wasn't sure if they couldn't actually change at runtime: see, the TrackedDevice is only an index in to OpenVRs list of devices. to my understandings controllers and trackers can be added/removed (ie. paired/unpaired) at runtime which i'd assume can change their order in the list. therefore i thought putting those in update is the saver option. 

if we are certain it can not happen, then of course it should go to the constructor. but then not only SerialNumber but also DeviceClass.

@xen2 otherwise yes, please.

tebjan:
@joreg ah you are right, if the index of the same physical device changes during runtime, it would create confusion. we could test that with the vive trackers and the controllers.

joreg:
@xen2 i just checked, and indeed, devices can be added at runtime, so i moved the serialnumber check back to update. interestingly though it seems they cannot be removed at runtime. once a slot is occupied it is occupied for runtime, even if the device is turned off.

also i allowed myself to expose another feature we need: VRDevice.SetTrackingSpace()

sources/engine/Xenko.VirtualReality/OpenVR/OpenVR.cs
```diff
@@ -150,6 +151,34 @@ public void Update()
...
+
+            const int StringBuilderSize = 64;
+            StringBuilder serialNumberStringBuilder = new StringBuilder(StringBuilderSize);
+            public void Update()
+            {
+                var error = ETrackedPropertyError.TrackedProp_Success;
+                serialNumberStringBuilder.Clear();
+                Valve.VR.OpenVR.System.GetStringTrackedDeviceProperty((uint)TrackerIndex, ETrackedDeviceProperty.Prop_SerialNumber_String, serialNumberStringBuilder, StringBuilderSize, ref error);
+                if (error == ETrackedPropertyError.TrackedProp_Success)
+                    SerialNumber = serialNumberStringBuilder.ToString();
```
xen2: Does that mean it will allocate every frame? (something we are trying to avoid)

joreg: what do you mean by "allocate every frame"? what those lines do, is accessing the serialnumber property of the trackeddevice with the index TrackerIndex. i don't see anything getting allocated there.

Kryptos-FR: `StringBuilder.ToString()` will allocate a new string at every update.

joreg: ok, any idea to improve that?

joreg:
ouright, will move SerialNumber, DeviceClass and Batteries into getters.

xen2:
Seems like there is some line ending issues in most edited files (I see the warning when opening https://github.com/vvvv/xenko/edit/VRTrackerSupport/sources/engine/Xenko.VirtualReality/OpenVR/OpenVR.cs)

joreg:
indeed, i've just moved to a new PC where this wasn't set correctly. i checked all files this PR modifies and only found an inconsistency in this one mentioned file though..

xen2:
@joreg sorry for the delay...
Let's merge this for 3.1.

sources/engine/Xenko.VirtualReality/DeviceEnums.cs
xen2: Don't mind so much with so small enums, but so far most of the code is one file per enum, so might be better to keep consistency.

sources/engine/Xenko.VirtualReality/DeviceEnums.cs
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    public enum DeviceClass
+    {
+        Invalid,            //There is no device at this index
+        HMD,                //The device at this index is an HMD
+        Controller,         //The device is a controller
+        GenericTracker,     //The device is a tracker
+        TrackingReference,  //The device is a camera, Lighthouse base station, or other device that supplies tracking ground truth.
+        DisplayRedirect     //Accessories that aren't necessarily tracked themselves, but may redirect video output from other tracked devices
```
xen2: Can you put those comments following the /// xml convention?
Since you bothered adding them, better make them visible to API users as well 😄

sources/engine/Xenko.VirtualReality/DeviceClass.cs
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) Xenko contributors (https://xenko.com) and Silicon Studio Corp. (https://www.siliconstudio.co.jp)
```
xen2: New files should only be `// Copyright (c) Xenko contributors (https://xenko.com)` without Silicon Studio

sources/engine/Xenko.VirtualReality/OpenVR/OpenVrTrackedDevice.cs
```diff
@@ -0,0 +1,63 @@
+// Copyright (c) Xenko contributors (https://xenko.com) and Silicon Studio Corp. (https://www.siliconstudio.co.jp)
```
xen2: New files should only be `// Copyright (c) Xenko contributors (https://xenko.com)` without Silicon Studio

sources/engine/Xenko.VirtualReality/TrackedItem.cs
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) Xenko contributors (https://xenko.com) and Silicon Studio Corp. (https://www.siliconstudio.co.jp)
```
xen2: New files should only be `// Copyright (c) Xenko contributors (https://xenko.com)` without Silicon Studio

sources/engine/Xenko.VirtualReality/TrackingSpace.cs
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Xenko contributors (https://xenko.com) and Silicon Studio Corp. (https://www.siliconstudio.co.jp)
```
xen2: New files should only be `// Copyright (c) Xenko contributors (https://xenko.com)` without Silicon Studio

xen2:
LGTM, merging
Thanks!

PR #218, opened 2018-09-26
[Input] console patch

Eideren:
Remove remaining log mentioned [here](https://github.com/xenko3d/xenko/issues/159#issuecomment-424399846).
Sorry about that !


xen2:
No problem, thanks!

PR #222, opened 2018-09-26
[Samples] Fix the SpaceEscape sample

dontpanic92:
The SpaceEscape sample is not working when using version 3.0.0.5:

default

This PR tries to fix it, though I haven't fully understand the big picture about how a template is initiated.

Other small modifications:

- `.gitignore`: added `samples/.vs` folder to skip VS files for `XenkoSamples.sln`
- `XenkoSamples.sln`: removed dup entries for `MaterialShader.Game/Windows` projects

### DIFF
--- a/samples/XenkoSamples.sln
+++ b/samples/XenkoSamples.sln
@@ -217,10 +217,6 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "MaterialShader.Windows", "G
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "MaterialShader.Game", "Graphics\MaterialShader\MaterialShader.Game\MaterialShader.Game.csproj", "{CBDAF756-3F4A-4CAC-975D-A76A2B196A01}"
 EndProject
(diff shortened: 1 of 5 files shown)


samples/XenkoSamples.sln
```diff
@@ -217,10 +217,6 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "MaterialShader.Windows", "G
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "MaterialShader.Game", "Graphics\MaterialShader\MaterialShader.Game\MaterialShader.Game.csproj", "{CBDAF756-3F4A-4CAC-975D-A76A2B196A01}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "MaterialShader.Windows", "Graphics\MaterialShader\MaterialShader.Windows\MaterialShader.Windows.csproj", "{F4F74EB3-D2B6-405B-B483-1835E81A3A40}"
```
Kryptos-FR: Changes in that file are unnecessary.

Kryptos-FR:
I discussed with @xen2 and another way to fix it would be to change the `RootRenderFeature` to be a property (which is preferred for non-private members), hence keeping the name casing.

xen2:
Fixed with ce3f3f9

PR #224, opened 2018-09-27
[UWP] Update window interface implementation

Eideren:
See [202](https://github.com/xenko3d/xenko/pull/202#issuecomment-424959027).
Implemented `visible` as well, not sure why it wasn't before.


xen2:
Merged, thanks!

PR #228, opened 2018-10-06
[Entity] Fetching components by interface implementation

Eideren:
A fair bit of copy/paste had to be done since c# doesn't have an interface type constraint, I could remove the copy paste entirely by forcing the cast to object on `EntityComponent`'s path but I doubt that's worth it, your call.
I took the time to update the old path as well.

### DIFF
--- a/sources/engine/Xenko.Engine/Engine/Entity.cs
+++ b/sources/engine/Xenko.Engine/Engine/Entity.cs
@@ -165,10 +165,21 @@ namespace Xenko.Engine
         /// </summary>
         /// <typeparam name="T">Type of the component</typeparam>
         /// <returns>The component or null if does no exist</returns>
-        [MethodImpl(MethodImplOptions.AggressiveInlining)]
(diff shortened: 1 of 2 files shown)


Kryptos-FR:
Sorry I fail to understand the point of this PR. Can you give concrete examples of what this would achieve that was not possible before?

Eideren:
@Kryptos-FR Let's say that you're implementing an interact function for your game, so you want to query an entity to see if that entity has any interactable component, the most sane way to do so would be to create an interface which houses the interaction function and add it to your components.

Currently, the get function only allows classes deriving from `EntityComponent`, if we had to implement the system above under this constraint you would either need to go through a list of all of the types that are interact-able, enumerate through the base class and check manually or derive all of your interactable from the same base class ( which while doable in this case might be a pain in the ass in other, more complex, scenarios ).

Eideren:
Okay, I'll just do it on my side then if that's an issue.

tebjan:
you could add it to: https://github.com/dfkeenan/XenkoToolkit

PR #229, opened 2018-10-07
[Winforms] Focus -> ContainsFocus

Eideren:
Issue raised by [EssenRoc](https://github.com/xenko3d/xenko/pull/202#discussion_r223173370), window is treated as unfocused while editing text. 
Winforms' input system uses a hack to manage text input, it spawns a text box which gets focus when we request text input, focus only works on the window itself, this PR changes it so it includes component of the window as well into the focus check.

### DIFF
--- a/sources/engine/Xenko.Games/Desktop/GameWindowWinforms.cs
+++ b/sources/engine/Xenko.Games/Desktop/GameWindowWinforms.cs
@@ -379,7 +379,7 @@ namespace Xenko.Games
             {
                 if (form != null)
                 {
-                    return form.Focused;
+                    return form.ContainsFocus;
                 }
                 // Check for non-form control
                 return false;


EssenRoc:
It seems that there are already [IsActive](https://github.com/xenko3d/xenko/blob/b04c4b52668ec096e018eb94d5987e5fd8a6cc7d/sources/engine/Xenko.Games/GameBase.cs#L247l) variables. It is always true.

EssenRoc:
It was annotated [here](https://github.com/xenko3d/xenko/blob/b04c4b52668ec096e018eb94d5987e5fd8a6cc7d/sources/engine/Xenko.Games/Desktop/GameWindowWinforms.cs#L156)

EssenRoc:
Annotated if it is because it is disabled [this](https://github.com/xenko3d/xenko/blob/b04c4b52668ec096e018eb94d5987e5fd8a6cc7d/sources/engine/Xenko.Games/GameBase.cs#L548)

Eideren:
What ?

xen2:
Good way to fix the underlying issue. Merged, thanks!

PR #230, opened 2018-10-07
[Graphics] Added SharedHandle to texture 

tebjan:
added SharedHandle to texture to create shared DirectX 11 textures.


tebjan:
first tests worked fine with only using the Shared option. the SharedNthandle part is optional.

PR #231, opened 2018-10-08
[Graphics] Added SharedHandle to texture

tebjan:
added SharedHandle to texture to create shared DirectX 11 textures. first tests worked fine with only using the Shared option. the SharedNthandle part is optional.

### DIFF
--- a/sources/engine/Xenko.Graphics/Direct3D/Texture.Direct3D.cs
+++ b/sources/engine/Xenko.Graphics/Direct3D/Texture.Direct3D.cs
@@ -165,12 +165,14 @@ namespace Xenko.Graphics
                     var sharedResource = NativeDeviceChild.QueryInterface<SharpDX.DXGI.Resource>();
                     SharedHandle = sharedResource.SharedHandle;
                     break;
-                case TextureOptions.SharedNthandle:
+#if XENKO_GRAPHICS_API_DIRECT3D11
+                case TextureOptions.SharedNthandle | TextureOptions.SharedKeyedmutex:
                     var sharedResource1 = NativeDeviceChild.QueryInterface<SharpDX.DXGI.Resource1>();
                     var uniqueName = "Xenko:" + Guid.NewGuid().ToString();
                     SharedHandle = sharedResource1.CreateSharedHandle(uniqueName, SharpDX.DXGI.SharedResourceFlags.Write);
                     SharedNtHandleName = uniqueName;
-                    break;
+                    break; 
+#endif
                 default:
                     throw new ArgumentOutOfRangeException("textureDescription.Options");
             }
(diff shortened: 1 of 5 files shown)


sources/engine/Xenko.Graphics/Direct3D/Texture.Direct3D.cs
```diff
@@ -154,7 +154,34 @@ private void InitializeFromImpl(DataBox[] dataBoxes = null)
...
+                        var uniqueName = "Xenko:" + Guid.NewGuid().ToString();
+                        SharedHandle = sharedResource1.CreateSharedHandle(uniqueName, SharpDX.DXGI.SharedResourceFlags.Write);
+                        SharedNtHandleName = uniqueName;
+                        break;
+                    default:
+                        SharedHandle = IntPtr.Zero;
+                        break;
+                }
+            }
+            catch
```
Kryptos-FR: If possible, try to avoid empty `catch` clause. It is better to crash and detect the issue than silently swallowing it.

sources/engine/Xenko.Graphics/Direct3D/Texture.Direct3D.cs
```diff
@@ -154,7 +154,34 @@ private void InitializeFromImpl(DataBox[] dataBoxes = null)
...
+                        SharedHandle = sharedResource.SharedHandle;
+                        break;
+                    case TextureOptions.SharedNthandle:
+                        var sharedResource1 = NativeDeviceChild.QueryInterface<SharpDX.DXGI.Resource1>();
+                        var uniqueName = "Xenko:" + Guid.NewGuid().ToString();
+                        SharedHandle = sharedResource1.CreateSharedHandle(uniqueName, SharpDX.DXGI.SharedResourceFlags.Write);
+                        SharedNtHandleName = uniqueName;
+                        break;
+                    default:
+                        SharedHandle = IntPtr.Zero;
```
Kryptos-FR: Maybe it's better to throw an exception (usually `ArgumentOutOfRangeException`) in the `default` clause.
It will help future development: if a value is added to the enum then we can detect where code need to be changed.

sources/engine/Xenko.Graphics/TextureOptions.cs
```diff
@@ -0,0 +1,77 @@
...
+    [Flags]
+    public enum TextureOptions
+    {
+        /// <summary>
+        /// None. The default.
+        /// </summary>
+        None = 0,
+
+        /// <summary>
+        /// Enables resource data sharing between two or more Direct3D devices. The only
```
Kryptos-FR: I think it is better to keep a single line in the` <summary>` and put everything else in `<remarks>`.

tebjan:
@Kryptos-FR thanks a lot for the review! latest commit addresses all points.

xen2:
Changes look good to me.
I still had one point I wanted solved before merging:

At a first glance, the way D3D expose this feature seems quite D3D/Windows specific (i.e. `SharedNtHandleName`).

So I was wondering if this should really be part of the common abstract API or only exists as extra functions in Xenko D3D11 version (then we would need to wrap its use inside `#if XENKO_GRAPHICS_PLATFORM_DIRECT3D11`).

Did you have a chance to take a look if the current design would work out with how other API (such as Vulkan and possibly others) might expose shared textures, if this feature even exists?

tebjan:
@xen2 yes, i had the same thoughts... in general all APIs support texture sharing and all would have a resource pointer to share. so that part is general. SharedKeyedmutex is also a general API to synchronize resource access with a lock. only the NtHandle name thing might be specific to windows, not to D3D thought. i could #IF that part, then all APIs would have the shared handle but only windows the shared NT handle name?

tebjan:
the nt handle code is now in an `#if` region. other graphics API solutions build fine apart from the tests.

xen2:
Thanks!

PR #242, opened 2018-10-25
[Backers] Mike Jones represents MLF

tebjan:
added link in BACKERS.md


xen2:
Thanks.
I'll probably have to keep a list of who is who in BACKERS.md otherwise I might be confused at some point when updating it :)

PR #245, opened 2018-10-27
[Graphics] fixed geometry shader with stream out 

tebjan:
the work was almost done, only someone forgot to build the proper stream out elements and add the matching buffer flag.

### DIFF
--- a/sources/engine/Xenko.Graphics/Buffer.Vertex.cs
+++ b/sources/engine/Xenko.Graphics/Buffer.Vertex.cs
@@ -33,19 +33,18 @@ namespace Xenko.Graphics
         public static class Vertex
         {
             /// <summary>
-            /// Creates a new Vertex buffer with <see cref="GraphicsResourceUsage.Default" /> uasge by default.
+            /// Creates a new Vertex buffer with <see cref="GraphicsResourceUsage.Default"/> uasge by default.
             /// </summary>
-            /// <param name="device">The <see cref="GraphicsDevice" />.</param>
+            /// <param name="device">The <see cref="GraphicsDevice"/>.</param>
             /// <param name="size">The size in bytes.</param>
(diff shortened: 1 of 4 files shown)


sources/engine/Xenko.Graphics/Buffer.Vertex.cs
```diff
@@ -33,15 +33,19 @@ public partial class Buffer
...
-            /// <returns>A Vertex buffer</returns>
-            public static Buffer New(GraphicsDevice device, int size, GraphicsResourceUsage usage = GraphicsResourceUsage.Default)
+            /// <param name="isStreamOutput">if set to <c>true</c> the buffer can be bound to the geometry shader stream output stage.</param>
+            /// <returns>
+            /// A Vertex buffer
+            /// </returns>
+            public static Buffer New(GraphicsDevice device, int size, GraphicsResourceUsage usage = GraphicsResourceUsage.Default, bool isStreamOutput = false)
             {
-                return Buffer.New(device, size, BufferFlags.VertexBuffer, usage);
+                var flags = isStreamOutput ? BufferFlags.VertexBuffer | BufferFlags.StreamOutput : BufferFlags.VertexBuffer;
```
xen2: Rather than a bool, let's use a `BufferFlags` parameter with sensible default: `BufferFlags flags = BufferFlags.VertexBuffer`
That way it's more explicit than a boolean when reading the invoking code.

tebjan: ah ok, i thought because this is in the Buffer.Vertex file the vertex buffer flag should always be set. but flags with vertex buffer as default is more flexible ofc.

xen2: Yes I agree with you, I was also hesitating because of the same reason, but let's go with more explicit choice.

PR #249, opened 2018-10-29
[Graphics] When mixing Dispatch and Draw calls in Direct3D11 the set shaders on the native device context got mixed up.

azeno:
For example effect A sets a hull shader, effect B a compute shader and effect C only vertex and pixel shaders - the hull shader from effect A was now still set on the native device context leading to a broken draw call because when applying effect B the draw shaders were not cleared and follow up comparison when applying C failed.


xen2:
Merged, thanks!

PR #256, opened 2018-11-06
[Serialization] Clarify some issues that users might encounter.

Eideren:
This PR implements a button on the log viewer's UI to show stacktraces and inner exceptions. It also wraps activator/constructor exceptions around a more generic exception to catch them more efficiently.

I'm also presenting a way to entirely ignore any constructor issues while still providing a workable object, see DefaultObjectFactory.cs line 118 and ClassDataSerializer.cs line 18.

Constructor exceptions crashes the editor while its open and block any attempt to open the project afterward. The log viewer doesn't explicitly tell the user what goes wrong when opening the project, this PR should clarify that issue.
If we *really* don't want to block the user from opening a project we could also use the solution mentioned above to bypass this issue and mention to the user that he should resolve his constructor issues.

I'm not including those lines in this first push as I can see how those could lead to some undesired complications, **I will include or remove those lines based on what you would prefer**.

### DIFF
--- a/sources/assets/Xenko.Core.Assets/Diagnostics/AssetLogMessage.cs
+++ b/sources/assets/Xenko.Core.Assets/Diagnostics/AssetLogMessage.cs
@@ -96,8 +96,6 @@ namespace Xenko.Core.Assets.Diagnostics
             {
                 assetLogMessage.Line = yamlException.Start.Line;
                 assetLogMessage.Character = yamlException.Start.Column;
-                // We've already got everything, no need to pollute log with stack trace of exception
-                assetLogMessage.Exception = null;
             }
 
             return assetLogMessage;
--- a/sources/assets/Xenko.Core.Assets/Yaml/CollectionWithIdsSerializerBase.cs
+++ b/sources/assets/Xenko.Core.Assets/Yaml/CollectionWithIdsSerializerBase.cs
@@ -162,7 +162,10 @@ namespace Xenko.Core.Yaml
                     if (objectContext.SerializerContext.AllowErrors)
                     {
                         var logger = objectContext.SerializerContext.Logger;
-                        logger?.Warning("Ignored dictionary item that could not be deserialized", ex);
+                        if(ex.InnerException is DefaultObjectFactory.InstanceCreationException ice)
+                            logger?.Warning($"Ignored dictionary item that could not be deserialized:\n{ice.Message}", ex);
+                        else
+                            logger?.Warning($"Ignored dictionary item that could not be deserialized:\n{ex.Message}", ex);
                         objectContext.Reader.Skip(currentDepth, objectContext.Reader.Parser.Current == startParsingEvent);
                     }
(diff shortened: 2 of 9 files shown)


Eideren:
@xen2 / @Kryptos-FR hey, could you tell me if I should go ahead and implement the solution proposed above or just ignore it and cleanup the PR so I can be done with this.

Eideren:
No idea where I was going with text interpolation there, let's just blame it on a lack of sleep.
Just to clarify, `GetUninitializedObject` in my mind would *only* be used in cases where `Activator` / `new ()` fails.
There are cases of duplicate messages: some logs copies the inner exception message inside their message, filtering that would require manipulating the string which doesn't seem worthwhile to me.

xen2:
I noticed some messages are now displayed twice when this feature is enabled. Is it something on the displaying/formatting side or just something to fix on code throwing exception?

in this case `(Lin: 85, Col: 54, Chr: 4484) - (Lin: 86, Col: 24, Chr: 4555): Unable to resolve tag [!MyGame50.BasicCameraController,MyGame50.Game] to type from tag resolution or registered assemblies` is there twice:

enabled:
image

disabled:
image

Same repro as before:
* Create MyGame
* Make BasicCameraController.cs file not compile properly
* Reload

xen2:
@Eideren Sorry, skimmed through it! Good then!

xen2:
Merged, thanks!

Eideren:
@xen2 Uuuh, I'm pretty sure you pushed the two commented lines we were supposed to discuss (see first post) into master, I still have to push at least once ( either to remove or implement those lines ) before this PR is ready.

xen2:
Ha, just to make sure I understood correctly, from what I saw it behaves same as before in terms of exception, right?

If yes, I merged current version and was keeping option to do separate discussion/PR for this specific point (as mentioned in https://github.com/xenko3d/xenko/pull/256#issuecomment-436991785).

xen2:
You mean those commented lines should be removed in the meantime?

Eideren:
Alright, as long as we keep the discussion going I'm fine with leaving them here, I just thought that you decided against implementing it and forgot to deal with them.

xen2:
I like the idea, however my main concern was that uninitialized objects might fail slightly later which is difficult to diagnose (getter/setters not working because underlying objects not created).

Anyway, as you said, I didn't want to block this PR because of this more contentious bit. Feel free to open a new issue if you want to discuss that part to see what we can do.

Note: we also have this mechanism to create fake proxy object when assemblies couldn't be compiled or object couldn't be deserialized.

Maybe it better go through the same mechanism to make sure that:
1. we don't loose any data on saving back
2. it's clearly visible to the user something went wrong
3. a single unified mechanism for error recovery (less maintenance)

It's possible that .ctor failure doesn't go into this case, I will test again if this still works at all (and add some unit tests to make sure all those cases including .ctor failure is also covered for now).

PR #258, opened 2018-11-09
Patch for #251

### LINKED ISSUE #251: Xenko.Core.Mathematics library
Color:
Color uses byte for its channels. But in functions like `public static Color Add(Color left, Color right)` just add them together which returns an integer. These integers are used in the construction of the result, but because of overloads the constructor that is actually called is `public Color(float red, float green, float blue, float alpha)`.
This constructor uses `R = ToByte(red);` which results in bad values.
So for every method casts to byte should be used.

A lot of other structs:
With the addition of the `in` keyword a lot of the methods of this library might be faster.  Additionally a lot of `ref` keywords which are used for that can be changed to `in` to have the const contract.

SleepyMode:
This pull request should fix #251, mostly.

One thing that was unable to be done was actually using the `in` modifier due to the fact that it wasn't added until C# 7.2 (while we're using C# 7.0).

Additionally, the negation operator (`public static Color operator -`) was left untouched due to the odd behaviour that was observed upon it's modification.

### DIFF
--- a/sources/core/Xenko.Core.Mathematics/Color.cs
+++ b/sources/core/Xenko.Core.Mathematics/Color.cs
@@ -475,7 +475,7 @@ namespace Xenko.Core.Mathematics
         /// <returns>The sum of the two colors.</returns>
         public static Color Add(Color left, Color right)
         {
-            return new Color(left.R + right.R, left.G + right.G, left.B + right.B, left.A + right.A);
+            return new Color((byte)(left.R + right.R), (byte)(left.G + right.G), (byte)(left.B + right.B), (byte)(left.A + right.A));
         }
 
         /// <summary>
@@ -500,7 +500,7 @@ namespace Xenko.Core.Mathematics
         /// <returns>The difference of the two colors.</returns>
         public static Color Subtract(Color left, Color right)
         {
-            return new Color(left.R - right.R, left.G - right.G, left.B - right.B, left.A - right.A);
+            return new Color((byte)(left.R - right.R), (byte)(left.G - right.G), (byte)(left.B - right.B), (byte)(left.A - right.A));
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


sources/core/Xenko.Core.Mathematics/Color.cs
```diff
@@ -525,7 +525,7 @@ public static void Modulate(ref Color left, ref Color right, out Color result)
         /// <returns>The modulated color.</returns>
         public static Color Modulate(Color left, Color right)
         {
-            return new Color(left.R * right.R, left.G * right.G, left.B * right.B, left.A * right.A);
+            return new Color((byte)(left.R * right.R), (byte)(left.G * right.G), (byte)(left.B * right.B), (byte)(left.A * right.A));
```
xen2: Shouldn't there be a division by 255 here? (i.e. max intensity red would be R = 255 x R = 255 and should result in R = 255)

sources/core/Xenko.Core.Mathematics/ColorBGRA.cs
```diff
@@ -499,7 +499,7 @@ public static void Modulate(ref ColorBGRA left, ref ColorBGRA right, out ColorBG
         /// <returns>The modulated color.</returns>
         public static ColorBGRA Modulate(ColorBGRA left, ColorBGRA right)
         {
-            return new ColorBGRA((left.R * right.R) >> 8, (left.G * right.G) >> 8, (left.B * right.B) >> 8, (left.A * right.A) >> 8);
+            return new ColorBGRA((byte)((left.R * right.R) >> 8), (byte)((left.G * right.G) >> 8), (byte)((left.B * right.B) >> 8), (byte)((left.A * right.A) >> 8));
```
xen2: Same

Kryptos-FR: That's what `>> 8` is supposed to do here. But it's wrong because it divides by 256 instead of 255.

xen2: Right, sorry I thought this was same mistake as in `Color.cs`.
Anywway, should be / 255 as you said.

SleepyMode: Hm, well, I could add it once I add the in keywords soon.
Let's wait for the C# 7.2 PR first. (Which I'm going to upload any moment now)

xen2:
@SleepyMode I will just go ahead and do the last fixes myself and merge it. Thanks!

xen2:
Merged in a separate commit, thanks!

PR #268, opened 2018-11-15
Revert "Update to C# 7.2 (#264)"

tebjan:
This reverts commit 2eb9a6f0b34b182026e7eda7255e86521419dba8.

had to revert this, because it crashes the shader compiler. the key files just contain an exception.
so this needs to be further tested by @SleepyMode or @xen2 before we can upgrade the solution to csharp 7.2.

in general i think it is better to upgrade only if there is a feature that is actually required. especially for big projects its always a risk to upgrade core parts.


xen2:
Sorry, I assumed this would have no impact (backward compatible) and didn't test it properly, my bad.
Thanks!

SleepyMode:
We assumed background compability(wrongly).
C#7.2 isn't needed any longer due to the issues being pointed out in Discord about the `in` keyword anyways.

Apologies.

tebjan:
no worries, maybe the fix is easy to do in the shader key file generator, just didn't have time to look at it.

PR #275, opened 2018-11-23
[Graphics] added 'options' parameter to texture 2d create methods

tebjan:
seem to have missed to adjust a few overloads on PR #230


xen2:
Merged, thanks!

Issue #34, opened 2018-08-04
Update to latest SharpDX


tebjan:
Should be easy and makes xenko engine compatible to other projects using latest SharpDX already.

tebjan:
when i wrote this issue i was under the impression that xenko was still using sharpdx 3.2 but xenko 3.0 is using sharpdx 4.0.1 which is fairly new. so not a high priority issue, probably..

Line-:
This should be probably be closed, there isn't a stable release past 4.0.1 that we would want to update to until 4.2.0 is released.

Kryptos-FR:
@Line- It's fine to close the pull request. But we can keep that issue open until a new stable version is released. It shows that we had a discussion and hence would prevent people from creating duplicates.

xen2:
Updated to SharpDX 4.2.0 in latest 3.1 beta (will be released in few minutes)

PR #277, opened 2018-11-24
[WIP] Update SharpDX dependency to version 4.2.0

Venthe:
Update to latest SharpDX #34

**It is not yet tested.**, give me a few days. Pull request made to signal that someone is working :)

### DIFF
--- a/Targets/Xenko.UWP.targets
+++ b/Targets/Xenko.UWP.targets
@@ -10,8 +10,8 @@
   <ItemGroup>
     <!-- Temporarily until dependencies are properly added to .nuspec file (note: will need https://github.com/NuGet/Home/issues/1660 to have different dependencies for different graphics platforms) -->
     <PackageReference Include="Microsoft.NETCore.UniversalWindowsPlatform" Version="6.0.6" />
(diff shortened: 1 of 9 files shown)


xen2:
Thanks! Added WIP in title to reflect current state.

Venthe:
@tebjan no, it's okay. I had a spare time; so I'll probably confirm it's working state & will wait for the new version for the merge

xen2:
@tebjan Would it be OK to upgrade now for 3.1? Or you think better to keep version as low as possible?

tebjan:
if the upgrade doesn't break anything, i think it's safe to go for it. it might well be the last SharpDX version since alexandre droppend support for it. so everyone will end up at this version sooner or later.

for us it would be a bit more easy to stay on the current version since we have third party dependencies that also use that one. but i can do a PR there as well to move up to 4.2. so go ahead, we will manage...

dfkeenan:
It looks as if SharpDX has been updated to 4.2.0. @Xen2 should this PR request be closed?

xen2:
Yes, thanks !
Updated to SharpDX 4.2.0 in latest 3.1 beta

PR #289, opened 2018-12-10
Update BulletSharp to latest version

Eideren:
_Content of this post where edited for future references, you can find previous records on the edit history located inside the header of this post._

Original Description:
Our version of BulletSharp is a modified, extremely old, pre-github version of https://github.com/AndresTraks/BulletSharpPInvoke , this PR should fix dependencies and the likes to allow us to pull from a clean and up to date nuget version of BulletSharpPInvoke instead.
Xenko's version contains a bunch of methods and fields which will have to be replaced with workarounds, some of which don't have any straightforward substitute.

Location of the sources, build scripts, tools and instructions used for this PR's libs located over here:
[https://github.com/Eideren/BulletSharpPInvoke](https://github.com/Eideren/BulletSharpPInvoke)


Kryptos-FR:
I think our version is based on https://github.com/sinkingsugar/BulletSharpPInvoke which was indeed a fork of AndresTraks code. We might also use the mobile version from https://github.com/sinkingsugar/bullet2-sharp-mobile.

sources/core/Xenko.Core.Mathematics/Matrix.cs
```diff
@@ -3477,5 +3478,41 @@ public override bool Equals(object value)
...
+
+        /// <summary>
+        /// Performs an implicit conversion from <see cref="Xenko.Core.Mathematics.Matrix"/> to <see cref="BulletSharp.Math.Matrix"/>.
+        /// </summary>
+        /// <param name="value">The value.</param>
+        /// <returns>The result of the conversion.</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static implicit operator BulletSharp.Math.Matrix(Matrix value)
+        {
+            unsafe { return *(BulletSharp.Math.Matrix*)&value; }
```
xen2: Looking at https://github.com/AndresTraks/BulletSharpPInvoke/blob/master/BulletSharp/Math/Matrix.cs#L58
it seems we have a different memory layout compared to Bullet matrix (M12 vs M21)

sinkingsugar:
Hmm the API is slightly different, many things were done towards Xenko perf/optimization.
Can't remember all but like @xen2 said, our matrices are different.
And `SharpMotionState` was a big one to optimize perf and collisions detection.
Lot's of changes were also in the C++ library itself!

This is just a wrapper, not sure you get any benefit by updating it unless the C++ part is also updated ( and has outstanding changes )

So unless the native library is also updated and properly tested on all platforms I would pay attention.

Eideren:
Well, no idea how I ended up looking at xenko's matrix and miss the difference in their order.

Just to clarify, this PR wasn't made *just* to stay up to date with the wrapper, the main push behind this PR is the fact that the version used within xenko doesn't support multiple features of bullet, the main one being convex mesh collision.

From what I gather you guys don't have the sources for the C# and C++ side anymore, I can't do much about that without them so unless you do end up finding them I'll just continue on and we'll see where we end up.

xen2:
Sorry, didn't realize you were looking for the old source (esp. after @Kryptos-FR mentionned a repository which was almost the proper one, I thought you had it and were more focusing on how to rebuild from new sources!)

Here you go, it was there all along!
https://github.com/xenko3d/bullet2-sharp-mobile/tree/master/src
It contains both C++ and C#

xen2:
Now I understand better why you did a compare with decompiled code (at first I thought you had the sources but wanted to see everything formatted the same way... should have understood that earlier!)
Sorry for the misunderstanding!

xen2:
BTW, while I still have it on top of my head, some remarks about how those native libs are compiled so far:

We were usually building the sources for libbulletc and some other native libraries that needs target OS in manually in various computer/VM (hard for people to do it themselves from scratch since those VM had lot of specific software installed).

It would be great if we were switching to something more reproducible and easier to test.
For example, something that can either (1) run in CI, or (2) something that can setup a full docker instance with all dependencies (for Linux/Android, or similar system for Mac), or a combination of both.

That's probably something we should do for all those dependencies that can't be built using our C++ native layer directly in Xenko projects (which support only a very limited subset of the standard library) but instead need to be built on the target OS.

tebjan:
> Xenko matrices are different.

that is indeed a bit problematic. since directx as well es opengl and i guess bullet have exactly the same memory layout. i wonder what was leading to that decision and which action was taken. my feeling is that there was a couple of wrong assumptions of how matrix transposition and coordinate system handedness works.

xen2:
@tebjan I think the idea at the time was to send data to HLSL as and be able to keep HLSL order `mul(W, V, P)` without having to slap `row_major` everywhere.
I think it was also generating slightly simpler HLSL code too but that was long time ago so don't quote me on that.

However, I would be totally fine to revisit this and have your opinion.
Would it be to switch back to M11/M12 and swap all shader matrix multiplications?

I personally don't like the M11/M21 memory layout either, it's opposite to what most engine do 😄

xen2:
PS: I think I could remember HLSL was compiled to dot3 line by line when doing `mul(point, World)` rather than something slightly more complex where it has to take values from non contiguous memory (if it was `mul(World, point)`.

xen2:
@tebjan moved discussion and my comments in #291

tebjan:
thanks @xen2 , will need some time to get deeper into it. hope i will have time for this soon.

Eideren:
Hey, thanks a bunch for the source @xen2 , 'BvhTriangleMeshShape' is already implemented on that repo, as well as a ton of other missing classes, could the dll included with xenko be out of date ?

xen2:
Doesn't seem to be included in the csproj:
https://github.com/xenko3d/bullet2-sharp-mobile/blob/master/src/BulletSharpPInvoke/BulletSharpPInvoke.csproj

Eideren:
So, I ended up choosing to work on optimizing the main repo of BulletSharpPInvoke instead of keeping our own up to date, I already have posted a PR over there and we'll see how it works out.

All contact data in xenko is in world space right ?

xen2:
@Eideren Took a quick look at our code and bullet code and yes, it looks like all `ContactPoint` data is in world coordinates.

tebjan:
@GutterLab bepuphysics doesn't look as mature as bullet, from their readme:

It notably does *not* yet include:
- Convex hulls, cylinders, cones, or other complex shapes 
- Dedicated heightmap terrain meshes
- Mesh-mesh collision detection
- Bounciness, other than the frequency/damping ratio tuning
- Continuous collision detection, other than the speculative margin
- Many useful constraint types

Eideren:
I'm still working on this but BulletSharpPInvoke's maintainer seems pretty busy right now, we might have to fork it instead, we'll see.

Flavelius:
To chime in on the different engine suggestions, PhysX could be a good consideration too, potentially providing some speedups though GPU support.

Kryptos-FR:
> potentially providing some speedups though GPU support

Isn't that only on NVidia graphic cards. If so, we'll have a kind of vendor-lock.

Flavelius:
Afaik, it is. But having that optional, according to 
https://store.steampowered.com/hwsurvey/videocard/ or 
https://www.statista.com/chart/15172/the-pc-graphics-card-market-share-among-gamers-on-steam/ 
it could atleast benefit a good share of users. Although not all, yes.

Eideren:
Could you please continue your discussion in another thread: create a new issue or talk about it on discord but let's keep this thread about the subject at hand, thanks !

Now, the maintainer actually got back to me shortly after my last message and fixed most of the issues I had with the current state of his implementation, I consider it now to be in a good enough state to use it instead of our current one.
I was thinking of including his repo (or a xenko-specific fork of it) as a submodule of this one, what do you think about that @xen2 ?

xen2:
@Eideren Sorry I didn't see that question.

Maybe you could follow the same template we usually do in the `deps` subfolder, where there is a `checkout.bat` and `build.bat` files:
https://github.com/xenko3d/xenko/tree/master/deps/Mono.Cecil
The idea is to avoid everybody having to checkout the submodule (most people won't rebuild it).

Note: there is probably a few improvements to do to this model in the future:
* Use commit ID for checkout (i.e. https://github.com/KhronosGroup/MoltenVK/tree/master/ExternalRevisions)
* Use docker toolchains to have easily reproducible builds for Linux/Android/iOS native libs (possibly coupled with a CI system to build those)
* Maybe once we reach that point where a clean VS could build those, it might be worth discussing if we want to still commit binaries (time/convenience vs more-proper-build tradeoff)

Eideren:
Update: I don't think I'll have time to work on it this month, if someone wants to take over go for it.
What's left to be done is including ``BulletSharpPInvoke``'s repo or binary into our solution (see xen's message above this one for more info), remove our old version and verify compatibility with our supported platforms (iOS, android, max, unix, etc.).

phr00t:
Making what appears to be significant progress on this, although things keep going deeper and deeper. Specifically, AndresTraks/BulletSharpPInvoke does NOT MAINTAIN MACOS BINARIES, so I have to build them myself! This means pulling the source for BulletSharpPInvoke, bullet & libbulletc. Apparently the bullet-2.82 doesn't have functions libbulletc is looking for, so I pulled 2.88 from the bullet github. Things are compiling, but the Makefiles don't line up with all of the cpp files that have changed... working on that part.

**I'm totally worried that after all this work, something won't line up on Xenko's end anymore... which means I'll probably need more help!**

phr00t:
I got bulletsharp updated (or so I believe) & have it "working" (a ball drops and hits the plane floor) with these additional changes from what @Eideren did:

Remove BulletSharp nuget references: https://github.com/xenko3d/xenko/commit/da35e31de9604d762dad08fee0ae10b3694bdea4

Update binaries and some code: https://github.com/xenko3d/xenko/commit/96a0a3a6aa2bf4270ae56a077addb472a28a77ca

I didn't update Android or UWP binaries, so those platforms might not work with this update. I haven't tested on Linux or Mac yet, but I'd love someone to do so! My branch unfortunately is kinda ugly, so I'm hoping @Eideren can cherry-pick the two above commits into his bulletsharp_nuget branch to update this pull request?

phr00t:
Just noting that I've been enjoying updated BulletSharp on my fork for awhile now. It is working very well. You should be able to just pull those changes into your branch to finish up this pull request.

Eideren:
Alright, everything I proposed has been implemented and works, there are a couple of hacks that I want to replace though.

ReferenceSerializer throws, here's the stacktrace : [ReferenceSerializer.txt](https://github.com/xenko3d/xenko/files/3167203/ReferenceSerializer.txt) here's the location in master : https://github.com/xenko3d/xenko/blob/582915a5b5ade7fcb4ac8b57d0de6f51fd7c978f/sources/core/Xenko.Core.Serialization/Serialization/Contents/ReferenceSerializer.cs#L113
When I [comment that exception out](https://github.com/xenko3d/xenko/pull/289/commits/3a96ee3f0da9c02e5f1221669745316567122e5e#diff-5a1502f071804fedaa74fbeacaf555a2R113) it doesn't crash, which is weird, and ends up working fine*.
*The shape is getting constantly re-created collision shape, might be related ? It spawns correctly in game and the collision shape is visible within the editor.

The CommandList used to extract mesh data from models is [done through reflection here](https://github.com/xenko3d/xenko/pull/289/commits/3a96ee3f0da9c02e5f1221669745316567122e5e#diff-326f6a72b6593dc71e3cb054048c4395R89),
should I just pass something which contains that (like a game instance) to [CreateShape and co.](https://github.com/xenko3d/xenko/blob/3a96ee3f0da9c02e5f1221669745316567122e5e/sources/engine/Xenko.Physics/Engine/PhysicsColliderShape.cs), that would be kind of messy, or should I still use ``mesh.Draw.IndexBuffer.Buffer.GraphicsDevice.InternalMainCommandList`` but with more explicit access ? Would you rather I did it in some other way ?

As a note, here's why I didn't end up doing it as an asset (like the convex hull system already in place) : 
- It doesn't require any processing before being useable at runtime.
- Creating asset for meshes that already exists within the project is redundant and time consuming.
- It would increase build size by duplicating mesh data.

Ethereal77:
Hi, @Eideren. I've been following this issue and hope it is integrated soon. Regarding the points you make:

- While it is important to be able to easily create collision geometry at runtime (for procedural models, for example), I think it is already important to be able to import collision geometry as asset.
- Imported collision geometry is not redundant in several interesting use cases, like creating optimized, simplified or custom collision meshes for static models (e.g. you have a very highly detailed mesh for rendering, but you only need a very simple low-poly collision mesh).
- Imported collision geometry also brings the possibility of preprocessing the mesh. That way, the collision mesh is a very slim and fast-loading data that only defines vertex positions and indices and nothing more.

As I said first, producing collision meshes at runtime is important and should be kept. But please do not discard the asset way.

Eideren:
Hey @Ethereal77 !
I wasn't very clear ; when I am referring to loading it as an asset I should have rather said having an asset type specifically designed to be consumed by static colliders and only by them, I'll edit that post to avoid this confusion.

This PR, right now, uses any Model asset that you provide in editor to create the static collider from.
You can totally import a low-poly mesh for the static collider while having another high quality model alongside it for rendering purposes.

If we where to process the collision mesh we might as well implement that on top of our Model asset as that might not only be useful for colliders but for LOD as well.

As for the loading speed, I might investigate that later, we should be able to load data from disk/bypass how Models upload their data directly to gpu to avoid that kind of issues.

xen2:
@Eideren 
**ReferenceSerializer crash**
It looks like DeepClone() is made only for simple objects and can't deal with objects containing references to other complex objects (such as Model).
Maybe it would be better to switch `PhysicsElementInfo.CloneDescs` to use `AssetCloner.Clone()` rather than `DeepClone()` method.

**CommandList**
It's not a good idea to access command list like this (i.e. on Vulkan I think this is not even set).
Also, command list are better passed to not hide the fact they are not threadsafe (i.e. in the code wouldn't run well depending on how/when it's executed).

However, I remember we had a branch trying to remove `CommandList` from some variants of `GetData/SetData` to better match newer API like DX12.
I am not able to take a look at it right away (computer is in a box until 15th) but will try to keep you posted about that.

Eideren:
``AssetCloner.Clone()`` doesn't clone Mesh properly, it returns null, I might not be using it right though ? (``var cloned = Core.Assets.AssetCloner.Clone(desc);`` instead of [this](https://github.com/xenko3d/xenko/blob/582915a5b5ade7fcb4ac8b57d0de6f51fd7c978f/sources/editor/Xenko.Assets.Presentation/AssetEditors/Gizmos/PhysicsGizmo.cs#L110)).

xen2:
@Eideren Please try again using latest master and `AssetClonerFlags.KeepReferences`.

Eideren:
Serialization exception thrown when launching a game on this branch right after merging with master, could not be reproduced on master.
[Exception.txt](https://github.com/xenko3d/xenko/files/3212955/Exception.txt)

Eideren:
@xen2
works fine @ [Assets] Added a new KeepReferences flag to AssetCloner
works fine @ [Physics] Fix initial collider scaling 
exception @ Merge branch 'master' into bulletsharp_nuget
exception @ [Physics] Use asset cloner for shape desc

All of that without using any static meshes.

xen2:
@Eideren OK thanks, will take care of that one

xen2:
@Eideren OK, it seems due to bump to latest Mono.Cecil.
I reverted that part for now (while keeping some various fixes done at the same time), and investigate later.
Please merge again latest master.

Eideren:
Tested it out, works fine !
Ready to be reviewed/merged.

Eideren:
Now we just have to deal with the CommandList issue

xen2:
For future reference, the Mono.Cecil issue is already fixed but not released yet.
Asked about it there: https://github.com/jbevain/cecil/pull/570

xen2:
Unfortunately I don't have access on my current computer to the branch mentioned earlier (with non-CommandList GetData).
It is likely on the other computer which will be delivered in roughly 10 days.

This PR will likely need further work anyway.

How do you want to proceed? Wait for that? Or maybe expose CommandList parameter as a temporary workaround, and then we could work on removing that restriction later?

Eideren:
@xen2 
People using Vulkan are going to crash when creating static mesh colliders, if you're okay with that I'll expose commandlist, replace the reflection calls and you can merge it.

Right now I am looking at how I could retrieve mesh data by reading the asset from disk instead, I'm kind of getting nowhere as I'm not familiar with that side of the API ( retrieve asset path from instance of model/mesh ? then retrieve raw mesh data from that path ?), I would still need to pass around services or something to have access to all of that ?

xen2:
@Eideren 
Just to clarify, the idea was to add `CommandList` as an explicit parameter of `NewSharpFromDesc(CommandList)`, not to expose the `InternalMainCommandList` (I say that because you said Vulkan wouldn't work, but it wouldn't be the case if `CommandList` is a parameter as in `GetData/SetData`.
However, I think this approach might not even be possible because you might not have a `CommandList` to pass to `NewSharpFromDesc(CommandList)` in some places (didn't check your code in detail yet though).

For second approach (loading data from disk), do you want to do that:
(1) when asset is loaded from disk, same time as Model is created
(2) afterwise when shape is created?

For third approach (remove CommandList from GetData, similar to ID3D12Resource::Map): unfortunately the PC that had the branch is still stuck at custom... I will look around if I have it somewhere on another HDD. Not 100% sure it would solve this specific problem yet either. I will keep you posted.

Eideren:
@xen2 Anytime when we end up transforming a model/mesh asset into a mesh collider.
The model's asset data must somehow be pulled from the disk into ram before being passed on to gpu/vram, I would like to take the same path but instead of pushing into vram pass it to the mesh collider.

xen2:
I wonder, wouldn't it be better if it was properly converted/compiled and stored as a separate blob?

I assume such physics data would never be large (otherwise it would get really slow anyway).
Also, I assume Bullet generates some kind of optimized data structure (a quick search seems to indicate BVH), which probably take some time to build. However it can be deserialized.

All that considered, wouldn't it make more sense to build the BVH at compile time, serialize it in a binary stream, and then load that at runtime?
`ConvexHullColliderShapeDesc` already works in a similar manner, so if you look at how it's done, it shouldn't be difficult to do the same. Please take a look at `ColliderShapeAssetCompiler` too.

Note: The way ConvexHull is done currently is quite hackish:
* Both design-time `Model` and runtime representation `ConvexHulls` and `ConvexHullsIndices` in the same object).
We should probably switch to proper separate objects at some point, but that could be investigated later.
* `ColliderShapeAssetCompiler` is in `Xenko.Assets`
Both of those issues are higly related to the fact `Xenko.Physics` doesn't have a design-time counterpart assembly (yet).

Eideren:
@xen2 I just tested out a legit game level made out of a single 1,800,000 triangle mesh. 
Using it within the FPS sample did not seem to significantly affect load time; the entire game loads in less than 8 seconds and, yes, it does work correctly as I could not pass through the mesh.

I do not think that the load time is significant enough to warrant looking into backing the collision data even more so when we consider the issues that it would bring:
Increase our build output size (the vert+indices of the mesh itself is only about 40MB but the BVH data might significantly increase the required size, we also might not be able to re-use BVHs so multiply all of that by the amount of colliders using that mesh), increase the compile time for those builds and waste plenty of our users time as they would need to explicitly ask/be prompted for such an asset to be created.

Again, lets not make this more complex than it needs to.

xen2:
@Eideren Sure, I was not trying to add complexity but rather trying to avoid adding some complexity in the loading area (reading the data twice is not so trivial as the buffer data is deep in the object tree, and there's no common states when loading).

Anyway, it seems there's no avoiding it. 

I will explain little bit the issues in details in another post, and then we could try to figure out a solution. Maybe there's a more clever way I couldn't think of yet.

Sorry it's taking some time to figure out, it's definitely not trivial as it sounds.
(also I have a lot going on at the same time -- almost settled in new place, moved in yesterday/today!)

xen2:
Quick update: after 3 long weeks of delay due to customs, I should receive all my stuff tomorrow.
Crossing finger that I have that "get/setdata with no command list" branch lying somewhere.

xen2:
OK, got the computer today!
I could extract the branch, rebase it (git didn't play so nice since it was on old Xenko repo without common ancestor and before all the renaming) and fix a few compile errors.
It really had the `GetData` and `SetData` without `CommandList`, so at least that part I remembered well!
Here it is: https://github.com/xenko3d/xenko/tree/d3d_multithreading

I didn't investigate (yet) the details of the PR as well as if & how to merge it.
Please note that there is a few things that are probably broken still (i.e. I didn't see a grid in the Game Studio).

Eideren:
@xen2 You didn't get back to me on the issue with extracting vertex data from disk, I did prod around a bit to see if I could figure it out myself and from what I understood the data is uploaded to the GPU as soon as it is deserialized ? Without any proper way to hook ourselves into the system ?

I would love to help you out with the commandlist stuff but I don't have much experience with graphics api, so I won't be able to contribute significantly.

I think I should put the editor-side mesh collider code in a separate PR so that we can finally merge this one if you're ready to do so ?

xen2:
@Eideren Yes, the issue is that data is loaded to GPU memory right away.
Technically one could hook deserialization of buffer, but at the point it's called, there is not enough context info to know if the data needs to be set aside or not for something else.

Command list stuff would help in that you could get data from buffer without a command list, so you could just read back the GPU memory just like reading CPU memory.

Also, if everything is somewhat usable without editor-side mesh collider, we should definitely split this in another PR and merge the other part. Sorry that I didn't realize it was OK without!

Eideren:
@xen2 There you go, I'll create the mesh collider PR from the master branch as soon as this one is merged in.

xen2:
@Eideren thanks, code looks good to me!
I will run a quick test to see if everything works good in practice and then merge it.

xen2:
I noticed a bug when trying the ThirdPersonPlatformer.
It happens only in specific area of the map.
Probably due to the raycast.
Here's a video of it:
https://youtu.be/DOEGVDvNUkI

Eideren:
@xen2 I'll take a look at it

Eideren:
Solved, the issue laid within ShapeSweep, that's what the TPP template uses by default for its camera collision, works fine now.

xen2:
Thanks.
I did some additional testing and everything looks good, LGTM for merge!

Maybe one last thing before doing so: could you add a short markdown file at `deps\BulletPhysics\README.md` that contains:
- link to repo for `BulletSharp.NetStandard.dll` and `libbulletc.dll` (https://github.com/Eideren/BulletSharpPInvoke if I understood correctly)
- which version is currently compiled (git commit or tag)
- any extra info that might be useful to remember when trying to build it again (I suppose most of it is already in your BulletSharpPInvoke repo, in which case it's all good, but if you think of anything else...)

Maybe you could also either update or delete the batch files (build.bat, checkout.bat and sign.bat) so that people are not confused trying to fetch different repo.

Eideren:
@xen2 Sure I meant to deal with those but kind of forgot, thanks for reminding me.
So, ``sign.bat`` is as up to date as I can do on my end, not sure how all of that works but the paths should all be mapped correctly ( ``paradox.snk`` was swapped to ``xenko.public.snk`` but I don't know if the functionality, if any, is retained ).

I think you should probably fork/copy my repo to xenko3d's account and then modify ``checkout.bat`` and the new ``README.md`` to point to that fork instead.

xen2:
Merged, thanks again and sorry it was delayed (esp. I didn't realize this could be split in two to take care of the more delicate part later)

Eideren:
@xen2 Nah, it's my fault for not mentioning it sooner :)

xen2:
@Eideren no worries, the delicate part should have been taken care of sooner too!
Anyway, I have made a fork of BulletSharpPInvoke in xenko org: https://github.com/xenko3d/BulletSharpPInvoke
I will give you full permissions on it.

Issue #20, opened 2018-08-03
Script and Task improvements Epic


xen2:
Various ideas this Epic covers:
* Make our task system alloc-free (possible with recent C# version) so that people are not afraid to create thousands of scripts with `await`
* Reorganize task system so that we can wait various points of engine. So far we only have Update/Draw, but I suspect many scripts wants much more control when they want to run (after physics, etc...).
Not sure yet if we want some kind of high-level sync point graph editor, or something more programmatic.

rikimaru0345:
@xen2 
> * Make our task system alloc-free (possible with recent C# version)

I would love that!!
But how is that possible? Which C# feature can help with that and in what way?
Doesn't calling methods that represent a state-machine (async; and iterator/enumerable metods) always have to allocate at least some space the statemachine (its state and its "locals")?

velhaco20000:
Is there anyone working on this task ?

velhaco20000:
You proposed 2 features:
- Minimize allocation in a task start, possibly with a minimum thread local storage;
- Create several points to start a script in order to achieve a precondition.

Is that right ?

xen2:
Yes, that's two separate features (probably done together in the same feature branch).
I started some prototyping on this task: https://github.com/xen2/xenko/tree/microthread2
I have made a simple library to test part of the concepts, next step is to decide exactly how to integrate it into Xenko and switch various part of Xenko to use it.

velhaco20000:
Do you want to create a mini batch system ?

velhaco20000:
I will read the code and make a proposal.

xen2:
FYI the code is still a prototype/playground, it is not yet integrated and lot of things will change.

velhaco20000:
Just a little update, and correct me if I'm wrong :
- You wrote a mini batch system , with a scheduler and a synchronization system;
- Normally, users don't access directly such systems. There is a layer in the engine breaking great tasks into little taks and this layer and the little tasks weight depends on system configuration (I.E. core counts, core capabilities, etc) ; 
- If those assumptions are correct, the next step is construct this layer to break great tasks in little tasks and create the synchronization points;
-DX12 and Vulkan helps with this job;
- I don't know if you are thinking on transparency on this layer, permitting the user/programmer to control it beyond creating custom sync points;
Let me hear your observations and corrections.

velhaco20000:
little Update 2: (correct me if I'm wrong) 
[...]

PR #294, opened 2018-12-21
[ScriptSystem] Allow ScriptSystem to be derived from.

Eideren:
Minor changes to allow for game-specific implementation of the 'ScriptSystem'. 
This enables users to implement custom ScriptComponent events, execution culling, etc. While requiring little to no maintenance on our part.

### DIFF
--- a/sources/engine/Xenko.Engine/Engine/Processors/ScriptSystem.cs
+++ b/sources/engine/Xenko.Engine/Engine/Processors/ScriptSystem.cs
@@ -17,8 +17,9 @@ namespace Xenko.Engine.Processors
 {
     /// <summary>
     /// The script system handles scripts scheduling in a game.
+    /// This class can be derived from, to do so you'll have to provide your derived type to the game's constructor.
     /// </summary>
-    public sealed class ScriptSystem : GameSystemBase
+    public class ScriptSystem : GameSystemBase
     {
         private const long UpdateBit = 1L << 32;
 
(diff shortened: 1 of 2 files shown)


sources/engine/Xenko.Engine/Engine/Processors/ScriptSystem.cs
```diff
@@ -127,7 +128,7 @@ public override void Update(GameTime gameTime)
         /// Allows to wait for next frame.
         /// </summary>
         /// <returns>ChannelMicroThreadAwaiter&lt;System.Int32&gt;.</returns>
-        public ChannelMicroThreadAwaiter<int> NextFrame()
+        public virtual ChannelMicroThreadAwaiter<int> NextFrame()
```
Kryptos-FR: Virtual methods are more expensive than non-virtual and can prevent inlining.

While making `Add` and `Remove` virtual will probably have a low impact, `NextFrame()` is called very often, hence its impact is non-trivial.

Eideren:
Okay.

xen2:
Few remarks:
* Since `NextFrame()` is something called from user script, it shouldn't be hard for you to create a custom type that provide all the sync points you want? (it could even use `NextFrame()` under the hood)
* You mention you want to do custom events, so it's not only about overriding `NextFrame()` but also providing new functions? (which can also be done in an outside class)
I mostly want to make sure you can do everything you want in an external class or if the unseal/virtual would make some new scenario possible.
* This code will be replaced/obsolete soon (#20). `NextFrame` was OK to start with but we need much more control.
Interested in your use case, should take them into account during #20 redesign.

Kryptos-FR:
> Kryptos-FR provided enough good reasons for me to drop this one off

@Eideren I am not always right and I could be missing some key points.. Don't be so prompt to delete your work until the discussion has reached a conclusion :wink:

PR #302, opened 2018-12-27
[Core] ThreadPool: shutdown idle threads

Eideren:
ThreadPool never close off the threads it spawns, see [Discord discussion](https://discordapp.com/channels/500285081265635328/500290856532574209/527465891529621514) for more details.

@Kryptos-FR I did a couple of changes, how would you deal with synchronization ?

### DIFF
--- a/sources/core/Xenko.Core/Threading/ThreadPool.cs
+++ b/sources/core/Xenko.Core/Threading/ThreadPool.cs
@@ -24,8 +24,9 @@ namespace Xenko.Core.Threading
         private readonly ManualResetEvent workAvailable = new ManualResetEvent(false);
 
         private SpinLock spinLock = new SpinLock();
-        private volatile int workingCount;
-        private volatile int aliveCount;
+        private int workingCount;
+        /// <summary> Usage only within <see cref="spinLock"/> </summary>
+        private int aliveCount;
 
         public void QueueWorkItem([NotNull] [Pooled] Action workItem)
         {
@@ -39,10 +40,9 @@ namespace Xenko.Core.Threading
                 workAvailable.Set();
 
                 var curWorkingCount = Interlocked.CompareExchange(ref workingCount, 0, 0);
-                var curAliveCount = Interlocked.CompareExchange(ref aliveCount, 0, 0);
-                if (curWorkingCount + 1 >= curAliveCount && curAliveCount < maxThreadCount)
+                if (curWorkingCount + 1 >= aliveCount && aliveCount < maxThreadCount)
                 {
-                    Interlocked.Increment(ref aliveCount);
+                    aliveCount++;
                     Task.Factory.StartNew(ProcessWorkItems, TaskCreationOptions.LongRunning);
                 }
             }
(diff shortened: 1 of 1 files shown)


Eideren:
Also, we just ignore exceptions created by the delegates it runs, shouldn't we handle them in some way ?

xen2:
Yes I also noticed this exception swallowing when we discussed about it in Discord channel, not good...

Any proposal?
What do we want for default behavior if exception, do we keep running or make the app crash?
Should we offer an event? (i.e.  `Dispatcher.UnhandledException`)
Or simply let it go to `Application.UnhandledException`?

xen2:
Also concerning synchronization, please keep the `SpinLock`.

This is highly performance critical (we use that for microtasks and spawn hundreds of them per frame).
In most case the wait should be very brief and `SpinLock` as many advantages for that (very fast if no lock, value type, etc...)
The engineer who worked on it spend quite some time doing measurements/tests.
We don't want to change that without further test/measurements and a good reason to do so.

sources/core/Xenko.Core/Threading/ThreadPool.cs
```diff
@@ -1,9 +1,10 @@
 // Copyright (c) Xenko contributors (https://xenko.com) and Silicon Studio Corp. (https://www.siliconstudio.co.jp)
 // Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
 using System;
-using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
+using System.Diagnostics;
```
Kryptos-FR: If possible, keep `using` sorted alphabetically.

sources/core/Xenko.Core/Threading/ThreadPool.cs
```diff
@@ -19,109 +20,71 @@ internal class ThreadPool
...
         private readonly int maxThreadCount = Environment.ProcessorCount + 2;
-        //private readonly int maxThreadCount = Environment.ProcessorCount * 2;
-        private readonly List<Task> workers = new List<Task>();
         private readonly Queue<Action> workItems = new Queue<Action>();
         private readonly ManualResetEvent workAvailable = new ManualResetEvent(false);
 
-        private SpinLock spinLock = new SpinLock();
-        private int activeThreadCount;
+        private object _lock = new object();
+        volatile private int activeThreadCount;
```
Kryptos-FR: Modifier order: `private volatile`. But anyway, `Interlocked` and `volatile` should not be combined, either use one or the other but not both.

sources/core/Xenko.Core/Threading/ThreadPool.cs
```diff
@@ -19,109 +20,71 @@ internal class ThreadPool
...
+        volatile private int aliveThreadCount;
 
         public void QueueWorkItem([NotNull] [Pooled] Action workItem)
         {
-            var lockTaken = false;
-            try
+            lock (_lock)
```
Kryptos-FR: Locking is very expensive here as @xen2 mentioned. `Spinlock` might be better, contrary to what I thought first.

sources/core/Xenko.Core/Threading/ThreadPool.cs
```diff
@@ -19,109 +20,71 @@ internal class ThreadPool
...
             {
-                spinLock.Enter(ref lockTaken);
-
                 PooledDelegateHelper.AddReference(workItem);
                 workItems.Enqueue(workItem);
-
-                if (activeThreadCount + 1 >= workers.Count && workers.Count < maxThreadCount)
+                workAvailable.Set();
+                if (activeThreadCount + 1 >= aliveThreadCount && aliveThreadCount < maxThreadCount)
```
Kryptos-FR: Usually, it is an issue to access a field with and without `Interlocked`.

Eideren: Wait, I thought that reading a 32bit value was guaranteed to be atomic ? I just made them volatile to guide the compiler ?

xen2:
BTW I can think of at least one race condition & potential deadlock:

If all threads are finishing, after the while loop but before the `Interlocked.Decrement(ref aliveCount)`.
`QueueWorkItem` won't add the item but then the other thread finishes and everything is stuck unless another task is queued.
I guess risk is higher if `maxThreadCount` is very low.

Probably unlikely, but multithreading program often tend to have those worst case you think are never possible once in a while given enough time :) (esp. since all the threads might be started at approx the same time)

Eideren:
I could also compute the result of the time delta comparison outside of the spinlock, might as well trade some precision with shorter spinlock duration, what do you guys think ?

sources/core/Xenko.Core/Threading/ThreadPool.cs
```diff
@@ -55,54 +55,52 @@ public void QueueWorkItem([NotNull] [Pooled] Action workItem)
...
+                        workItem = workItems.Dequeue();
+                        if (workItems.Count == 0)
+                            workAvailable.Reset();
                     }
-                    finally
+                    else if(Utilities.ConvertRawToTimestamp(Stopwatch.GetTimestamp() - lastWork) < maxIdleTime)
                     {
-                        if (lockTaken)
-                            spinLock.Exit(true);
+                        aliveCount++;
```
xen2: aliveCount--

xen2:
@Eideren 
If you move the comparison and aliveCount-- outside of the spinLock, it will go back to previous situation with the race condition.
I suppose you have something slightly different in mind? (i.e. release lock to compute and take the look again if it looks like we are really really exiting and do the test again?)

Eideren:
I wasn't as clear as I should have, [here's what I meant in my previous message](https://github.com/xenko3d/xenko/pull/302/commits/f8f8ac0baa1fb63464c1b676c50e1e6f3495e1c6#diff-af5fab9ef0e39887467591292ac7076eR64)

xen2:
Merged, thanks!

Issue #273, opened 2018-11-21
View culling of sprites broken if the sprite is within a certain rotation range


justStand:
**Release Type**: Official Release

**Version**: 3.0.0.6

**Platform(s)**: Windows

**Describe the bug**

If the rotation is set to -1 radians, whenever the center of the sprite goes off screen, the entire sprite will stop being rendered. 

**To Reproduce**
1. Start with default project. 
2. Set camera projection to orthographic.
3. Set position to 0,0,5
4. Set rotation to 0,0,0
5. Remove the plane / sphere.
6. Create a new sprite. Add any texture that you want to it.
7. Set the rotation in the inspector to 0, 0, -57.2958
8. Run the project and use the camera controls q / e to get the sprite towards the edge of the view port.

**Expected behavior**
Sprite should be rendered if partially visible.

**Screenshots**
![rotbug](https://user-images.githubusercontent.com/6134159/48846191-52adf200-ed53-11e8-9109-57a91fd5deef.gif)

justStand:
Here's a gif demonstrating the issue a bit better.

![rotbug2](https://user-images.githubusercontent.com/6134159/49536028-70627780-f87a-11e8-96be-49d7b2dff6bf.gif)

dklompmaker:
I'll take this one. It should be relatively easy to solve.

PR #322, opened 2019-01-11
Issue #273 sprite culling

dklompmaker:
This fixes the issue #273. It also provides two added helper math functions for `params float[]` args of min & max. Along with an optimization to only calculate bounding box upon changes to worldMatrix or sprite size.

### DIFF
--- a/sources/core/Xenko.Core.Mathematics/MathUtil.cs
+++ b/sources/core/Xenko.Core.Mathematics/MathUtil.cs
@@ -651,5 +651,49 @@ namespace Xenko.Core.Mathematics
                 (float)Math.Round((value.Z / gap), MidpointRounding.AwayFromZero) * gap,
                 (float)Math.Round((value.W / gap), MidpointRounding.AwayFromZero) * gap);
         }
+
+        /// <summary>
+        /// Returns the minimum value from the passed params of float.
+        /// </summary>
+        /// <param name="args"></param>
+        /// <returns></returns>
+        public static float Min(params float[] args)
+        {
+            if (args.Length == 0)
+            {
+                return default(float);
+            }
+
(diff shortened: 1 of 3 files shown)


Kryptos-FR:
I don't see the helper methods being used in the changed code.

If not, maybe it is better not include it in that pull-request. I personally prefer to only have changes related to the issue in one pull-request.

For the actual fix @xen2 is the one with the knowledge to review it :wink:

dklompmaker:
@Kryptos-FR The `Max` method is being used on change 36 of RenderSprite. However, you are correct that `Min` isn't being used. I added it as I imagine we'd run into other scenarios where we need to min or max against 3+ numbers at once. Of course these methods aren't required, but are more elegant than say: `Math.Min(num0, Math.Min(num1, Math.Min(num2, num3)));`

@xen2 I'll keep my eye out for any questions or suggestions.

xen2:
Also did a few edit in the caching code.

dklompmaker:
@xen2 Thanks. Fair points about the vararg. My original intention was to cache the `float[]` and pass that in instead of the individuals. However, it's not a requirement for this bug!

xen2:
@dklompmaker I gave this PR a try and my suspicions are confirmed: this code does the same as the old one. I can still repro the bug even after.

I will take a look and try to fix it now that I started debugging it 😄

Kryptos-FR:
Typical triangle inequality situation (https://en.wikipedia.org/wiki/Triangle_inequality).
```
‖x + y‖ ≤ ‖x‖ + ‖y‖
```

xen2:
@Kryptos-FR Yes exactly. So it broke down as soon as there was some rotation.

PR #327, opened 2019-01-18
Added issue templates for bugs and feature requests

indigozero:


Kryptos-FR:
There is already one for bug report though: https://github.com/xenko3d/xenko/blob/master/docs/ISSUE_TEMPLATE.md

xen2:
Yes that old docs/ISSUE_TEMPLATE.md should be removed.

indigozero:
Probably wouldn't be too bad of an Idea, that way I wouldn't have a blank description ;)

indigozero:
Probably would be a good idea for Pull Requests, that is.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,3 +1,12 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: ''
```
xen2: Maybe we should add bug as default label and add another (almost empty) template for "Questions" and/or "Others"?

indigozero: I like the idea of a template for "'Questions' and/or 'Others'". Will automatically adding "bug" as default confuse us as to which issues have been verified as and actual bug as opposed to something like user error?

indigozero: I thought that "Questions and/or Comments" sounded better, but I'd be open to changing it. I also have the template set to auto label with 'question'.

xen2: I would say it's OK if bug is set by default, we can work the opposite way (close issue and/or remove bug tag if it wasn't a bug).

.github/ISSUE_TEMPLATE/questions_request.md
```diff
@@ -0,0 +1,9 @@
+---
+name: Questions and/or Comment
+about: Have a questions that needs to be answered?
```
Kryptos-FR: We might need to inform the user that general purpose questions can also be answered on our forums and discord community. So they could first try there.

Especially questions such as "how to create XXX?" or "how do I use YYY?"

.github/ISSUE_TEMPLATE/bug_report.md
indigozero: bug label added

.github/ISSUE_TEMPLATE/questions_request.md
indigozero: @Kryptos-FR Let me know if my latest commit works for you as a solutions and if any rewording

indigozero:
@xen2 @Kryptos-FR I want to make sure the [pull_request_template](https://github.com/indigozero/xenko/commit/ee58ffecf0b53afd1d80d2c3569da1fae3aa47a5) works for you guys also.

Kryptos-FR:
@indigozero LGTM

.github/ISSUE_TEMPLATE/questions_request.md
```diff
@@ -0,0 +1,12 @@
...
```
xen2: Small typo: questions => question

xen2: Never mind I can edit it inline, I will just do that.

xen2:
LGTM, merging! Thanks!

PR #332, opened 2019-01-19
reverted vs2019 version support from visual studio plugin installer

tebjan:
makes visual studio plugin installer work again

### DIFF
--- a/sources/tools/Xenko.VisualStudio.Package/source.extension.vsixmanifest
+++ b/sources/tools/Xenko.VisualStudio.Package/source.extension.vsixmanifest
@@ -7,7 +7,7 @@
     <Icon>Logo.ico</Icon>
   </Metadata>
   <Installation InstalledByMsi="false">
-    <InstallationTarget Version="[15.0,17.0)" Id="Microsoft.VisualStudio.Community" />
+    <InstallationTarget Version="[14.0,16.0)" Id="Microsoft.VisualStudio.Community" />
   </Installation>
   <Dependencies>
     <Dependency Id="Microsoft.Framework.NDP" DisplayName="Microsoft .NET Framework" d:Source="Manual" Version="[4.7.2,)" />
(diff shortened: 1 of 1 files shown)


tebjan:
there might be a better fix in order to support all versions, this however works if someone needs it.

Kryptos-FR:
@tebjan I don't have any issue with the plugin and VS 2017. Are you sure this revert is needed?

tebjan:
still wondering why it would fail only because i have 2015 on my machine... that might need some investigation.

xen2:
Did you get the "install log"? (I could see there was one in the failure messagebox in your discord screenshot)

tebjan:
yes, here it is:
image
[vs install log.txt](https://github.com/xenko3d/xenko/files/2776827/vs.install.log.txt)

xen2:
Thanks.
What version of Visual Studio 2017 do you have? Community or another one?

xen2:
Maybe you can show me the Visual Studio version selection screen you have with your custom vsix you made? (the one where you can choose which VS to install to)

tebjan:
image

this is current master including this PR:
image

xen2:
OK, last thing: when this vsix is running, you should see a `VSIXInstaller.exe` process running. What's its path?
Also, what happens if you try to run the .vsix directly? (located in `%UserProfile%\.nuget\packages\xenko.visualstudio.package\3.0.3\tools\`)

tebjan:
this is the patch when running the compiled one:
"C:\Program Files (x86)\Microsoft Visual Studio\Installer\resources\app\ServiceHub\Services\Microsoft.VisualStudio.Setup.Service\VSIXInstaller.exe"

when running from xenko launcher i get:
"C:\Program Files (x86)\Microsoft Visual Studio 14.0\Common7\IDE\VSIXInstaller.exe"

interestingly, my .nuget\packages\xenko.visualstudio.package\3.0.3\tools is empty...

xen2:
OK I think the fix is finding the proper VSIXInstaller.
Either use 15+ from now on, or use the global one from Microsoft.VisualStudio.Setup.Service just like when running vsix by hand.

xen2:
Thanks for the investigation.

tebjan:
yes, it also seems to run this one after the first error:
"C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\Common7\IDE\VSIXInstaller.exe"

tebjan:
confirmed, when opening with "C:\Program Files (x86)\Microsoft Visual Studio\Installer\resources\app\ServiceHub\Services\Microsoft.VisualStudio.Setup.Service\VSIXInstaller.exe" it asks for products and only shows 2017 with a checkbox.

tebjan:
yes, that also works.

Issue #331, opened 2019-01-19
dotnet publish doesn't copy proper native/xenko dependencies


xen2:
**Release Type**: GitHub

**Version**: master

**Describe the bug**
Running `dotnet publish`  does not properly copy dependencies.
It seems `dotnet publish` uses slightly different targets than what's done for standard build.
(note: heard it might be unified in .NET Core 3.0)

**To Reproduce**
1. Create a new game with Linux or macOS
2. Build Linux or macOS (might be optional)
3. Run `dotnet publish`

**Expected behavior**
`Bin\Linux\Debug\linux-x64\publish` should contain all the native files (that are pulled by .ssdeps) just like in `Bin\Linux\Debug\linux-x64` but it doesnt.

The fix likely needs to be done in https://github.com/xenko3d/xenko/blob/master/sources/core/Xenko.Core/build/Xenko.Core.targets

xen2:
I think we need a target with `AfterTargets="Publish"`, but not so sure on that.

jazzay:
@xen2 for the build step (optional) are you doing that via **dotnet build**?

xen2:
I did it with VS.
Didn't try `dotnet build`, does it work?

jazzay:
Not sure, did not try yet. Probably publish is similar to dotnet build, so discrepancy might be between VS and dotnet?

xen2:
I think `dotnet build` is OK and discrepency is between `dotnet build` and `dotnet publish`, but please double check.

jazzay:
Explicitly adding this reference to the Windows project fixes the problems above:
```
    <PackageReference Include="Xenko" Version="3.1.0.1-beta01" PrivateAssets="contentfiles;analyzers" />
```
Which makes sense I guess since that Project is where the runtime.json file is. And now I see that the MacOS project contains that already.

jazzay:
If I try to run a Windows published build from cmd line i get this:
```
D:\dev\gamedev\Xenko.TestProjects\TestMacPublishBug\Bin\Windows\Release\win-x64\publish
λ TestMacPublishBug.Windows.exe

[...]

PR #334, opened 2019-01-20
publish - fix native deps  and data folder failing to copy

jazzay:
Here's a solution to #331 

- Adjusts all game projects to support copying native deps properly, and a post publish target to copy the data build folder.
- Adds macOS to the AllPlatforms.bat

Note: probably we can find a solution to the data copy from within Xenko target files. But not sure best approach there. I tried within Xenko.Core.Assets.CompilerApp.targets, but it did not work.

Also: I switched the Windows game project to using NetCore by default. I believe this will help us track down netcore related problems that affect all platforms, like this issue did. Not sure if you agree! :)

### DIFF
--- a/sources/assets/Xenko.Core.Assets.CompilerApp/build/Xenko.Core.Assets.CompilerApp.targets
+++ b/sources/assets/Xenko.Core.Assets.CompilerApp/build/Xenko.Core.Assets.CompilerApp.targets
@@ -135,4 +135,14 @@
       </None>
     </ItemGroup>
   </Target>
+
+  <!--Feels like we could do this here, but it does not appear to work-->
+  <!--Also we only really want this per Game not for each dependent dll-->
+  <!--<Target Name="CopyDataFolder" AfterTargets="Publish">
+    <ItemGroup>  
+        <SourceDataFiles Include="$(OutDir)data\**\*.*"/>  
+    </ItemGroup>  
+    <Copy SourceFiles="@(SourceDataFiles)" DestinationFiles="@(SourceDataFiles->'$(PublishDir)data\%(RecursiveDir)%(Filename)%(Extension)')" />
+  </Target>-->
+
 </Project>
--- a/sources/editor/Xenko.Assets.Presentation/Templates/Core/ProjectExecutable.Linux/$ProjectName$.csproj.t4
+++ b/sources/editor/Xenko.Assets.Presentation/Templates/Core/ProjectExecutable.Linux/$ProjectName$.csproj.t4
@@ -8,6 +8,9 @@
     <OutputType>WinExe</OutputType>
     <RootNamespace><#= Properties.Namespace #></RootNamespace>
 
+    <!-- This resolves native dep publishing -->
+    <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
+
     <OutputPath>..\Bin\Linux\$(Configuration)\</OutputPath>
     <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
 
@@ -20,5 +23,13 @@
     <!-- Needed for custom runtime.json -->
     <PackageReference Include="Xenko" Version="<#= Xenko.Assets.XenkoConfig.GetLatestPackageDependency().Version #>" PrivateAssets="contentfiles;analyzers" />
   </ItemGroup>
+
+  <!-- Would be nice to do this from Xenko targets somehow, but I am unclear how to... -->
+  <Target Name="CopyData" AfterTargets="Publish">
+    <ItemGroup>  
+        <SourceDataFiles Include="$(OutDir)data\**\*.*"/>  
+    </ItemGroup>  
+    <Copy SourceFiles="@(SourceDataFiles)" DestinationFiles="@(SourceDataFiles->'$(PublishDir)data\%(RecursiveDir)%(Filename)%(Extension)')" />
+  </Target>
 
 </Project>
--- a/sources/editor/Xenko.Assets.Presentation/Templates/Core/ProjectExecutable.Windows/$ProjectName$.csproj.t4
+++ b/sources/editor/Xenko.Assets.Presentation/Templates/Core/ProjectExecutable.Windows/$ProjectName$.csproj.t4
@@ -2,10 +2,16 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net461</TargetFramework>
+    <!-- We should switch Windows Game to netCore as well. Would help us find xplatform issues quicker -->
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <RuntimeIdentifier>win-x64</RuntimeIdentifier>
+
     <ApplicationIcon>Resources\Icon.ico</ApplicationIcon>
     <OutputType>WinExe</OutputType>
     <RootNamespace><#= Properties.Namespace #></RootNamespace>
+
+    <!-- This resolves native dep publishing -->
+    <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
 
     <OutputPath>..\Bin\Windows\$(Configuration)\</OutputPath>
     <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
@@ -16,6 +22,16 @@
 
   <ItemGroup>
     <ProjectReference Include="..\<#= Properties.ProjectGameRelativePath #>" />
+    <!-- Needed for custom runtime.json -->
+    <PackageReference Include="Xenko" Version="3.1.0.1-beta01" PrivateAssets="contentfiles;analyzers" />
   </ItemGroup>
+
+  <!-- Would be nice to do this from Xenko targets somehow, but I am unclear how to... -->
+  <Target Name="CopyData" AfterTargets="Publish">
+    <ItemGroup>  
+        <SourceDataFiles Include="$(OutDir)data\**\*.*"/>  
+    </ItemGroup>  
+    <Copy SourceFiles="@(SourceDataFiles)" DestinationFiles="@(SourceDataFiles->'$(PublishDir)data\%(RecursiveDir)%(Filename)%(Extension)')" />
+  </Target>
 
 </Project>
--- a/build/Xenko.AllPlatforms.bat
+++ b/build/Xenko.AllPlatforms.bat
@@ -1,2 +1,2 @@
-set XenkoPlatforms=Windows;UWP;Android;iOS;Linux
+set XenkoPlatforms=Windows;UWP;Android;iOS;Linux;macOS
 Xenko.sln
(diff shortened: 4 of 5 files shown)


sources/editor/Xenko.Assets.Presentation/Templates/Core/ProjectExecutable.Windows/$ProjectName$.csproj.t4
```diff
@@ -2,11 +2,17 @@
...
+    <!-- We should switch Windows Game to netCore as well. Would help us find xplatform issues quicker -->
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <RuntimeIdentifier>win-x64</RuntimeIdentifier>
+
     <ApplicationIcon>Resources\Icon.ico</ApplicationIcon>
     <OutputType>WinExe</OutputType>
     <RootNamespace><#= Properties.Namespace #></RootNamespace>
 
+    <!-- This resolves native dep publishing -->
+    <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
```
xen2: This will copy all assemblies to output folder (in non-publish mode).
Windows is working without this (dotnet resolves assemblies from NuGet and native dependencies are loaded from within NuGet).
With your other `CopyData` fix, is it really necessary?

sources/editor/Xenko.Assets.Presentation/Templates/Core/ProjectExecutable.Linux/$ProjectName$.csproj.t4
```diff
@@ -8,6 +8,9 @@
     <OutputType>WinExe</OutputType>
     <RootNamespace><#= Properties.Namespace #></RootNamespace>
 
+    <!-- This resolves native dep publishing -->
+    <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
```
xen2: This will copy all assemblies to output folder (in non-publish mode).
I think `dotnet` is supposed to work without this (`dotnet` resolves assemblies from NuGet and native dependencies are loaded from within NuGet). At least it does under Windows.
IMHO Linux/macOS needs explicit publish so only publish should be fixed.
So I was wondering, with your other `CopyData` fix, isn't publish fixed already?

sources/editor/Xenko.Assets.Presentation/Templates/Core/ProjectExecutable.Windows/$ProjectName$.csproj.t4
```diff
@@ -2,11 +2,17 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net461</TargetFramework>
+    <!-- We should switch Windows Game to netCore as well. Would help us find xplatform issues quicker -->
```
xen2: Let's keep Windows as is.
It's a big breaking change to convert all users Windows project from .NET Framework to .NET Core.
Note that I am not against the concept. However, we'll likely have to do this in a separate PR with some backward compat options (i.e. offer a .NET Core/Framework dropdown when creating the Windows platform).

sources/assets/Xenko.Core.Assets.CompilerApp/build/Xenko.Core.Assets.CompilerApp.targets
```diff
@@ -135,4 +135,14 @@
       </None>
     </ItemGroup>
   </Target>
+
+  <!--Feels like we could do this here, but it does not appear to work-->
```
xen2: Please remove that from here (it's a batch file, never going to work)

sources/editor/Xenko.Assets.Presentation/Templates/Core/ProjectExecutable.Linux/$ProjectName$.csproj.t4
```diff
@@ -21,4 +24,12 @@
     <PackageReference Include="Xenko" Version="<#= Xenko.Assets.XenkoConfig.GetLatestPackageDependency().Version #>" PrivateAssets="contentfiles;analyzers" />
   </ItemGroup>
 
+  <!-- Would be nice to do this from Xenko targets somehow, but I am unclear how to... -->
+  <Target Name="CopyData" AfterTargets="Publish">
```
xen2: If this works well, we should move it to `sources\core\Xenko.Core\build\Xenko.Core.targets` so that it doesn't pollute user projects.
Otherwise it will be harder to maintain/update since outside of our control (user can remove/tweak it), and Visual Studio default projects won't work out of the box.

sources/editor/Xenko.Assets.Presentation/Templates/Core/ProjectExecutable.Linux/$ProjectName$.csproj.t4
xen2: Sorry, I noticed afterwise `CopyData` is only for data folder.
We should use a similar fix for native libraries. It should probably be located in `Xenko.Core\build\Xenko.Core.targets`. There is already a bunch of targets like `_XenkoSetupNativeLibrariesWindows` and `_XenkoSetupNativeLibrariesUWP` and so on. We probably want a new one `_XenkoSetupNativeLibrariesDotNetPublish`.

sources/editor/Xenko.Assets.Presentation/Templates/Core/ProjectExecutable.Windows/$ProjectName$.csproj.t4
Kryptos-FR: Is there a way to have both configuration target generated, and keep Windows as default?

Something like `<TargetFrameworks>net461;netcoreapp2.1</TargetFrameworks>` should work but I think it builds both targets by default (?), which we might want to avoid.

xen2: I think it's fine to have two separate projects (either simultaneously with an additional Xenko platform or just a dropdown choice for Windows platform).

Otherwise it would lead to lot of complexity: since the project have some different prorperties, there would be lot of Condition depending on the `TargetFramework` (that will get destroyed once people edit their project in VS).
And I don't see much advantages over having explicit simple projects that the user can generate/regenerate at will.

So my opinion: let's keep it simple (no MSBuild target-hell like on Xenko dev side), and similar to what VS generates out of the box.

sources/editor/Xenko.Assets.Presentation/Templates/Core/ProjectExecutable.Linux/$ProjectName$.csproj.t4
jazzay: Yes I missed that this impacts non-publish scenario. That's not good. Will revisit this.

sources/editor/Xenko.Assets.Presentation/Templates/Core/ProjectExecutable.Windows/$ProjectName$.csproj.t4
jazzay: Since this is just the template for new projects this should not break existing projects right? 

My main concern here is that by not having netcore as the default moving forward we do not see/fix issues on that platform (which is key for Mac/Linux) until later, and increases maintenance burden for us. A question perhaps for the Developer meetup is how to move forward? My vote is to go 100% NetCore on platforms that can support that. There's many advantages to that. And with NetCore3 on the horizon there is improvements for GUI/Window apps.

Regarding short term solution: What about a new Platform in the Project creation that is - NetCore, this could then support all 3 desktop platforms via Multi-targeting. Then we leave existing Windows (.NetFramework) alone. How does that sound? Then user can decide, which or both if they want. I think this what you guys are alluding to above.

sources/assets/Xenko.Core.Assets.CompilerApp/build/Xenko.Core.Assets.CompilerApp.targets
jazzay: Left this for discussion only to see your thoughts.

sources/editor/Xenko.Assets.Presentation/Templates/Core/ProjectExecutable.Linux/$ProjectName$.csproj.t4
jazzay: Hmm. This comment appears to conflict with your comment above where I tried to add it to the CompilerApp.targets. Maybe I am confused?

sources/editor/Xenko.Assets.Presentation/Templates/Core/ProjectExecutable.Windows/$ProjectName$.csproj.t4
xen2: Sure, I am also all for switching to .NET Core as soon as possible.
However, this probably warrants a discussion on how/when to do it so I would prefer to keep that out of this PR to be able to merge that one quickly.

sources/editor/Xenko.Assets.Presentation/Templates/Core/ProjectExecutable.Linux/$ProjectName$.csproj.t4
xen2: Sorry for the confusion.
There is two targets to write:
* Copying asset compiler output data (currently called `CopyData`). This should go in `Xenko.Core.Assets.CompilerApp`
* Making .ssdeps (native libs) work. This should go in `Xenko.Core`

Btw, please make both target names starts with either Xenko or _Xenko (depending if public or not, likely _Xenko in your case)

jazzay: After further investigation I can see that this is a custom file format generated from the xenko target file. Will try to figure out how to feed that into a publish copy operation. Any suggestions? Should it be built into an existing C# tool? PackageBuilder for example, new MSBuild Task?

jazzay: Also it feels like we should leverage the built in functionality of nuget/dotnet. Other nuget 3rd party packages accomplish native packaging with no custom code, what is different about Xenko's process?

xen2: Unfortunately there is no built-in functionality.
The point is not to copy but also to set those libs as native lib.
Each platform **had** a different path for that (`AndroidNativeLibrary` for Android, Copy for UWP but need to choose proper CPU, `NativeReference` + `MtouchExtraArgs` for iOS, etc...).
The idea was to hide those differences so that the csproj just add a single `XenkoNativeLibrary` and it automatically do the proper stuff for each platform.
However, this was done some time ago and it's possible there is easier way to do this (highly doubt it though). Ideally at some point I should write a spec and submit PR for every platform (Xamarin.iOS/Android, dotnet and MSBuild), but I think it would be a lot of work!

xen2:
Thanks to your work, I could use some of your research and ideas to make a proper fix.
Closing this PR.

PR #362, opened 2019-02-01
Enable the feature that generates complex convex hulls

EternalTamago:
# PR Details

Add Generator property to ConvexHullColliderShapeDesc instead of some properties.
Update version of ColliderShapeAsset from 2.0.0.0 to 3.0.0.0.

## Description

Xenko can already generate complex convex hulls with VHACD but that is not used in Game Studio.
This exposes it to use in Game Studio.
Computing them will take a lot of time (easily a minute or more by the default).
image

## Related Issue

https://github.com/xenko3d/xenko/issues/81#issuecomment-455246809

## Motivation and Context

We would like to create collider shapes that fit for Model.

## Types of changes

## Checklist


xen2:
Thanks for the PR.
A few suggestions, after checking little bit better how it works.

I had a quick misunderstanding, I thought the VHACD lib was used exclusively in this case, but it seems used even when not doing decomposition.
As a result, it is just a bunch of optional parameters. The point to make an interface was to be able to replace it, but this is not necessary.

Considering this, my proposal would be to:
* Rename type `SimpleConvexHullGenerator` to `DecompositionParameters`
* Rename property `ConvexHullColliderShapeDesc.Generator` to `ConvexHullColliderShapeDesc.DecompositionParameters`
* Remove the interfaces (you can use `DecompositionParameters` type directly)
* Add a `public bool Enabled { get; set; }` in type `DecompositionParameters`
This will make property grid behave nicely with a Enable check box for the whole decomposition.
Then we can use `SimpleHull = convexHullDesc.DecompositionParameters.Enabled`.

Hope that makes sense.
Please tell me what you think.

EternalTamago:
I totally agree.
There is a difference between the name and the function of `Generator`.
About interfaces, I thought that those were needed to be able to select items in Game Studio.
However, I overlooked `Enabled `property.
That will be enough to be able to select them in Game Studio.
I will try it.

Thanks!

xen2:
Merged, thanks!

PR #364, opened 2019-02-03
Fixing 3D text rendering

phr00t:
# PR Details

Fixes rendering of 3D text when not fullscreen or buildboard.

## Description

Instead of just calculating one world view projection matrix at the start of the string, I calculate a world view projection matrix for each glyph. This isn't needed in 2D rendering, so it should check somewhere if it is 3D text or not for performance reasons. There are probably other better ways to go about this, but this fixes the function problem in hopes of a better production solution.

I also have to set RealVirtualResolutionRatio to (1f, 1f), or else scaling is all messed up in 3D. I'm sure this value is useful in 2D somewhere, though, so again -- it needs to know when it is rendering in 3D (so it can just set this Vector2 to (1f, 1f) or in 2D and calculate it).

## Related Issue

https://github.com/xenko3d/xenko/issues/314

## Motivation and Context

3D text simply doesn't work in Xenko currently. VR UIs would be impossible without a fix like this.

## Types of changes

- [ x ] Bug fix (non-breaking change which fixes an issue)
- [ x ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

### DIFF
--- a/sources/engine/Xenko.Graphics/UIBatch.cs
+++ b/sources/engine/Xenko.Graphics/UIBatch.cs
@@ -26,11 +26,7 @@ namespace Xenko.Graphics
         /// <summary>
         /// The view projection matrix that will be used for the current begin/end draw calls.
         /// </summary>
-        private Matrix viewProjectionMatrix;
-
-        public Matrix getViewProjectionMatrix() {
-            return viewProjectionMatrix;
-        }
+        public Matrix viewProjectionMatrix;
 
         // Cached states
         private BlendStateDescription? currentBlendState;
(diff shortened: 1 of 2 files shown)


phr00t:
OK, I think this is "cleaned up" & does some checking to see if it is rendering in 3D or 2D. I made viewProjectionMatrix public within UIBatch to be quicker to access for calculating world view matrices for each glyph. Hoping this is ready for merging, or at least very close.

In other news, I really think the blinking cursor is an existing bug unrelated to my work -- but easily overlooked or worked around by making the "edit text" field no higher than a single line of the font. Although, multi-line edit text fields won't work...

sources/engine/Xenko.Graphics/UIBatch.cs
```diff
@@ -26,7 +26,7 @@ public class UIBatch : BatchBase<UIBatch.UIImageDrawInfo>
         /// <summary>
         /// The view projection matrix that will be used for the current begin/end draw calls.
         /// </summary>
-        private Matrix viewProjectionMatrix;
+        public Matrix viewProjectionMatrix;
```
xen2: Please keep this private and pass it to `SpriteFont.InternalUIDrawGlyph` as a `ref Matrix`

fdrobidoux:
Hello there, feature I was trying to merge since Monday.

phr00t:
Noting that there were more issues relating to 3D text which I moved on to fix in my fork:

https://github.com/phr00t/FocusEngine/issues/84
https://github.com/phr00t/FocusEngine/issues/97

PR #381, opened 2019-02-16
[Localization] Fix build for path containing spaces

Kryptos-FR:
# PR Details

If the path to the project contains any spaces, then the build steps for the localization files will fail.

## Description

Escape all paths in the command line. Special case for`$(TargetDir)` that needs an extra `.` since otherwise the last quote is escaped as well.

## Related Issue

None.

## Motivation and Context

Building from sources will fail for users that have a space in the path where the sources are located. Especially annoying if the space is in the username.

## Types of changes

## Checklist

### DIFF
--- a/docs/technical/editor-localization.md
+++ b/docs/technical/editor-localization.md
@@ -418,7 +418,7 @@ In Xenko's projects file, a command line similar to the following one is used:
 
 ```
 Path=$(MSBuildBinPath)\Roslyn;$(Path)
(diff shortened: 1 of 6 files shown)


dfkeenan:
You should probably also update "\docs\technical\editor-localization.md".

xen2:
Curious, shouldn't it be `&quot;` rather than `"`?

Also, note that if I create a new csproj, it uses `<Target Name="PreBuild" BeforeTargets="PreBuildEvent">` rather than `PostBuildEvent`. Xml values vs attribute might have different escape rules (esp. attribute you can use single quote as main delimiter so that double quote works as is).

dfkeenan:
According to the XML spec `&quot;` is only required for attribute values.

xen2:
@dfkeenan ok thanks for the clarification (not a XML expert).

I will probably merge this PR as is then.

dfkeenan:
I'm no expert either. But I did some googling and everything I found only mentioned `&quot;` in regards to attributes.

There were already `"` in there just not on all file paths. And seeing how it builds (I assume) I don't think it will make things worse. :)

xen2:
Thanks, merged!

PR #416, opened 2019-03-19
fix vulkan texture rendering

phr00t:
# PR Details

Allows textures to render in Vulkan

## Related Issue

https://github.com/xenko3d/xenko/issues/415

## Types of changes

## Checklist


xen2:
Thanks! I'll test and merge once back from GDC in a few days.

xen2:
Just checking, is this ready for review? (I remember seeing on Discord there was some cases still not covered, but not sure if this was before or after this PR)

phr00t:
I've been keeping this pull request up to date with my fixes. The only issue I'm aware of is point lights don't provide shadows in Vulkan. I've had some rare startup crashes, but I haven't noticed them yet with the latest commit (specific bool created yadda yadda). However, I haven't been stress testing it. I'd say this is ready for review.

sources/engine/Xenko.Graphics/Vulkan/CommandList.Vulkan.cs
```diff
@@ -24,6 +24,7 @@ public partial class CommandList
         private int framebufferAttachmentCount;
         private bool framebufferDirty = true;
         private Framebuffer activeFramebuffer;
+        private Texture dummyTexture;
```
xen2: Shouldn't this be part of `GraphicsDevice.Vulkan.cs` instead?

sources/engine/Xenko.Graphics/Vulkan/CommandList.Vulkan.cs
```diff
@@ -42,6 +43,7 @@ public static CommandList New(GraphicsDevice device)
         private CommandList(GraphicsDevice device) : base(device)
         {
             Recreate();
+            dummyTexture = Texture.New2D(GraphicsDevice, 1, 1, PixelFormat.R8G8B8A8_UNorm_SRgb, TextureFlags.ShaderResource);
```
xen2: Shouldn't this be part of `GraphicsDevice.Vulkan.cs` instead?

xen2: Also, need to add a line to destroy it when `GraphicsDevice` is destroyed.

sources/engine/Xenko.Graphics/Vulkan/CommandList.Vulkan.cs
phr00t: Yes, it should. Working on another issue with Vulkan (https://github.com/phr00t/xenko/issues/49), so I'm not able to get to this right at the moment.

sources/engine/Xenko.Graphics/Vulkan/CommandList.Vulkan.cs
phr00t: Yes, it should. Working on another issue with Vulkan (https://github.com/phr00t/xenko/issues/49), so I'm not able to get to this right at the moment.

sources/engine/Xenko.Graphics/Vulkan/CommandList.Vulkan.cs
phr00t: Did it when I was trying to fix that other issue. Help there would be greatly appreciated.

sources/engine/Xenko.Graphics/Vulkan/CommandList.Vulkan.cs
phr00t: This should be done.

sources/engine/Xenko.Graphics/Vulkan/GraphicsDevice.Vulkan.cs
```diff
@@ -471,6 +478,7 @@ private unsafe void ReleaseDevice()
             // Release fenced resources
             nativeResourceCollector.Dispose();
             DescriptorPools.Dispose();
+            if( dummyTexture != null ) dummyTexture.Dispose();
```
xen2: Space around if.
Please split it into two lines.

sources/engine/Xenko.Graphics/Vulkan/PipelineState.Vulkan.cs
```diff
@@ -9,6 +9,7 @@
 using Xenko.Core.Collections;
 using Xenko.Core.Serialization;
 using Xenko.Shaders;
+using System.Threading;
```
xen2: Usually `using` are ordered (System then custom such as Xenko, alphabetically is a bonus then)

sources/engine/Xenko.Graphics/Vulkan/CommandList.Vulkan.cs
```diff
@@ -424,6 +422,9 @@ public unsafe void ResourceBarrierTransition(GraphicsResource resource, Graphics
                     texture = texture.ParentTexture;
 
                 // TODO VULKAN: Check for change
+                
+                // wait, are we ready to use this texture yet?
+                if (texture.ReadyForSampling == false) texture = GraphicsDevice.dummyTexture;
```
xen2: Please split into two lines.

sources/engine/Xenko.Graphics/Vulkan/GraphicsDevice.Vulkan.cs
```diff
@@ -85,6 +85,27 @@ public BufferInfo(long fenceValue, SharpVulkan.Buffer buffer, DeviceMemory memor
...
+        /// <summary>
+        /// A minimal 1x1 texture used while bigger textures are loading.
+        /// </summary>
+        public Texture dummyTexture
+        {
+            get
+            {
+                while( _dummyTexture.NativeImageView == ImageView.Null ) 
+                {
+                    Thread.Sleep(1); // wait until this dummy texture is ready to be used
```
xen2: Why is this necessary?
In which case would it be accessed before being ready?
Having a while/sleep loop looks like a recipe for disaster (deadlock, etc.)

`dummyTexture` should be a simple private field and have deterministic lifetime (initialization/usage/destruction).

phr00t: It doesn't seem to be necessary on NVIDIA hardware, but it appears to be on AMD. I suspect even dummyTexture tries to get used before it is fully initialized on AMD hardware, leading to a memory access violation. This sleep holds off returning it until dummyTexture is ready, so a memory access violation should never happen on it. If you can think of a better way of handling this, please let me know... but I'm going with what works. Getting Vulkan to work on different hardware and driver implementations is dirty business, apparently.

xen2: Not sure how this has anything to do with GPU.

The initialization is done on CPU side with code `dummyTexture = Texture.New2D(this, 1, 1, PixelFormat.R8G8B8A8_UNorm_SRgb, TextureFlags.ShaderResource);`

If nothing is using it before that, this while should never be used.
If it us used before that, it means the initialization should be done earlier.

phr00t: The crash isn't from a null dummyTexture, but a memory access violation likely from dummyTexxture.NativeImageView == null. Other textures were not having a NativeImageView set soon enough before sampling (which is why we have a dummyTexture in the first place) and it appears on AMD that dummyTexture's NativeImageView was null for too long after its CPU initialization. Hard to verify this, though, my main development machine is Intel / NVIDIA. After making this change, things worked on different AMD test machines -- so this is all the theory on why this fix worked.

xen2: `Texture.New2D` should call `InitializeFromImpl` which should set the `NativeImageView`, so `NativeImageView` should be set as soon as texture is created.

I think this specific problem (memory access violation) is very likely due to something else (i.e. CPU/GPU sync or async resource allocation too late/freeing too early) and should be fixed very differently.

phr00t: If NativeImageView is set as soon as the texture is created, then this Thread.Sleep should never be called, and behavior wouldn't have changed due to this, and AMD hardware would still crash.

The only other change that could plausibly explain working AMD hardware is checking if VK_NV_external_memory_capabilities exists before requesting it, which is an OpenVR-required Vulkan extension. It should exist in the latest AMD drivers (although it didn't in old AMD drivers) & add it just as if it was always being added in older (and still crashing) Xenko builds...

phr00t: Hold up.... VK_NV_external_memory_capabilities

I'm guessing "NV" in that extension is referring to NVIDIA. Which means AMD wouldn't have it...

xen2: Yes, `VK_NV_*` extensions are NVIDIA only

xen2:
Closing, will need to break this down into smaller PR.

Issue #259, opened 2018-11-11
Color math should ideally always be in floats


aaronfranke:
**Version**: All, master

**Platform(s)**: All, but mostly affects newer devices

Color math should ideally be done with `float` to support arbitrary precision. It's somewhat common to have 10-bpc and 12-bpc displays, and HDR color math often needs 16 bits per channel. `float` aka Single precision floats have 23 bits of significant binary digits which is plenty for colors.

Floats also avoid the problem of clipping, as colors above 1 still work fine. Even if they clip when displayed, if the brightness is reduced for the whole image, the brightness differences in bright areas is preserved when using floats.

Floats are also more friendly to the user. It's easier to understand that "0.5" is 50% strength than "127".

Xenko already has `Color3` and `Color4` classes which already work with floats. However, there's not much point to `Color3` in my opinion. You can just have constructors and methods that take the alpha value as an optional parameter for use cases that don't need it. 

If it were up to me, I'd do the following:

1. Modify `Color4` to make the alpha value optional. Consider deprecating `Color3`. This isn't a breaking change.

2. Rename `Color` to `Color8` and consider deprecating it. This is a breaking change.

3. Rename `Color4` to `Color`. This is a breaking change.

These don't all necessarily have to be done at once. In fact it may be preferable to leave a time gap between 2 and 3 to avoid people using "Color" in their code and having the implementation change without them knowing.

Obviously it's the Xenko team's decision of what to do, but I do think that something should be done, since color math being done with bytes is not ideal. Having a "Color" class with nothing after the name implies that it's the recommended / default choice.

tebjan:
hi @aaronfranke the low bit/less channel color types match the pixel format for different texture formats. so if you need to create packed memory that you can copy into a texture you need those color types since they don't need a conversion step to match the format. also many operations on bytes are much faster than with floats. so the color types make sense and i would keep them...

Qibbi:
An issue would also be that the suggested Color8 uses the number to denote the bits per channel, while Color4 uses the number to denote the number of channels. This would result in an inconsistent naming scheme.
Additionally the current naming is what people who come from other engines expect them to be.

Kryptos-FR:
> Modify Color4 to make the alpha value optional. Consider deprecating Color3. This isn't a breaking change.

[...]

PR #418, opened 2019-03-21
Color improvements (VertexPositionNormalColor fix & optional Color4.A)

phr00t:
# PR Details

Color is used improperly in VertexPositionNormalColor, when it should be Color4. This is a breaking change, but VertexPositionNormalColor was already broken. To also promote an easier, unified color structure (like that used in VertexPositionNormalColor), Color4's alpha argument is now optional.

## Related Issue

https://github.com/xenko3d/xenko/issues/259

## Motivation and Context

Unity has a float-operated Color structure with optional alpha which has been problem-free in thousands of use cases. Xenko should have one too. Ideally Color4 would be the default "Color", but Color (with byte parameters) is already in use in many places, unfortunately.

## Types of changes

## Checklist

### DIFF
--- a/sources/core/Xenko.Core.Mathematics/Color4.cs
+++ b/sources/core/Xenko.Core.Mathematics/Color4.cs
@@ -92,7 +92,7 @@ namespace Xenko.Core.Mathematics
         /// <param name="green">The green component of the color.</param>
         /// <param name="blue">The blue component of the color.</param>
         /// <param name="alpha">The alpha component of the color.</param>
-        public Color4(float red, float green, float blue, float alpha)
+        public Color4(float red, float green, float blue, float alpha = 1f)
         {
             R = red;
             G = green;
@@ -117,7 +117,7 @@ namespace Xenko.Core.Mathematics
         /// </summary>
         /// <param name="value">The red, green, and blue components of the color.</param>
         /// <param name="alpha">The alpha component of the color.</param>
-        public Color4(Vector3 value, float alpha)
+        public Color4(Vector3 value, float alpha = 1f)
         {
             R = value.X;
             G = value.Y;
@@ -159,13 +159,13 @@ namespace Xenko.Core.Mathematics
         {
             if (values == null)
                 throw new ArgumentNullException(nameof(values));
-            if (values.Length != 4)
-                throw new ArgumentOutOfRangeException(nameof(values), "There must be four and only four input values for Color4.");
+            if (values.Length < 3)
+                throw new ArgumentOutOfRangeException(nameof(values), "There must be at least 3 float[] values for Color4.");
 
             R = values[0];
             G = values[1];
             B = values[2];
-            A = values[3];
+            A = values.Length >= 4 ? values[3] : 1f;
         }
 
         /// <summary>
@@ -209,7 +209,7 @@ namespace Xenko.Core.Mathematics
         /// </summary>
         /// <param name="color"><see cref="Color3"/> used to initialize the color.</param>
         /// <param name="alpha">The alpha component of the color.</param>
-        public Color4(Color3 color, float alpha)
+        public Color4(Color3 color, float alpha = 1f)
         {
             R = color.R;
             G = color.G;
--- a/sources/core/Xenko.Core.Mathematics/Color4.cs
+++ b/sources/core/Xenko.Core.Mathematics/Color4.cs
@@ -159,8 +159,8 @@ namespace Xenko.Core.Mathematics
         {
             if (values == null)
                 throw new ArgumentNullException(nameof(values));
-            if (values.Length < 3)
-                throw new ArgumentOutOfRangeException(nameof(values), "There must be at least 3 float[] values for Color4.");
+            if (values.Length != 3 && values.Length != 4)
+                throw new ArgumentOutOfRangeException(nameof(values), "There must be 3 or 4 float[] values for Color4.");
 
             R = values[0];
             G = values[1];
--- a/sources/editor/Xenko.Assets.Presentation/AssetEditors/EntityHierarchyEditor/Game/EditorGameLightProbeGizmoService.cs
+++ b/sources/editor/Xenko.Assets.Presentation/AssetEditors/EntityHierarchyEditor/Game/EditorGameLightProbeGizmoService.cs
@@ -326,7 +326,7 @@ namespace Xenko.Assets.Presentation.AssetEditors.EntityHierarchyEditor.Game
             var vertices = new VertexPositionNormalColor[lightProbeRuntimeData.LightProbes.Length];
             for (var i = 0; i < lightProbeRuntimeData.LightProbes.Length; i++)
             {
-                vertices[i] = new VertexPositionNormalColor(lightProbeRuntimeData.Vertices[i], Vector3.Zero, Color.White);
+                vertices[i] = new VertexPositionNormalColor(lightProbeRuntimeData.Vertices[i], Vector3.Zero, Color4.White);
             }
 
             // Generate data for index buffer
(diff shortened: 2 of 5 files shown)


sources/core/Xenko.Core.Mathematics/Color4.cs
```diff
@@ -159,13 +159,13 @@ public Color4(float[] values)
         {
             if (values == null)
                 throw new ArgumentNullException(nameof(values));
-            if (values.Length != 4)
-                throw new ArgumentOutOfRangeException(nameof(values), "There must be four and only four input values for Color4.");
+            if (values.Length < 3)
+                throw new ArgumentOutOfRangeException(nameof(values), "There must be at least 3 float[] values for Color4.");
```
aaronfranke: Seems like a bad idea to allow non-3 non-4 size arrays.

phr00t: Made the change. My original thought was, there was enough values for the function to work, so I'd let the function work.

phr00t:
Updated the pull request to only allow float of sizes 3 & 4. For what it is worth, I believed developers might find uses for having bigger float[] arrays with the first 4 being color values -- hard to say larger arrays being passed would always be considered a bug that required an exception.

aaronfranke:
`Color4` looks good, but you should also squash the commits to clean the history. `git rebase -i HEAD~2`

xen2:
Except for HDR (>1.0f) colors, is there any reason to not use `Color`?
If there is some issues where it's wrongly interpreted (255 in byte should be 1.0f in float), that's another issue that should be fixed.
`Color` is much more compact. 4 byte instead of 16 byte, which could start to matter a lot of large meshes, i.e. 1 millions vertices => 12 additional MB.

phr00t:
VertexPositionNormalColor used in conjunction with a vertex color stream shader is where the color values are misinterpreted. If you make the shader (or wherever the misinterpretation is happening) accept 0-255 values, we will have to change VertexPositionNormalColor back to using Color (which does have memory benefits).

xen2:
`Color` should use a `R8G8B8A8_UNorm` semantic which means 255 in memory means 1.0f in the shader.

Taken from https://docs.microsoft.com/en-us/windows/desktop/api/dxgiformat/ne-dxgiformat-dxgi_format#format-modifiers

_UNORM | Unsigned normalized integer; which is interpreted in a resource as an unsigned integer, and is interpreted in a shader as an unsigned normalized floating-point value in the range [0, 1]. All 0's maps to 0.0f, and all 1's maps to 1.0f. A sequence of evenly spaced floating-point values from 0.0f to 1.0f are represented. For instance, a 2-bit UNORM represents 0.0f, 1/3, 2/3, and 1.0f.
-- | --

If you receive 255.0f in the shader, it means something is wrong and should be fixed elsewhere.
Is that what you saw? If yes, was it when trying Vulkan, or does it also happen with D3D?

phr00t:
I think the root of the problem is here: https://github.com/xenko3d/xenko/blob/master/sources/engine/Xenko.Rendering/Rendering/Materials/ComputeColors/ComputeColor.cs#L91

There is no consideration for a "Color" object with byte values, only float structures (e.g. Color3, Color4, Vector3...)

xen2:
On the shader side it's always going to be exposed float3 or float4.

How it is on CPU side is a different matter. The `VertexDeclaration` describes how values uploaded in GPU memory should be interpreted when read from a shader.

To explain again, in memory you will have 0xFF00FFFF with `D3D_FORMAT_R8G8B8A8_UNorm`  and on GPU you will receive it as `float4(1.0f, 0.0f, 1.0f, 1.0f)`. This has always been working well (at least in general when using D3D/OpenGL, need to double check Xenko -- it might be a Vulkan only issue too, due to incorrect API usage).

xen2:
Link to a simple support example in C#:
https://github.com/sharpdx/SharpDX-Samples/blob/ed7e80cdcac145cf1864c6b9358a558e5fe8deb6/StoreApp/OldSamplesToBeBackPorted/SimpleMiniCube/Program.cs#L119

Some other interesting link: https://github.com/Microsoft/DirectXTK/wiki/VertexTypes#visual-studio-content-pipeline

phr00t:
This appears to happen with D3D too -- not just a Vulkan problem. Tried running my test app without a win-vulkan RuntimeIdentifier & saw bad colors...

xen2:
@phr00t Could you provide a small repro example?

phr00t:
Ideally yes, I'd make a repro example... but my current "example" is my main project. Hoping to solve this without making a separate project... where is the vertex data packed & passed into a shader...? I'm guessing that is where the problem lies, but haven't been able to track down that spot yet...

phr00t:
It looks like other Vertex structures have been storing colors as Color4, take for example: https://github.com/xenko3d/xenko/blob/master/sources/engine/Xenko.Graphics/VertexPositionColorTextureSwizzle.cs#L41

... when others are not: https://github.com/xenko3d/xenko/blob/master/sources/engine/Xenko.Graphics/VertexPositionColorTexture.cs#L37

It seems Color4 is the one that "works." All of these vertex structures need to switch to "Color" (to get memory benefits) or they need to be "Color4" (to maintain float precision and currently work properly). The inconsistency is probably what led to this issue.

I'm moving ahead with my project using Color4, because it works and considering other Vertex objects used Color4, I'm not sure it was originally designed to use anything but Color4. It is probably a good idea to convert the system to use Color in these structures, but that is a bit deeper than I want to go at the moment.

aaronfranke:
@xen2 In addition to HDR, `Color4` also has much more precision, so it works well for high-color displays such as 10-bpc and 12-bpc.

xen2:
Yes it could be useful in some (rare) cases and I wouldn't mind adding a separate Vertex type that contains a float color.

Why is it rarely useful?
In shaders, interpolation is done with floats already (flat gradient would look nice even if input is in RGB 8 bit, intermediate colors between vertices would be linearly interpolated using 10 or 12-bpc if the render target is in this format).
Also, when doing PBR most maps are in [0..1] range (except emissive)
So from the top of my head, float colors are useful only for very limited cases such as emissive color in HDR, which is a quite limited case. Or maybe if doing some kind of custom rendering (but then it's usually more "custom vertex data" than colors).

With that being said, we still need to fix the underlying problem with `Color`, which should work fine (we should always end up with a 1.0f in shader, never a 255.0f).
And then we could see about maybe adding a `VertexPositionNormalColor` with float colors if there is a real need/use case.

aaronfranke:
> In shaders, interpolation is done with floats already

Wouldn't this mean that the code wouldn't have to do as many conversions if it uses floats everywhere?

I completely understand that it's bad to use more space without benefit, but also one reason I opened #259 is that the "default" Color struct should be using floats since it's capable for all situations. Saving into RGB(A) 8-bpc should only be for saving and not for intermediate code IMO.

I thought I'd mention, in Godot, there is only one `Color` struct that uses floats, but you can also save the color as four 8-bit values packed into an `int` or four 16-bit values packed into a `long`. I don't know if there's actually a difference, but it might be better for "long-term" storage to save as a simple `int` rather than a full struct. (Xenko already has `int` packing methods, but not `long` ones, and I'm not aware of how often they're used in the code).

xen2:
@aaronfranke Yes agree with what you said.
IMHO `VertexPositionNormalColor`-like types map how data is uploaded to GPU so those should have the compact format. For the same reason, various other values could be in a more compact format in those types (i.e. float16 instead of float32, etc.).

Nothing prevents you from using `Color4` in your scripts/computations (it's probably recommended as soon as you need accurate intermediate computations) and then use the implicit conversion to `Color` when filling GPU-buffers.

xen2:
I cherry-picked in `master` the commits related to making `Color4.A` optional in constructors.
Thanks!

Won't merge the rest for now, so closing this PR.

Issue #420, opened 2019-03-24
Could not find a compatible version of MSBuild: File Not found


boriscallens:
**Release Type**: Official Release
**Version**: 3.1.0.1-beta02-05432
**Platform(s)**: Windows 10

**Describe the bug**
When pressing the big red Start 3.1.0.1-beta02-05432 buttton in the xenko launcher a dialogue " Could not find a compatible version of MSBuild" prevents me from going any further. 

**To Reproduce**
1. Fresh machine install.
2. Installed and used successfuly for work VS17 (and vs2019 preview, vscode):
  - Visual Studio core editor
  - .NET desktop development
  - Universal Windows Platform development
  - ASP.NET and web development
  - Azure development
  - .NET Core cross-platform development
3. Installed Xenko from homepage (http://xenko.com/files/XenkoSetup.exe)
4. Launched Xenko Launcher
5. Pressed the big red button

**Expected behavior**
Xenko Game Editor launches

**Screenshots**
If applicable, add screenshots to help explain your problem (use GitHub drag & drop)

**Log and callstacks**
  > Could not find a compatible version of MSBuild.
  > Check that you have a valid installation with the required workloads, or go to
  > www.visualstudio.com/downloads to install a new one.
  > System.IO.FileNotFoundException: Could not load file or assembly
  > 'Microsoft.Build, Version=15.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
  > or one of its dependencies.
  > The system cannot find the file specified. File name: 
  > 'Microsoft.Build, Version=15.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
  > at Xenko.Core.Assets.PackageSessionPublicHelper.CheckMSBuildToolset()
  > at Xenko.GameStudio.Program.<Startup>d__12.MoveNext()
  > in D:\BuildAgent\work\8cafb933a54b01d0\sources\editor\Xenko.GameStudio\Program.cs:line 241
  > WRN: Assembly binding logging is turned OFF etc..

**Additional context**
The file version at C:\Program Files (x86)\Microsoft Visual Studio\2017\Professional\MSBuild\15.0\Bin\msbuild.exe is 15.9.21.664

Seemingly related issue: https://github.com/xenko3d/xenko/issues/47

boriscallens:
Anything else I can test? Not sure how I can make a minimal reproduction. Maybe with a VM?

erictuvesson:
[...]

PR #421, opened 2019-03-24
Assets Compiler Support MSBuild Toolset Version 16.0 

erictuvesson:
# PR Details

Added support for more MSBuild Toolset Versions.

## Related Issue

#420

## Motivation and Context

MSBuild Toolset Version 15.0 is not available anymore.

## Types of changes

## Checklist

### DIFF
--- a/sources/assets/Xenko.Core.Assets/PackageSessionPublicHelper.cs
+++ b/sources/assets/Xenko.Core.Assets/PackageSessionPublicHelper.cs
@@ -65,7 +65,7 @@ namespace Xenko.Core.Assets
             {
                 if (!projectCollection.Toolsets.Any(x => new Version(x.ToolsVersion).Major >= 15))
                 {
-                    throw new InvalidOperationException("Could not find a supported MSBuild toolset version");
+                    throw new InvalidOperationException("Could not find a supported MSBuild toolset version (expected 15.0 or later)");
                 }
             }
         }
--- a/sources/assets/Xenko.Core.Assets/PackageSessionPublicHelper.cs
+++ b/sources/assets/Xenko.Core.Assets/PackageSessionPublicHelper.cs
@@ -65,9 +65,11 @@ namespace Xenko.Core.Assets
             {
                 if (projectCollection.Toolsets.Any(x => new Version(x.ToolsVersion).Major >= 15))
                 {
-                    throw new InvalidOperationException("Could not find a supported MSBuild toolset version");
+                    return;
                 }
             }
+
+            throw new InvalidOperationException("Could not find a supported MSBuild toolset version");
         }
     }
 }
--- a/sources/assets/Xenko.Core.Assets/PackageSessionPublicHelper.cs
+++ b/sources/assets/Xenko.Core.Assets/PackageSessionPublicHelper.cs
@@ -63,13 +63,11 @@ namespace Xenko.Core.Assets
             // Check that we can create a project
             using (var projectCollection = new Microsoft.Build.Evaluation.ProjectCollection())
             {
-                if (projectCollection.Toolsets.Any(x => new Version(x.ToolsVersion).Major >= 15))
+                if (!projectCollection.Toolsets.Any(x => new Version(x.ToolsVersion).Major >= 15))
                 {
-                    return;
+                    throw new InvalidOperationException("Could not find a supported MSBuild toolset version");
                 }
             }
-
-            throw new InvalidOperationException("Could not find a supported MSBuild toolset version");
         }
     }
 }
--- a/sources/assets/Xenko.Core.Assets/PackageSessionPublicHelper.cs
+++ b/sources/assets/Xenko.Core.Assets/PackageSessionPublicHelper.cs
@@ -71,7 +71,7 @@ namespace Xenko.Core.Assets
             {
                 if (s_msBuildToolVersions.Select(ver => projectCollection.GetToolset(ver)).Any())
                 {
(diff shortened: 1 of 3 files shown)


sources/assets/Xenko.Core.Assets/PackageSessionPublicHelper.cs
```diff
@@ -64,8 +69,10 @@ private static void CheckMSBuildToolset()
             // Check that we can create a project
             using (var projectCollection = new Microsoft.Build.Evaluation.ProjectCollection())
             {
-                if (projectCollection.GetToolset("15.0") == null)
-                    throw new InvalidOperationException("Could not find MSBuild toolset 15.0");
+                if (s_msBuildToolVersions.Select(ver => projectCollection.GetToolset(ver)).Any())
```
wgraham17: Is this check correct? Seems like it should be checking if there are *not* any.

erictuvesson: I had the same after thought before I went to bed last night, just wasn't sure.

I will correct it soon.

sources/assets/Xenko.Core.Assets/PackageSessionPublicHelper.cs
```diff
@@ -64,8 +69,10 @@ private static void CheckMSBuildToolset()
             // Check that we can create a project
             using (var projectCollection = new Microsoft.Build.Evaluation.ProjectCollection())
             {
-                if (projectCollection.GetToolset("15.0") == null)
-                    throw new InvalidOperationException("Could not find MSBuild toolset 15.0");
+                if (!s_msBuildToolVersions.Select(ver => projectCollection.GetToolset(ver)).Any())
```
xen2: At this point, to be futureproof, shouldn't we just use `projectCollection.Toolsets` and accept if any of them is higher or equal to 15.0?
Something like `projectCollection.Toolsets.Any(x => new Version(x.ToolsVersion).Major >= 15)` (note: might need some testing to check what ToolsVersion contains exactly)

erictuvesson: I will look at it tonight. 👍

erictuvesson: I changed the code and verified that the format is "15.0" and using Version works fine.

sources/assets/Xenko.Core.Assets/PackageSessionPublicHelper.cs
```diff
@@ -65,9 +65,11 @@ private static void CheckMSBuildToolset()
             {
                 if (projectCollection.Toolsets.Any(x => new Version(x.ToolsVersion).Major >= 15))
```
xen2: If you don't mind, I would prefer the previous form to keep the flow simpler to understand, and easier to put additional checks following this one later.
You can and add a ! to negate the `Any`.

sources/assets/Xenko.Core.Assets/PackageSessionPublicHelper.cs
```diff
@@ -63,13 +63,11 @@ private static void CheckMSBuildToolset()
             // Check that we can create a project
             using (var projectCollection = new Microsoft.Build.Evaluation.ProjectCollection())
             {
-                if (projectCollection.Toolsets.Any(x => new Version(x.ToolsVersion).Major >= 15))
+                if (!projectCollection.Toolsets.Any(x => new Version(x.ToolsVersion).Major >= 15))
```
xen2: Cool!
Thinking about it, since it's a string, I wouldn't want it to break the second they add a semantic name or use something that's not a proper version (`new Version()` throws an exception if invalid input).

What about using `PackageVersion.TryParse(x.ToolsVersion, out var version) && version.Version.Major >= 15` as condition instead?

sources/assets/Xenko.Core.Assets/PackageSessionPublicHelper.cs
```diff
@@ -64,8 +63,10 @@ private static void CheckMSBuildToolset()
             // Check that we can create a project
             using (var projectCollection = new Microsoft.Build.Evaluation.ProjectCollection())
             {
-                if (projectCollection.GetToolset("15.0") == null)
-                    throw new InvalidOperationException("Could not find MSBuild toolset 15.0");
+                if (!projectCollection.Toolsets.Any(x => new Version(x.ToolsVersion).Major >= 15))
+                {
+                    throw new InvalidOperationException("Could not find a supported MSBuild toolset version");
```
xen2: Let's add detail about expected version (15.0 or later)

erictuvesson: Would this be good?
"Could not find a supported MSBuild toolset version (expected 15.0 or later)"

xen2: Yes, thanks

phr00t:
Is this ready to be pulled in? If so, I may proactively pull it into my fork

erictuvesson:
I have tested it locally and it works for me. I installed VS2017 Build tools so the old also works for me, but I use this when I debug the asset compiler.

xen2:
I have done some testing and ended up adding two commits. Should be OK to merge.

xen2:
Thanks @erictuvesson !

myxibrium:
Thanks a ton for fixing this! As this is blocking me from installing Xenko, I would like to know when this version is available via the Launcher, or if there's some other way I can install it. Thanks!

PR #424, opened 2019-03-26
Physics improvements

phr00t:
# PR Details

Improves the ability for Xenko to find libbulletc.dll

Also adds important helper functions to setting a physical position, which also resulted in helper functions for getting world information for transforms.

## Motivation and Context

Setting positions and working with transforms is too unintuitive, making developers come up with weird hacks to do very common things. This provides developers an out-of-the-box method for doing common things. 

## Types of changes

## Checklist

### DIFF
--- a/sources/core/Xenko.Core.Mathematics/Matrix.cs
+++ b/sources/core/Xenko.Core.Mathematics/Matrix.cs
@@ -746,6 +746,17 @@ namespace Xenko.Core.Mathematics
         }
 
         /// <summary>
+        /// Decomposes a matrix into a scale.
+        /// </summary>
+        /// <param name="scale">When the method completes, contains the scale component of the decomposed matrix.</param>
+        public void GetScale(out Vector3 scale) {
+            //Scaling is the length of the rows.
+            scale.X = (float)Math.Sqrt((M11 * M11) + (M12 * M12) + (M13 * M13));
+            scale.Y = (float)Math.Sqrt((M21 * M21) + (M22 * M22) + (M23 * M23));
+            scale.Z = (float)Math.Sqrt((M31 * M31) + (M32 * M32) + (M33 * M33));
+        }
+
+        /// <summary>
         /// Decomposes a matrix into a rotation.
         /// </summary>
         /// <param name="q">When the method completes, contains the rotation component of the decomposed matrix.</param>
--- a/sources/core/Xenko.Core.Mathematics/Matrix.cs
+++ b/sources/core/Xenko.Core.Mathematics/Matrix.cs
@@ -694,7 +694,7 @@ namespace Xenko.Core.Mathematics
         /// Decomposes a matrix into a scale, rotation, and translation.
         /// </summary>
         /// <param name="scale">When the method completes, contains the scaling component of the decomposed matrix.</param>
-        /// <param name="rotation">When the method completes, contains the rtoation component of the decomposed matrix.</param>
+        /// <param name="rotation">When the method completes, contains the rotation component of the decomposed matrix.</param>
         /// <param name="translation">When the method completes, contains the translation component of the decomposed matrix.</param>
         /// <remarks>
         /// This method is designed to decompose an SRT transformation matrix only.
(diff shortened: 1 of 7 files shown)


phr00t:
"New" math functions are just copied & pasted from existing math functions -- just the math needed for what is being asked is done.

sources/core/Xenko.Core.Mathematics/Matrix.cs
```diff
@@ -707,11 +707,98 @@ public bool Decompose(out Vector3 scale, out Quaternion rotation, out Vector3 tr
...
+            matrix.Backward = at;
+
+            return true;
+        }
+
+        /// <summary>
+        /// Decomposes a matrix into a scale.
+        /// </summary>
+        /// <param name="scale">When the method completes, contains the scale component of the decomposed matrix.</param>
+        public void GetScale(out Vector3 scale) {
```
xen2: We already have `ScaleVector` which does something different.
This might be confusing to have both like this. Minimum would be to find different names for each.

sources/core/Xenko.Core.Mathematics/Matrix.cs
```diff
@@ -707,11 +707,98 @@ public bool Decompose(out Vector3 scale, out Quaternion rotation, out Vector3 tr
...
         }
 
+        /// <summary>
+        /// Decomposes a matrix into a rotation.
+        /// </summary>
+        /// <param name="matrix">When the method completes, contains the rotation component of the decomposed matrix.</param>
+        /// <remarks>
+        /// This method is designed to decompose an SRT transformation matrix only.
+        /// </remarks>
+        public bool GetRotationMatrix(out Matrix matrix)
```
xen2: We have `Matrix.Decompose`.
I suppose the reason to add one that deals with rotation or scaling only is to avoid computing the other parts?

If yes, maybe we should at least rename it in a similar way (i.e. `DecomposeRotation` and `DecomposeScale`)?

sources/editor/Xenko.Assets.Presentation/Templates/Assets/Scripts/Physics/DebugShapesRender.cs
```diff
@@ -35,23 +34,23 @@ public override async Task Execute()
...
             var simulation = this.GetSimulation();
-            if (simulation != null)
-                simulation.ColliderShapesRenderGroup = RenderGroup;
 
             var enabled = false;
-            while (Game.IsRunning)
-            {
-                if (Input.IsKeyDown(Keys.LeftShift) && Input.IsKeyDown(Keys.LeftCtrl) && Input.IsKeyReleased(Keys.P))
-                {
+            while (Game.IsRunning) {
```
xen2: Code style (curly braces)
Applies to full method.

sources/engine/Xenko.Physics/Engine/PhysicsComponent.cs
```diff
@@ -449,70 +460,47 @@ public void RemoveDebugEntity(Scene scene)
...
-            Vector3 translation;
-            Vector3 scale;
-            Entity.Transform.WorldMatrix.Decompose(out scale, out rotation, out translation);
+            fromMatrix.Decompose(out Vector3 scale, out Matrix rotation, out Vector3 translation);
+            DerivePhysicsTransform(translation, rotation, scale, out outMatrix);
+        }
 
-            var translationMatrix = Matrix.Translation(translation);
-            Matrix.Multiply(ref rotation, ref translationMatrix, out derivedTransformation);
+        internal void DerivePhysicsTransform(Vector3? worldPosition, Matrix? worldRotation, Vector3? worldScale, out Matrix outMatrix)
```
xen2: Curious, wouldn't it be nicer to work with `Quaternion`  here?

phr00t: Yes, but Decompose outputs a rotation matrix, not a quaternion. To avoid the conversion, I just work with the Matrix.

xen2: There's an overload of `Matrix.Decompose` which works with `Quaternion`:
https://github.com/xenko3d/xenko/blob/releases/3.0.0.6/sources/core/Xenko.Core.Mathematics/Matrix.cs#L702

sources/core/Xenko.Core.Mathematics/Matrix.cs
phr00t: Yes, if we only need a rotation or scale, I don't want to compute the other parts.

Renaming them to Decompose(type) isn't a bad idea, although "decompose" isn't an intuitive as "get", in my humble opinion... but consistency is more important.

sources/core/Xenko.Core.Mathematics/Matrix.cs
phr00t: I initially tried to use ScaleVector and was very confused when it did unexpected things. That function should be renamed to better convey what I'm actually getting. I still don't really understand what it is trying to do to properly name it.

sources/engine/Xenko.Physics/Engine/PhysicsComponent.cs
phr00t: Yes, but that does some extra math to do the conversion, which can be avoided internally if I just work with a Matrix (if I'm understanding this correctly).

PR #427, opened 2019-04-01
Fix memleak on GraphicsDevce.DirectD3.Begin()

ryancheung:
# PR Details
Fix memleak on GraphicsDevce.DirectD3.Begin()

`GraphicsDevce.DirectD3.Begin()` method keep new creating `SharpDX.DirectD3D11.Query` object after
every frame and created objects do not get freed.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Xenko.Graphics/Direct3D/GraphicsDevice.Direct3D.cs
+++ b/sources/engine/Xenko.Graphics/Direct3D/GraphicsDevice.Direct3D.cs
@@ -113,19 +113,18 @@ namespace Xenko.Graphics
             FrameTriangleCount = 0;
             FrameDrawCalls = 0;
 
-            // Try to read back the oldest disjoint query and reuse it. If not ready, create a new one.
-            if (disjointQueries.Count > 0 && NativeDeviceContext.GetData(disjointQueries.Peek(), out QueryDataTimestampDisjoint result))
-            {
-                TimestampFrequency = result.Frequency;
-                currentDisjointQuery = disjointQueries.Dequeue();
-            }
-            else
+            if (currentDisjointQuery == null)
             {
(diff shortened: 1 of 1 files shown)


sources/engine/Xenko.Graphics/Direct3D/GraphicsDevice.Direct3D.cs
```diff
@@ -113,19 +113,18 @@ public void Begin()
...
-            else
+            if (currentDisjointQuery == null)
             {
                 var disjointQueryDiscription = new QueryDescription { Type = SharpDX.Direct3D11.QueryType.TimestampDisjoint };
                 currentDisjointQuery = new Query(NativeDevice, disjointQueryDiscription);
             }
 
-            disjointQueries.Enqueue(currentDisjointQuery);
+            // Try to read back the oldest disjoint query and reuse it. If not ready, create a new one.
+            if (NativeDeviceContext.GetData(currentDisjointQuery, out QueryDataTimestampDisjoint result))
```
xen2: Data is supposed to come back asynchronously, typically a few frame later.
As a result, this `GetData` calls probably always return false?

This is the reason we need a pool of them. It shouldn't grow much more than 2/3 objects (GPU is behind a few frames, but not much more usually).
Is this the place that keeps allocating regularly?
We need to find out why it keeps allocating them rather than always reusing a pool of 2/3.

ryancheung: Thanks for engaging in this issue. Could you take a loot at https://github.com/xenko3d/xenko/issues/428?

This memleak was found when I'm trying to fix low performance issue with Xenko.

xen2:
I investigated and found the root reason.
Will commit a fix for this very shortly.
Closing this PR.

PR #450, opened 2019-05-18
added NameChanged event to ComponentBase

tebjan:
# PR Details

added NameChanged event to ComponentBase

## Motivation and Context

makes updating the names in a scene graph tree view more easy.

## Types of changes
New feature (non-breaking change which adds functionality)


tebjan:
there is a better way to do this without changing the runtime, not needed anymore.

Issue #440, opened 2019-04-22
Errors when importing a model


oschakravarthi:
1. When I try to import a model (those models are attached to this post), I am getting the below error.

Error:An exception occurred while generating the template.
Object reference not set to an instance of an object.
   at Xenko.Assets.Models.ModelAssetImporter.Import(UFile localPath, AssetImporterParameters importParameters)
   at Xenko.Assets.Presentation.Templates.ModelFromFileTemplateGenerator.CreateAssets(AssetTemplateGeneratorParameters parameters)
   at Xenko.Assets.Presentation.Templates.AssetTemplateGenerator.Run(AssetTemplateGeneratorParameters parameters)
   at Xenko.Core.Assets.Editor.Components.TemplateDescriptions.TemplateGeneratorHelper.<>c__DisplayClass0_0`1.<RunTemplateGeneratorSafe>b__0()

Info:Nothing to import.
[models.zip](https://github.com/xenko3d/xenko/files/3103634/models.zip)

Eideren:
Tested importing the given model and it works fine, probably fixed through #458 .

Issue #336, opened 2019-01-21
Impossible to open blend files created in Blender 2.8


SeleDreams:
**Release Type**: Official Release

**Version**: 3.1.0.1-beta-01-0359
**Platform(s)**: Windows 10 64 bits

**Describe the bug**
When importing a .blend file created in Blender 2.8, the error "Error:An exception occurred while generating the template." appears and nothing ends up imported.

**To Reproduce**
1. Create a project in blender 2.8
2. Save it as a .blend
3. Load it in xenko by importing a model as .blend
4. The error will appear when importing

**Expected behavior**
The model should open properly without errors

**Screenshots**
![image](https://user-images.githubusercontent.com/16335601/51452877-d88eef00-1d3c-11e9-853b-4bd78d1fad56.png)

**Log and callstacks**
```
Error:An exception occurred while generating the template.
Info:Nothing to import.
```

**Additional context**
It happens on any kind of blender 2.8 projects, I've tested with a project containing a complex model and with a simple new project only containing the basic cube, it always happens, trying with blender 2.79, the projects are imported properly, so something probably changed in the format between 2.79 and 2.8

RUSshy:
.blend format changed for 2.8, since xenko uses assimp, issue should be reported there i think: https://github.com/assimp/assimp

If you want to stick with blender 2.8 try to export as .gltf

GutterLab:
You might want to export to fbx if you're using animations. I don't think they're supported with gltf imports yet: https://github.com/xenko3d/xenko/issues/96

Since you currently can't import multiple animations in a single fbx, somebody made a blender script to split the animations and export them as separate fbx files: https://github.com/GutterLab/BlenderToXenkoAnimationSeparator

There was some discussion in the discord about adding support for importing multiple animations from a single fbx where someone was told to create an issue: https://github.com/xenko3d/xenko/issues/328 I assume it'll be implemented soon.

pikachuk:
I see
about assimp though since I don't tend to use this library, I don't know how I should report the bug there

pikachuk:
> 
> 
> You might want to export to fbx if you're using animations. I don't think they're supported with gltf imports yet: #96
> 
> Since you currently can't import multiple animations in a single fbx, somebody made a blender script to split the animations and export them as separate fbx files: https://github.com/GutterLab/BlenderToXenkoAnimationSeparator
> 
[...]

Issue #96, opened 2018-08-11
glTF Asset Importer


tebjan:
Just because it's a recent hype and it looks like a nice open standard not dominated by a certain 3d software company:
https://github.com/KhronosGroup/glTF

mafiesto4:
Xenko uses Assimp which supports importing some data from glTF (geometry and textures but animations are still WIP).

xen2:
What needs to be done:
- allow `.glb` extension in model importer
- check that the version of assimp we use is up to date enough to support this format (and upgrade assimp to latest version if too old)
- test it

Eideren:
Mostly supported, I'll close this one of and leave #603 open to continue over there

PR #458, opened 2019-05-23
Assimp 4.1.0 update and glTF support

tebjan:
## Description

* updated assimp native references to official 4.1.0 binaries which supports glTF out of the box: https://github.com/assimp/assimp/releases/tag/v4.1.0
* unified place for supported file extensions
* added glTF and STL to supported formats

## Related Issue

glTF importer: #96 
the "Nothing to Import" issue is still present and not in scope of this PR: #440 and #336

## Motivation and Context

loading models in new glTF and STL formats

## Types of changes

## Checklist


profan:
Does this handle the PBR material stuff in gltf and set that up on the xenko end at all or is that part of assimps interface and already magically works since xenko had hooked into it earlier? 🤔

I'd test it and see (but I won't be near a computer for a week) but I'm mostly curious as that's one of the biggest advantages of gltf, if not we can always make it the next step 👀👍

Nice either way!

xen2:
@profan Nothing changed for materials, some additional work needed in this area.
@tebjan Merged, along with some additional testing/error reporting. User should now have a much better idea has an idea of what went wrong rather than cryptic NullReferenceException as reported in #440 and #336.

tebjan:
good to hear that errors are now shown in a better way. thanks!

PR #472, opened 2019-06-16
Remove cast to Game in AudioSystem

Hyperpred:
# PR Details

Added GameSettings to the services and then removed the cast to Game in AudioSystem. This is beginning the work of decoupling the Systems from Game to allow for a Headless Game to be created.

## Description

As above, just added GameSettings to the Services so AudioSystem (and other classes as needed) can access them without casting to Game.

In AudioSystem specifically, the only reference to Game is to get the settings. Other areas such as SyncScript would still be able to access the settings directly.

I did not add GameSettings to IGame because of it being unavailable to that project. Another possibility would be to call Content.Load<GameSettings> in Audio system and then unload after the settings have been used.

## Related Issue

None - if needed I can create an issue as I go about decoupling systems if this is even a path that we want to take.

## Motivation and Context

Decoupling the classes will help allow for a Headless game which will be needed for using the engine as a dedicated server. Specifically, in this case, the only coupling is to load the settings and this was the most straight forward way to break that coupling.

## Types of changes

## Checklist


sources/engine/Xenko.Engine/Audio/AudioSystem.cs
```diff
@@ -53,8 +54,8 @@ public override void Initialize()
             lock (AudioEngineStaticLock)
             {
                 if (audioEngineSingleton == null)
-                {
-                    var settings = ((Game)Game)?.Settings?.Configurations?.Get<AudioEngineSettings>();
+                {
+                    var settings = Services.GetService<GameSettings>()?.Configurations?.Get<AudioEngineSettings>();
```
xen2: What about adding `GameSettings Settings { get; }` to `IGame` and `GameBase` (as abstract)? `GameSettings` doesn't really sound like something that should be a service (maybe a `IGameSettingsProvider`, but simply adding it to `IGame` sounds simpler and good enough for now).

Hyperpred: Adding GameSeetings to IGame was my first intention. However, `GameSettings `is in Xenko.Engine and `IGame ` is in Xenko.Game. I think that it might be worthwhile to just move GameSettings to Xenko.Game but I don't know what the side effects of that would be.

If `IGameSettingsProvider `only has a `GameSettings {get;}` I don't know if that kind of abstraction would make sense or not. However, I'm open to do whichever you prefer.

xen2: `Xenko.Engine` references `Xenko.Games` and `AudioSystem` already has a member `Game`. So simply adding `IGame.Settings` member should be enough to be able to do `Game.Settings` within `AudioSystem` class.

Hyperpred: image
The problem is that Xenko.Games does not reference Xenko.Engine. So right now GameSettings is in Xenko.Engine and IGame is in Xenko.Games. That is why I was unable to add GameSettings to IGame. Another alternative if we don't move GameSettings might be to call Content.Load in AudioSystem?

xen2: My bad, sorry for the misunderstanding!
In that case, let's go for a `IGameSettingsService` with a `GameSettings { get; }` member.
The `Game` class could implement this interface directly and register it in the list of services (so no need to create a new class implementing it).

Hyperpred: I've made the requested update. I wasn't sure if you wanted me to squash the commits or just add the commit on top of the first merge request. 

If you need me to do anything different please let me know. But, I have added the IGameSettingsService and it is implemented by Game now.

xen2:
@Hyperpred I think you have a problem with your CRLF settings.
Various files are comitted with CRLF but they should only have LF.
git config should have `core.autocrlf=true`. I suspect you have it set to `false` or `input`.

Hyperpred:
@xen2  You are correct, I didn't have that setting so it was defaulted to false. I have added it to my Xenko config.

xen2:
@Hyperpred just so you know, it's usually recommended to have it set to true globally on Windows and most project follow that convention since it's the default for Git for Windows.
That way it won't happen again if you clone a separate xenko or another repo later.

Kryptos-FR:
@xen2 It is even better to set it on the `.gitattributes` file. I will open a pull-request.

```
# Set the default behavior, in case people don't have core.autocrlf set.
* text=auto

# Explicitly declare text files you want to always be normalized and converted
# to native line endings on checkout.
*.c       text
*.cpp     text
*.cs      text
*.h       text

# Declare files that will always have CRLF line endings on checkout.
*.bat     text eol=crlf
*.csproj  text eol=crlf
*.sln     text eol=crlf

# Declare files that will always have LF line endings on checkout.
*.sh      text eol=lf
```

PR #478, opened 2019-06-24
Add output merger UAV binding.

WhyPenguins:
# PR Details

Previously, unordered access views could only be bound to the compute stage. This patch lets them be bound to the output merger stage, so pixel shaders can read and write to them.

## Description

Much of the patch is based on how UAV binding for the compute stage works - there is an array of currently bound UAVs, and in SetUnorderedAccessView this is checked against and the UAV is bound if necessary.

However due to how Direct3D ties render targets and UAVs together (see [here](https://docs.microsoft.com/en-us/windows/desktop/api/d3d11/nf-d3d11-id3d11devicecontext-omsetrendertargetsandunorderedaccessviews)), its also necessary to keep track of the number of bound render targets, and to re-bind all the UAVs whenever one is changed. The patch prioritises render targets, in that if the targets and UAV registers are interleaved for whatever reason, it'll ensure all the render targets stay bound. Hopefully that makes sense haha.

## Related Issue
I originally posted about this here: https://github.com/xenko3d/xenko/issues/471

## Motivation and Context

There are several techniques that need the pixel shader to be able to write to arbitrary locations, such as voxelization, or order independent transparency. 

## Types of changes

## Checklist

Haha I'll admit I'm not really sure what to do regarding proper tests. At the very least I can say I've bound multiple UAVs (buffers and textures), and have successfully written and read from them. I imagine something like that'd be a good test to write? I'll look into getting that done.

### DIFF
--- a/sources/engine/Xenko.Graphics/Direct3D/CommandList.Direct3D.cs
+++ b/sources/engine/Xenko.Graphics/Direct3D/CommandList.Direct3D.cs
@@ -270,7 +270,7 @@ namespace Xenko.Graphics
         /// <param name="shaderResourceView">The shader resource view.</param>
         /// <param name="view">The native unordered access view.</param>
         /// <param name="uavInitialOffset">The Append/Consume buffer offset. See SetUnorderedAccessView for more details.</param>
-        internal void OMSetSingleUnorderedAccssView(int slot, SharpDX.Direct3D11.UnorderedAccessView view, int uavInitialOffset)
+        internal void OMSetSingleUnorderedAccessView(int slot, SharpDX.Direct3D11.UnorderedAccessView view, int uavInitialOffset)
         {
             currentUARenderTargetViews[slot] = view;
 
@@ -316,7 +316,7 @@ namespace Xenko.Graphics
             {
                 if (currentUARenderTargetViews[slot] != view)
                 {
-                    OMSetSingleUnorderedAccssView(slot, view, uavInitialOffset);
+                    OMSetSingleUnorderedAccessView(slot, view, uavInitialOffset);
                 }
             }
         }
@@ -343,7 +343,7 @@ namespace Xenko.Graphics
             {
                 if (currentUARenderTargetViews[slot] == view)
                 {
-                    OMSetSingleUnorderedAccssView(slot, null, -1);
+                    OMSetSingleUnorderedAccessView(slot, null, -1);
                 }
             }
         }
--- a/sources/engine/Xenko.Graphics/Direct3D/CommandList.Direct3D.cs
+++ b/sources/engine/Xenko.Graphics/Direct3D/CommandList.Direct3D.cs
@@ -26,6 +26,8 @@ namespace Xenko.Graphics
         private SharpDX.Direct3D11.OutputMergerStage outputMerger;
 
         private readonly SharpDX.Direct3D11.RenderTargetView[] currentRenderTargetViews = new SharpDX.Direct3D11.RenderTargetView[SimultaneousRenderTargetCount];
+        private          int currentRenderTargetViewsActiveCount = 0;
+        private readonly SharpDX.Direct3D11.UnorderedAccessView[] currentUARenderTargetViews = new SharpDX.Direct3D11.UnorderedAccessView[SimultaneousRenderTargetCount];
         private readonly SharpDX.Direct3D11.CommonShaderStage[] shaderStages = new SharpDX.Direct3D11.CommonShaderStage[StageCount];
         private readonly Buffer[] constantBuffers = new Buffer[StageCount * ConstantBufferCount];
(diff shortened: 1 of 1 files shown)


sources/engine/Xenko.Graphics/Direct3D/CommandList.Direct3D.cs
```diff
@@ -255,6 +263,30 @@ internal void SetShaderResourceView(ShaderStage stage, int slot, GraphicsResourc
...
+
+            var uavs = new SharpDX.Direct3D11.UnorderedAccessView[remainingSlots];
+            Array.Copy(currentUARenderTargetViews, currentRenderTargetViewsActiveCount, uavs, 0, remainingSlots);
+
+            var uavInitialCounts = new int[remainingSlots];
+            for (int i = 0; i < remainingSlots; i++)
+                uavInitialCounts[i] = -1;
+            uavInitialCounts[slot - currentRenderTargetViewsActiveCount] = uavInitialOffset;
+
+            outputMerger.SetUnorderedAccessViews(currentRenderTargetViewsActiveCount, uavs, uavInitialCounts);
```
xen2: Note: double-checked into SharpDX code and it seems they use `D3D11_KEEP_RENDER_TARGETS_AND_DEPTH_STENCIL` which is what we want. All good!

sources/engine/Xenko.Graphics/Direct3D/CommandList.Direct3D.cs
```diff
@@ -255,6 +263,30 @@ internal void SetShaderResourceView(ShaderStage stage, int slot, GraphicsResourc
...
+        /// <param name="shaderResourceView">The shader resource view.</param>
+        /// <param name="view">The native unordered access view.</param>
+        /// <param name="uavInitialOffset">The Append/Consume buffer offset. See SetUnorderedAccessView for more details.</param>
+        internal void OMSetSingleUnorderedAccssView(int slot, SharpDX.Direct3D11.UnorderedAccessView view, int uavInitialOffset)
+        {
+            currentUARenderTargetViews[slot] = view;
+
+            int remainingSlots = currentUARenderTargetViews.Length - currentRenderTargetViewsActiveCount;
+
+            var uavs = new SharpDX.Direct3D11.UnorderedAccessView[remainingSlots];
```
xen2: Let's preallocate this (as a class private field) to avoid allocation during typical frame.

sources/engine/Xenko.Graphics/Direct3D/CommandList.Direct3D.cs
```diff
@@ -255,6 +263,30 @@ internal void SetShaderResourceView(ShaderStage stage, int slot, GraphicsResourc
...
+            Array.Copy(currentUARenderTargetViews, currentRenderTargetViewsActiveCount, uavs, 0, remainingSlots);
+
+            var uavInitialCounts = new int[remainingSlots];
```
xen2: Let's preallocate this (as a class private field) to avoid allocation during typical frame.

WhyPenguins:
Ah sorry for replying so late! Thanks so much for the review, will correct the patch (that typo is embarrassing, sorry).

My C# skills being as rusty as they are though, do you have any tips on how to pre-allocate those arrays?

The reason I currently allocate, is so I can pass a slice of the currentUARenderTargetViews array to SetUnorderedAccessViews. The only way I know how to do this in C# without an allocation is with ArraySegment, but I can't pass one to SetUnorderedAccessViews. Any ideas?

xen2:
@WhyPenguins No problem for the typo, we had/have many of them as well!

For preallocating, the idea is to create a field which preallocate the array:
`private readonly SharpDX.Direct3D11.UnorderedAccessView[] temporaryUAVs = new SharpDX.Direct3D11.UnorderedAccessView[SimultaneousRenderTargetCount];`
Do same for initial counts.
Then just use this array inside `OMSetSingleUnorderedAccessView` rather than creating a new one every call.

xen2:
@WhyPenguins Concerning preallocation, it might not be even possible to do with current SharpDX API (it only takes start offset but not count -- which is not going to be equal to array length if start offset is not 0).

WhyPenguins:
Yeah that was my concern, sorry if I wasn't clear enough. It's unfortunate since there's an internal method that does end up taking a pointer and count, but yeah it can't be used.

I suppose if the pre-allocation is important, a workaround could be an array of arrays with increasing sizes. Then the desired slice of the currentUARenderTargetViews array could be copied into the array with correct count, and that could be passed in?

Also would you prefer the commits to be cleaned up via rebasing (I'd squash the original and typo commits together and rebase to the current master) or do you prefer keeping the original history?

xen2:
@WhyPenguins Let's keep it like this, it's fine in the meantime. We better fix it SharpDX side later rather than making a big ugly temporary workaround with array of arrays.
No problem for the commits, I can rebase+squash them directly from GitHub merge button.

PR #482, opened 2019-07-01
[SceneManagement] Assign Entity.sceneValue before firing collection events

Eideren:
# PR Details
Fixes a bug occurring when removing an entity, which can end up firing code which also tries to remove that entity, generating an exception as the collection does not expect such a state.

## Description
Forced ``Entity.sceneValue`` and ``Scene.Parent`` to be taken care of before firing events related to them.

## Related Issue
None reported.

## Motivation and Context
Here's an example of the kind of issues where it would throw:
- Remove entityA from scene
	- collection fires remove events [lifo](https://github.com/xenko3d/xenko/blob/5cee86cd3d8807c39446c07d0481348afd7c4131/sources/core/Xenko.Core/Collections/TrackingCollection.cs#L27), so first:
		- SceneInstance event:[fires processors](https://github.com/xenko3d/xenko/blob/5cee86cd3d8807c39446c07d0481348afd7c4131/sources/engine/Xenko.Engine/Engine/SceneInstance.cs#L142)
			- *some user side code removes entityA from its scene*
				- collection fires remove events
					- processors are mostly fine
					- Scene event:[replace Entity.sceneValue](https://github.com/xenko3d/xenko/blob/5cee86cd3d8807c39446c07d0481348afd7c4131/sources/engine/Xenko.Engine/Engine/Scene.cs#L32)
		- Scene event:[replace Entity.sceneValue](https://github.com/xenko3d/xenko/blob/5cee86cd3d8807c39446c07d0481348afd7c4131/sources/engine/Xenko.Engine/Engine/Scene.cs#L32), *woops*, value does not match as it has been replaced above, [throws right here](https://github.com/xenko3d/xenko/blob/5cee86cd3d8807c39446c07d0481348afd7c4131/sources/engine/Xenko.Engine/Engine/Scene.cs#L173)

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Xenko.Engine/Engine/Scene.cs
+++ b/sources/engine/Xenko.Engine/Engine/Scene.cs
@@ -2,7 +2,6 @@
 // Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
 
 using System;
-using System.Collections.Specialized;
 using Xenko.Core;
 using Xenko.Core.Annotations;
 using Xenko.Core.Collections;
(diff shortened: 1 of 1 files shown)


Kryptos-FR:
Can we check if this still works well with undo/redo in the editor?

Eideren:
@Kryptos-FR Just checked, undo/redo system still works.

PR #483, opened 2019-07-03
Update .gitattributes

Kryptos-FR:
Add explicit rules for line endings.

# PR Details

## Description

Add rules in the `.gitattributes` file so that source code files are checked out with native line endings, while some other files have explicit ones.

## Related Issue

Was mentioned in the discussion of #472.

## Motivation and Context

Should solve the issues when people haven't set their autocrlf config accordingly.

## Types of changes

## Checklist

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,21 @@
+# Set the default behavior, in case people don't have core.autocrlf set.
+* text=auto
+
+# Explicitly declare text files you want to always be normalized and converted
+# to native line endings on checkout.
+*.c       text
+*.cpp     text
+*.cs      text
+*.h       text
+
+# Declare files that will always have CRLF line endings on checkout.
+*.bat     text eol=crlf
+*.csproj  text eol=crlf
+*.sln     text eol=crlf
+
+# Declare files that will always have LF line endings on checkout.
+*.sh      text eol=lf
+
 # binaries
 *.dll filter=lfs diff=lfs merge=lfs -text
 *.exe filter=lfs diff=lfs merge=lfs -text


xen2:
Looked little bit into it, from what I understood, I think this will also force Unix/macOS checkout to have CRLF then?

This might be OK, but need to double-check how MSBuild for Linux/OSX and typical C# editor (i.e. VSCode and VS for Mac) deal with that.

Kryptos-FR:
@xen2 `* text=auto` means that line endings will be native on each platform: CRLF on Windows, LF on Linux and CR or LF on Mac.

The only file type where we force CRLF are for batch file and Visual Studio project and Solution because that's how they should be on all platforms. And Unix shell scripts should be LF. All other types will be "auto".

xen2:
@Kryptos-FR when I read the doc/stackoverflow, I really understand `text=auto` will also put CRLF on Linux.
Even got a confirmation there: https://groups.google.com/forum/#!topic/git-users/ck_pKFGCYKc
`auto` is little bit misleading in that it doesn't mean **auto** adjust according to OS. Instead, it seems to simply mean **auto** convert LF to CRLF, whatever the OS is.

Note: didn't test on Linux myself, just my understanding from doc.

Kryptos-FR:
From the documentation https://git-scm.com/docs/gitattributes

Setting `text=auto` means that it will follow whatever `core.eol` value is set on the local config. In Windows it's default is CRLF while in Linux it should be LF. My guess is whatever said it doesn't work in Linux had the wrong setting on their local copy. Or they had `core.autocrlf` set to `true` (which overrides `core.eol`).

xen2:
@Kryptos-FR Good find, thanks for the research!

xen2:
Did a quick test on a small repo and it seems to work fine, thanks for the contrib!

PR #485, opened 2019-07-03
[SceneManagement] Fix loop skipping items when collection is modified

Eideren:
# PR Details
The collection can be changed through user code which might make the enum skip over items, this allocates and fills a temp array to store the collection state as is.

## Description
See above.

## Related Issue
None reported.

## Motivation and Context
Ex: A component removing another entity after being removed itself, the collection's enum implementation doesn't check back if the collection changed so it just continues from the index it left off on which might not be accurate anymore if the removed item's index lays at or before the enum's index.
The default .Net implementation would probably have thrown here, should we perhaps implement safeguards as well ?

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Xenko.Engine/Engine/SceneInstance.cs
+++ b/sources/engine/Xenko.Engine/Engine/SceneInstance.cs
@@ -131,12 +131,67 @@ namespace Xenko.Engine
 
         private void Add(Scene scene)
         {
-            foreach (var entity in scene.Entities)
-                Add(entity);
+            if (scene.Entities.Count > 0)
+            {
+                var entitiesToAdd = new FastList<Entity>();
+                // Reverse order, we're adding and removing from the tail to 
+                // avoid forcing the list to move all items when removing at [0]
+                for (int i = scene.Entities.Count -1; i >= 0; i-- )
+                    entitiesToAdd.Add(scene.Entities[i]);
 
-            // Listen to future changes in Scene.Entities
-            foreach (var childScene in scene.Children)
-                Add(childScene);
+                scene.Entities.CollectionChanged += DealWithTempChanges;
+                while (entitiesToAdd.Count > 0)
+                {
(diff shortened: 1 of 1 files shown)


Kryptos-FR:
I think we should implement the same safeguard as the default .Net collection, i.e. keep a "version" field that is incremented every time a change occurs. If an iterator is resumed on a different version, throw.

Most of the time, it is a coding error to try to modify a collection while it is being iterated. If we attempt to hide that error, we will have some inconsistencies.

But having that safeguard adds a small cost (i.e. incrementing and checking against an integer). So we need some more analysis to determine the impact.

xen2:
@Kryptos-FR this could also be activated only in debug build (assert-like).
If still there in production release, I doubt it can be recovered from anyway (something very wrong).

@Eideren Wouldn't it make sense to do it in the `Add` too?

Eideren:
@xen2 Well, for ``Add()`` it's not as simple as that, how do we handle new items / insertions in the source list while the temp one is being processed through ``Add(Entity)`` or ``Add(aSubScene)``.
The second commit I just pushed shows a sub-optimal implementation to deal with this issue, how would you deal with this ?

Eideren:
I would rather have a version on the collection and while loop ``Add()`` new items as long as the version doesn't stabilize.

xen2:
@Eideren makes sense, thanks (didn't think about the details when I asked)!

xen2:
And just to know, do you think we should merge this or better to have collection "version" right away instead?
Wondering because we are going to allow this behavior and maybe forbid it again later...

Eideren:
@xen2 Well, if you can't think of a better way for that ``Add()`` implementation I did, 'guess we should merge. Collection version might take a lot of time to properly implement as it might break part of the engine using collection enums.

xen2:
OK understood, thanks for the insightful feedbacks!

Issue #488, opened 2019-07-22
Rotation and scale should be from selection center


nesrak1:
**Release Type**: GitHub

**Version**: Master 3.1.0.1-beta02

**Platform(s)**: Windows

**Describe the bug**
The rotate tool rotates from the last selected object and scale scales all objects from their own center instead of the selection center

**Expected behavior**
Rotation and scale should pivot from selection center instead

**Screenshots**
Expected:
![](https://cdn.discordapp.com/attachments/500285081265635330/602696484240228372/2019-07-21_21-56-21.gif)

Current:
![](https://cdn.discordapp.com/attachments/500285081265635330/602696503278043137/2019-07-21_21-55-05.gif)

Aggror:
Sometimes rotating on the last selected object is really wanted. I do this to perform pivot based rotations. An interface button to toggle the selection pivot mode between Center or per last selected object would be nice.

PR #489, opened 2019-07-24
[WIP] Add selection center rotation and scaling

nesrak1:
# PR Details

Adds rotation and scaling from the center or from the last selected object

## Description

You can now scale from the center from a selection as well as transforming away from the center rather than from the origin of each object. The selection can also be rotated from the center rather than only the last selected object.

## Related Issue

#488 

## Motivation and Context

Sometimes it makes more sense to scale from the center rather than the last selected object. For example, something that takes up multiple objects like a room would be hard to scale otherwise.

## Types of changes

## Checklist

### DIFF
--- a/sources/editor/Xenko.Assets.Presentation/AssetEditors/Gizmos/ScaleGizmo.cs
+++ b/sources/editor/Xenko.Assets.Presentation/AssetEditors/Gizmos/ScaleGizmo.cs
@@ -125,9 +125,6 @@ namespace Xenko.Assets.Presentation.AssetEditors.Gizmos
         {
             const float MaxScaleValue = 1f / MathUtil.ZeroTolerance;
 
-            if (UseSnap)
-                translation = MathUtil.Snap(translation, SnapValue);
-
             return Math.Max(-MaxScaleValue, Math.Min(MaxScaleValue, translation));
         }
 
--- a/sources/editor/Xenko.Assets.Presentation/AssetEditors/Gizmos/ScaleGizmo.cs
+++ b/sources/editor/Xenko.Assets.Presentation/AssetEditors/Gizmos/ScaleGizmo.cs
@@ -125,13 +125,10 @@ namespace Xenko.Assets.Presentation.AssetEditors.Gizmos
         {
             const float MaxScaleValue = 1f / MathUtil.ZeroTolerance;
 
-            // snap the value if needed BEFORE the scale function
             if (UseSnap)
                 translation = MathUtil.Snap(translation, SnapValue);
 
-            //var scaleValue = translation > 0 ? (float)Math.Exp(translation) : 1 / ((float)Math.Exp(-translation));
-
-            return Math.Max(MathUtil.ZeroTolerance, Math.Min(MaxScaleValue, translation));
+            return Math.Max(-MaxScaleValue, Math.Min(MaxScaleValue, translation));
         }
 
         protected override void UpdateColors()
--- a/sources/editor/Xenko.Assets.Presentation/AssetEditors/Gizmos/ScaleGizmo.cs
+++ b/sources/editor/Xenko.Assets.Presentation/AssetEditors/Gizmos/ScaleGizmo.cs
@@ -105,7 +105,7 @@ namespace Xenko.Assets.Presentation.AssetEditors.Gizmos
         protected override InitialTransformation CalculateTransformation()
         {
             var transform = base.CalculateTransformation();
-
+            
             if (TransformationAxes == GizmoTransformationAxes.XYZ) // special transformation mode
             {
                 var mouseDrag = Input.MousePosition - StartMousePosition;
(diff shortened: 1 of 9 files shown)


Eideren:
Hey, thanks a bunch for this PR @nesrak1 
I found a couple of things to fix/clarify, I'll list them below and in review.

Scaling down doesn't seem to work at the moment, is this PR ready yet ? If it isn't please mark this PR as WIP and I'll come back to it once it isn't, thanks !

sources/editor/Xenko.Assets.Presentation/AssetEditors/Gizmos/ScaleGizmo.cs
```diff
@@ -125,13 +125,13 @@ private float ComputeScaleFactorFromTranslation(float translation)
         {
             const float MaxScaleValue = 1f / MathUtil.ZeroTolerance;
 
-            // snap the value if needed BEFORE the exp function
+            // snap the value if needed BEFORE the scale function
             if (UseSnap)
                 translation = MathUtil.Snap(translation, SnapValue);
 
-            var scaleValue = translation > 0 ? (float)Math.Exp(translation) : 1 / ((float)Math.Exp(-translation));
+            //var scaleValue = translation > 0 ? (float)Math.Exp(translation) : 1 / ((float)Math.Exp(-translation));
```
Eideren: You can omit commented-out lines from PRs, we can always rollback if necessary.

Eideren (changes requested):
Hey @nesrak1 , thanks a bunch for this PR.

Scaling lower than the scale pre-transformation doesn't seem to work on this PR right now, is this ready or should we wait ? If it isn't please mark this as WIP and we'll get back to it once it isn't, thanks !

sources/editor/Xenko.Assets.Presentation/AssetEditors/Gizmos/ScaleGizmo.cs
nesrak1: This was for me, I just forgot to remove it

nesrak1:
I already had these fixes but forgot to commit, sorry!

nesrak1:
Scaling negative works which may or may not be a good thing, since it seems to break lighting but can also flip the model.

Eideren:
It's quite alright :) 
Doesn't quite seem to work correctly if object don't have zeroed rotation / are inside parents

sources/editor/Xenko.Assets.Presentation/AssetEditors/Gizmos/TransformationGizmo.cs
```diff
@@ -434,6 +504,27 @@ private async Task TransformSceneEntityBase()
...
+            vec.Z = GetScaledAxis(position.Z, scaleOrigin.Z, scale.Z);
+            return vec;
+        }
+
+        protected float GetScaledAxis(float position, float scaleOrigin, float scale)
+        {
+            return scaleOrigin + (position - scaleOrigin) * scale;
+        }
+
+        protected Vector3 Transform(Vector3 vector, Matrix transform)
```
nesrak1: The only Transform without refs returns a Vector4. I could either put a new method in Vector3 called something like `TransformVec3` or convert the Vector4 to Vector3, not sure which one is better.

Eideren: I don't really see a reason to implement that function, you could use ``Transform(ref initialTranslation, ref inverseScaleSpace, out var position);``, on line 484 and others, I don't see a problem with that ?
The vector API uses refs to pass matrix and vectors more efficiently, passing large structs as method parameters incur a fairly substantial cost if not done by reference.

nesrak1: I didn't think the performance hit would be significant enough and it seemed to be consistent with the functions around along with the stack of local variables I had earlier.

nesrak1: In that case, should `GetScaledLocation` use refs as well?

Eideren: Nah you can keep using non-refs, you don't have to force yourself, I was just clarifying why the API guides you towards using refs. The cost is just large enough to warrant using an already existing ref function instead of its normal counterpart if the readability of your code does not suffer from it, which, in that case, seemed like a perfect fit.

xen2:
@nesrak1 just checking, is the PR ready/finished?

nesrak1:
Should be ready unless you find another bug I haven't seen yet

xen2:
Ran a few tests and noticed some issues.

Here's what happens when I:
- Create entity with some rotation (i.e. 90 degrees Y axis)
- Put a sphere as children of this entity with some translation (i.e. Z = 2)

The gizmo is not at the expected location:
image

Here is what's expected:
image

Eideren:
@nesrak1 Hey, this PR hasn't been very active those past few months, are you still working on this or should we close this off ? If someone wants to take it from here please do so but leaving this active might prevent other people from tackling this issue.

nesrak1:
I plan to pick progress back up soon but I don't know when I will finish it.

xen2:
Closing due to lack of activity.
Please reopen if planning to finish/merge it.

PR #496, opened 2019-08-11
Update README.md

mostafa901:
Hi, Just wondering if the Prerequisites says VS 2019 then I should open Xenko.sln via 2019

# PR Details

## Types of changes

## Checklist


Kryptos-FR:
Good point.
I think we can even remove the mention of the year and just keep "VIsual Studio". That way, when a new version comes and we need to upgrade, we will just have to change the one lin in the pre-requisites section.

mostafa901:
Yes that will be more efficient indeed

PR #497, opened 2019-08-13
[Game] GameBase refactor

Eideren:
# PR Details
Refactored a large part of the code to be less obtuse, clarified and extended a bunch of property summaries, removed and implemented a couple of minor functions and allowed internal tick producer to be overridden.

## Description
- Cleaned up logic and split functionnality of ``TickInternal`` into ``Tick``, virtual ``RawTickProducer`` and ``RawTick``.
- Extended and clarified a bunch of summaries, see changes.
- Fixed IsMouseVisible setting itself back to true when the window is created.
- Provided public read access to ``IsExiting``.
- Provides the ability to force drawing even when the window is minimized (gameplay relying on gpu compute, forcing wait for vsync, etc.)
- Moved related functions near each other.

_Removed the following features_, if you don't agree with those changes, please reply back. I'll keep them in if they actually are useful someone.
- ``SlowDownDrawCalls``:
  - This variable is internal and not used anywhere in the source code.
- ``IsRunningSlowly``:
  - This should be implemented on a case by case basis on the user's side, it doesn't make a lot of sense to embed it in the game's core functionality and within ``GameTime`` as well.
- ``EarlyExit``:
  - Redundant, implementation acts in the same way as ``IsExiting`` does.
- ``State``:
  - Not used anywhere in source nor samples, looks like legacy code.
- ``PlayTime``:
  - Not used anywhere in source nor samples, extremely similar to ``UpdateTime``, non-essential, can easily be implemented user-side if necessary.

## Motivation and Context
I need more control over how updates and draw are scheduled, while cleaning up the logic to be overridden I decided to deal with the rest of the file as well.

## Types of changes

## Checklist

## Notes
Are the todos at ``GraphicsDeviceService_DeviceReset``, ``GraphicsDeviceService_DeviceResetting`` and ``GraphicsDeviceService_DeviceDisposing`` implemented ? 'looks like they are ? Can I remove those comments ?


sources/engine/Xenko.Games/GameBase.cs
```diff
@@ -165,50 +140,27 @@ protected GameBase()
         #region Public Properties
 
         /// <summary>
-        /// Gets the current update time from the start of the game.
+        /// The current time from the start of the game, this one should be used for any logic running within the update loop.
```
Kryptos-FR: Put the second part of the sentence either in a second sentence or under `<remarks>`.

sources/engine/Xenko.Games/GameBase.cs
```diff
@@ -165,50 +140,27 @@ protected GameBase()
...
-            get
-            {
-                return updateTime;
-            }
-        }
+        public GameTime UpdateTime { get; }
 
         /// <summary>
-        /// Gets the current draw time from the start of the game.
+        /// The current time from the start of the game, this one should be used for any logic running within the draw loop.
```
Kryptos-FR: Put the second part of the sentence either in a second sentence or under `<remarks>`.

sources/engine/Xenko.Games/GameBase.cs
```diff
@@ -274,9 +231,8 @@ public bool IsMouseVisible
         {
             get
             {
-                return isMouseVisible;
+                return Window == null ? isMouseVisible : Window.IsMouseVisible;
```
Kryptos-FR:
```suggestion
                return Window?.IsMouseVisible ?? isMouseVisible;
```

Using the null-coalesce operator prevent race condition where `Window` could change from `null` to `not-null` (or opposite) during the evaluation. And I personally think it is clearer 😉

sources/engine/Xenko.Games/GameBase.cs
Eideren: That part of the description is *essential* to use those object properly. I reduced the wordiness instead, how does that fare ?

sources/engine/Xenko.Games/GameBase.cs
Eideren: See conversation above.

sources/engine/Xenko.Games/GameBase.cs
Eideren: Yep, thanks.

Eideren:
@fdrobidoux The game can be setup to run in fixed time step, the issue I have with it is that it abstracts away too much information, if someone is running in fixed time step he would rather know how many steps are queued up, sample that across time, build his own threshold to deal with this and handle them based on all of this data.

xen2:
@Eideren Since there are some breaking change, it might be a good opportunity to switch to 3.2 (which is long overdue) rather than being in 3.1. What do you think?
If yes, I will branch `master-3.1`, mark it non-beta and merge this in `master`

Eideren:
@xen2 Sure.
Are the todos at ``GraphicsDeviceService_DeviceReset``, ``GraphicsDeviceService_DeviceResetting`` and ``GraphicsDeviceService_DeviceDisposing`` out of date ?

xen2:
@Eideren I have just splitted `master` & `master-3.1`, so I can merge anytime.
The todos seems out of date, yes.

Eideren:
@xen2 I don't have anything else to add, go ahead when you are ready :)

xen2:
Sorry... better late than never!
Fixed some conflicts, think it should be OK (`if (gamePlatform.IsBlockingRun)` moved together with the new location of calls to `Throttle`).

PR #502, opened 2019-08-29
 [Threading] ThreadPool refactor, reduce allocation, fix wrong argument 

Eideren:
# PR Details
Title.
This PR has been benchmarked and, as expected given those changes, does not produce any performance difference besides faster initial run and slightly less allocation.

## Description
``Task.Factory.StartNew(ProcessWorkItems, TaskCreationOptions.LongRunning);`` maps to ``StartNew(Action(object), object)`` function signature instead of ``StartNew(Action(object), TaskCreationOptions)`` which, of course, didn't actually set the task as long running.
Cache the delegate we pass onto the task factory to reduce allocation.
Push as much stuff as possible outside of the lock.
Let exceptions fall into unhandled instead of ignoring them, if you have any better idea feel free to comment/start a review.

## Related Issue
#302 

## Motivation and Context
Bug, alloc.

## Types of changes

## Checklist

### DIFF
--- a/sources/core/Xenko.Core/Threading/ThreadPool.cs
+++ b/sources/core/Xenko.Core/Threading/ThreadPool.cs
@@ -17,22 +17,31 @@ namespace Xenko.Core.Threading
     /// </remarks>
     internal class ThreadPool
     {
-        private const long MaxIdleTimeInTicks = 5 * TimeSpan.TicksPerSecond;
+        private const int MaxIdleTimeInMS = 5000;
+        private readonly long MaxIdleTimeTS = (long)((double)Stopwatch.Frequency / 1000 * MaxIdleTimeInMS);
 
         public static readonly ThreadPool Instance = new ThreadPool();
+
+        private readonly Action<object> cachedTaskLoop;
 
         private readonly int maxThreadCount = Environment.ProcessorCount + 2;
         private readonly Queue<Action> workItems = new Queue<Action>();
         private readonly ManualResetEvent workAvailable = new ManualResetEvent(false);
 
         private SpinLock spinLock = new SpinLock();
-        private int workingCount;
-        /// <summary> Usage only within <see cref="spinLock"/> </summary>
+        private int busyCount;
         private int aliveCount;
+
+        public ThreadPool()
+        {
+            // Cache delegate to avoid pointless allocation
+            cachedTaskLoop = (o) => ProcessWorkItems();
+        }
 
         public void QueueWorkItem([NotNull] [Pooled] Action workItem)
         {
-            var lockTaken = false;
+            bool lockTaken = false;
+            bool startNewTask = false;
             try
             {
                 spinLock.Enter(ref lockTaken);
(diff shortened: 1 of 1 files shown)


Kryptos-FR:
`Task.Factory.StartNew(ProcessWorkItems, TaskCreationOptions.LongRunning)` is bad anyway, so better get rid of it entirely and use `Task.Run()` instead.

See https://blog.stephencleary.com/2013/08/startnew-is-dangerous.html

sources/core/Xenko.Core/Threading/ThreadPool.cs
```diff
@@ -17,93 +17,126 @@ namespace Xenko.Core.Threading
...
         private SpinLock spinLock = new SpinLock();
-        private int workingCount;
-        /// <summary> Usage only within <see cref="spinLock"/> </summary>
+        private int busyCount;
         private int aliveCount;
 
+        public ThreadPool()
+        {
+            // Cache delegate to avoid pointless allocation
+            cachedTaskLoop = (o) => ProcessWorkItems();
```
Kryptos-FR: I am not convinced that you are reducing allocation that way. Delegate are usually cached by the runtime, so the previous code using the method group was most likely not creating a delegate at every call.

Do you have profiling supporting your change?

Eideren: Yes, 64 bytes allocated on every call for both ``(o) => ProcessWorkItems()`` and just passing ``ProcessWorkItems`` when ProcessWorkItems' signature has an object parameter.

sources/core/Xenko.Core/Threading/ThreadPool.cs
```diff
@@ -17,93 +17,126 @@ namespace Xenko.Core.Threading
...
             {
                 if (lockTaken)
+                {
                     spinLock.Exit(true);
+                }
+            }
+            // No point in wasting spins on the lock while creating the task
+            if (startNewTask)
+            {
+                new Task(cachedTaskLoop, null, TaskCreationOptions.LongRunning).Start();
```
Kryptos-FR: Why that change? It is probably even worst than using TaskFactory.StartNew().

A long running task is a task that is blocking (e.g. using a Thread.Join() or Thread.Sleep()). That is not the case here. Over subscribing the managed thread pool is a recipe for disaster.

Kryptos-FR:
After careful reading, I don't even know why we have that custom ThreadPool implementation, I don't find any benefit to it. The same limiting effect could be achieved by a more simpler producer-consumer queue using semaphores (or even by using [Task.DataFlow](https://docs.microsoft.com/en-us/dotnet/standard/parallel-programming/dataflow-task-parallel-library)), or by limiting the number of threads in the framework thread pool (see [ThreadPool.SetMaxThreads(Int32, Int32)](https://docs.microsoft.com/en-us/dotnet/api/system.threading.threadpool.setmaxthreads)).

@xen2 do you remember why we have this?

sources/core/Xenko.Core/Threading/ThreadPool.cs
Eideren: That task *is* blocking, either through ``workAvailable.WaitOne`` or the spinwait. That task lives for at least 5 seconds and in most case for the lifetime of the application, it is long running.
What we're doing here is effectively the same thing as starting a thread when we have too much work to deal with for the current amount of thread and let them block while waiting for work.

Eideren:
@Kryptos-FR The official ThreadPool is *extremely* slow compared to our implementation, ours can start working as soon as ~0.008ms on average after being given a job while the ThreadPool is closer to ~0.07ms *minimum* (tested on net core, fw is around 4 times slower than that).
Task.DataFlow will waste cpu time and introduce more allocations, those are really nice constructs for enterprise but aren't suited to how we're using it: tons of random jobs that lasts for far less than a millisecond.

sources/core/Xenko.Core/Threading/ThreadPool.cs
Kryptos-FR: OK for that remark. I should have read more carefully.

Still, the `new Task().Start()` needs to go away as it introduces unnecessary locking (as per the link I provided).

Kryptos-FR:
@Eideren, we could fix that by just starting all those threads at the start of the game and still use the framework ThreadPool. It is a common occurrence in web server application.

Since, as you said, the threads are most likely waiting on the *workAvailable* `EvantHandle`, they won't be consuming any resources, until they make some job progress.

There are two ways of doing that: either starting the threads manually, or using `ThreadPool.SetMinThreads(Int32, Int32)` to set a higher value, so that the framework thread pool reserves/allocates a bit more threads at the beginning. We could even do a combination of both (probably the best approach).

I'm concerned, that:
1. our implementation might be missing some corner cases that could degrade performance
2. we won't benefit from future native/CLR/JIT optimizations on how tasks, threads, or parallelization in general are managed in .NET.
3. We have to maintain that code.

Eideren:
@Kryptos-FR What do you mean by 
>we could fix that by just starting all those threads at the start of the game and still use the framework ThreadPool.
Replacing ``new Task().Start()`` by ``ThreadPool.QueueUserWorkItem`` ? I don't think that the ThreadPool handles long running jobs well, users might also use it which would block their job, same thing goes for ``ThreadPool.SetMinThreads``, users might also fiddle with that which might have repercussions on how the engine operates without them knowing so.
I'm probably off the mark here, could you expand on what you where mentioning above ?

Kryptos-FR:
You do realize that starting a Task here always goes to the `ThreadPool`, don't you?
That is because the default scheduler for task is pushing all the tasks in the thread pool (and we are not providing a different scheduler here, unless I'm missing something). 

So `TaskFactory.StartNew()`, `new Task().Start()` or `ThreadPool.QueueUserWorkItem` (btw I recommend using `ThreadPool.UnsafeQueueUserWorkItem` instead) all end-up using the `ThreadPool`.  So again, what's the point of our implementation?

Kryptos-FR:
Why even use a `Task` if:
1. we bypass the task scheduler logic by using `TaskCreationOptions.LongRunning`
2. we only care about the number of threads running concurrently
3. we don;t support any of the `Task` feature such as continuation or cancellation

Won't it be more efficient to just start dedicated threads manually? Since that's what the default task scheduler ends-up doing anyway ([source](https://referencesource.microsoft.com/#mscorlib/system/threading/Tasks/ThreadPoolTaskScheduler.c)). We do pay the cost of using tasks, which has non-trivial execution logic ([source](https://referencesource.microsoft.com/#mscorlib/system/threading/Tasks/Task.cs)), even when ran from a dedicated thread.

Eideren:
@Kryptos-FR What about the ``ThreadPool`` I thought that was what you wanted to go for even if I don't fully understand why.
I personally would have gone with threads since #302 , I kept it as is to avoid having to defend that position as it ultimately doesn't really matter when that code path is only hit 10 / 18 times over a whole game.

I'll let you create the PR to implement it however you see fit.

xen2:
We didn't use `Thread` because it wasn't available in first versions of UWP/.NET Standard, so we had to work around it with long running tasks.
(it might be worth adding a small comment so that people are informed of those historical reasons when reading the codebase)

It would be fine to use `Thread` it and I would accept a PR in that sense (if properly tested).
However there's probably plenty of other more important things to do :)

PR #504, opened 2019-08-30
Windows Multitouch support via SDL

dominikjancik:
# PR Details

## Description

Added handling of SDL events regarding multitouch (SDL_FINGERDOWN/UP/MOTION) and created a new device representing this input - FingerSDL (possibly should be called TouchSDL?).

Pointer IDing logic is consistent with PointeriOS.

Device GUID is a new random GUID generated using VS GUID tool.

Tested on a Windows 10 touchscreen enabled notebook using a slightly modified TouchInput sample. Composite gesture recognition seems to work although it's a bit finicky (fingers have to touch the surface one after another, not at once). I would guess this is an issue with the Gesture recognizer.
[Test Video](https://youtu.be/h81-m0jnaRU)

Known issues:
Touch also generates mouse events. There seems to be an option in SDL to disable mouse emulation but this might also be done on OS level. Also it's not always a bad thing, just something to keep in mind.

## Questions

- Is FingerSDL a good name or an alternative like TouchSDL should be used?
- Should some form of Touchscreen detection be added so games can adjust? SDL this might be possible using SDL_GetNumTouchDevices and then adding something like HasTouchscreen to the InputManager.

## Motivation and Context

Although multitouch is supported in Xenko for iOS and Android devices, Windows is not supported despite the growing number of touchscreen on the market. Also upcoming vvvv gamma integration might bring more people requiring PC multitouch and related gestures to be supported for their creative projects.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Xenko.Graphics/SDL/Application.cs
+++ b/sources/engine/Xenko.Graphics/SDL/Application.cs
@@ -123,6 +123,9 @@ namespace Xenko.Graphics.SDL
             }
         }
 
+        // Used for FINGER* events that do not provide their own windowID
+        static Window lastCtrl = null;
+        
         /// <summary>
         /// Process a single event and dispatch it to the right window.
         /// </summary>
@@ -160,6 +163,12 @@ namespace Xenko.Graphics.SDL
                     ctrl = WindowFromSdlHandle(SDL.SDL_GetWindowFromID(e.text.windowID));
                     break;
 
+                case SDL.SDL_EventType.SDL_FINGERMOTION:
+                case SDL.SDL_EventType.SDL_FINGERDOWN:
+                case SDL.SDL_EventType.SDL_FINGERUP:
+                    ctrl = lastCtrl;
+                    break;
+
                 case SDL.SDL_EventType.SDL_WINDOWEVENT:
                 {
                     ctrl = WindowFromSdlHandle(SDL.SDL_GetWindowFromID(e.window.windowID));
@@ -182,6 +191,9 @@ namespace Xenko.Graphics.SDL
                     break;
             }
             ctrl?.ProcessEvent(e);
+
+            // TODO this is a bit of a mess but FINGER* events do not provide a proper windowID - how to handle?
+            if (ctrl != null) lastCtrl = ctrl;
         }
 
         /// <summary>
--- a/sources/engine/Xenko.Graphics/SDL/Application.cs
+++ b/sources/engine/Xenko.Graphics/SDL/Application.cs
@@ -123,9 +123,6 @@ namespace Xenko.Graphics.SDL
             }
         }
 
-        // Used for FINGER* events that do not provide their own windowID
-        static Window lastCtrl = null;
-        
         /// <summary>
         /// Process a single event and dispatch it to the right window.
         /// </summary>
@@ -166,7 +163,7 @@ namespace Xenko.Graphics.SDL
                 case SDL.SDL_EventType.SDL_FINGERMOTION:
                 case SDL.SDL_EventType.SDL_FINGERDOWN:
                 case SDL.SDL_EventType.SDL_FINGERUP:
-                    ctrl = lastCtrl;
+                    ctrl = WindowWithFocus;
                     break;
 
                 case SDL.SDL_EventType.SDL_WINDOWEVENT:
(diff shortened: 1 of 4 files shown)


Eideren:
Thanks a lot for this PR !
I'll go over the questions you made and we'll see where to go from there.
> Is FingerSDL a good name or an alternative like TouchSDL should be used?

Touch input for other platforms are named ``Pointer[Platform]``, even UWP so we should perhaps keep the same naming scheme even if it is somewhat confusing.

> Touch also generates mouse events. There seems to be an option in SDL to disable mouse emulation but this might also be done on OS level. Also it's not always a bad thing, just something to keep in mind.

I would have said that we should try to normalize the behavior across platforms, emulate if android and iOS does so but you should probably disable it instead: devs could be using touch input and mouse input as independent input sources, ex: touch for player one / mouse for player two.
Let's leave that responsibility on the dev using xenko or the OS running the application.

As for the tittle of this PR, this is not really/only windows but rather other unixes since those are run under SDL while windows by default uses winforms/UWP afaik, perhaps change your title to reflect that ? I'm really nitpicking here but it would help if people where to look/search for this PR in the future.

It looks like this PR is ready but just in case, did you deal with the issue you mentioned over on Discord ? If it's not quite ready yet, please mark it as WIP and we'll come back to it when you're ready !

dominikjancik:
> Touch input for other platforms are named `Pointer[Platform]`, even UWP so we should perhaps keep the same naming scheme even if it is somewhat confusing.

Latest commit renames FingerSDL to PointerSDL (along with other occurences).

> I would have said that we should try to normalize the behavior across platforms, emulate if android and iOS does so but you should probably disable it instead: devs could be using touch input and mouse input as independent input sources, ex: touch for player one / mouse for player two.
> Let's leave that responsibility on the dev using xenko or the OS running the application.

Thing is I'm not entirely sure how to disable this behavior. In SDL the SDL_touch.c contains mouse synthesis code which should react to the `SDL_HINT_TOUCH_MOUSE_EVENTS` hint but in my testing this made no difference. This hint is not even documented in the SDL wiki.

Another way to disable this seems to be directly setting `mouse->touch_mouse_events` field of SDL mouse. This is however low level functionality not wrapped by the C# wrapper. It's also seems to be possible to recompile SDL without this code completely by removing the `SYNTHESIZE_TOUCH_TO_MOUSE` and then use the output instead of the original dll. However I didn't test either of these so I don't even know if it changes anything.

In my test game I'm filtering the touch generated pointer events using `event.Device.Name` which is however platform specific. The gesture recognizer is probably getting all inputs - touch and overlaid synthesised mouse - which might be slightly throwing it off.

May be useful to add some cross-platform source tag enum to IPointerDevice/IInputDevice to allow easier filtering?
> As for the tittle of this PR, this is not really/only windows but rather other unixes since those are run under SDL while windows by default uses winforms/UWP afaik, perhaps change your title to reflect that ? I'm really nitpicking here but it would help if people where to look/search for this PR in the future.

Changed.

> It looks like this PR is ready but just in case, did you deal with the issue you mentioned over on Discord ? If it's not quite ready yet, please mark it as WIP and we'll come back to it when you're ready !

If you mean the lastCtrl Window reference stuff - that is replaced by using the WindowWithFocus reference which is working well as far as I can tell.

Eideren:
> May be useful to add some cross-platform source tag enum to IPointerDevice/IInputDevice to allow easier filtering?

Let's go with that for now.
It would be nice if we could implement that missing feature over SDL-CS' repo and pull those changes from there but I totally understand if you don't have the time to do so.

phr00t:
Hey guys, now that you've renamed it to PointerSDL -- think of VirtualButton.Pointer:

dominikjancik:
> Let's go with that for now.
> It would be nice if we could implement that missing feature over SDL-CS' repo and pull those changes from there but I totally understand if you don't have the time to do so.

Turns out it's not necessary to implement anything in SDL2-CS, I was just using the nuget version in my test app that is actually really outdated.

This line disables the propagation of touch to mouse events:
`SDL.SDL_SetHint(SDL.SDL_HINT_TOUCH_MOUSE_EVENTS, "false");`

The gesture recognition also becomes much more reliable with this.

In the latest commit I placed this inside the PointerSDL constructor, logic being that whenever that is used the events shouldn't be passed to the mouse anymore. Question is whether to let game developers choose if they want dedicated touch handling or not.

> VirtualButton Pointer should use this Pointer, or they should have different names.

This VirtualButton Pointer should in theory see mouse and touch inputs just the same. They're both inheriting from `PointerDeviceBase` that implements `IPointerDevice` and keeps track of the pressed and released pointers. InputManager should provide the VirtualButton with all Pointers.

xen2:
@Eideren @dominikjancik @phr00t If I understood correctly, this one is ready to merge?

Eideren:
@xen2 Yup

dominikjancik:
🎉🎉🎉

PR #528, opened 2019-10-15
Master 3.0 zn

DogGodGit:
＃ 中文翻译

<！---主要是对po文件翻译，增加zh文件夹->

## 说明

<！---主要是对po文件翻译，增加zh文件夹->

-[]我的更改需要更改文档。
-[]我添加了测试以涵盖我的更改。
-[]所有新的和现有的测试都通过了。


Kryptos-FR:
@DogGodGit Is that a joke? Did you really expect that Xenko will suddenly stop supporting English and convert everyone to Chinese?

If your goal was to contribute Chinese translation of the editor, I'm afraid that is not the right way.

Eideren:
@Kryptos-FR Okay, let's not jump the boat, hanlon's razor, the author might not have intended to replace english by chinese.

@DogGodGit 
Your PR is not usable as the changes you have pushed contains old versions of those files, others that shouldn't be localized and also contains a lot of other resource files like PNGs that definitely don't need to be changed as part of localization. But fixing those issues won't help, the right way to contribute to the localization of Xenko is by contributing through weblate, here's how you can do so https://github.com/xenko3d/xenko/blob/master/docs/localization.md

Kryptos-FR:
@Eideren you're right. I should avoid sarcasm. Not everyone might take it well.

As for the incorrect file versions, it might be because the branch was created out of an older 3.0 branch. But it targets master, maybe there would be less conflicts if it was targeting master-3.0 instead as the title suggests? Anyway can't take it as-is, unfortunately.

xen2:
Closing (cf @Eideren explanation)

PR #535, opened 2019-10-23
[LIB] Test executable PATH when preloading libraries

Eideren:
# PR Details
If the given library wasn't found through the default paths, go through each paths set in the local executable's PATH variable.

## Description
See above.

## Related Issue
None.

## Motivation and Context
This mirrors how windows loads missing dlls and enables custom directory layouts (ex:one executable in root, dotnet dlls inside ``bin/`` and native libs inside ``bin/x64`` / ``bin/x86`` by modifying PATH at runtime).

## Types of changes

## Checklist

### DIFF
--- a/sources/core/Xenko.Core/Native/NativeLibrary.cs
+++ b/sources/core/Xenko.Core/Native/NativeLibrary.cs
@@ -133,11 +133,13 @@ namespace Xenko.Core
 
         static void NormalizeLibName(ref string libName)
         {
+#if XENKO_PLATFORM_WINDOWS_DESKTOP
             libName = libName.ToLowerInvariant();
             if (libName.EndsWith(".dll") == false)
             {
                 libName += ".dll";
             }
+#endif
         }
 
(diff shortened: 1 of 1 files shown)


sources/core/Xenko.Core/Native/NativeLibrary.cs
```diff
@@ -108,6 +131,17 @@ public static void UnLoadAll()
 #endif
         }
 
+        static void NormalizeLibName(ref string libName)
```
Kryptos-FR: Missing `private` access modifier.

sources/core/Xenko.Core/Native/NativeLibrary.cs
```diff
@@ -108,6 +131,17 @@ public static void UnLoadAll()
...
+        {
+#if XENKO_PLATFORM_WINDOWS_DESKTOP
```
Kryptos-FR: Let's exclude the whole method from non-Windows build. That will remove one entry from the type metadata.

PR #546, opened 2019-10-29
Workaround for Exception which caused Crash when Entering an invalid value in fields

Usman4D:
…olor value in the Diffuse Map https://github.com/xenko3d/xenko/issues/534

# PR Details

## Types of changes

## Checklist


phr00t:
Thank you for not only pointing this out, but providing at least a temporary workaround. We should encourage assistance here, even if it isn't ideal or complete solutions. Yes, validating input it a great idea and should be on the "to do" list. Until then, I don't want my GameStudio to crash unexpectedly from a simple typo and have worked this into my fork in the meantime.

Issue #494, opened 2019-08-08
Heightfield collider in Game Studio


EternalTamago:
**Is your feature request related to a problem? Please describe.**
HeightfieldColliderShape can be created only at runtime.
It's not available in Game Studio and Navigation.

**Describe the solution you'd like**
Implement HeightfieldColliderShapeDesc like other collider shape descs.
I tried it and I think that there are some points to implement it.

+ A. What type of source should we use to initialize the heights?
New asset or texture?

+ B. Which type of desc should it be, asset type or inline type?
Asset type(IAssetColliderShapeDesc):
It can use texture asset as initial heights like ConvexHullColliderShapeDesc does but can not be scaled in a scene.
Inline type(IInlineColliderShapeDesc):
It can be scaled in a scene but can not use texture as initial heights because to read texture at runtime needs CommandList.

+ C. How to update NavigationMeshCachedObject with updated heights?
When heightfield's heights were updated at runtime, NavigationMeshCachedObject should be updated at rebuilding dynamic navigation mesh.
However, NavigationMeshBuilder does not have a way to know whether or not the heights were updated.

IMHO,

+ A. Add new asset named HeightmapAsset that uses an image file as a source. (use R channel).

+ B. Inline type(IInlineColliderShapeDesc).

+ C. ~~Add a method named Refresh to update hashcode of HeightfieldCollliderShapeDesc.
Call it before rebuilding dynamic navigation mesh.
This might be a temporary solution but seems to work.~~

Here is [a branch](https://github.com/EternalTamago/xenko/tree/feature_3.1_heightfield_in_game_studio) that is based on the above.
I would like to pull-request it if the above has no problems.
I'm fine if the branch is thrown away by other ideas.

**Describe alternatives you've considered**
As first step, we can implement HeightfieldColliderShapeDesc without supporting initial heights and dynamic navigation mesh.

**Additional context**
I'm not sure that we can say "HeightfieldColliderShape is stable."
For example, I found a problem with collision between heightfield and cylinder.
Some cylinder(small? thin?) passes through the heightfield.
Using other collider shape type instead of cylinder is only way to fix it, as far as I know.
Anyway, it should be posted as new issue.

meriaizen86:
Cool! I really hope @xen2 merge your branch! Excellent work!

Aggror:
A pull request is more likely to be looked at then an 'enhancement issue.

EternalTamago:
Thank you.
I will pull-request that.

PR #561, opened 2019-11-28
Expose HeightfieldColliderShape to use it in Game Studio

EternalTamago:
Updated 2020/03/31

# PR Details

Expose HeightfieldColliderShape to use it in Game Studio.

scene_heightfield

* Add HeightmapAsset used to set initial heights of heightfield.
* Add HeightfieldColliderShapeDesc to expose HeightfieldColliderShape in Game Studio.
* Add AlwaysUpdateNaviMeshCache property to StaticColliderComponent to update the cache at rebuilding navigation mesh.

### HeightmapAsset
Used to initialize heights of heightfield.

add_heightmap

heightmap_asset_thumbnail

heightmap_asset_property

#### Source
An image file used to generate heightmap from its R channels.

#### Floating Point Component Range
Determine the range of the floating-point component when the R component is the floating point component.

#### Is Symmetric Short Component
Determine the range of the signed short integer component when the R component is the signed short integer component.
If enabled, [-32767 .. 32767], otherwise [-32768 .. 32767].

#### s RGB sampling
Enable if needed to load the image file as sRGB.

#### Height
Type of height.

##### Height Range
X is min height and Y is max height. "height * HeightScale" should be in this range. In Byte height type, the range can't include both of positive and negative.

##### Height Scale
In Short type or in Byte type, "height * HeightScale" is used for collisions in physics. In Float type, heights are used as they are.

|item|description|
|:---|------|
|Auto|HeightScale Calculated from HeightRange|
|Custom|Custom HeightScale|

#### Resize
New size of the heightmap. X is width and Y is length. If disabled, the size is same to the image size(width * height). Width and length should be greater than or equal to 2.

#### Scale To Height Range
If enabled, scale each of the heights to the height range before they are stored as heightmap. By the default, they are considered to be in [-1 .. 1] when the R component is floating-point component format. Edit FloatingPointComponentRange to actual floating point component range(e.g. [0 .. 1]), if needed.

### HeightfieldColliderShapeDesc

Used to create HeightfieldColliderShape.

heightfield_inline_property

#### Source
The source to initialize heights.

##### Heightmap
The heightmap to initialize heights.

##### Float, Short and Byte
Type of height.

###### Size
Size of the heightfield. X is width and Y is length. Width and length should be greater than or equal to 2. The Size should be 65 * 65 when you want 64 * 64 size in a scene.

###### Height Range
X is min height and Y is max height. "height * HeightScale" should be in this range. In Byte height type, the range can't include both of positive and negative.

###### Height Scale
In Short type or in Byte type, "height * HeightScale" is used for collisions in physics. In Float type, heights are used as they are.

|item|description|
|:---|------|
|Auto|HeightScale Calculated from HeightRange|
|Custom|Custom HeightScale|

#### Flip Quad Edges
Change how triangles forming heightfield collider shape are generated.(I'm not sure about this)

#### Centering
Add offset to ColliderShape.LocalOffset in order to center specific height.

### StaticColliderComponent.AlwaysUpdateNaviMeshCache
Enable if you want to update the cache of the collider shape at rebuilding the NavigationMesh.

Use one of the following arrays to update the heights at runtime.
|HeightType|Heights of heightfield|
|:---|--|
|Float|HeightfieldColliderShape.FloatArray|
|Short|HeightfieldColliderShape.ShortArray|
|Byte|HeightfieldColliderShape.ByteArray|

## Related Issue

#494

## Motivation and Context

Currently there is HeightfieldColliderShape but it can be created only at runtime. It's not available in Game Studio and NavigationMesh.

## Types of changes

## Checklist


HeadClot:
I love this PR.

sources/engine/Xenko.Physics/Engine/Heightmap.cs
```diff
@@ -0,0 +1,82 @@
...
+            var type = data.GetType();
+
+            if (type == typeof(float[]))
+            {
+                return new Heightmap
+                {
+                    HeightfieldType = HeightfieldTypes.Float,
+                    Width = width,
+                    Length = length,
+                    Floats = new List<float>(data as float[]),
```
xen2: Wouldn't it make more sense to make `Heightmap.Floats`, `Heightmap.Shorts` and such directly as arrays, and avoid the two extra copy? (one copy here and one copy in `CreateHeights` because of `ToArray()`)

sources/engine/Xenko.Physics/Shapes/HeightfieldColliderShape.cs
```diff
@@ -48,7 +68,7 @@ bool flipQuadEdges
             HeightStickWidth = heightStickWidth;
             HeightStickLength = heightStickLength;
             HeightType = heightType;
-            HeightScale = heightScale;
+            HeightScale = (HeightType == HeightfieldTypes.Float) ? 1f : heightScale;
```
xen2: Seems strange/misleading to ignore this parameter only in case of float?

sources/engine/Xenko.Physics/Engine/Heightmap.cs
EternalTamago: Yes, those should be arrays.

sources/engine/Xenko.Physics/Shapes/HeightfieldColliderShape.cs
EternalTamago: I revert that.
That is unnecessary because HeightfieldColliderShapeDesc does it.

xen2:
Tried with a heightmap but couldn't make it work properly:

image

I tried float/short/byte (set to same value on both component collider shape's list and heightmap asset).
Anything I might have forget?

Also small remark, maybe `HeightmapAsset.Size` might be better to default to something meaningful like 1024 rather than 0? (even if not enabled)

EternalTamago:
It seems that Heightmap size is not same to HeightStickSize.
Does that happen even if those are matched? (Enable Resize of HeightmapAsset and set it to 80x80)
Heightfield uses heightmap as it is, that doesn't try to match the size.

>HeightmapAsset.Size

Yes, I think so too.
It might be better that that is same to default value of HeightStickSize?

EternalTamago:
Would it better to change the list of the property to this?

another_property_type


BTW, I realized that HeightfieldColliderShape.FloatArray/ShortArray/ByteArray should be locked while building the navigation mesh.

xen2:
@EternalTamago By this change, all the size/type info would come from the collider shape instead of being in two places, right?
If yes, that would be ideal! (I was going to suggest this too so that it is more user-friendly)

EternalTamago:
@xen2 I didn't quite understand what "from the collider shape" means.
HeightStickSize and HeightType will be from the Heightmap when the Heightmap is selected at InitialHeights.

I am currently encountering an access violation exception that I do not know the cause of.
The exception occurs in btCollisionWorld_rayTest (called in the PlayerInput script of the TopDownRPG project) after rebuilding the DynamicNavigationMesh.
Or the game will be killed without that exception at similar situation.
It does not always happen.
So far, this only happens when there is a CompoundColliderShape that contains a HeightfieldColliderShape.

EternalTamago:
I think that the problem has been fixed.
The cause was that the collider shape was recreated at rebuilding NavigaitonMesh.

xen2:
Works much better for me in new version!

Noticed a few things that either didn't work or that I might have misunderstood:
* `HeightfieldColliderShapeDesc.HeightScale` doesn't work for float
* `HeightfieldColliderShapeDesc.HeightScale` behave strangely for non-float: enabled and not-enabled with size 1 result in different (I think enabled results in 255)
* If I change `HeightfieldColliderShapeDesc.HeightRange` to [0..10], I see everything under the Y axis ground plane (Y < 0) where I would expect it to be between Y=0 and Y=10.

EternalTamago:
@xen2 

1. HeightfieldColliderShapeDesc.HeightScale is used for Short or Byte.
When Float, HeightfieldColliderShapeDesc.HeightScale is not used because heightfield uses FloatArray as it is.
**Update**: Changed display name of HeightfieldColliderShapeDesc.HeightScale.

2. When disabled, HeightScale was set internally.
**Update**: Changed type of HeightScale of desc to be better.

3. Recentering happens when the HeightRange is not symmetric.
**Update**: Added an option to add extra offset to the local offset to counteract recentering.

EternalTamago:
Sorry to change many times after review request.

I would like to move `HeightfieldColliderShapeDesc.HeightRange` and `HeightfieldColliderShapeDesc.HeightScale` to `HeightmapAsset`.
That might make it more friendly.
What do you think about that?

xen2:
@EternalTamago I think it makes sense UX-wise to move those two properties on the asset itself.

EternalTamago:
@xen2 Thanks for the reply.
I will modify this once more.

EternalTamago:
@xen2 Sorry for the delay!
I've finished that.

EternalTamago:
I have started thinking GetStaticColliderCacheSettings is bad method.

To decide whether to update the navigation mesh cache object at runtime,
it might be better to add something like AlwaysUpdateNaviMeshCache property to StaticColliderComponent instead of adding StaticColliderCacheSettings to StaticColliderData.
However, the property will be useless for the physics and that will be used only at building/rebuilding NavigationMesh.

EternalTamago:
@xen2 I'm sorry to bother you again.

I will try to add AlwaysUpdateNaviMeshCache to StaticColliderComponent in order to decide wthether to update the cache at rebuilding NavigationMesh.
And GetStaticColliderCacheSettings will be removed.

Also I'd like to update parts of HeightfieldColliderShapeDesc in order to be easy to understand and to be able to set initial height value that might be needed when the height range is not include 0.

xen2:
@EternalTamago Hey, in few days Xenko will be renamed to Stride so it's easier to merge PRs before that. Let me know if you have a working version that can be merged already.
Otherwise no worries, I will guide you on how to make sure PR don't conflict with the renaming whenever you are ready.

EternalTamago:
@xen2 Yes. Sorry to late response!

EternalTamago:
@xen2 The current state is everything. My words were not enough. Sorry if you are blocked by me.

xen2:
@EternalTamago 
Thanks for clarification! No worries, I understood it was complete with previous messages.
I wanted to test it quickly before merging, sorry for the delay.

Aggror:
@EternalTamago Hi EternalTamago,
You wouldn't happen to have a nice video recording that demonstrates the heightmap field with a nice example scene? I want to compile a nice video containing the cool new features for the Stride 4.0 release. 
If you have anything resource wise that can help out with showcasing your heightfielf PR, you can contact me on Discord #Aggror#6358 or here on Gitbhub.

EternalTamago:
Hi, @Aggror 
I'll try to record simple video with Stride sample assets etc in a few days.
(You don't have to wait for it when we don't have the time.)

Kryptos-FR:
@EternalTamago @Aggror 
Digging some old tweets 😉 :
https://twitter.com/EternalTamago/status/1045021643620274176
https://twitter.com/EternalTamago/status/1045742093786787840

Aggror:
Note that we will now just use small videos in the release notes instead of one long video.

EternalTamago:
@Aggror 
I have sent the video to you via DM in Discord.

@Kryptos-FR 
New video may be same as the above mostly except that bitrate 😄

EternalTamago:
@xen2 
Recently I posted separately #734 from #494.
And I'd like to know what you think about #734 and the heightfield.
That is related to the stability of Stride.

For me, the heightfield collider shape in Game Studio is useful even if there is that issue.
However, the issue might be critical for some users that try to use the heightfield.
What do you think about this?

I know that it's too late to talk, but I think that we should get clarity on this before the release.
I'm sorry to bother you.

PR #562, opened 2019-12-05
[Threading] Reduce dispatcher allocations

Eideren:
# PR Details
One day I'll have to stop fiddling with the ``ThreadPool``, but not today !
Greatly reduces allocations of calls to the ``Dispatcher``, overall per frame allocation on phroot's test scene where reduced from 22k to 6k. Also slight increase in performances overall.

I had to change most function signatures to handle those changes, which might force users of ``Dispatcher`` to adapt to those new ones, I'm not sure whether to classify that as a breaking change.

## Related Issue
None.

## Motivation and Context
Performances.

## Types of changes
- [x] Breaking change ? See PR Details (fix or feature that would cause existing functionality to change)

## Checklist

### DIFF
--- a/sources/core/Xenko.Core/Threading/ConcurrentFixedPool.cs
+++ b/sources/core/Xenko.Core/Threading/ConcurrentFixedPool.cs
@@ -34,7 +34,7 @@ namespace Xenko.Core.Threading
                 Items = new T[size],
                 Mask = size - 1,
             };
-            Push(new T());
+            TryPush(new T());
         }
 
         public T Pop()
@@ -52,12 +52,12 @@ namespace Xenko.Core.Threading
             // The entire logic works on the guarantee that there is
             // always at least one item within the collection.
             if (iterator <= 0)
-                Push(new T());
+                TryPush(new T());
 
             return item;
         }
 
-        public void Push(T item)
+        public bool TryPush(T item)
         {
             var localHead = head;
             int iterator;
@@ -67,7 +67,7 @@ namespace Xenko.Core.Threading
             {
                 iterator = Volatile.Read(ref localHead.Iterator);
                 if (iterator >= localHead.Items.Length)
-                    return;
+                    return false;
                 if (Interlocked.CompareExchange(ref localHead.Iterator, iterator + 1, iterator) == iterator)
                     break;
                 spin.SpinOnce();
@@ -78,6 +78,7 @@ namespace Xenko.Core.Threading
             {
                 spin.SpinOnce();
             }
+            return true;
         }
     }
 }
--- a/sources/core/Xenko.Core/Threading/ConcurrentFixedPool.cs
+++ b/sources/core/Xenko.Core/Threading/ConcurrentFixedPool.cs
@@ -65,7 +65,7 @@ namespace Xenko.Core.Threading
             var spin = new SpinWait();
             do
             {
-                iterator = Volatile.Read(ref localHead.Iterator);
+                iterator = localHead.Iterator;
                 if (iterator >= localHead.Items.Length)
                     return false;
                 if (Interlocked.CompareExchange(ref localHead.Iterator, iterator + 1, iterator) == iterator)
--- a/sources/core/Xenko.Core/Threading/ConcurrentFixedPool.cs
+++ b/sources/core/Xenko.Core/Threading/ConcurrentFixedPool.cs
@@ -1,0 +1,83 @@
+// Copyright (c) Xenko contributors (https://xenko.com)
+// Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using System;
+using System.Threading;
+
+namespace Xenko.Core.Threading
+{
+    /// <summary>
+    /// A stack of fixed size, that size must be a power of two,
+    /// always contains at least one item.
+    /// Fastest collection type under heavy concurrency.
+    /// </summary>
+    public class ConcurrentFixedPool<T> where T : class, new()
+    {
+        private class Segment
+        {
+            public T[] Items;
+            public int Mask;
+            public int Iterator;
+        }
+
+        private readonly Segment head;
+
+        /// <exception cref="ArgumentOutOfRangeException"><paramref name="size"/> must be a power of two</exception>
+        public ConcurrentFixedPool(int size)
+        {
+            // Size must be a power of two for modulo and overflow of read/write indices to behave correctly
+            if (size <= 0 || ((size & (size - 1)) != 0))
+                throw new ArgumentOutOfRangeException(nameof(size), "Must be power of two");
+
+            head = new Segment
+            {
+                Items = new T[size],
+                Mask = size - 1,
+            };
+            Push(new T());
+        }
+
+        public T Pop()
+        {
+            var spin = new SpinWait();
+            var localHead = head;
+            var iterator = Interlocked.Decrement(ref localHead.Iterator);
+
+            var index = iterator & localHead.Mask;
+
+            T item;
+            while ((item = Interlocked.Exchange(ref localHead.Items[index], null)) == null)
+                spin.SpinOnce();
+            
+            // The entire logic works on the guarantee that there is
+            // always at least one item within the collection.
+            if (iterator <= 0)
+                Push(new T());
+
+            return item;
+        }
+
+        public void Push(T item)
+        {
+            var localHead = head;
+            int iterator;
+
+            var spin = new SpinWait();
+            do
+            {
+                iterator = Volatile.Read(ref localHead.Iterator);
+                if (iterator >= localHead.Items.Length)
+                    return;
+                if (Interlocked.CompareExchange(ref localHead.Iterator, iterator + 1, iterator) == iterator)
+                    break;
+                spin.SpinOnce();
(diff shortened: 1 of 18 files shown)


sources/core/Xenko.Core/Threading/ConcurrentFixedPool.cs
```diff
@@ -0,0 +1,84 @@
...
+using System.Threading;
+
+namespace Xenko.Core.Threading
+{
+    /// <summary>
+    /// A stack of fixed size, that size must be a power of two,
+    /// always contains at least one item.
+    /// Fastest collection type under heavy concurrency.
+    /// </summary>
+    public class ConcurrentFixedPool<T> where T : class, new()
```
Kryptos-FR: Wondering if `new()` generic constraint should be removed. Its only purpose here is to be able to do `new T()` on lines 37 and 55.

As far as I can remember, that constraint is translated by the compiler with using `Activator.CreateInstance` which uses reflection and is slower. The following article seems to confirm that: https://devblogs.microsoft.com/premier-developer/dissecting-the-new-constraint-in-c-a-perfect-example-of-a-leaky-abstraction/

@Eideren can you try to remove it and instead pass a *factory* method to the ConcurrentFixedPool constructor such as:

```csharp
// constructor signature
public ConcurrentFixedPool(int size, Func<T> factory) { }

// call
var pool = new ConcurrentFixedPool<T>(42, () => new T());
```

Then check whether performance and/or allocations are changed between the two versions.

sources/core/Xenko.Core/Threading/ConcurrentFixedPool.cs
```diff
@@ -0,0 +1,84 @@
...
+
+        public bool TryPush(T item)
+        {
+            var localHead = head;
+            int iterator;
+
+            var spin = new SpinWait();
+            do
+            {
+                iterator = Volatile.Read(ref localHead.Iterator);
```
Kryptos-FR: `Interlocked.CompareExchange` used below guarantees that this read is not re-ordered. I don't know if the compiler or JIT is smart enough to not emit the corresponding memory barrier so maybe removing it would be better.

Reference: http://www.albahari.com/threading/part5.aspx#_SpinLock_and_SpinWait

sources/core/Xenko.Core/Threading/ConcurrentFixedPool.cs
Eideren: It's actually far slower, I tested it before and that's why I defaulted to new() instead

sources/core/Xenko.Core/Threading/ConcurrentFixedPool.cs
Eideren: You're right, I didn't really think this one through, thanks !

sources/core/Xenko.Core/Threading/ConcurrentFixedPool.cs
Kryptos-FR: Weird. The article from the Microsoft engineer seems to indicate the opposite.

Eideren: I'll try it again just in case I missed something.

Eideren: I'm pretty sure I didn't test it under release at the time, it's hard to loose the habit of just F5'ing after any changes. Anyway, in debug mode the average runtime of a pop->push is about 2% slower with a factory. The run to run variant is not even half a percent so that 2% is not within margin of error.
In release, factory is about 6% faster. I don't test instance creation specifically because it is supposed to be an exceptional case, this type is supposed to always have more items than required so that branch is not supposed to be hit, it is only there as a safeguard to keep the logic working. I'll push those changes asap.

xen2:
@Eideren should be ready for merge now, or some work still pending?

Eideren:
Sorry, no, I have a lot of fixes that have not yet been pushed here yet. I'll ping you once it's ready

Eideren:
@xen2 Okay, I'm done, you can review/merge.

xen2:
@Eideren I was looking into the changes you did to the dispatcher calls.
However I am not sure which allocations you are trying to reduce. For example, here is `RootEffectRenderFeature.PrepareEffectPermutations` disassembled (notice the `<pool>.Acquire()`):
```
	<>c__DisplayClass55_0 <>c__DisplayClass55_ = <>c__DisplayClass55_0.<pool>.Acquire();
	<>c__DisplayClass55_.AddReference();
	<>c__DisplayClass55_.<>4__this = this;
	base.PrepareEffectPermutations(context);
	<>c__DisplayClass55_.renderEffects = RenderData.GetData(RenderEffectKey);
	<>c__DisplayClass55_.effectSlotCount = EffectPermutationSlotCount;
	Dispatcher.ForEach(base.RenderSystem.Views, <>c__DisplayClass55_.<delegate><PrepareEffectPermutations>b__0);
	PrepareEffectPermutationsImpl(context);
	PrepareRenderTargetExtensionsMixins(context);
	<>c__DisplayClass55_.currentTime = DateTime.UtcNow;
	Dispatcher.ForEach(RenderObjects, <>c__DisplayClass55_.<delegate><PrepareEffectPermutations>b__1);
```
The assembly processor already get the lambda class from a pool (logic in `DispatcherProcess.cs`).

I probably missed something and you focused on some other allocations maybe?

Eideren:
@xen2 That's what it is trying to do, sure, but I don't think it actually works afaict:
Here's vs alloc monitoring for pre-PR:
Untitled
Same but after applying this PR:
Untitled2
Note that I had to re-install visual studio between the two runs, that's why the interfaces differ.
Unless I'm interpreting that data incorrectly the pool fails somewhere to properly avoid allocation for whatever reason.

Now we could try to figure out why the current pooling system doesn't work but I don't yet have the kind of experience it would require to efficiently debug this issue so I can't really help you if you want to take that path instead.

Eideren:
@xen2 From a glance you seem to have fixed most of the issues through 9a6f630. I'll test them out and adapt this PR accordingly. 
Even if the allocs are fixed I still want to test out whether the way I schedule jobs in the dispatcher is faster and if the threadpool changes help.

xen2:
It seems `DispatcherProcessor` has trouble to pool delegates in some specific cases.
I have fixed part of the issues with https://github.com/xenko3d/xenko/commit/9a6f63004d7f653718b86cd4baedfd1ab64f4f77

There are still various other cases of `Dispatcher` lambda alloc that still are not optimized properly.
One example of that is when there is a cast to another Action/Func type.
But I think they should/can probably fixed in `DispatcherProcessor` as well.

xen2:
@Eideren Cool, I was wondering if those `ThreadPool` needed to be separate.

Eideren:
@xen2 Does the ``DispatcherProcessor`` handle delegates that captures both ``this`` and other external variables that are called within the function, that's a fairly common scenario across the code base and it doesn't seem to work ?
Ex: 
[RenderSystem.Draw()](https://github.com/stride3d/xenko/blob/878ddc01b15e04de2162b5d3e613c6a539725e97/sources/engine/Xenko.Rendering/Rendering/RenderSystem.cs#L368) generates a closure, probably because of ``batchSize``.

[RootEffectRenderFeature.Prepare()](https://github.com/stride3d/xenko/blob/878ddc01b15e04de2162b5d3e613c6a539725e97/sources/engine/Xenko.Rendering/Rendering/RootEffectRenderFeature.cs#L675) also generates a closure, probably for ``renderEffects`` ``view``, and/or ``viewFeature`` and a ``Func<RootEffectRenderFeature.PrepareThreadContext>``

[ForwardLightingRenderFeature.Prepare()](https://github.com/stride3d/xenko/blob/878ddc01b15e04de2162b5d3e613c6a539725e97/sources/engine/Xenko.Rendering/Rendering/Lights/ForwardLightingRenderFeature.cs#L408) closure probably for ``viewIndex``.

Do you think those issues (along with generics support and other todos within the ``DispatcherProcessor``) can be resolved through that system, if so we'll leave the rest of those issues on the back-burner for now, I'll just remove the code that avoids allocations in this PR, test out the rest and hopefully merge.
Or will those issues require too much effort and so we should perhaps consider using the system in this PR even if it is fairly clunky ?

xen2:
@Eideren `RootEffectRenderFeature.Prepare()` is the one I started to try to optimize after and the pattern was little bit more complex (it creates two internal classes and one reference another through a field). Hopefully the two others are probably similar.

I think it should be feasible (not 100% sure but almost). However, I realized this case was little bit more subtle than I though so I had to put it aside for now (I tried quickly but realized it was starting to delay the renaming which is top priority).

I would prefer much more if we did it in the assembly processor automatically so that users can use any kind of lambda easily without signature/capture restrictions (which would be difficult to detect/enforce anyway).
Also I realized it's probably possible to emit a warning for the cases the system can't optimize, so that we are at least aware of those much more easily (and user can temporarily avoid them if user-code).

Eideren:
@xen2 Sure, I'll start filtering out those changes tomorrow then, thanks !

Eideren:
I'll close this PR and continue this issue through PR #623 instead.

Issue #393, opened 2019-02-24
Typing the name of an incompatible asset type in a property editor field causes a crash.


dfkeenan:
**Release Type**: Official Release & GitHub

**Version**: >= 3.0 (Possibly all)

**Platform(s)**: Windows?

**Describe the bug**
When typing the name of an asset in the text box of an asset property in the property inspector window that is of a different type than the one expected it causes a crash.

**To Reproduce**
1. Create a script with a property of type `Scene` Called "NextScene".
2. Add script to an entity.
3. Create a model asset called "Sphere".
4. Type the name "Sphere" in the textbox of the "NextScene" property.
5. CRASH

**Expected behavior**
It should handle it gracefully. The same way it does if you try to Drag and Drop an asset. Which sets the property to null.

**Screenshots**
If applicable, add screenshots to help explain your problem (use GitHub drag & drop)

**Log and callstacks**
If any, please attach here any log or callstack (preferably in a `.txt` file using GitHub drag and drop)

**Additional context**
The exception is thrown at https://github.com/xenko3d/xenko/blob/293d60a2bdc4b6d2773a1da914fd802483828288/sources/presentation/Xenko.Core.Presentation.Quantum/ViewModels/NodeViewModel.cs#L538

It should possibly return null if it cannot convert it.  I tried this fix and it gave the expected behavior as above. I am not sure if this would cause any other issues though.

profan:
Can confirm I ran into this just yesterday when pasting a color code into a box on a material property:
![image](https://user-images.githubusercontent.com/2816340/53297592-bf3bff80-3820-11e9-99a8-25652385e031.png)

(typing in an invalid property there, like just taking the same text and omitting the # causes a crash)

profan:
Any idea @xen2 if solving this by returning null if it can't convert it would cause any far-reaching consequences?

Just ran into this again today:
![image](https://user-images.githubusercontent.com/2816340/57339960-1836e680-7134-11e9-80e4-34055157b790.png)

xen2:
@profan Agree it definitely shouldn't crash the editor.
Maybe a better approach would be to display no editor (or an error message explaining it can't be edited, possibly with a button to get more details/actual exception).
That way we are still aware when it's broken, and editor shouldn't crash anymore.

xen2:
@profan need to investigate if `null` has no side effect.
If it doesn't set the `null` value and stop right there, that's probably fine.
[...]

PR #391, opened 2019-02-23
Add UrlReference

dfkeenan:
# PR Details

This is an attempt to implement #390.

## Description

Add new type `UrlReference` and related engine and Game Studio changes.

Things to-do:

- [x] Drag and Drop.
- [x] Select an asset button.
- [x] Clear reference button.
- [x] Select the referenced asset button.
- [ ] Make collections of `UrlReference` work. May need to change `AddNewItemCommand`. Or making it so `UrlReference` have a default constructor, which I will do if we end-up revmoving `AssetId` from it.
- [ ] Using the text box to type the name of assets causes an exception. Bug also exists for existing asset functionality. See #393 .
- [ ] `UrlReference` being treated as a real asset reference so asset gets included in build and things like renaming assets works the same as for existing asset functionality.
- [ ] Unit tests.
- [ ] Documentation.

## Related Issue

#390

## Motivation and Context

See #390.

## Types of changes

- [ ] Docs change / refactoring / dependency upgrade
- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [X] My change requires a change to the documentation.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

Issue #390, opened 2019-02-23
Create a UrlReference type so Asset urls can be set in Game Studio


dfkeenan:
**Is your feature request related to a problem? Please describe.**
Currently when using asset URLs to load content (i.e. `Content.Load<Texture>("MyTexture")`) you have to rely on getting the URL correct yourself and have to make the asset as include as root. This can make it easy to make mistakes.

**Describe the solution you'd like**
As discussed, we would like to add a new type `UrlReference`/`UrlReference<T>` and accompanying `ContentManager` methods to give the same support as adding assets directly to components.

**Describe alternatives you've considered**
Use the current `string` URL methods.

**Additional context**

PR #564, opened 2019-12-07
Add UrlReference attempt number 2

dfkeenan:
# PR Details

This is an attempt to implement #390. Replacing PR #391.

## Description

Add new type `UrlReference` and related engine and Game Studio changes.

Things to-do:

- [x] Drag and Drop.
- [x] Select an asset button.
- [x] Clear reference button.
- [x] Select the referenced asset button.
- [x] Make collections of `UrlReference` work.
- [ ] Using the text box to type the name of assets causes an exception. Bug also exists for existing asset functionality. See #393 .
- [x] `UrlReference` being treated as a real asset reference so asset gets included in build and things like renaming assets works the same as for existing asset functionality.
- [ ] Unit tests.
- [x] Documentation. See PR xenko3d/xenko-docs#25.

## Related Issue

#390

## Motivation and Context

See #390.

## Types of changes

## Checklist


dfkeenan:
For the extension methods I did for `UrlReference` I used `IContentManager`.  There are a couple of methods on the `ContentManager` class not on the `IContentManager` interface that I think would be good to make an extension method for. As far as I can tell there is only one implementation of this interface so I was wondering if it would be ok to change the interface?

The methods I am referring to are:
* `public T Get<T>(string url)`
* `public bool IsLoaded(string url, bool loadedManuallyOnly = false)`
* `public void Unload(string url)`

xen2:
Thanks for working again on this feature!

It might be good to also upgrade samples. I did a quick `git grep` of `Content.Load` (and also noticed a `EffectSystem.LoadEffect` that might benefit from it).
Note: maybe some of them can't be easily switched, I didn't look in details.
```
Games/JumpyJet/JumpyJet.Game/PipesScript.cs:            var pipeSetPrefab = Content.Load<Prefab>("Pipe Set");
Graphics/CustomEffect/CustomEffect.Game/CustomEffectRenderer.cs:            customEffect = EffectSystem.LoadEffect("Effect").WaitForResult();
Templates/TopDownRPG/TopDownRPG/TopDownRPG.Game/Gameplay/SceneStreaming.cs:                        Instance = Content.Load<Scene>(Url);
Templates/TopDownRPG/TopDownRPG/TopDownRPG.Game/Gameplay/SceneStreaming.cs:                        var localLoadingTask = loadingTask = Content.LoadAsync<Scene>(Url);
Tutorials/CSharpBeginner/CSharpBeginner/CSharpBeginner.Game/Code/TutorialUI.cs:            tutorialScene = Content.Load<Scene>("Scenes/Basics/" + newTutorialScene.Value);
UI/GameMenu/GameMenu.Game/SplashScript.cs:                SceneSystem.SceneInstance.RootScene = Content.Load<Scene>("MainScene");
UI/UIElementLink/UIElementLink.Game/SplashScript.cs:            SceneSystem.SceneInstance.RootScene = Content.Load<Scene>(NextScene);
```

dfkeenan:
You are welcome. It has been a bit of a challenge but worth it I think. I have been wanting this feature for quite some time.

I should be able to upgrade the samples some time this week. Might not be until the weekend.

Not sure about `EffectSystem.LoadEffect` though. I had a quick look it not sure how this would work. 

Regarding my proposed changes to `IContentManager` as stated above. Would this also be acceptable to include?

xen2:
Seems working well as far as I tested.
Glad you could figure it out -- and you found out a way that didn't require much change, all the better!
I will take a closer look at implementation but this should be good to go soon!

No problem to add some more method in `IContentManager`.

sources/core/Xenko.Core.Serialization/Serialization/UrlReference.cs
```diff
@@ -0,0 +1,94 @@
...
+namespace Xenko.Core.Serialization
+{
+    /// <summary>
+    /// Represents a Url to an asset.
+    /// </summary>
+    [DataContract("urlref", Inherited = true)]
+    [DataStyle(DataStyle.Compact)]
+    [ReferenceSerializer]
+    [DataSerializer(typeof(UrlReferenceDataSerializer))]
+    public class UrlReference
```
xen2: I was wondering if it wouldn't be better as a `struct`?
Especially since we also have to check for `IsEmpty` so there is two ways to have invalid values currently.
Could remove lot of null check as well.
Not sure if all the converters and editor code will like it, but might be worth a try don't you think?

sources/core/Xenko.Core.Serialization/Serialization/UrlReference.cs
```diff
@@ -0,0 +1,94 @@
...
+
+    }
+
+    /// <summary>
+    /// Represents a Url to an asset of type <see cref="T"/>.
+    /// </summary>
+    /// <typeparam name="T">The type off asset.</typeparam>
+    [DataStyle(DataStyle.Compact)]
+    [DataSerializer(typeof(UrlReferenceDataSerializer<>), Mode = DataSerializerGenericMode.GenericArguments)]
+    public sealed class UrlReference<T> : UrlReference
```
xen2: Same goes for this (`struct`).
Might need a `IUrlReference` with `Url` and `IsEmpty` to act as common base.

sources/core/Xenko.Core.Serialization/Serialization/UrlReference.cs
```diff
@@ -0,0 +1,94 @@
...
+                if (string.IsNullOrEmpty(value))
+                {
+                    throw new ArgumentNullException(nameof(value));
+                }
+                url = value;
+            }
+        }
+
+        [DataMemberIgnore]
+        public bool IsEmpty => string.IsNullOrEmpty(url);
```
xen2: When reading the code (ctor and Url setter), it seems that `Url` can't be null or Empty, so `IsEmpty` would always be false?

dfkeenan:
I don't think we can use a `struct`. That wouldn't work with the attached reference stuff. You wouldn't be able to use a `struct` as a key in `ConditionalWeakTable`.

sources/core/Xenko.Core.Serialization/Serialization/UrlReference.cs
dfkeenan: `AttachedReferenceManager` requires `objects` because it uses a `ConditionalWeakTable` to attach references. Ideally the `UrlReference` types would not have a default constructor so you could not instantiate "empty" ones. But various parts of the asset stuff and UI require a default constructor for them to work.  It would require a lot more changes to accomplish this without default constructors.

I have created an interface and a base class so the `UrlReference` and `UrlReference<T>` types can both be `sealed`. I have also tidied up a bunch of things:
* Fix some formatting (remove blank likes etc.)
* Fixed/added doc comments
* Fixed some exceptions being thrown (Had parameters around the wrong way).
* Removed/sorted using statements
* Added License notice to the tops of files

sources/core/Xenko.Core.Serialization/Serialization/UrlReference.cs
dfkeenan: Please see reply above.

dfkeenan:
As suggested I have updated some of the samples/templates. I have updated:
* JumpyJet
* TopDownRPG
* GameMenu
* UIElementLink
* And the `SceneStreaming` script template.

Updating the `SceneStreaming` script template probably means I should also update the "Load scenes" documentation.

(I did not look at the CSharpBeginner tutorials.)

I have noticed some strange behaviour. Using the `UrlReference` on an asset does make it as referenced and includes it and it's dependencies in the build but they do not all seem to get the Green indicator:

image


In this example the "Pipe Set" Prefab is referenced as a url and it includes the "Pipe" Prefab. But is not being marked as green.

xen2:
@dfkeenan this is expected behavior. A prefab is typically design-time only since there's a full copy of it in the scene.
Thanks for the samples update.
Feel free to make the PR non-draft when you believe it's ready for last review/merge.

dfkeenan:
OK. I just never played with Prefabs before I guess. I just tried a small test in 3.1 and it does this. So I didn't break it, or at least not as bad as I thought.

I will change it to ready for review now. But I really should create unit tests. 

Have you had a chance to look at the documentation PR? I had a couple of questions in there.

sources/core/Xenko.Core.Serialization/Serialization/UrlReference.cs
xen2: Good point about attached reference, makes sense. Thanks!

Wondering, could we do it the other way around and forbid to use null/empty string then?
i.e. if not set in editor, `UrlReference` instance itself is null. If instance is set, it has to have an actual valid `Url` value?

Note: if annoying/complex, no need to do that now.

dfkeenan: I did have a couple of solutions that didn't expose the default constructor that worked mostly. For example, make the default constructor `internal` and change how instantiation code works. I had a couple of different versions of this that "worked" but I wasn't too happy with the bits I changed. 

Though I might have just thought of an option....

But even still, doing it this way meant that using collection properties i.e. `List<UrlReference>` don't work because the UI code cannot instantiate an instance without a default constructor. So would either have to leave this feature out or change the UI code.

xen2: That's true, let's keep it like this for now, thanks.

samples/UI/GameMenu/Assets/Shared/SplashScene.xkscene
```diff
@@ -19,14 +19,18 @@ Hierarchy:
...
                     f6b6b1a8d48a8045bd940370455f3450: !UIComponent
                         Id: a8b1b6f6-8ad4-4580-bd94-0370455f3450
                         Page: 62c493c2-13ea-4fac-bbaf-578afafac104:SplashUI
                         Resolution: {X: 640.0, Y: 1136.0, Z: 1000.0}
                         Size: {X: 1.0, Y: 1.0, Z: 1.0}
                         ResolutionStretch: FixedWidthFixedHeight
+                    fd54344b6ac8b2ed232d62dc08820239: !GameMenu.SplashScript,GameMenu.Game
+                        Id: c7c92d4b-2c04-446f-b8c5-59e9d2b989cd
+                        NextSceneUrl: 9294c606-6173-4215-a8dc-0d5043d68333:MainScene
+                    2f704fab49b226562874d765af6b6918: !GameMenu.SplashScript,GameMenu.Game
```
xen2: I suppose it's been added twice as mistake?

xen2:
Love how the samples now look, with automatic type inference in the `Load` and the proper asset id stored in the `xkscene`!

samples/UI/GameMenu/Assets/Shared/SplashScene.xkscene
dfkeenan: Hmm, not sure about that. Hope I haven't introduced some bug.

dfkeenan: I do kind of remember having some issues while updating the samples. Game Studio wasn't playing nice when I was changing code. Maybe I tried removing/adding the component while trying to do it. I would just remove it but you all ready merged the change.

PR #512, opened 2019-09-09
[UI/INPUT] bugfix & commenting UIRenderFeature.UIElementUnderMouseCursor

CharlesWoodhill:
# PR Details

previous determination of the element was in a loop, so the last loop won (loop over UIComponents?).
now all loops are taken into account, and the last processed UIComponent that had a hit, will be taken. 
also created xml comment for the property.

## Description

moved the assignment of the public property outside and after the loop over UIComponents.

## Related Issue
#509

## Motivation and Context

it was buggy, it had to be fixed.

## Types of changes

- [ ] Docs change / refactoring / dependency upgrade
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [ ] My change requires a change to the documentation.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #573, opened 2019-12-21
Fix mouse over state

aikixd:
# PR Details

Fixed two instances where MouseOverState wasn't properly updated.

## Description

There were two instances where MouseOverState wasn't properly updated.
1. When an overred element was removed from the hierarchy (button click removes it from a panel) - With the element removed, it wasn't possible to update state of its parents and to invoke the hit test for the current overred element. Now, when element is removed from a panel and was overred, the parent state is updated to MouseOverNone. And, when when an element measure is invalidated, it is marked to require a hit test, in case it is the last overred element. This will also cover cases when the elements are animated under the cursor.
2. When more than a single UI was present in the scene - The issue was that the value for current overred element was shared in a field between different UIs, so the last rendered UI would always overwrite this value for other UIs. This is now a local variable.

**I'm not sure how to write tests for this fix.**

## Related Issue

https://github.com/xenko3d/xenko/issues/565

## Motivation and Context

Bug fixing

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Xenko.UI/Rendering/UI/UIRenderFeature.Picking.cs
+++ b/sources/engine/Xenko.UI/Rendering/UI/UIRenderFeature.Picking.cs
@@ -19,8 +19,6 @@ namespace Xenko.Rendering.UI
 
         private readonly List<PointerEvent> compactedPointerEvents = new List<PointerEvent>();
 
-        public UIElement UIElementUnderMouseCursor { get; private set; }
-
         partial void PickingUpdate(RenderUIElement renderUIElement, Viewport viewport, ref Matrix worldViewProj, GameTime drawTime)
         {
             if (renderUIElement.Page?.RootElement == null)
@@ -241,6 +239,8 @@ namespace Xenko.Rendering.UI
             var rootElement = state.Page.RootElement;
             var lastMouseOverElement = state.LastMouseOverElement;
 
+            UIElement uIElementUnderMouseCursor = lastMouseOverElement;
+
             // determine currently overred element.
             if (mousePosition != state.LastMousePosition
                 || (lastMouseOverElement?.RequiresMouseOverUpdate ?? false))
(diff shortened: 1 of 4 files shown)


Eideren:
Hey, thanks for taking the time to create this PR !

sources/engine/Xenko.UI/UIElement.Events.cs
```diff
@@ -98,6 +98,9 @@ internal set
             }
         }
 
+        [DataMemberIgnore]
+        public bool RequiresMouseOverUpdate { get; set; }
```
Eideren: Is this field useful to users ? If not you could change it to internal. If it is useful could you add a summary to it ?

sources/engine/Xenko.UI/Rendering/UI/UIRenderFeature.Picking.cs
```diff
@@ -19,8 +19,6 @@ public partial class UIRenderFeature
 
         private readonly List<PointerEvent> compactedPointerEvents = new List<PointerEvent>();
 
-        public UIElement UIElementUnderMouseCursor { get; private set; }
```
Eideren: Isn't this a breaking change ? I haven't used the UI much so perhaps its already not recommended ?

sources/engine/Xenko.UI/UIElement.Events.cs
aikixd: Added summary

sources/engine/Xenko.UI/Rendering/UI/UIRenderFeature.Picking.cs
aikixd: This is a render feature, I really doubt someone actually have used this property. I didn't know it exist until I was fixing those bugs.
Anyhow, I've returned the property and marked it as obsolete.

Eideren:
Looks good to me

tebjan:
This has created a build error somehow... 
image

tebjan:
@Ethereal77 @aikixd did you accidentally push commits here?

Ethereal77:
I've not pushed any commits here. I don't know why it says I have a commit that references this PR. I have only merged master onto my own.
And in the commit history there is no commit from me that I can see.

tebjan:
Might be a merge error then. Can any of you tell me what's the correct state of the file?

Ethereal77:
Mine compiles fine and it is the same as master's. Here [you can see](https://github.com/Ethereal77/xenko/blob/master/sources/engine/Xenko.UI/Rendering/UI/UIRenderFeature.Picking.cs).

Ethereal77:
Maybe this is what changed? #512 was merged several hours ago.

PR #579, opened 2020-01-08
Initial proposal

Eideren:
# PR Details
Throw restore issue instead of silently failing.

## Description
The VS extension failed to parse shaders, generate shader keys and such. The only hint given was a somewhat random nullref exception. The issue was resolved through building ``Xenko.VisualStudio.Commands.csproj`` which I could have done had I known that this was the issue before starting to debug this whole thing.
This PR is for just that, I don't know how you would want to implement this @xen2 so I'm starting a draft and will implement whatever you would rather have.

## Related Issue
See Description

## Motivation and Context
Fix.

## Types of changes

## Checklist

### DIFF
--- a/sources/tools/Xenko.VisualStudio.Package/Commands/XenkoCommandsProxy.cs
+++ b/sources/tools/Xenko.VisualStudio.Package/Commands/XenkoCommandsProxy.cs
@@ -337,7 +337,7 @@ namespace Xenko.VisualStudio.Commands
                     }
                     else
                     {
-                        // TODO: Report error from log
+                        throw new InvalidOperationException( $"Could not restore {packageName} {packageInfo.ExpectedVersion}, build it or pull it from nugget manually." );
                     }
                 }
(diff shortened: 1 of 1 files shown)


xen2:
Thanks, important fix for developers!
Focus was mostly on consumer version, dev version still has various broken or not-well-explained behavior like this one (which we easily shared within the team at Silicon Studio, but it shouldn't have been like that to start with).

The way your fixed it looks good for now. I assume it displays the error in the VS error log, right?
If yes, I can merge it as is.

Note that official version should never have this problem since it actually does try to install the package if not downloaded yet. It should only happen on local dev version if the .nupkg file has never been built with that version (it should be in the `bin\packages` folder which is a nuget repository).
Maybe message should only mention that case and explain it well? (i.e. something like "if this is a locally built version, please make sure to build project {packageName}")?

Eideren:
This should do it

Issue #360, opened 2019-02-01
Cross-platform build for Xenko projects


xen2:
Idea is to be able to build Xenko projects (incl. asset compilation) on Linux/macOS.

Some important steps:
- [ ] Remove managed c++ code (esp. FBX/Assimp importers)
- [ ] Find/recompile/remove various native dependencies for Linux/macOS
- [ ] Convert projects to .NET Standard or .NET Core (to decide per project)

This is a necessary step before [cross-platform editor](https://github.com/xenko3d/xenko/issues/8)

Kryptos-FR:
> Convert projects to .NET Core

.NET Standard should be enough for some of them, see my early attempts in https://github.com/Kryptos-FR/Avalonia.GameStudio

xen2:
@Kryptos-FR yes sorry that's what I meant, editing

dfkeenan:
That "cross-platform editor" link doesn't seem to go anywhere.

SleepyMode:
> 
> 
> That "cross-platform editor" link doesn't seem to go anywhere.

It's a link to another issue.
#8

xen2:
Yes my bad, it was broken before (I tried `[cross-platform editor](#8)` but it made a link on same page with `#8` in URL). Fixed it a few days ago.

xen2:
More detailed steps:

**Runtime**
- [x] Update to latest `System` assemblies
- [x] Convert `Xenko.Games` and `Xenko.Input` to use .NET Core 3.0 WinForms & WPF, like on .NET Framework

** Compiler tools (.NET Core 3 on Windows)**
- [ ] Managed code (Assimp/FBX): if not converted to pure C/C#, need to do at least multi TFM build for .NET Core and .NET FW
- [ ] Review various use of EnvDTE (mostly used for debugger reattaching)
- [ ] Remove WCF code (maybe convert to [ServiceWire](https://github.com/tylerjensen/ServiceWire)?)

** Compiler tools (multiplatform) **
- [ ] Managed code (Assimp/FBX) needs to be converted to pure C/C#
- [ ] Rebuild various dependencies (Texture libraries, VHACD, msdfgen, etc.) for Linux/OSX

** Game Studio (.NET Core 3.0 on Windows) **
- [x] Update RoslynPad to latest version (which supports .NET Core 3.0)
- [x] Recompile QuickGraph and GraphX with support for .NET Standard or .NET Core 3.0
[...]

PR #587, opened 2020-02-02
Convert WCF usage to ServiceWire (for .net core)

makotech222:
# PR Details

This is a PR for converting xenko's usage of WCF named pipes to ServiceWire. The purpose is to help prepare Xenko for switch to .Net Core 3.1 on Windows (#360 )

## Description
I've so far just replaced usage in two of the projects in Xenko so far. Changes are somewhat difficult to test, as i'm not 100% sure where they are used. I'm also unsure of bidirectional communication, like in ExecServer using the client's logger instance, so I could use some insight there.


xen2:
I did some research/tests back then and `ServiceWire` was one of the most promising alternative I found to replace WCF.
I even had a branch with some projects converted to it, but sadly I don't have access to the computer having it until early March...
I remember some parts were easy and some were more tricky (that's when I decided to prioritize my time on other things).

Note: there are several classes using WCF that are not really used currently, so no need to bother converting to ServiceWire for them.
I am not totally sure (need to double check more carefully) but I think `BuildThreadMonitor` class, `GameDebuggerHost` class and `Asset.CompilerClient` project are among those.

makotech222:
Thanks @xen2. Ill will continue on and skip those you mentioned. Let me know if you ever get a second look. Its hard to tell where and when these projects are used in the xenko context.

xen2:
Just to check, I remember you said you were not sure how to test some specific parts, so I wanted to check which part now using ServiceWire did you test, and if there's any part I should make sure still work properly?

makotech222:
I've tested the AssetCompiler pipeline for logging, and that is working fine now. I think thats pretty much the only use of wcf now.

sources/core/Xenko.Core/Diagnostics/Logger.cs
```diff
@@ -10,6 +10,7 @@ namespace Xenko.Core.Diagnostics
     /// </summary>
     public abstract partial class Logger : ILogger
     {
+        private static object _lock = new object();
```
Eideren: Is static necessary here ?

sources/editor/Xenko.GameStudio/Debugging/DebugHost.cs
```diff
@@ -74,7 +73,7 @@ public void Stop()
             }
             if (ServiceHost != null)
             {
-                ServiceHost.Abort();
+                ServiceHost.Close();
```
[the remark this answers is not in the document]
makotech222: The Close() call actually does Dispose() internally. I didn't realize that until later in the development. I will remove the call to Dispose().

sources/core/Xenko.Core/Diagnostics/Logger.cs
makotech222: There are other instances of logger, and the purpose of the lock is to make sure calls to log are single threaded across all instances.

sources/buildengine/Xenko.Core.BuildEngine.Common/Command.cs
```diff
@@ -15,7 +15,7 @@
 
 namespace Xenko.Core.BuildEngine
 {
-    [DataContract(Inherited = true)]
+    [DataContract(Inherited = true), Serializable]
```
xen2: I would prefer to use only our serialization (which is much more tested and already used in many places) rather than having a third way (`BinarySerialization` and Yaml) to serialize our type that will likely almost never be tested and might behave differently for various specific case (i.e. we use attribute on properties to express a bunch of stuff).

Is there a way, like WCF, to use a custom serialization engine?
After a quick check, it looks like it's not possible easily (https://github.com/tylerjensen/ServiceWire/wiki/Fast-Serialization-Overview mentions you have to use byte array if you want to avoid `BinaryFormatter`...)

Assuming we have to stick with `Serializable` still, did you test if this part work?
As far as I remember, building in separate process is used only for FBX so far (because their API is not threadsafe), so testing with a project having FBX is necessary if you didn't do it so far.

xen2: Took a closer look at ServiceWire, it looks like we can do the same by implementing a custom `ServiceWire.ISerializer`  and passing it as parameter to `NpHost` ctor.
You can look at WCF serialization implementation `XenkoXmlObjectSerializer.EncodeObject` and `XenkoXmlObjectSerializer.DecodeObject`, the new `ServiceWire.ISerializer.Serialize` and `Deserialize` should be very close.

makotech222: Okay I'll give that a try. Yes i originally was using JSON serialization but decided to drop it to remove dependencies. I'll do the xml serialize instead.

xen2: Just to be clear, it won't use XML at all (the class name might have misled you: WCF is using XML to communicate but Xenko generates `byte[]`, so we were wrapping our `byte[]` and wrapped it in XML data using base64 encoding).
If you check methods of last message, you will see that `ServiceWire.ISerializer` expects `byte[]` and the method to use as reference (`XenkoXmlObjectSerializer.EncodeObject` and `DecodeObject` are also working with `byte[]`.
Hopefully it should be trivial, let me know if you need any help.

makotech222:
Okay I've made commits to address the issues. I have reimplemented the original serializer, and put it in use where it was previously used.

xen2:
@makotech222 I have some issues with FBX:
image


Did you make sure `BuildSlave` code was running in your test?
Maybe GameStudio built the FBX asset.
What I did to avoid that:
- Open project in Visual Studio
- Close Game Studio
- Delete `MyGameX.Windows\obj\xenko\assetbuild` folder
- Build again (then you can copy-paste failing parameters for `Xenko.Core.Assets.CompilerApp` in Xenko solution for debugging)

makotech222:
Interesting. Okay ill give that a try to reproduce and fix it.

xen2:
Tried to debug again and noticed some issues in ServiceWire (custom serializer not being used to transfer custom parameters).
Submitted PR https://github.com/tylerjensen/ServiceWire/pull/27

makotech222:
Oh wow, interesting find!

xen2:
Rebased/squashed (after Stride rename) and merged with 7a4bdfa
We currently use our own custom https://www.nuget.org/packages/Stride.ServiceWire/5.3.2 until tylerjensen/ServiceWire#27 is merged

PR #588, opened 2020-02-04
Adding Entity pooling system

meriaizen86:
# PR Details

Entity pooling system for reusing prefabs/entities in games.

## Description

Easy and fast memory saving manner to spawn lots of prefabs or entities.

Example:
<code>
YourEntity = EntityPool.Spawn(YourPrefab.PackToEntity(), Vector3 pos, Quaternion rotation );

Entity.Scene.Entities.Add(YourEntity);
</code>

For more info check Phroot's PR:

https://github.com/phr00t/FocusEngine/commit/01c4f2baf29232993e942433a5c21920e93b312d#diff-b9a6f25f07f0f44e603fd200d7551ef7

https://github.com/phr00t/FocusEngine/commit/7666a55511abc7a1a2efba5875f0f4f58f4fb5be#diff-b9a6f25f07f0f44e603fd200d7551ef7

## Types of changes

## Checklist

- [x ] I have added tests to cover my changes.
- [x ] All new and existing tests passed.


Eideren:
Hey thanks a bunch @meriaizen86 !

I think that this would actually work better as a component + processor combo, that logic fits better as a xenko service; you don't have to add stuff to the ``Scene``'s logic and you could use it within multiple running games ( the logic within ``EntityPool`` is static and not thread safe, when changing it to a processor you can remove statics and it'll be safe to use across concurrently running games since each one will have one unique instance of that processor ), what do you think ?

SebastianStehle:
Btw: There is already a pool implementation: https://www.nuget.org/packages/Microsoft.Extensions.ObjectPool/

Not sure if you can reuse it, but it is already used a lot in kestrel, serialization and networking in general. So in places where performance matters.

meriaizen86:
> Hey thanks a bunch @meriaizen86 !
> 
> I think that this would actually work better as a component + processor combo, that logic fits better as a xenko service; you don't have to add stuff to the `Entity` and `EntityManager`'s logic and you could use it within multiple running games ( the logic within `EntityPool` is static and not thread safe, when changing it to a processor you can remove statics and it'll be safe to use across concurrently running games since each one will have one unique instance of that processor ), what do you think ?

Feel free to add changes in order to improve it. Thanks to you.

tebjan:
That Microsoft.Extensions.ObjectPool nuget looks proper. it even manages disposable objects... 

I also think that it's cleaner to register it as a service on the game instance instead of a static class.

Eideren:
@meriaizen86 If you don't intend to make changes, besides those last couple of unrelated ones, we should probably close this PR ?

xen2:
@Eideren @meriaizen86 agree with closing. Feel free to reopen if somebody is willing to do the proper changes.

PR #589, opened 2020-02-05
[Editor] Remember expand/collapse states in property window

### LINKED ISSUE #358: Rigidbody component collapses in the property grid after Ctrl+S (File >> Save) is pressed
**Release Type**: GitHub

**Version**: 3.1.0.1-beta01

**Platform(s)**: Windows

**To Reproduce**
1. Add a Rigidbody component to the entity, expand it properties in the property grid.
2. Press Ctrl+S. 

**Expected behavior**
It collapses and doesn't remember it's "expanded" state.

This is a huge workflow killer, this component has many parameters to tweak. After pressing "build the project and start the game button", the scene will be inevitably saved and all rigid body component properties will collapse. 

**This doesn't apply only for rigid body component.** Many components, like particle system have a lot of nested properties. When typically working with it, the property grid becomes long and ideally you want to have the vertical scroll bar state preserved after save/game launch, so you could go back later and quickly adjust a specific particle system parameter.

Basewq:
# PR Details

Component properties with `ExpandRule.Auto` or `ExpandRule.Once` will remember their expand/collapsed states for the whole session.

## Description

The private property path `Hashset`s are changed to static so it's retained for the entire session.

## Related Issue

One minor issue is if there's duplicate path names, they will affect each other. In the short term, this is probably not a big issue. In the long term, you'll probably want to do something like prefix all the paths will something unique?

## Motivation and Context

Fixes #358

## Types of changes

## Checklist

### DIFF
--- a/sources/editor/Xenko.Core.Assets.Editor/View/Behaviors/PropertyViewAutoExpandNodesBehavior.cs
+++ b/sources/editor/Xenko.Core.Assets.Editor/View/Behaviors/PropertyViewAutoExpandNodesBehavior.cs
@@ -21,14 +21,15 @@ namespace Xenko.Core.Assets.Editor.View.Behaviors
     public class PropertyViewAutoExpandNodesBehavior : Behavior<PropertyView>
     {
         private readonly List<PropertyViewItem> expandedItems = new List<PropertyViewItem>();
-        private readonly HashSet<string> expandedPropertyPaths = new HashSet<string>();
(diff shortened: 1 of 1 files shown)


tebjan:
Fantastic, this was a major annoyance. Thanks for looking into it!

xen2:
Great improvement!
Even though it's not perfect (duplicate path name issue & static should be replaced by something stored in a session-specific view model), I will merge this version as is since it has the merit of being very simple and good enough for the time being!

Issue #534, opened 2019-10-23
Entering an invalid color value in the Diffuse Map crashes GameStudio


Acissathar:
**Release Type**: GitHub 

**Version**: master

**Platform(s)**: Windows (unsure of others but assume so)

**Describe the bug**
Enter an invalid color value (such as CCCCCCC) into the color value section. Once leaving the field GameStudio will crash.

**To Reproduce**
1. Go to any material.
2. Change the Diffuse color value to anything that isn't a valid value.
3. Leave the field.

**Expected behavior**
1. Not a crash at the very least, ideally revert to previous valid value.

**Screenshots**
![image](https://user-images.githubusercontent.com/10227954/67346247-5619a280-f4fb-11e9-8131-285dfd834a63.png)

**Log and callstacks**
System.InvalidOperationException: 'Can not convert value to the required type'
![image](https://user-images.githubusercontent.com/10227954/67346365-b3155880-f4fb-11e9-96c5-5ba2b2e138f6.png)

PR #591, opened 2020-02-07
Prevent application crash when invalid property value is entered

Basewq:
# PR Details

The primary focus is to prevent the app from crashing due to invalid type conversion, and give the user an indication the data is not valid.

## Description

When the data conversion fails (eg. invalid color), it'll flash a red border three times and revert back (this reverting is done through some magic in the binding).
I don't have any expertise in WPF so this is the best I could, rather than try to do something like a tooltip message or toast.
While this is probably more of an excuse than an argument, Unity just reverts when invalid data is entered and has zero indicators.

I have changed `InvalidOperationException` to `InvalidCastException` because I believe this is the correct exception type for what the code is actually trying to do.

## Related Issue

See #393, #534 for issues which should no longer crash with this fix.

## Additional Notes
This is not a full on validation solution, and any existing behaviors/quirks remain in place.
eg. numeric values out of range are automatically reverted without any indication.
eg2. if asset paths are incorrectly typed it'll clear the reference instead of reverting.
eg3. invalid color hex are still "accepted" (but usually turns into something like `#000000`)
I think a proper validation 'fix' will require some breaking changes to `TextBoxBase.Validate` and it's tie with the inherited controls like `NumericTextBox`.

I apologize if this is rather hacky! Even if it's not merged, I hope this provides a good starting point for which files to look at when trying to how to fix this!

## Types of changes

## Checklist

### DIFF
--- a/sources/editor/Xenko.Core.Assets.Editor/View/DefaultPropertyTemplateProviders.xaml
+++ b/sources/editor/Xenko.Core.Assets.Editor/View/DefaultPropertyTemplateProviders.xaml
@@ -13,12 +13,20 @@
                     xmlns:io="clr-namespace:Xenko.Core.IO;assembly=Xenko.Core.Design"
                     xmlns:edvw="clr-namespace:Xenko.Core.Assets.Editor.View"
                     xmlns:viewModel="clr-namespace:Xenko.Core.Assets.Editor.ViewModel"
+                    xmlns:adr="clr-namespace:Xenko.Core.Presentation.Adorners;assembly=Xenko.Core.Presentation"
                     xmlns:qvm="clr-namespace:Xenko.Core.Presentation.Quantum.ViewModels;assembly=Xenko.Core.Presentation.Quantum"
                     xmlns:assetCommands="clr-namespace:Xenko.Core.Assets.Editor.Quantum.NodePresenters.Commands"
                     mc:Ignorable="d">
   <ResourceDictionary.MergedDictionaries>
     <ResourceDictionary Source="CommonResources.xaml"/>
   </ResourceDictionary.MergedDictionaries>
+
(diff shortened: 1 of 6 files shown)


sources/editor/Xenko.Core.Assets.Editor/View/DefaultPropertyTemplateProviders.xaml
```diff
@@ -684,9 +698,10 @@
                          DecimalPlaces="{Binding DecimalPlaces, Converter={xk:Chained {cvt:DifferentValuesToNull}, {xk:ToDouble}}, FallbackValue=6}"
                          WatermarkContentTemplate="{StaticResource DifferentValuesWatermarkTemplate}"
                          DisplayUpDownButtons="False">
-      <!--<i:Interaction.Behaviors>
-        <xk:NumericTextBoxTransactionalRepeatButtonsBehavior UndoRedoService="{Binding Session.UndoRedoService}"/>
-      </i:Interaction.Behaviors>-->
+      <i:Interaction.Behaviors>
+        <!--<xk:NumericTextBoxTransactionalRepeatButtonsBehavior UndoRedoService="{Binding Session.UndoRedoService}"/>-->
```
Kryptos-FR: Wouldn't that change make Undo/Redo not work?

Basewq: No, all those were already commented out (the entire `<i:Interaction.Behaviors>` was commented out), I've just left it in there but still kept the specific tag commented out.

xen2:
Great UX improvement, thanks!
I will give it a quick try early next week and merge it ASAP assuming no big issue!

xen2:
Tested it, works nicely!
It even works on individual component of vector/matrix, sweet!

Issue #553, opened 2019-11-14
Android Resource error


sognatore99:
Xenko Version 3.1.0.1
Using Physics Sample and opening in Visual Studio 2019 Xamarin.  When I compile (either Debug or Release) for Android I am receiving this error:

Severity	Code	Description	Project	File	Line	Suppression State	Suppression State
Error	CS0433	The type 'Resource' exists in both 'Xenko.Engine, Version=3.1.0.1, Culture=neutral, PublicKeyToken=null' and 'Xenko, Version=3.1.0.1, Culture=neutral, PublicKeyToken=null'	PhysicsSample.Android	C:\Users\xxxx\Documents\Xenko Projects\PhysicsSample\PhysicsSample.Android\Resources\Resource.Designer.cs	29		Active

I am using Visual Studio 2019 Version 16.3.8.  I have read that VS 2019 has some issue with resources but I could not load the solution produced in VS 2017 as I get a waring that "the solution is higher than the solution version supported"

Any help would be appreciated.  I have done the standard clean, rebuild, restart VS2019, reboot computer, screamed loudly :) etc.

Thanks

WernhervonBraun:
I have this error on any Android project. Even empty.

RodrigoMedeirosRS:
Any workaround to this error?

RodrigoMedeirosRS:
![Error](https://user-images.githubusercontent.com/7956669/71946287-213af480-31a8-11ea-93d6-d30e8a8bcfcc.png)

yangguosdxl:
This Error is caused by auto generate Resource.Designer.cs.

Xenko.Engine.csproj's RootNamespace is Xenko, so it's namespace is Xenko in generated Resource.Designer.cs. this namespace is the same as Xenko.csproj's Resource.Designer.cs


I change the RootNamespace property to Xenko.Engine in project Xenko.Engine, it generate the Resource class to Xenko.Engine. This fix the namesapce conflict error. but cause another error
“d:\home\workspace\test\XenkoSrc\build\xenko.build”(BuildAndroidXenkoEngine target) (1) ->
“d:\home\workspace\test\XenkoSrc\sources\engine\Xenko.Engine\Xenko.Engine.csproj”(Build target) (2) ->
“d:\home\workspace\test\XenkoSrc\sources\engine\Xenko.Engine\Xenko.Engine.csproj”(Build target) (2:5) ->
(CoreCompile target) ->
[...]

PR #592, opened 2020-02-07
fix bug: Android Resource error #553

yangguosdxl:
# PR Details

fix bug: issue 

## Description

Xenko.Engine.csproj RootNamespace is set error, cause generate error Resource.designer.cs
this pr is fix this bug

## Related Issue
#553 

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Xenko.Engine/Starter/AndroidXenkoActivity.cs
+++ b/sources/engine/Xenko.Engine/Starter/AndroidXenkoActivity.cs
@@ -21,6 +21,8 @@ using Xenko.Graphics.OpenGL;
 
 namespace Xenko.Starter
 {
+    using Resource = Xenko.Engine.Resource;
+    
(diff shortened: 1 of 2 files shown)


xen2:
Thanks for investigating this!

Is there a way to change Android namespace without changing default namespace?

Even though assembly is `Xenko.Engine`, it actually defines classes in all the `Xenko` namespace and the file hierarchy is done as such. If we change the `RootNamespace`, any newly created file will have the wrong namespace, so that's something better to avoid if possible.

yangguosdxl:
> 
> 
> Thanks for investigating this!
> 
> Is there a way to change Android namespace without changing default namespace?
> 
> Even though assembly is `Xenko.Engine`, it actually defines classes in all the `Xenko` namespace and the file hierarchy is done as such. If we change the `RootNamespace`, any newly created file will have the wrong namespace, so that's something better to avoid if possible.

Currently, I haven't find the method ...

yangguosdxl:
maybe just set rootnamespace to Xenko.Engine when building ?

PR #593, opened 2020-02-07
[Games] Added Closing event to GameWindow

tebjan:
# PR Details

Added an event that happens when the game window gets closed by the user.

## Description

The event happens before the window gets disposed. Implemented for Winforms, OpenTK, and SDL.

## Motivation and Context

It allows the developer to react to the closing window in order to clean up resources before the window gets disposed.

### DIFF
--- a/sources/engine/Xenko.Games/GameWindow.cs
+++ b/sources/engine/Xenko.Games/GameWindow.cs
@@ -53,7 +53,7 @@ namespace Xenko.Games
         public event EventHandler<EventArgs> Activated;
 
         /// <summary>
-        /// Occurs, when device client size is changed.
+        /// Occurs when device client size is changed.
         /// </summary>
         public event EventHandler<EventArgs> ClientSizeChanged;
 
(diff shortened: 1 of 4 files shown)


sources/engine/Xenko.Games/GameWindow.cs
```diff
@@ -228,6 +233,12 @@ protected void OnFullscreenToggle(object source, EventArgs e)
             handler?.Invoke(this, e);
         }
 
+        protected void OnClosing(object source, EventArgs e)
+        {
+            var handler = Closing;
+            handler?.Invoke(this, e);
```
Kryptos-FR: Should be inlined. `?` operator already does a `dup` in IL code.

Let's do it everywhere in that file.

tebjan: yes, i thought about this... but then decided against it, since there are many other places where this could be changed and i thought it's not the scope of this PR. but a general event code cleanup should be easy and quick to do with some code analyzer...

PR #606, opened 2020-02-15
Fix compile issue in master.

makotech222:
This just fixes the compile error that is currently in master branch. Please make sure that the logic is correct, I just made assumption about what its supposed to be.


tebjan:
Great, thanks! @CharlesWoodhill and @aikixd could any of you have a quick look and check whether the logic is correct?

xen2:
I'll merge it in the meantime since it breaks the build.
@CharlesWoodhill and @aikixd can still confirm if it was OK, or otherwise submit a PR to adjust this.

PR #607, opened 2020-02-15
Reduce object allocation/boxing

Basewq:
# PR Details

These commits are aimed at reducing memory allocations, which in turn reduces the garbage collector being called to clean up the objects.

## Description

* Reduce the number of structs that get boxed due to missing IEquatable inheritance when used as a `Dictionary` key.
* Avoid structs being boxed (ie. another object allocation) in the object array in `ParameterCollection` when the currently stored value is the same as the incoming value.
* ~Reduce the number of times `RenderMesh` is created by fixing the mesh count logic check.~
* Use concrete classes when calling enumerator to get the struct enumerable to avoid object allocation.

## Additional Notes
For the `ParameterCollection` fix, I am unsure whether the second check is required or if it could just go into the same logic as the equality test.
To play it safe, if `<T>` is an object type (ie. a proper object reference), it'll always be assigned because there might be some edge case where if the object is internally equal but *not* reference equals, then a programmer manipulates the newly assigned object after it is set, then the programmer expects it to be changed.
However, this does pay a cost of having the second check.

~You are free to reject the `RenderMesh` commit as it currently behaves on "broken" logic (and essentially trades one quirk for another).
Currently, it is broken for models that have materials that do multiple passes (eg. opaque pass + transparent pass) and it'll always go through the RegenerateMeshes code path because mesh count logic always fails.
This change ensures it counts the meshes correctly, however if a user somehow replaces the meshes but retains the same count, it will not render with the new mesh.
Perhaps this is ok, and we need an explicit force regenerate flag or processor method?~

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Xenko.Engine/Updater/UpdateEngine.cs
+++ b/sources/engine/Xenko.Engine/Updater/UpdateEngine.cs
@@ -723,7 +723,7 @@ namespace Xenko.Updater
         /// <summary>
         /// Internally used as key to register members.
         /// </summary>
-        private struct UpdateKey
+        private struct UpdateKey : IEquatable<UpdateKey>
         {
             public readonly Type Owner;
             public readonly string Name;
@@ -737,6 +737,26 @@ namespace Xenko.Updater
             public override string ToString()
             {
                 return $"{Owner.Name}.{Name}";
+            }
+
+            public bool Equals(UpdateKey other)
+            {
+                return Owner == other.Owner && Name == other.Name;
+            }
+
+            public override bool Equals(object obj)
+            {
+                return obj is UpdateKey key && Equals(key);
+            }
+
+            public override int GetHashCode()
+            {
+                unchecked
+                {
+                    int hashCode = Owner.GetHashCode();
+                    hashCode = (hashCode * 397) ^ Name.GetHashCode();
+                    return hashCode;
+                }
             }
         }
 
--- a/sources/engine/Xenko.Rendering/Extensions/IndexExtensions.cs
+++ b/sources/engine/Xenko.Rendering/Extensions/IndexExtensions.cs
@@ -293,7 +293,7 @@ namespace Xenko.Extensions
             meshData.PrimitiveType = PrimitiveType.PatchList.ControlPointCount(12);
         }
 
-        private struct EdgeKeyAEN
+        private struct EdgeKeyAEN : IEquatable<EdgeKeyAEN>
         {
             public readonly int PositionIndex0;
             public readonly int PositionIndex1;
@@ -307,6 +307,16 @@ namespace Xenko.Extensions
             public EdgeKeyAEN(EdgeAEN edge)
                 : this(edge.PositionIndex0, edge.PositionIndex1)
             {
+            }
+
+            public bool Equals(EdgeKeyAEN other)
+            {
+                return PositionIndex0 == other.PositionIndex0 && PositionIndex1 == other.PositionIndex1;
+            }
+
+            public override bool Equals(object obj)
+            {
+                return obj is EdgeKeyAEN key && Equals(key);
             }
 
             public override int GetHashCode()
--- a/sources/core/Xenko.Core/Utilities.cs
+++ b/sources/core/Xenko.Core/Utilities.cs
@@ -633,6 +633,31 @@ namespace Xenko.Core
             return second.Keys.All(first.ContainsKey);
         }
 
+        /// <summary>
+        /// Compares two collection, element by elements.
+        /// </summary>
+        /// <param name="first">The collection to compare from.</param>
+        /// <param name="second">The colllection to compare to.</param>
+        /// <returns>True if lists are identical (but not necessarily in the same order). False otherwise.</returns>
+        /// <remarks>Concrete SortedList is favored over interface to avoid enumerator object allocation.</remarks>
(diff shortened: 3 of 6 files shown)


sources/engine/Xenko.Engine/Updater/UpdateEngine.cs
```diff
@@ -738,6 +738,27 @@ public override string ToString()
...
+
+            public bool Equals(UpdateKey other)
+            {
+                return Owner == other.Owner && Name == other.Name;
+            }
+
+            public override bool Equals(object obj)
+            {
+                if (ReferenceEquals(null, obj)) return false;
+                return obj is UpdateKey && Equals((UpdateKey)obj);
```
Eideren: Can be simplified to ``return obj is UpdateKey key && Equals(key);`` the null test above isn't necessary as the is test already returns false for nulls.

sources/engine/Xenko.Rendering/Extensions/IndexExtensions.cs
```diff
@@ -309,6 +309,17 @@ public EdgeKeyAEN(EdgeAEN edge)
...
 
+            public bool Equals(EdgeKeyAEN other)
+            {
+                return PositionIndex0 == other.PositionIndex0 && PositionIndex1 == other.PositionIndex1;
+            }
+
+            public override bool Equals(object obj)
+            {
+                if (ReferenceEquals(null, obj)) return false;
+                return obj is EdgeKeyAEN && Equals((EdgeKeyAEN)obj);
```
Eideren: Same here.

sources/engine/Xenko.Engine/Rendering/ModelRenderProcessor.cs
```diff
@@ -144,7 +144,7 @@ private void CheckMeshes(ModelComponent modelComponent, RenderModel renderModel)
                 if (model != null)
                 {
                     // Number of meshes changed in the model?
-                    if (model.Meshes.Count != renderModel.Meshes.Length)
+                    if (model.Meshes.Count != renderModel.UniqueMeshCount)
```
Eideren: From what I understood of the issue you mentioned on the post this would be resolved by comparing the two collections but you would rather not pay the perf cost of doing so ? Perhaps this could be resolved by introducing a collection with a ``version`` variable, you would test against the ``version`` variable instead of ``UniqueMeshCount`` and, if we really want to be thorough, do a collection comparison afterwards ?
If you think this would resolve the issue, @xen2 could we add a ``version`` to FastList or would you rather have another collection type for this purpose ?

sources/engine/Xenko.Engine/Updater/UpdateEngine.cs
Basewq: Ok, I copied it from an existing code. I assume this means we can use C# 7 features (I think I saw some in use already). Will make this change.

sources/engine/Xenko.Engine/Rendering/ModelRenderProcessor.cs
Basewq: Yes, the current faulty logic is trying to detect changes in the Meshes list, I think.
If you're willing for a proper breaking change, I think changing Meshes from a `List` to a `TrackingCollection` on the `Model` and having some sort of `RequiresMeshRegen` bool will suffice.
Even if they end up adding the exact same elements, I don't think it matters too much if it regenerates the mesh (especially since they're probably going through the regen code currently and it shouldn't happen every update frame anyway).
However, I think this will need some sort of serialization upgrade code or something? All I know is Model is one of the most complicated asset types and I do not dare to mess with that, haha.

sources/engine/Xenko.Engine/Updater/UpdateEngine.cs
Basewq: Please advise if you'd prefer me to amend the existing commit and force push instead of making another commit.

sources/engine/Xenko.Engine/Rendering/ModelRenderProcessor.cs
Eideren: Let's see what @xen2 has to say

xen2: I think even if we were to compare collection, it wouldn't be enough as `Mesh` and `Model` is not immutable (which would help a lot for some specific classes such as `Mesh` to simplify a lot of those clunky checks, however it would require non-trivial changes to runtime/editor).

For example, `Mesh.MaterialIndex` could still change afterwise, as well as `Model.Materials`.
That's why the second step is not checking not the `Mesh` (either as a class reference or all its content), but only the important part requiring a rebuild (actual material with all the indirections resolved).

Basewq: Ok, perhaps this change is not safe enough to make at this point in time.
I am ok with this not included in the pull.
Do you want me to remove this commit from this pull, or leave every thing as is and you'll sort it out from the rest of the commits?

Eideren: Please remove it, I'll merge afterwards if xen doesn't do it before I have time to :)

Basewq:
Ok, removed the mesh regen commit, squashed Eideren & Kryptos-FR's advices into their respective commits and rebased.

Eideren:
Thanks !

PR #610, opened 2020-02-17
Fix WPF resource dictionary errors.

makotech222:
This fixes a wpf compile error that was introduced when the DataGrid changes were merged.


tebjan:
Cool, thanks for looking into it!

PR #611, opened 2020-02-18
Add note to Matrix.Multiply methods to not use a parameter as both ref and out

Basewq:
EDIT: This pull has been changed to note the quirk with `Matrix.Multiply` rather than change it.
# PR Details

Add documentation to not use a `ref` variable as the `out` parameter for `Matrix.Multiply`, ie.
```
var matrix1 = // Some matrix
var matrix2 = // Some other matrix
Matrix.Multiply(ref matrix1, ref matrix2, out matrix2); // This will not correctly calculate matrix1 * matrix2
```

## Description

Add a `<remarks>` tag to `Matrix.Multiply`.
Fix code where `Matrix.Multiply` was used in the way above.

## Motivation and Context

We cannot fix `Matrix.Multiply` to calculate the new matrix defensively because of performance degradation, so we tell them not to do it.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Xenko.Engine/Engine/TransformComponent.cs
+++ b/sources/engine/Xenko.Engine/Engine/TransformComponent.cs
@@ -254,8 +254,7 @@ namespace Xenko.Engine
                 var scene = Entity?.Scene;
                 if (scene != null)
                 {
-                    var inverseSceneTransform = scene.WorldMatrix;
-                    inverseSceneTransform.Invert();
+                    Matrix.Invert(ref scene.WorldMatrix, out var inverseSceneTransform);
                     Matrix.Multiply(ref WorldMatrix, ref inverseSceneTransform, out LocalMatrix);
                 }
                 else
@@ -266,8 +265,7 @@ namespace Xenko.Engine
             else
             {
                 //We are not root so we need to derive the local matrix as well
-                var inverseParent = Parent.WorldMatrix;
-                inverseParent.Invert();
+                Matrix.Invert(ref Parent.WorldMatrix, out var inverseParent);
                 Matrix.Multiply(ref WorldMatrix, ref inverseParent, out LocalMatrix);
             }
         }
@@ -299,8 +297,6 @@ namespace Xenko.Engine
             }
             else
             {
-                WorldMatrix = LocalMatrix;
-
                 var scene = Entity?.Scene;
                 if (scene != null)
                 {
@@ -309,8 +305,11 @@ namespace Xenko.Engine
                         scene.UpdateWorldMatrix();
                     }
 
-                    var curWorldMatrix = WorldMatrix;   // Must make a copy to use as the ref parameter, otherwise matrix will not be calculated correctly
-                    Matrix.Multiply(ref curWorldMatrix, ref scene.WorldMatrix, out WorldMatrix);
+                    Matrix.Multiply(ref LocalMatrix, ref scene.WorldMatrix, out WorldMatrix);
+                }
+                else
+                {
+                    WorldMatrix = LocalMatrix;
                 }
             }
 
--- a/sources/engine/Xenko.Engine/Engine/TransformComponent.cs
+++ b/sources/engine/Xenko.Engine/Engine/TransformComponent.cs
@@ -309,7 +309,8 @@ namespace Xenko.Engine
                         scene.UpdateWorldMatrix();
                     }
 
-                    Matrix.Multiply(ref WorldMatrix, ref scene.WorldMatrix, out WorldMatrix);
+                    var curWorldMatrix = WorldMatrix;   // Must make a copy to use as the ref parameter, otherwise matrix will not be calculated correctly
+                    Matrix.Multiply(ref curWorldMatrix, ref scene.WorldMatrix, out WorldMatrix);
                 }
             }
 
--- a/sources/core/Xenko.Core.Mathematics/Matrix.cs
+++ b/sources/core/Xenko.Core.Mathematics/Matrix.cs
@@ -956,14 +956,12 @@ namespace Xenko.Core.Mathematics
 
         /// <summary>
         /// Determines the product of two matrices.
+        /// Variables passed as <paramref name="left"/> or <paramref name="right"/> must not be used as the out parameter
+        /// <paramref name="result"/>, because <paramref name="result"/> is calculated in-place.
         /// </summary>
         /// <param name="left">The first matrix to multiply.</param>
         /// <param name="right">The second matrix to multiply.</param>
         /// <param name="result">The product of the two matrices.</param>
-        /// <remarks>
-        /// Variables passed as <paramref name="left"/> or <paramref name="right"/> must not be used as the out parameter
-        /// <paramref name="result"/>, because <paramref name="result"/> is calculated in-place.
-        /// </remarks>
         public static void MultiplyTo(ref Matrix left, ref Matrix right, out Matrix result)
         {
             result.M11 = (left.M11 * right.M11) + (left.M12 * right.M21) + (left.M13 * right.M31) + (left.M14 * right.M41);
@@ -986,14 +984,12 @@ namespace Xenko.Core.Mathematics
 
         /// <summary>
         /// Determines the product of two matrices.
+        /// Variables passed as <paramref name="left"/> or <paramref name="right"/> must not be used as the out parameter
+        /// <paramref name="result"/>, because <paramref name="result"/> is calculated in-place.
         /// </summary>
         /// <param name="left">The first matrix to multiply.</param>
         /// <param name="right">The second matrix to multiply.</param>
         /// <param name="result">The product of the two matrices.</param>
-        /// <remarks>
-        /// Variables passed as <paramref name="left"/> or <paramref name="right"/> must not be used as the out parameter
-        /// <paramref name="result"/>, because <paramref name="result"/> is calculated in-place.
-        /// </remarks>
         public static void Multiply(ref Matrix left, ref Matrix right, out Matrix result)
         {
             result.M11 = (left.M11 * right.M11) + (left.M12 * right.M21) + (left.M13 * right.M31) + (left.M14 * right.M41);
@@ -1016,14 +1012,12 @@ namespace Xenko.Core.Mathematics
 
         /// <summary>
         /// Determines the product of two matrices.
+        /// Variables passed as <paramref name="left"/> or <paramref name="right"/> must not be used as the out parameter
+        /// <paramref name="result"/>, because <paramref name="result"/> is calculated in-place.
         /// </summary>
         /// <param name="left">The first matrix to multiply.</param>
         /// <param name="right">The second matrix to multiply.</param>
         /// <param name="result">The product of the two matrices.</param>
-        /// <remarks>
-        /// Variables passed as <paramref name="left"/> or <paramref name="right"/> must not be used as the out parameter
-        /// <paramref name="result"/>, because <paramref name="result"/> is calculated in-place.
-        /// </remarks>
         public static void MultiplyRef(ref Matrix left, ref Matrix right, ref Matrix result)
         {
             result.M11 = (left.M11 * right.M11) + (left.M12 * right.M21) + (left.M13 * right.M31) + (left.M14 * right.M41);
--- a/sources/core/Xenko.Core.Mathematics/Matrix.cs
+++ b/sources/core/Xenko.Core.Mathematics/Matrix.cs
@@ -7,17 +7,17 @@
 // -----------------------------------------------------------------------------
 /*
 * Copyright (c) 2007-2011 SlimDX Group
-* 
+*
 * Permission is hereby granted, free of charge, to any person obtaining a copy
 * of this software and associated documentation files (the "Software"), to deal
 * in the Software without restriction, including without limitation the rights
 * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
 * copies of the Software, and to permit persons to whom the Software is
 * furnished to do so, subject to the following conditions:
-* 
+*
 * The above copyright notice and this permission notice shall be included in
 * all copies or substantial portions of the Software.
-* 
+*
 * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
 * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
 * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
@@ -77,7 +77,7 @@ namespace Xenko.Core.Mathematics
         /// Value at row 4 column 1 of the matrix.
         /// </summary>
         public float M41;
-        
+
         /// <summary>
         /// Value at row 1 column 2 of the matrix.
         /// </summary>
@@ -102,7 +102,7 @@ namespace Xenko.Core.Mathematics
         /// Value at row 1 column 3 of the matrix.
         /// </summary>
         public float M13;
-        
+
         /// <summary>
         /// Value at row 2 column 3 of the matrix.
         /// </summary>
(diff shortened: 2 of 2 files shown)


Kryptos-FR:
Unless I am missing something, this PR does the exact opposite of what it claims to fix.

In other words, it always does a copy where in some case we could prevent it.

Consider:

    Matrix m1 = ...;
    Matrix m2 = ...;
    Matrix m3;

    Matrix.Multiply(ref m1, ref m2, out m3);

    // use m3...

    matrix1...  // change something

    // No copy in previous code, unnecessary copy in new code
    Matrix.Multiply(ref m1, ref m2, out m3);

What we should do instead is document that matrix3 must be a different reference from matrix1 or matrix2. But adding a temporary variable causes a copy in all cases.

What you are proposing is equivalent to (without change):

    Matrix m1, m2, m3;
    Matrix.Multiply(ref m1, ref m2, out m3);
    m2 = m3;

I would prefer the copy to be explicit at the calling site instead of hidden in the method.

Basewq:
I might be a little confused on your example.
As it stands having 3 separate Matrix variables works, regardless of this fix or not.
Having only 2 Matrix, where one is also the output, currently does not work because it is partially writing over itself then the overwritten fields are then being used the in subsequent calculations.

Yes, what I'm proposing is nearly equivalent to your bottom example, but technically `m3` will not exist.

No, matrix3 should not be documented to be different that `matrix1` & `matrix2`.

A real simple example (this is from Xenko's source code) is changing the code from:
`physicsTransform = Matrix.Multiply(ColliderShape.PositiveCenterMatrix, physicsTransform);`
to
`Matrix.Multiply(ref ColliderShape.PositiveCenterMatrix, ref physicsTransform, out physicsTransform);`

However one primary example of using it is for chaining your calculations:
```
Matrix.Multiply(ref positionMatrix, ref translationMatrix, out var finalTransformMatrix);
Matrix.Multiply(ref finalTransformMatrix, ref scaleMatrix, out finalTransformMatrix);
Matrix.Multiply(ref finalTransformMatrix, ref rotationMatrix, out finalTransformMatrix);
```
(or whatever the correct order of matrix operations).


As a side note, XNA/Monogame supports doing this. The whole purpose of these `ref` versions were because copying large structs around performed badly and people wanted a faster version.
Yes, the code may look stupid and uglier but that's a limitation of the C# language.

Kryptos-FR:
What I meant is that your change doesn't remove the copy, it just hides it. There is a copy happening inside the method now : the temp variable.

But it does add a copy for other code paths that did not need chaining calls.

Creating worse performance for convenience is not a good trade off in my opinion.

Basewq:
Ok, I understand. Yes, it's essentially creating a defensive copy in that sense.
And yes, it degrades the performance slightly because of this.

I would say that the results of these should all be equivalent, but they are not:
```
m1 *= m2;
m1 = Matrix.Multiply(m1, m2);
Matrix.Multiply(ref m1, ref m2, out m1);
```
This goes against the behavior of every other method that follow similar signatures.

If you still disagree, feel free to reject this.

Basewq:
> 
> > No, matrix3 should not be documented to be different that matrix1 & matrix2.
> 
> Yes it should. Passing the same variable to both a ref and out argument is considered bad practice and a code smell.

I don't consider it bad practice. It reduces code noise by not requiring a dummy variable while still using the `ref` feature.
FYI, the code base has a few of these already in place, and I find it perfectly readable, eg.

https://github.com/xenko3d/xenko/blob/master/sources/engine/Xenko.Graphics/SpriteBatch.cs#L506

https://github.com/xenko3d/xenko/blob/master/sources/engine/Xenko.Graphics/SpriteFont.cs#L236-L237

https://github.com/xenko3d/xenko/blob/master/sources/engine/Xenko.Graphics/Sprite3DBatch.cs#L170-L175

https://github.com/xenko3d/xenko/blob/28245aff55cea3ceacaef2ea3902657cd06f81b6/sources/engine/Xenko.Engine/Engine/EntityTransformExtensions.cs#L157-L160

Amusingly enough, I discover there is one single code that used `Matrix.Multiply` in the way you'd think it should work, but it turns out it doesn't:

https://github.com/xenko3d/xenko/blob/080774745151b858794111ac128a9a473b517d19/sources/engine/Xenko.Engine/Engine/TransformComponent.cs#L312

In the end, no hard feelings if this shouldn't be merged.

tebjan:
How about adding this as a new operation? something like `Matrix.MultiplyInPlace` with a proper help text what the difference to the other is.

Basewq:
Personally, I think the safer version should be the default version to be consistent with how the other `ref` -> `out`, and this also makes any XNA/Monogame developers not get caught out if they ported their code.
However, I understand if this this not to be done, since the majority of `Matrix.Multiply` used in the current code base would have to change to the unsafer version.

As a note, there's actually **three** versions, with the *exact* same code underneath:
```
public static void MultiplyTo(ref Matrix left, ref Matrix right, out Matrix result)
public static void Multiply(ref Matrix left, ref Matrix right, out Matrix result)
public static void MultiplyRef(ref Matrix left, ref Matrix right, ref Matrix result)
```

The history log shows nothing, but I suspect either `MultiplyTo` or `Multiply` used to be the safe version, but a dev didn't notice the subtle difference and thought both could be the same.

I don't understand why `MultiplyRef` exists, maybe the person didn't think `ref` & `out` were the same from an output perspective?

Basewq:
I did some (poor) benchmarking to see if there's any alternative to defensive copying.
It seems doing it like Monogame where you calculate each field as a free variable, then assign them all back to the out struct was faster than a temp variable, however it still performs badly against the current version.

I tested looping through an array with 1,000,000 items and multiplying once, and the current version beat the free variable assignment version by ~15%.
Testing it with chaining multiply matrices (which seems to prevents a lot of the optimization?), the difference was closer to 3%.

Anyway, I've changed this pull to just add a note to not do the inline trick. It probably deserves to be in the summary tag so it'll show up in intellisense, since it's behavior deviates from the rest, but for the time it's just left in the remarks tag.

Also fixed TransformComponent and added a note to ensure no one tries to inline it.

Looking through the code base, there's a few places which will benefit from changing their matrix multiplication from non-ref to ref version, but it'll need to declare a dummy variable. More code clutter, but there should still be some performance boost.

sources/engine/Xenko.Engine/Engine/TransformComponent.cs
```diff
@@ -309,7 +309,8 @@ internal void UpdateWorldMatrixInternal(bool recursive)
                         scene.UpdateWorldMatrix();
                     }
 
-                    Matrix.Multiply(ref WorldMatrix, ref scene.WorldMatrix, out WorldMatrix);
+                    var curWorldMatrix = WorldMatrix;   // Must make a copy to use as the ref parameter, otherwise matrix will not be calculated correctly
```
tebjan: wow, just out of interest, was this a bug all along? do you remember how it affected the scene?
I had some transformation bugs and wonder whether it came from this.

Basewq: Surprisingly, this bug doesn't manifest in game.
I haven't rigorously tested this, but I tested a few matrix multiplication, and simple translations & scaling as the second matrix still multiply correctly.
Once you have rotation, it'll calculation incorrectly.
Looking at Game Studio, it seems you can only do translation offsets to scenes, so you can't encounter this bug through this code (as far as I can tell).

Eideren:
While we're on the topic of matrix and optimizations, I pulled the simd implementation of dotnet's ``matrix * matrix``, changed it slightly and it works really well as long as we don't use their type; their method signatures forces defensive copies everywhere and our matrices are column major while dotnet's are row major.
```
public static void Matrix4x4Mult<T>(ref T left, ref T right, out T outV) where T : unmanaged
{
    unsafe
    {
        if( sizeof(T) != sizeof(float) * 4 * 4 )
            throw new ArgumentException( $"{typeof(T)}'s size does not match the size of standard 4x4 sp matrix" );
        
        fixed( void* lPtr = &left )
        fixed( void* rPtr = &right )
        fixed( void* outPtr = &outV )
        {
            // Swap left-right if row major
            float* leftPtr = (float*)lPtr;
            float* rightPtr = (float*)rPtr;
            float* outM = (float*)outPtr;
            
            var row = Sse.LoadVector128(rightPtr);
            Sse.Store(outM,
                Sse.Add(Sse.Add(Sse.Multiply(Sse.Shuffle(row, row, 0x00), Sse.LoadVector128(leftPtr)),
                                Sse.Multiply(Sse.Shuffle(row, row, 0x55), Sse.LoadVector128(leftPtr+4))),
                        Sse.Add(Sse.Multiply(Sse.Shuffle(row, row, 0xAA), Sse.LoadVector128(leftPtr+8)),
                                Sse.Multiply(Sse.Shuffle(row, row, 0xFF), Sse.LoadVector128(leftPtr+12)))));

            row = Sse.LoadVector128(rightPtr+4);
            Sse.Store(outM+4,
                Sse.Add(Sse.Add(Sse.Multiply(Sse.Shuffle(row, row, 0x00), Sse.LoadVector128(leftPtr)),
                                Sse.Multiply(Sse.Shuffle(row, row, 0x55), Sse.LoadVector128(leftPtr+4))),
                        Sse.Add(Sse.Multiply(Sse.Shuffle(row, row, 0xAA), Sse.LoadVector128(leftPtr+8)),
                                Sse.Multiply(Sse.Shuffle(row, row, 0xFF), Sse.LoadVector128(leftPtr+12)))));

            row = Sse.LoadVector128(rightPtr+8);
            Sse.Store(outM+8,
                Sse.Add(Sse.Add(Sse.Multiply(Sse.Shuffle(row, row, 0x00), Sse.LoadVector128(leftPtr)),
                                Sse.Multiply(Sse.Shuffle(row, row, 0x55), Sse.LoadVector128(leftPtr+4))),
                        Sse.Add(Sse.Multiply(Sse.Shuffle(row, row, 0xAA), Sse.LoadVector128(leftPtr+8)),
                                Sse.Multiply(Sse.Shuffle(row, row, 0xFF), Sse.LoadVector128(leftPtr+12)))));

            row = Sse.LoadVector128(rightPtr+12);
            Sse.Store(outM+12,
                Sse.Add(Sse.Add(Sse.Multiply(Sse.Shuffle(row, row, 0x00), Sse.LoadVector128(leftPtr)),
                                Sse.Multiply(Sse.Shuffle(row, row, 0x55), Sse.LoadVector128(leftPtr+4))),
                        Sse.Add(Sse.Multiply(Sse.Shuffle(row, row, 0xAA), Sse.LoadVector128(leftPtr+8)),
                                Sse.Multiply(Sse.Shuffle(row, row, 0xFF), Sse.LoadVector128(leftPtr+12)))));
        }
    }
}
```
BenchmarkDotNet:
```
| Method |      Mean |    Error |   StdDev | Rank |
|------- |----------:|---------:|---------:|-----:|
|   SIMD |  59.09 us | 0.300 us | 0.266 us |    1 |
|  Xenko | 179.11 us | 0.413 us | 0.345 us |    2 |
```
Dotnet's multiply with all of the defensive copies is around 150μ.
The average difference between the output from xenko and simd is ~1/3000000, it's perfectly safe to use.
[And this requires dotnet core 3.0+ though, so we can't use this yet](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.intrinsics.x86.sse.add?view=netcore-3.1).

sources/engine/Xenko.Engine/Engine/TransformComponent.cs
Eideren: We could perhaps assert/check in debug mode for cases like these, fix and compare pointers of refs and outs, if they are the same throw ?

tebjan: that's a pretty good idea!

Basewq: I am willing to make a commit for this (which can be accepted or rejected), though this is a heavily used method, so it might have an impact a user's debug build.
As far as I can tell, it needs to be written like this?
```
unsafe
{
    fixed (void* leftPtr = &left)
    fixed (void* rightPtr = &right)
    fixed (void* resultPtr = &result)
    {
        System.Diagnostics.Debug.Assert(leftPtr != resultPtr, "This method does not support 'left' parameter being the same as 'result' parameter.");
        System.Diagnostics.Debug.Assert(rightPtr != resultPtr, "This method does not support 'right' parameter being the same as 'result' parameter.");
    }
}
```

How do I expose this to propagate a user's build?
I think `#if DEBUG` doesn't work for user consumption?
Do I wrap it in another method but tag it `[Conditional("DEBUG")]`?

Eideren: The perf hit in debug mode is around 33% of the base cost of the function regardless of if we're using an exception or an assert and there's no elegant way to conditionally run logic based on the entry assembly's build type; it can't work properly with users debug build if we want optimal perfs in release so I don't think it's worthwhile to have this check in place in the end, sorry about that.

Eideren:
LGTM, If you don't have anything to add we'll merge this

Basewq:
All done on my end, nothing else to add. Thanks.

sources/engine/Xenko.Engine/Engine/TransformComponent.cs
tebjan: @Eideren ok, that's a lot for a method that get's used all over the place.
@Basewq is the comment now in the summary or remarks? since it's important info, the programmer should see it in IntelliSense, i'd say.

tebjan: > it can't work properly with users debug build if we want optimal perfs

is `#ifdef DEBUG` not working here?

Eideren: Preprocessors are taken care of in the compilation process, the JIT doesn't take care of this kind of things so release dlls are already stripped out of any ``#debug``. I don't think that we can provide both debug and release versions within nugget packages and conditionally include one or the other depending on the application's build mode. 

Users would have to explicitly take the debug dlls to ever have these kinds of checks.

You can detect whether the entry assembly runs in debug or release mode but that's kind of a hack since we don't know whether the user intended the engine to run in debug mode based solely on the entry assembly's build mode or if he just has a weird project structure. 
We would also have to pay the price of having at least one branch in release mode, I don't think the issue here warrants it.
Perhaps once c# has function pointers we could take a look at this again.

tebjan: yes, that would require to ship the debug dlls... true. not important enough. but I would move the info into the summary. @Basewq

Basewq: Ok, promoted the note to summary tag.

Regarding the debug note, I've been reading up on the `Conditional` attribute, and I think this allows for the kind of branching you're looking for.
[eg. this example show](https://www.dotnetperls.com/conditional) methods are always included in every dll, but only runs when certain symbols are declared. `Debug.Write` is probably the best example.
[The IL do not get emitted](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.conditionalattribute?view=netframework-4.8) if the symbol isn't declared.
We could technically define a new symbol for users to opt in to get this extra validation, however looking at the code base, there isn't anything like this for public use already so this would be a completely new feature introduced (for advanced users...).
That said, let's leave this alone for the time being.

Basewq:
Sorry, after looking at the TransformComponent again, I've decided to just avoid the quirk entirely and removed the defensive variable. Probably needs a fresh pair of eyes to check this really is the same logic, but I think it is.
Also reduced the `Matrix.Invert` usage to one line which should also be faster since it's not copying the struct before inverting.

This really is the last change, assuming there isn't anything wrong in the logic.

xen2:
LGTM, thanks!

tebjan:
@Basewq thanks for spotting this! this issue was the cause for a bug I was hunting on and off for a few weeks now. when we switched from SharpDX to Xenko math, we had a color transformation method that was using a matrix multiplication chain that had this problem. 

today I was debugging again and saw the multiplication with ref parameters used in a chain and remembered this PR. otherwise, I would probably have searched much longer...

PR #612, opened 2020-02-23
GameStudio to suspend game sub-windows when they are hidden in tab control

Basewq:
# PR Details

Throttle the game update and prevent draw calls for scene/prefab editors and the preview asset when they are hidden inside a tab in the editor.

## Description

Added `EditorServiceGame.IsEditorHidden` field to throttle the game updating to once per second, and stop the draw code.

~I tried to keep close to the current architecture.~ (Yeah, I don't think I can claim that anymore...)

A dockable sub-window's visibility can't be detected with `IsVisible`, no idea why...
What's actually required is testing `IsSelected`. This should work whether it's split or floating.
Unfortunately, the Asset Preview sub-window requires a bigger hack, due to `DockingLayoutManager` essentially destroying and recreating the `DockingManager` control. This broke every `<i:Interaction.Behaviors>` inside `DockingManager`, which I've restored however it turns out the bindings are actually broken on all these Behaviors already. This can be investigated and fixed later.

Regarding the Asset Preview sub-window, I wanted to use `OnEventCommandBehavior` in `GameStudioWindow.xaml` instead of using the code behind, but binding to `CommandParameter` requires binding to the `LayoutAnchorable` control (for `IsSelected`), but the fix to the `Behaviors` stated above still doesn't work on it, probably because the binding locks on to 'old' control rather than update itself to 'current' control.

Naming improvements and/or other suggestions to keep it consistent are welcome.

## Motivation and Context

Scenes/Prefabs/Preview Asset windows continue running/drawing when not visible, which may slow down the Game Studio if too many tabs are open.

## Types of changes

## Checklist

### DIFF
--- a/sources/editor/Xenko.Assets.Presentation/AssetEditors/GameEditor/Services/EditorGameController.cs
+++ b/sources/editor/Xenko.Assets.Presentation/AssetEditors/GameEditor/Services/EditorGameController.cs
@@ -231,6 +231,16 @@ namespace Xenko.Assets.Presentation.AssetEditors.GameEditor.Services
             return true;
         }
 
+        public void OnHideGame()
+        {
+            Game.IsEditorHidden = true;
+        }
+
+        public void OnShowGame()
+        {
+            Game.IsEditorHidden = false;
+        }
+
         public Vector3 GetMousePositionInScene(bool lastRightClick)
         {
             EnsureNotDestroyed();
--- a/sources/editor/Xenko.Assets.Presentation/AssetEditors/GameEditor/Services/IEditorGameController.cs
+++ b/sources/editor/Xenko.Assets.Presentation/AssetEditors/GameEditor/Services/IEditorGameController.cs
@@ -46,6 +46,16 @@ namespace Xenko.Assets.Presentation.AssetEditors.GameEditor.Services
         Task<bool> CreateScene();
 
         /// <summary>
+        /// Stops the game from rendering and throttle game updates.
+        /// </summary>
+        void OnHideGame();
+
+        /// <summary>
+        /// Resumes the game rendering and updating and .
+        /// </summary>
+        void OnShowGame();
+
+        /// <summary>
         /// Finds the game-side instance corresponding to the part with the given id, if it exists.
         /// </summary>
         /// <param name="partId">The id of the part to look for.</param>
--- a/sources/editor/Xenko.Assets.Presentation/AssetEditors/GameEditor/ViewModels/GameEditorViewModel.cs
+++ b/sources/editor/Xenko.Assets.Presentation/AssetEditors/GameEditor/ViewModels/GameEditorViewModel.cs
@@ -29,7 +29,7 @@ namespace Xenko.Assets.Presentation.AssetEditors.GameEditor.ViewModels
         {
             Controller = controllerFactory(this);
             CopyErrorToClipboardCommand = new AnonymousCommand(ServiceProvider, CopyErrorToClipboard);
-            ResumeCommand = new AnonymousCommand(ServiceProvider, Resume);
+            ResumeCommand = new AnonymousCommand(ServiceProvider, ResumeFromError);
         }
 
         /// <summary>
@@ -77,6 +77,16 @@ namespace Xenko.Assets.Presentation.AssetEditors.GameEditor.ViewModels
             base.Destroy();
         }
 
+        public void HideGame()
+        {
+            Controller.OnHideGame();
+        }
+
+        public void ShowGame()
+        {
+            Controller.OnShowGame();
+        }
+
         protected virtual async Task<bool> InitializeEditor()
         {
             Dispatcher.EnsureAccess();
@@ -112,7 +122,7 @@ namespace Xenko.Assets.Presentation.AssetEditors.GameEditor.ViewModels
                 SafeClipboard.SetText(log);
         }
 
-        private void Resume()
+        private void ResumeFromError()
         {
             Controller.GetService<IEditorGameRecoveryViewModelService>()?.Resume();
         }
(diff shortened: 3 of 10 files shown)


Basewq:
Ok, this is isn't working for the Asset Preview tab. It turns out `DockingLayoutManager` removes all tab elements then re-adds saved ones which effectively removes any event handlers declared in the xaml.
I'll need to think of a new strategy around this...

Basewq:
> 
> 
> Have you tested this with UI tabs? I did some very quick testing (not this commit, but up to the last one) with 4 tabs (1 scene, 1 prefab + 2 UI tabs), and things were pretty slow -- even after closing all tabs the slowness remained. I reverted the changes and although things still got slow after the 4 tabs, closing the tabs resolved the slowness.

Thanks for looking, I didn't but have now. It looks ok to me, but I've only tested by putting a `Debug.WriteLine` in the `EditorServiceGame.Draw`/`EditorServiceGame.Update`, so there could be some other hidden things running that I'm not aware about.

Can you open your 4 tabs again, then open something like Graphics Compositor (which doesn't run an `EditorServiceGame`), and check if `EditorServiceGame.Draw` is still running (after the `IsSuspended` check). If it is still running, then the suspend logic isn't being detected properly and I might need to rethink how to detect the tab state. You'll probably need to test with the latest commits since the asset preview was still running on the older version.

Also, I'm aware you have your own fork, this is done off the latest version, and it uses `Extended.Wpf.Toolkit` Version="3.8.1" (in `Xenko.Core.Assets.Editor.csproj`). I don't know if you're on a different version or not? The docking control may or may not have changed behaviors if you're on an older version.

Basewq:
Thanks for the heads up phr00t, it might not related to the docking control.
Preliminary testing shows it's not disposing the resources (calling `UnloadContent`) when closing the tab, which I suspect is due to the `IsSuspended` logic bypassing something that's needed on the inside `Update`. If true, I think this bug also occurs if a user closes the scene when it has faulted, well.
Will look into finding out how to fix this.

Basewq:
Unfortunately, I couldn't come up with a decent solution to closing the sub-window, and it required yet another hack. If someone knows a better way, feel free to pitch in.
The problem is the game's exit detection placed in the `ScriptSystem`, so it requires a game update to execute.

As I suspected, even the Faulted case, where Game Studio blocks the scene with a messaging saying there's fix the error before continuing, will not exit (and release its memory) if a user just decides to close the window rather than resolving it.
As a side note, even in the 'standard' case, a scene releases most of its memory, however there might be some memory leak, as it still slowly increments overall when you repeatedly open, close and open a scene.

xen2:
I was going to suggest that even if game is hidden, you still want to run update once in a while (i.e. 1 every second).

This is necessary because a bunch of actions have impact in-game (we post them as action to run during next script update). and lot of UI operations might even be stuck while awaiting for results to be back.
It seems you already run into one of those issue with the exit detection not being called, but it could happen to lot of other thing (any asset edit that's being used in the live scene/game will also have each pending edit being queued to be replicated in the game).

Basewq:
Thanks for your input, I'll look at changing it to a throttled update instead.

As a side note, looking at the hack I tried to do (forcing a single update), it wouldn't fix the Faulted case anyway, as the exit script is executed last (which would probably be caught in another exception before attempting to run the exit script). I'll create an issue for that.

Basewq:
Commit has been redone to skip rendering when flagged hidden and throttle updates to once per second to ensure the exit command is detected.

phr00t:
Apologizes for not getting back to this quickly! I commented on the commit and didn't get a notification on this pull request. Yes, I do need to update my Wpf to 3.8.1 (doing that now). It was part of a commit I tried but ended up reverting. I'll try out your latest changes when I get a chance! You've got a bunch of interesting stuff coming really fast, which I like!

phr00t:
I integrated the Wpf 3.8.1 changes and tried out this latest commit. I will say performance returned when closing tabs, which is an improvement. However, performance was much worse with multiple tabs open vs. without this commit (at least on my fork). I had 3 scenes open and 2 significant UI tabs open, which slowed things down quite a bit with or without this latest commit. However, I noticed much less responsiveness (and zooming in and out of the UI tab caused huge jumps) with this commit. I've reverted it again. Perhaps try testing with more heavy scenes and UI? Might be related to something different on my fork, too.

Basewq:
> 
> 
> I integrated the Wpf 3.8.1 changes and tried out this latest commit. I will say performance returned when closing tabs, which is an improvement. However, performance was much worse with multiple tabs open vs. without this commit (at least on my fork). I had 3 scenes open and 2 significant UI tabs open, which slowed things down quite a bit with or without this latest commit. However, I noticed much less responsiveness (and zooming in and out of the UI tab caused huge jumps) with this commit. I've reverted it again. Perhaps try testing with more heavy scenes and UI? Might be related to something different on my fork, too.

Thanks for trying again. I probably should have edited my previous message, you didn't actually need to update your WPF, as I note in a later message that it was an issue on my end! I won't be so quick on thinking it's a difference in the fork, but thinking of some underlying issue in my code which needs to be uncovered.
In the latest version, I changed the logic so it only throttled the updates, meaning all "games" will still run, but the rendering shouldn't be happening at all. The fact that all "games" still exist in memory while they exist in the opened tabs means there's still a memory limit on your own machine (not saying your machine specifically, I'm talking in general).
Off the top of my head, I suspect there's probably some catch up logic in the underlying game update, which might be triggered since I throttle outside the main logic, so I'll investigate if that's the case.

Basewq:
@phr00t I've forced pushed a new version of the commit, which you can test if you like.
I suspect it was the `BeginDraw` logic, where it'd still try to "prepare" the graphics device which might be expensive if multiple games are doing it (and also not actually rendering with it).

I don't know about your hardware specs, but there is also a possibility that opening too many scenes/UI editor is making you hit your ram limit, however that's probably not be your case since you're saying the editor is working ok without this commit applied.

xen2:
@Basewq Should I still consider merging this without his feedbacks?

phr00t:
I haven't gotten to testing this and unsure if I will get to it, so please test and consider merging without my feedback.

Basewq:
@xen2 Yep, you have phr00t and my ok to test/merge. If there is any additional performance issues I'm a bit stump on where/how it manifests, but hopefully this is ok as is.

xen2:
@Basewq merged, thanks again for yet another perf-improvement PR!

PR #616, opened 2020-02-26
Minor performance improvement

Basewq:
# PR Details

Minor performance improvement in the game engine.

## Description

`Dictionary.ContainsKey` then `Dictionary[indexer]` is slower than `Dictionary.TryGetValue` since it'll go through twice to get the value so code is changed to single retrieval.

`UIBatch` has a `static readonly` `PrimitiveTypeToIndices` which doesn't need to be a `List` since it's only set up once.

Repeated array access code cleaned up to be declared once. There shouldn't be any performance gain or loss from this, only cleaner code (in my opinion).
Deeper array access (eg. `elementToStripDefinitions[dim][element][i]`) might see some performance gain (don't know how smart the Jitter is for this type of code).

Use `Matrix` methods with `ref/out` parameters to reduce the number of struct copies when doing calculations.

Introduce `LightPointShadowMapRendererCubeMap.ViewFaceRotationMatrices` to cache a matrix calculation to be used and avoid the switch statement (ie. a branching logic).

## Motivation and Context

Hopefully gain a little bit more fps.

## Types of changes

## Checklist

### DIFF
--- a/sources/editor/Xenko.Assets.Presentation/AssetEditors/Gizmos/ParticleSystemGizmo.cs
+++ b/sources/editor/Xenko.Assets.Presentation/AssetEditors/Gizmos/ParticleSystemGizmo.cs
@@ -60,9 +60,9 @@ namespace Xenko.Assets.Presentation.AssetEditors.Gizmos
                 return;
             }
 
-            var pos = new Vector3(0, 0, 0);
-            var rot = new Quaternion(0, 0, 0, 1);
-            var scl = new Vector3(1, 1, 1);
+            var pos = Vector3.Zero;
+            var rot = Quaternion.Identity;
+            var scl = Vector3.One;
             DebugDrawShape drawShape = DebugDrawShape.None;
 
             if (Component.ParticleSystem.TryGetDebugDrawShape(ref drawShape, ref pos, ref rot, ref scl))
--- a/sources/assets/Xenko.Core.Assets/Analysis/AssetDependencies.cs
+++ b/sources/assets/Xenko.Core.Assets/Analysis/AssetDependencies.cs
@@ -295,8 +295,8 @@ namespace Xenko.Core.Assets.Analysis
                 dictionary = new Dictionary<AssetId, AssetLink>();
 
             var id = contentLink.Element.Id;
-            if (dictionary.ContainsKey(id))
-                contentLink.Type |= dictionary[id].Type;
+            if (dictionary.TryGetValue(id, out var existingLink))
+                contentLink.Type |= existingLink.Type;
 
             dictionary[id] = contentLink;
         }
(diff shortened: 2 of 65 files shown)


Kryptos-FR:
I quickly reviewed the `ContainsKey` versus `TryGetValue` cases. I'm OK with those.

I'm not familiar with the `ref` implications so I will let that part for someone else to review.

sources/editor/Xenko.Assets.Presentation/AssetEditors/Gizmos/NavigationBoundingBoxGizmo.cs
```diff
@@ -87,7 +87,7 @@ public void Build()
 
                 vertices[0] = new VertexPositionNormalTexture(new Vector3(-1, 1, -1), Vector3.UnitY, Vector2.Zero);
                 vertices[1] = new VertexPositionNormalTexture(new Vector3(-1, 1, 1), Vector3.UnitY, Vector2.Zero);
-                vertices[2] = new VertexPositionNormalTexture(new Vector3(1, 1, 1), Vector3.UnitY, Vector2.Zero);
+                vertices[2] = new VertexPositionNormalTexture(Vector3.One, Vector3.UnitY, Vector2.Zero);
```
Kryptos-FR: I think the idea here was to be symmetric with the above and below lines.
Using `Vector.One` here doesn't provide any performance improvement since it is a `struct`, under the hood it is basically doing the same

sources/editor/Xenko.Assets.Presentation/AssetEditors/Gizmos/ParticleSystemGizmo.cs
```diff
@@ -60,9 +60,9 @@ public override void Update()
                 return;
             }
 
-            var pos = new Vector3(0, 0, 0);
-            var rot = new Quaternion(0, 0, 0, 1);
-            var scl = new Vector3(1, 1, 1);
+            var pos = Vector3.Zero;
+            var rot = Quaternion.Identity;
+            var scl = Vector3.One;
```
Kryptos-FR: On the other hand, here it makes it more clear 😉

sources/editor/Xenko.Assets.Presentation/AssetEditors/Gizmos/NavigationBoundingBoxGizmo.cs
Basewq: Good catch, thanks. I did a bulk find and replace and tried to undo the ones that looked more sensible in the old form, and clearly I didn't missed undoing some. I'll fix this, along with any other if you spot them.

Eideren:
LGTM, thanks again @Basewq

xen2:
LGTM to me too, merging. Thanks for those small numerous improvements!

PR #618, opened 2020-02-27
[Graphics] Slightly faster UIBatch and Sprite3DBatch

Basewq:
# PR Details

These changes should make the UI slightly faster.

## Description

Mainly tried to figure out which variables could be declared upfront and pre-calculated so the loops are more straight forward.
Also, trying to strike a balance between code readability vs performance, so the methods like
`Vector4 Vector4Add(ref Vector4 v1, ref Vector4 v2)` is used instead of the standard `ref, ref, out` version.

I've only tested against the Menu sample template.
Here's the FPS of the game without this change:
uibatch_prev - Copy

Here's the FPS of the game with the changes applied:
uibatch_new - Copy

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Xenko.Graphics/Sprite3DBatch.cs
+++ b/sources/engine/Xenko.Graphics/Sprite3DBatch.cs
@@ -2,6 +2,7 @@
 // Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
 
 using System;
+using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 
 using Xenko.Core.Mathematics;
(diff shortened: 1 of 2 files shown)


sources/engine/Xenko.Graphics/Sprite3DBatch.cs
```diff
@@ -142,41 +143,58 @@ protected override unsafe void UpdateBufferValuesFromElementInfo(ref ElementInfo
...
+                    vertex->ColorAdd = drawInfo->ColorAdd;
+                    vertex->Swizzle = swizzle;
+                    vertex->TextureCoordinate = texCoords[i];
+                    vertex->Position = vertexPositions[i];
                 }
             }
         }
-         
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
Kryptos-FR: Not sure it does anything. Inlining rules are pretty strict and most of the time this has no effect. Unless we can prove it does something, we should probably leave it out as it otherwise would be misleading.

According to some sources, it seems to be the case for methods having structs as parameter. But here, since the values are given by reference, I don't know if it still applies. By even if it did, I think the method body is still too big to be considered for inlining.

sources/engine/Xenko.Graphics/UIBatch.cs
```diff
@@ -517,28 +518,32 @@ protected override unsafe void UpdateBufferValuesFromElementInfo(ref ElementInfo
...
                     {
-                        vertex->ColorScale = drawInfo->ColorScale;
-                        vertex->ColorAdd = drawInfo->ColorAdd;
-
-                        vertex->Swizzle = (int)drawInfo->Swizzle;
-                        vertex->TextureCoordinate.X = 0; // cubes are used only for color
-                        vertex->TextureCoordinate.Y = 0; // cubes are used only for color
+                        vertex->ColorScale = colorScale;
+                        vertex->ColorAdd = colorAdd;
+                        vertex->Swizzle = swizzle;
```
Kryptos-FR: Looks like the value was clamped to an `int` before, but not anymore. Was it on purpose?

Basewq: To be honest, the swizzle lines confused me a little.
`drawInfo.Swizzle` is an enum (on both UIBatch & Sprite3DBatch), whereas the ones on the vertex data is a float, so there is a final conversion to float.
I'd like to think casting enum to int doesn't do anything, but starting to doubt myself.
I can convert it to int first if needed.

sources/engine/Xenko.Graphics/Sprite3DBatch.cs
Basewq: I think you're right. I tried to benchmark one with AggressiveInlining and one without and I think the numbers were generally the same. I've removed it now.

sources/engine/Xenko.Graphics/UIBatch.cs
Kryptos-FR: I only reviewed it by sight :wink:. I haven't used Xenko for quite a while, so I'll let the experts figure this out.

Good job trying to improve the performance in any case.

PR #621, opened 2020-03-07
[Rendering] Replaced Sin and Cos calls of same angle with sincos()

tebjan:
# PR Details

Just some minor performance improvement found by browsing shader code.

## Description

Changed sin/cos calls of the same angle with the sincos method. Also tabs -> spaces.

### DIFF
--- a/sources/engine/Xenko.Particles/Shaders/ParticleUtilities.xksl
+++ b/sources/engine/Xenko.Particles/Shaders/ParticleUtilities.xksl
@@ -24,9 +24,9 @@ shader ParticleUtilities
 
     stage float GetLinearDepth(float z)
     {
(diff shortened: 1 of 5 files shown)


xen2:
Thanks!
Could you check if HLSL compiler is not doing that kind of optimization already?
You can try to run fxc on the logs/XXX.hlsl files and see the resulting bytecode (if you see any sincos or not).

tebjan:
Good question, didn't check that. Will have a look and report back.

tebjan:
Yes, it actually does it! These compilers are hard to beat... I also found out that there isn't even sin or cos as individual functions. No matter what you use, it always calls sincos(), often with one argument as null, if it isn't needed.

PR #622, opened 2020-03-09
Pointersimulation

sebllll:
# PR Details

just adds adds PointerSimulated.cs. That class is analogue to MouseSimulated.cs

## Description

This is still a WIP and also my first PR to stride, so please provide feedback on what to improve or change.
One thing i was unsure is the GUID: 
i created the GUI via VS and just inserted it here. Maybe something more has to be done with this GUID like registering it somewhere else.

## Motivation and Context

It helps me to create pointer events from various sources. Imagine a special input-device like a touchframe, that doesn't come with a touch-driver but sends its events via network. With this class, you can create an easy adapter, that translates the incoming network events to pointer-events.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Xenko.Input/Simulated/InputSourceSimulated.cs
+++ b/sources/engine/Xenko.Input/Simulated/InputSourceSimulated.cs
@@ -15,11 +15,14 @@ namespace Xenko.Input
         private List<GamePadSimulated> gamePads = new List<GamePadSimulated>();
         private List<MouseSimulated> mice = new List<MouseSimulated>();
         private List<KeyboardSimulated> keyboards = new List<KeyboardSimulated>();
-        
+        private List<PointerSimulated> pointerDevices = new List<PointerSimulated>();
+
+
         public IReadOnlyList<KeyboardSimulated> Keyboards => keyboards;
         public IReadOnlyList<MouseSimulated> Mice => mice;
         public IReadOnlyList<GamePadSimulated> GamePads => gamePads;
-        
+        public IReadOnlyList<PointerSimulated> PointerDevices => pointerDevices;
+
         public override void Initialize(InputManager inputManager)
         {
         }
@@ -30,6 +33,7 @@ namespace Xenko.Input
             keyboards.Clear();
             mice.Clear();
             gamePads.Clear();
+            pointerDevices.Clear();
         }
 
         public GamePadSimulated AddGamePad()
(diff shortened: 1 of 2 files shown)


sources/engine/Xenko.Input/Simulated/InputSourceSimulated.cs
```diff
@@ -15,11 +15,14 @@ public class InputSourceSimulated : InputSourceBase
...
         private List<KeyboardSimulated> keyboards = new List<KeyboardSimulated>();
-        
+        private List<PointerSimulated> pointerDevices = new List<PointerSimulated>();
+
+
         public IReadOnlyList<KeyboardSimulated> Keyboards => keyboards;
         public IReadOnlyList<MouseSimulated> Mice => mice;
         public IReadOnlyList<GamePadSimulated> GamePads => gamePads;
-        
+        public IReadOnlyList<PointerSimulated> PointerDevices => pointerDevices;
```
xen2: If we want to follow name convention of other devices, maybe `PointerDevices` should be renamed `Pointers`.

sources/engine/Xenko.Input/Simulated/InputSourceSimulated.cs
```diff
@@ -15,11 +15,14 @@ public class InputSourceSimulated : InputSourceBase
         private List<GamePadSimulated> gamePads = new List<GamePadSimulated>();
         private List<MouseSimulated> mice = new List<MouseSimulated>();
         private List<KeyboardSimulated> keyboards = new List<KeyboardSimulated>();
-        
+        private List<PointerSimulated> pointerDevices = new List<PointerSimulated>();
```
xen2: If we want to follow name convention of other devices, maybe `pointerDevices` should be renamed `pointers`.

sources/engine/Xenko.Input/Simulated/InputSourceSimulated.cs
```diff
@@ -101,6 +105,29 @@ public void RemoveAllKeyboards()
             keyboards.Clear();
         }
 
+        public PointerSimulated AddPointerDevice()
```
xen2: Same, `AddPointer`, `RemovePointer`, etc.

sources/engine/Xenko.Input/Simulated/PointerSimulated.cs
```diff
@@ -0,0 +1,61 @@
+// Copyright (c) Xenko contributors (https://xenko.com) and Silicon Studio Corp. (https://www.siliconstudio.co.jp)
```
xen2: New file headers should not contain Silicon Studio Corp copyright, only xenko part

sebllll:
ok thanks, I did the changes.

also removed this unnecessary enum from InputSourceSimulated.cs:
```csharp
private enum DeviceEventType
{
    Add,
    Remove,
}
```

xen2:
Merged, thanks!

PR #623, opened 2020-03-12
[Threading] Dispatcher scheduling (cont. #562)

Eideren:
# PR Details
This is a second attempt to implement some of #562 features, the logic implemented to reduce allocations from that PR has not been transferred over to this one, see the discussion [here](https://github.com/stride3d/xenko/pull/562#issuecomment-594200461) for more info.

## Description
- ThreadPool: Remove the last couple of allocations that the internal 'queue' might create when scheduling work.
- Dispatcher: Schedule work for other threads sooner and in bulk.
- Dispatcher: Continue through the frame as soon as work as been completed instead of waiting for other threads to wake up and signal before continuing even though there aren't any work left for them to deal with.

## Related Issue
None.

## Motivation and Context
Performances.

## Types of changes

## Checklist

Note:
Previously the PR also took care of refactoring the dispatcher's logic into something more maintainable, less logic duplication etc. I couldn't handle that as doing so would prevent the ``DispatchProcessor`` in its current state from caching delegates. I/We'll take care of the rest once the issues with the ``DispatchProcessor`` have been taken care of.

### DIFF
--- a/sources/core/Xenko.Core/Threading/Dispatcher.cs
+++ b/sources/core/Xenko.Core/Threading/Dispatcher.cs
@@ -42,7 +42,7 @@ namespace Xenko.Core.Threading
                 else
                 {
                     var state = BatchState.Acquire();
-                    state.StartInclusive = fromInclusive;
+                    state.WorkDone = state.StartInclusive = fromInclusive;
 
                     try
                     {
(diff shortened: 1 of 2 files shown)


Eideren:
Well, this works but it could be faster if threads allocate a small class each time they wake up, not sure if I should go for that instead.

xen2:
@Eideren 
I will soon merge the renaming branch.
Can I merge this one before?

Eideren:
@xen2 Sure

PR #624, opened 2020-03-12
More minor performance improvement

Basewq:
# PR Details

Minor performance improvement in the game engine.

## Description

~Reduce the number of constructor calls when copying the struct directly is applicable.~
Reduce redundant calculations where possible.
Pass structs by ref where applicable.
Foreach-loop on `VisualChildrenCollection` to avoid enumerator object allocations.

~Note that I've created new `Color3.Black` & `Color3.White` public fields. Not sure if you want this exposed, or keep it internal.~

## Motivation and Context

A little more FPS can't hurt.

## Types of changes

## Checklist

### DIFF
--- a/sources/core/Xenko.Core.Mathematics/Matrix.cs
+++ b/sources/core/Xenko.Core.Mathematics/Matrix.cs
@@ -1454,14 +1454,23 @@ namespace Xenko.Core.Mathematics
             //By separating the above algorithm into multiple lines, we actually increase accuracy.
             result = value;
 
-            result.Row2 = result.Row2 - (Vector4.Dot(result.Row1, result.Row2) / Vector4.Dot(result.Row1, result.Row1)) * result.Row1;
+            var row1 = result.Row1;
+            var row2 = result.Row2;
+            var row3 = result.Row3;
+            var row4 = result.Row4;
 
-            result.Row3 = result.Row3 - (Vector4.Dot(result.Row1, result.Row3) / Vector4.Dot(result.Row1, result.Row1)) * result.Row1;
-            result.Row3 = result.Row3 - (Vector4.Dot(result.Row2, result.Row3) / Vector4.Dot(result.Row2, result.Row2)) * result.Row2;
+            row2 = row2 - (Vector4.Dot(row1, row2) / Vector4.Dot(row1, row1)) * row1;
 
-            result.Row4 = result.Row4 - (Vector4.Dot(result.Row1, result.Row4) / Vector4.Dot(result.Row1, result.Row1)) * result.Row1;
-            result.Row4 = result.Row4 - (Vector4.Dot(result.Row2, result.Row4) / Vector4.Dot(result.Row2, result.Row2)) * result.Row2;
-            result.Row4 = result.Row4 - (Vector4.Dot(result.Row3, result.Row4) / Vector4.Dot(result.Row3, result.Row3)) * result.Row3;
+            row3 = row3 - (Vector4.Dot(row1, row3) / Vector4.Dot(row1, row1)) * row1;
+            row3 = row3 - (Vector4.Dot(row2, row3) / Vector4.Dot(row2, row2)) * row2;
+
+            row4 = row4 - (Vector4.Dot(row1, row4) / Vector4.Dot(row1, row1)) * row1;
+            row4 = row4 - (Vector4.Dot(row2, row4) / Vector4.Dot(row2, row2)) * row2;
+            row4 = row4 - (Vector4.Dot(row3, row4) / Vector4.Dot(row3, row3)) * row3;
+
+            result.Row2 = row2;
+            result.Row3 = row3;
+            result.Row4 = row4;
         }
 
         /// <summary>
@@ -1516,21 +1525,30 @@ namespace Xenko.Core.Mathematics
             //q4 = (m4 - (q1 ⋅ m4) * q1 - (q2 ⋅ m4) * q2 - (q3 ⋅ m4) * q3) / |m4 - (q1 ⋅ m4) * q1 - (q2 ⋅ m4) * q2 - (q3 ⋅ m4) * q3|
 
             //By separating the above algorithm into multiple lines, we actually increase accuracy.
-            result = value;
+            var row1 = value.Row1;
+            var row2 = value.Row2;
+            var row3 = value.Row3;
+            var row4 = value.Row4;
 
-            result.Row1 = Vector4.Normalize(result.Row1);
+            row1.Normalize();
 
-            result.Row2 = result.Row2 - Vector4.Dot(result.Row1, result.Row2) * result.Row1;
-            result.Row2 = Vector4.Normalize(result.Row2);
+            row2 = row2 - Vector4.Dot(row1, row2) * row1;
+            row2.Normalize();
 
-            result.Row3 = result.Row3 - Vector4.Dot(result.Row1, result.Row3) * result.Row1;
-            result.Row3 = result.Row3 - Vector4.Dot(result.Row2, result.Row3) * result.Row2;
-            result.Row3 = Vector4.Normalize(result.Row3);
+            row3 = row3 - Vector4.Dot(row1, row3) * row1;
+            row3 = row3 - Vector4.Dot(row2, row3) * row2;
+            row3.Normalize();
 
-            result.Row4 = result.Row4 - Vector4.Dot(result.Row1, result.Row4) * result.Row1;
-            result.Row4 = result.Row4 - Vector4.Dot(result.Row2, result.Row4) * result.Row2;
-            result.Row4 = result.Row4 - Vector4.Dot(result.Row3, result.Row4) * result.Row3;
-            result.Row4 = Vector4.Normalize(result.Row4);
+            row4 = row4 - Vector4.Dot(row1, row4) * row1;
+            row4 = row4 - Vector4.Dot(row2, row4) * row2;
+            row4 = row4 - Vector4.Dot(row3, row4) * row3;
+            row4.Normalize();
+
+            result = default;
+            result.Row1 = row1;
+            result.Row2 = row2;
+            result.Row3 = row3;
+            result.Row4 = row4;
         }
 
         /// <summary>
--- a/sources/core/Xenko.Core.Mathematics/Matrix.cs
+++ b/sources/core/Xenko.Core.Mathematics/Matrix.cs
@@ -2838,8 +2838,7 @@ namespace Xenko.Core.Mathematics
         /// <param name="result">When the method completes, contains the created rotation matrix.</param>
         public static void RotationYawPitchRoll(float yaw, float pitch, float roll, out Matrix result)
         {
-            Quaternion quaternion;
-            Quaternion.RotationYawPitchRoll(yaw, pitch, roll, out quaternion);
+            Quaternion.RotationYawPitchRoll(yaw, pitch, roll, out var quaternion);
             RotationQuaternion(ref quaternion, out result);
         }
 
--- a/sources/core/Xenko.Core.Mathematics/Matrix.cs
+++ b/sources/core/Xenko.Core.Mathematics/Matrix.cs
@@ -694,7 +694,7 @@ namespace Xenko.Core.Mathematics
         /// Decomposes a matrix into a scale, rotation, and translation.
         /// </summary>
         /// <param name="scale">When the method completes, contains the scaling component of the decomposed matrix.</param>
-        /// <param name="rotation">When the method completes, contains the rtoation component of the decomposed matrix.</param>
+        /// <param name="rotation">When the method completes, contains the rotation component of the decomposed matrix.</param>
         /// <param name="translation">When the method completes, contains the translation component of the decomposed matrix.</param>
         /// <remarks>
         /// This method is designed to decompose an SRT transformation matrix only.
@@ -711,7 +711,7 @@ namespace Xenko.Core.Mathematics
         /// Decomposes a matrix into a scale, rotation, and translation.
         /// </summary>
         /// <param name="scale">When the method completes, contains the scaling component of the decomposed matrix.</param>
-        /// <param name="rotation">When the method completes, contains the rtoation component of the decomposed matrix.</param>
+        /// <param name="rotation">When the method completes, contains the rotation component of the decomposed matrix.</param>
         /// <param name="translation">When the method completes, contains the translation component of the decomposed matrix.</param>
         /// <remarks>
         /// This method is designed to decompose an SRT transformation matrix only.
@@ -2820,7 +2820,7 @@ namespace Xenko.Core.Mathematics
         /// <param name="result">When the method completes, contains the created rotation matrix.</param>
         public static void RotationYawPitchRoll(float yaw, float pitch, float roll, out Matrix result)
         {
-            Quaternion quaternion = new Quaternion();
+            Quaternion quaternion;
             Quaternion.RotationYawPitchRoll(yaw, pitch, roll, out quaternion);
             RotationQuaternion(ref quaternion, out result);
         }
--- a/sources/core/Xenko.Core.Mathematics/Color3.cs
+++ b/sources/core/Xenko.Core.Mathematics/Color3.cs
@@ -66,7 +66,9 @@ namespace Xenko.Core.Mathematics
         /// <param name="value">The value that will be assigned to all components.</param>
         public Color3(float value)
         {
-            R = G = B = value;
+            R = value;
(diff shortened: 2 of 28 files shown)


Kryptos-FR:
Would be nice to have some benchmarks to be sure there is an actual gain.

Basewq:
Fair enough, I'll look at making a formal-ish benchmark project later.

Basewq:
Here's the [link to the project](https://github.com/Basewq/XenkoCodeTestBenchmarks) to how I wrote the benchmarks, so it's open to critique if I've done the benchmark wrong.

Ran on a Win7 machine, Intel Core i5-3570K CPU 3.40GHz.

Oddly, I can't run BenchmarkDotNet in .NET 4.6.1 (which is what I would have preferred to test under).
Targeting net472, BenchmarkDotNet runs under:
Runtime=.NET Framework 4.8 (4.8.3928.0), X64 RyuJIT
Targeting netcoreapp2.0, BenchmarkDotNet runs under:
Runtime=.NET Core 2.1.15 (CoreCLR 4.6.28325.01, CoreFX 4.6.28327.02), X64 RyuJIT


These benchmarks are relatively close to the commits in this pull.
Anyway, onto the results:

**Caching a Vector3's property via its indexer:**
```
net472
|                                           Method |     Mean |    Error |   StdDev |
|------------------------------------------------- |---------:|---------:|---------:|
| Canvas_ComputeAbsolutePinPosition_InlineProperty | 41.77 ms | 0.696 ms | 0.651 ms |
|  Canvas_ComputeAbsolutePinPosition_CacheProperty | 29.02 ms | 0.192 ms | 0.179 ms |

netcoreapp2.0
|                                           Method |     Mean |    Error |   StdDev |
|------------------------------------------------- |---------:|---------:|---------:|
| Canvas_ComputeAbsolutePinPosition_InlineProperty | 31.99 ms | 0.080 ms | 0.071 ms |
|  Canvas_ComputeAbsolutePinPosition_CacheProperty | 29.59 ms | 0.181 ms | 0.169 ms |

```
No surprises here, there's a switch case inside the Vector's property accessor.

**Color3 constructors vs copying static readonly field:**
```
net472
|                   Method |     Mean |     Error |    StdDev |
|------------------------- |---------:|----------:|----------:|
|   Color_EmptyConstructor | 1.803 ms | 0.0291 ms | 0.0272 ms |
| Color_ConstructorArgZero | 3.022 ms | 0.0048 ms | 0.0042 ms |
|              Color_Black | 2.094 ms | 0.0227 ms | 0.0212 ms |
|  Color_ConstructorArgOne | 3.123 ms | 0.0101 ms | 0.0089 ms |
|              Color_White | 2.057 ms | 0.0077 ms | 0.0068 ms |

netcoreapp2.0
|                   Method |     Mean |     Error |    StdDev |
|------------------------- |---------:|----------:|----------:|
|   Color_EmptyConstructor | 1.808 ms | 0.0105 ms | 0.0098 ms |
| Color_ConstructorArgZero | 2.983 ms | 0.0201 ms | 0.0188 ms |
|              Color_Black | 1.989 ms | 0.0150 ms | 0.0125 ms |
|  Color_ConstructorArgOne | 3.071 ms | 0.0317 ms | 0.0296 ms |
|              Color_White | 1.999 ms | 0.0374 ms | 0.0367 ms |

```
Ok, a little surprising (or maybe not), the default constructor is faster than copying, though using the "real" constructor being slower than copying is what I expected.

**Hoisting the 'lerpedAttenuation' in LightStreak.DrawCore (with its default properties)**
```
net472
|                     Method |    Mean |    Error |   StdDev |
|--------------------------- |--------:|---------:|---------:|
| DrawCore_InlineAttenuation | 4.490 s | 0.0108 s | 0.0101 s |
|  DrawCore_HoistAttenuation | 4.377 s | 0.0164 s | 0.0145 s |

netcoreapp2.0
|                     Method |    Mean |    Error |   StdDev |
|--------------------------- |--------:|---------:|---------:|
| DrawCore_InlineAttenuation | 4.076 s | 0.0374 s | 0.0350 s |
|  DrawCore_HoistAttenuation | 3.947 s | 0.0117 s | 0.0109 s |

```
Razor thin gain. Still think it's worth it, though!

**Matrix methods performance:**
```
net472
|                        Method |      Mean |    Error |   StdDev |
|------------------------------ |----------:|---------:|---------:|
|       Invert_AssignThenInvert |  23.42 ms | 0.097 ms | 0.081 ms |
|         Invert_ByValAndReturn |  24.60 ms | 0.192 ms | 0.180 ms |
|              Invert_RefAndOut |  21.27 ms | 0.141 ms | 0.132 ms |
|  Orthogonalize_NoTempVariable | 313.36 ms | 0.618 ms | 0.548 ms |
| Orthogonalize_UseTempVariable |  66.44 ms | 0.447 ms | 0.418 ms |

netcoreapp2.0
|                        Method |      Mean |    Error |   StdDev |
|------------------------------ |----------:|---------:|---------:|
|       Invert_AssignThenInvert |  23.81 ms | 0.042 ms | 0.040 ms |
|         Invert_ByValAndReturn |  23.41 ms | 0.030 ms | 0.028 ms |
|              Invert_RefAndOut |  20.82 ms | 0.029 ms | 0.027 ms |
|  Orthogonalize_NoTempVariable | 308.71 ms | 0.445 ms | 0.417 ms |
| Orthogonalize_UseTempVariable |  65.33 ms | 0.172 ms | 0.161 ms |

```
No surprises here.

**ModelComponent for-loop accessing 'nested' fields through list -> array**
```
net472
|                      Method |     Mean |   Error |  StdDev |
|---------------------------- |---------:|--------:|--------:|
|  Update_HasNestedReferences | 830.2 ms | 3.07 ms | 2.88 ms |
|   Update_NoNestedReferences | 797.8 ms | 3.30 ms | 3.09 ms |
| Update_HasNestedReferences2 | 825.1 ms | 2.45 ms | 2.29 ms |
|  Update_NoNestedReferences2 | 802.1 ms | 2.18 ms | 2.04 ms |

netcoreapp2.0
|                      Method |     Mean |   Error |  StdDev |
|---------------------------- |---------:|--------:|--------:|
|  Update_HasNestedReferences | 824.1 ms | 3.45 ms | 3.23 ms |
|   Update_NoNestedReferences | 796.2 ms | 5.29 ms | 4.94 ms |
| Update_HasNestedReferences2 | 811.4 ms | 0.90 ms | 0.79 ms |
|  Update_NoNestedReferences2 | 794.4 ms | 1.03 ms | 0.97 ms |

```
Note I run the methods twice because the benchmark hits some weird stalling behavior on the first runs on my machine (as seen by the large Error/StdDev). Probably because it didn't like me instantiating 100000 Model objects.

Basewq:
Reverted all the `Color.Black` back to use default constructor, expect for one in the constructor of `Vignetting` which shouldn't hurt since it's not a frequently instantiated object.

tebjan:
great work on benchmarking everything, thanks!

Kryptos-FR:
> I can't run BenchmarkDotNet in .NET 4.6.1 (which is what I would have preferred to test under).

Why not? I use BenchmarkDotNet often at work targeting 4.5.2. So there shouldn't be any issue to do it for 4.6.1.

Just use an older version such as https://www.nuget.org/packages/BenchmarkDotNet/0.11.3

Basewq:
> 
> Why not? I use BenchmarkDotNet often at work targeting 4.5.2. So there shouldn't be any issue to do it for 4.6.1. Just use an older version such as https://www.nuget.org/packages/BenchmarkDotNet/0.11.3

Do you know if you're using the old csproj format or new format at work?

My issue shows this type of error https://github.com/dotnet/standard/issues/704
I've tried using 0.11.3, but same issue. Anyway, I'll play around.

Kryptos-FR:
@Basewq Thanks for the results. That should be enough.

From that, I take that `Color.Black` is not needed. I am not surprised since the default (empty) constructor for a struct is basically a no-op: it just sets all fields to their default value from a "zero" memory initialization.

`Orthogonalize_UseTempVariable` shows a real gain, the rest is marginal.

Eideren:
@Kryptos-FR 
> I take that Color.Black is not needed
I don't agree with that, the rest of our (and other) API still define concrete variables for common default values ( Vector*.Zero, int*.Zero, Point.Zero, Matrix.Zero, ... ) even if that path is always slower.
We should either remove all *.Zero from our other types or add this one to keep our API consistent.

> the rest is marginal.
Perhaps but it still improves performances without degrading readability/clarity, I say we merge this in, I don't see any reason not to do it.

Kryptos-FR:
I slightly disagree. We should indeed remove all the constant for Zero. There are not needed and have worst performance. Considering that now the language as a simple `default` keyword:

```csharp
// Slow (does a copy)
Vector3 v = Vector3.Zero;
// Slower (initializes fields one by one)
Vector3 v = new Vector3(0, 0, 0);
// As fast as can be
Vector3 v = default;
```

Basewq:
First, I have to apology for my crappy benchmarking approach.
A while back I was convinced that copying a static readonly struct was faster than calling a constructor.
Turns out it was **wrong**.
I expanded the constructor tests to test Color3, Vector3 & Vector4 and got the following:

```
|                      Method |       Runtime |     Mean |    Error |   StdDev |
|---------------------------- |-------------- |---------:|---------:|---------:|
|     Color3_EmptyConstructor |    .NET 4.6.1 | 21.45 ms | 0.166 ms | 0.155 ms |
|    Color3_EmptyConstructor2 |    .NET 4.6.1 | 21.36 ms | 0.083 ms | 0.064 ms |
|              Color3_Default |    .NET 4.6.1 | 21.43 ms | 0.078 ms | 0.065 ms |
|             Color3_Default2 |    .NET 4.6.1 | 21.38 ms | 0.084 ms | 0.070 ms |
|    Color3_ConstructorArgOne |    .NET 4.6.1 | 28.82 ms | 0.078 ms | 0.069 ms |
|   Color3_ConstructorArgOne2 |    .NET 4.6.1 | 21.49 ms | 0.076 ms | 0.063 ms |
|  Color3_ConstructorArgThree |    .NET 4.6.1 | 21.48 ms | 0.077 ms | 0.064 ms |
| Color3_ConstructorArgThree2 |    .NET 4.6.1 | 21.46 ms | 0.055 ms | 0.046 ms |
|                Color3_White |    .NET 4.6.1 | 22.11 ms | 0.427 ms | 0.539 ms |
|               Color3_White2 |    .NET 4.6.1 | 22.00 ms | 0.433 ms | 0.532 ms |
|     Color3_EmptyConstructor | .NET Core 2.0 | 22.05 ms | 0.489 ms | 0.582 ms |
|    Color3_EmptyConstructor2 | .NET Core 2.0 | 21.70 ms | 0.290 ms | 0.257 ms |
|              Color3_Default | .NET Core 2.0 | 21.81 ms | 0.414 ms | 0.387 ms |
|             Color3_Default2 | .NET Core 2.0 | 21.60 ms | 0.243 ms | 0.215 ms |
|    Color3_ConstructorArgOne | .NET Core 2.0 | 28.95 ms | 0.122 ms | 0.114 ms |
|   Color3_ConstructorArgOne2 | .NET Core 2.0 | 21.52 ms | 0.137 ms | 0.114 ms |
|  Color3_ConstructorArgThree | .NET Core 2.0 | 21.56 ms | 0.155 ms | 0.137 ms |
| Color3_ConstructorArgThree2 | .NET Core 2.0 | 21.81 ms | 0.273 ms | 0.228 ms |
|                Color3_White | .NET Core 2.0 | 21.47 ms | 0.307 ms | 0.256 ms |
|               Color3_White2 | .NET Core 2.0 | 21.59 ms | 0.306 ms | 0.255 ms |

|                      Method |       Runtime |     Mean |    Error |   StdDev |
|---------------------------- |-------------- |---------:|---------:|---------:|
|    Vector3_EmptyConstructor |    .NET 4.6.1 | 21.48 ms | 0.084 ms | 0.070 ms |
|             Vector3_Default |    .NET 4.6.1 | 21.69 ms | 0.426 ms | 0.437 ms |
|   Vector3_ConstructorArgOne |    .NET 4.6.1 | 22.36 ms | 0.505 ms | 0.910 ms |
| Vector3_ConstructorArgThree |    .NET 4.6.1 | 22.51 ms | 0.446 ms | 0.438 ms |
|          Vector3_Vector3One |    .NET 4.6.1 | 22.42 ms | 0.507 ms | 0.774 ms |
|    Vector3_EmptyConstructor | .NET Core 2.0 | 24.28 ms | 0.388 ms | 0.363 ms |
|             Vector3_Default | .NET Core 2.0 | 22.79 ms | 0.453 ms | 0.378 ms |
|   Vector3_ConstructorArgOne | .NET Core 2.0 | 21.83 ms | 0.430 ms | 0.403 ms |
| Vector3_ConstructorArgThree | .NET Core 2.0 | 21.45 ms | 0.060 ms | 0.050 ms |
|          Vector3_Vector3One | .NET Core 2.0 | 21.91 ms | 0.528 ms | 0.519 ms |

|                      Method |       Runtime |     Mean |    Error |   StdDev |
|---------------------------- |-------------- |---------:|---------:|---------:|
|    Vector4_EmptyConstructor |    .NET 4.6.1 | 28.62 ms | 0.377 ms | 0.353 ms |
|             Vector4_Default |    .NET 4.6.1 | 28.45 ms | 0.169 ms | 0.158 ms |
|   Vector4_ConstructorArgOne |    .NET 4.6.1 | 28.69 ms | 0.095 ms | 0.089 ms |
| Vector4_ConstructorArgThree |    .NET 4.6.1 | 29.07 ms | 0.570 ms | 0.560 ms |
|          Vector4_Vector4One |    .NET 4.6.1 | 28.31 ms | 0.060 ms | 0.053 ms |
|    Vector4_EmptyConstructor | .NET Core 2.0 | 28.33 ms | 0.032 ms | 0.027 ms |
|             Vector4_Default | .NET Core 2.0 | 28.32 ms | 0.050 ms | 0.041 ms |
|   Vector4_ConstructorArgOne | .NET Core 2.0 | 28.54 ms | 0.044 ms | 0.039 ms |
| Vector4_ConstructorArgThree | .NET Core 2.0 | 28.56 ms | 0.037 ms | 0.031 ms |
|          Vector4_Vector4One | .NET Core 2.0 | 28.89 ms | 0.557 ms | 0.547 ms |
```

I've upped the count to 10,000,000 items, I don't know if that actually makes more accurate or overly optimized by the jitter...
I've ran the benchmark multiple times, but I suspect these code is too small to judge properly, and probably not wise to overly judge the times + StdDev.
eg. Vector3_EmptyConstructor in .NET Core 2.0 says it's slower than the rest, even with StdDev but Vector4_EmptyConstructor is about the same as Vector4_Default (which is what I'd expect).
I think the final answer to the copy prefined vs constructor is...they're probably same, or at least it's not conclusively faster, accounting for the StdDev.

So where did I get my faulty assumption?
You'll notice Color3_ConstructorArgOne actually stands out at 28ms whereas the rest is in 21/22ms range. Back in a previous commit, this was what I benchmarked, and assumed the rest of the constructors performance would be the same as that single arg constructor. I was wrong.
The single arg constructor for Color3 was the real outlier.
It turns out `R = G = B = value;` was the thing dragging down the performance (in fact this is also visible in the IL).
The test Color3_ConstructorArgOne2 is changing the constructor so each R/G/B is assigned to `value` individually, showing it brings the performance back in line with the rest.

Kryptos-FR:
Ok. Maybe we are going a bit too far here. I share that blame with my comments.

Do what you think is best. The two cases where the difference was noticeable need to be merged. For the rest maybe we are out of our leagues, like you said the JIT might do some optimizations anyway.

Thanks for taking the time to do all those tests. I feel like we all learned something here anyway.

Basewq:
Ok, I've decided to ditch the predefined Color3's. Probably doesn't matter in the grand scheme of things, at least for the time being, unless I can find a more convincing benchmark.
Also removed the caching the lerpedAttenuation (though I was a little surprised the jitter was ok with that, but I guess it detected the Attenuation property was 'simple').
The rest should be fine as is.

sources/core/Xenko.Core.Mathematics/Matrix.cs
```diff
@@ -2820,7 +2838,7 @@ public static Matrix RotationQuaternion(Quaternion rotation)
         /// <param name="result">When the method completes, contains the created rotation matrix.</param>
         public static void RotationYawPitchRoll(float yaw, float pitch, float roll, out Matrix result)
         {
-            Quaternion quaternion = new Quaternion();
+            Quaternion quaternion;
```
xen2: Could even do a `out var quaternion` here.

xen2:
Merged, thanks for the detailed investigation!

PR #626, opened 2020-03-17
Fix Game Studio adding duplicate asset templates

Basewq:
# PR Details

Fixes a special case where I have explicitly included a Xenko package in my project, and Game Studio adds its own packages (in `ExtraPackages`) causing duplicate templates to appear (in the 'Add Asset' menu).

## Description

This change checks if any of the `ExtraPackage`s is already defined in the current session's packages and ignores re-adding them.

## Motivation and Context

The truth is that all I want is access to `GizmoComponentAttribute` and `EntityGizmo` so that I can get my custom components in Game Studio to be clickable within the scene, but currently this requires referencing `Xenko.Assets.Presentation` to get to these classes, and I'm not really willing to make any massive changes to get this to work.
`ExtraPackages` also defines `Xenko.Assets.Presentation` so the concatenation duplicates the asset templates.

I'm aware this is more of a hidden "feature" so feel free to reject this pull.

## Types of changes

## Checklist

### DIFF
--- a/sources/assets/Xenko.Core.Assets/Templates/TemplateManager.cs
+++ b/sources/assets/Xenko.Core.Assets/Templates/TemplateManager.cs
@@ -59,7 +59,12 @@ namespace Xenko.Core.Assets.Templates
         /// <returns>A sequence containing all registered template descriptions.</returns>
         public static IEnumerable<TemplateDescription> FindTemplates(PackageSession session = null)
         {
-            var packages = session?.Packages.Concat(ExtraPackages) ?? ExtraPackages;
+            IEnumerable<Package> packages = session?.Packages ?? ExtraPackages;
+            if (packages != ExtraPackages)
+            {
+                var extraPackages = ExtraPackages.Where(x => !packages.Any(sessionPkg => sessionPkg.FullPath == x.FullPath)).ToArray();
+                packages = packages.Concat(extraPackages);
+            }
             // TODO this will not work if the same package has different versions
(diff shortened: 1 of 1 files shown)


sources/assets/Xenko.Core.Assets/Templates/TemplateManager.cs
```diff
@@ -59,7 +59,12 @@ public static void Unregister(ITemplateGenerator generator)
         /// <returns>A sequence containing all registered template descriptions.</returns>
         public static IEnumerable<TemplateDescription> FindTemplates(PackageSession session = null)
         {
-            var packages = session?.Packages.Concat(ExtraPackages) ?? ExtraPackages;
+            IEnumerable<Package> packages = session?.Packages ?? ExtraPackages;
+            if (packages != ExtraPackages)
```
xen2: I am afraid the extra `if` branch (depending on which part of ternary operator is taken) is not so easy to understand at a quick glance.

What about using `Distinct()` to remove duplicates?
`var packages = session?.Packages.Concat(ExtraPackages).Distinct() ?? ExtraPackages;`

Basewq: Done, however it needs a comparer since the packages are not object equals (in fact the `ExtraPackages` packages have empty `Meta` data.
I avoided Distinct the first time round as I'm semi-conscious of the hashset it builds behind the scenes, however since I think this isn't called frequently it should be ok.

PR #642, opened 2020-03-30
[Shaders] Added ShaderClassString Step 1

tebjan:
* Added `ShaderClassString` and a common base class `ShaderClassCode`
* Where possible, the common base class is used
* `LoadShaderSource` can now also take a source code string as input
* Added package upgrader and increased `MaterialAssetCompiler` to version 5
* Added `TestShaderLoadingString`

This PR doesn't change functionality for existing shaders but lays the foundation to create shaders from a string.

Creating shaders from string is not yet fully tested. The step `ExtendLibrary` currently fails. I would need some input in order to fix that.


xen2:
Even though it's not functional, I will merge it to make it easier with the renaming.

Issue #111, opened 2018-08-15
Editor loses padding value set on buttons in UI library after restart.


jayrulez:
I created a button in a UI library setting the padding to 0 all around.

When the editor restarts the value of the padding is changed from zeros to 10 10 5 7.

jayrulez:
I don't know much about the UI framework yet but this would seem to reset the padding everytime a button is constructed: https://github.com/xenko3d/xenko/blob/master/sources/engine/Xenko.UI/Controls/Button.cs#L31

jayrulez:
Any luck here @Kryptos-FR ? If you point me where to look I can take a stab and  possibly push a fix.

Issue #153, opened 2018-08-26
Default value in YAML are ignored if set in derived class constructor


Kryptos-FR:
In YAML, default values are not explicitly saved. That makes the YAML file smaller and easier to read: only properties with values that are different than their default are visible.

It usually works well when deserializing the YAML file but there is a situation when it does not.

If a property is defined in class `A` with a certain default value, but the derived class `B` set the property with a different value in its constructor. Then if the actual saved value is the default one (as defined in `A`), then after deserialization it will have `B`'s value instead.

A good example is the `ButtonBase` class. It sets the `CanBeHitByUser` to `true`. However, the default value in the base class `UIElement` is `false`. If a user creates a UI page or UI library and sets the value to `false`. Then the next time the asset is deserialized, the value is reset to `true` instead.

PR #643, opened 2020-04-01
Fix YAML serialization when control subclasses override the default values

Basewq:
# PR Details

Fix YAML bug where derived controls change the default value, which doesn't get detected by the serializer.

## Description

Added a new attribute to allow for overriding default values via a `DataContractMetadataTypeAttribute`.
This is similar to .NET's [MetadataTypeAttribute](
https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.metadatatypeattribute?view=netframework-4.8), however our rules differ slightly. eg. We allow all sub-classes to declare the metadata attribute.
The metadata property is matched on both field/property name AND the data type. This is because a user might hide a property with a new one in the sub class, but it has a different data type (unlikely scenario...?).

The other fix is to push the 'parent' property that generated the value and check the default value from that (if it exists).

## Related Issue

This is my proposal on how to fix issue #111 & #153
There are two issues at hand:
#111 occurs because the control is defining a default value for an entire struct that differs from the default values in the underlying properties. eg. For a `Button.Padding`, the default is expected to be `(10, 5, 10, 7)`, but `Thickness` states the default for each dimension is `0`. The serializer does not account for the 'higher' class' declaration of the default value, so explicitly setting `(0, 0, 0, 0)` is treated as 'do not serialize'.

#153 occurs because only the property with the `DefaultValueAttribute` is considered for the default value during serialization. It is not possible in C# to get a sub-class 'override' the attribute itself.
Inspired by `MetadataTypeAttribute`, this solution allows for sub-classes to 'override' the attribute by declaring it on a metadata class.

## Motivation and Context

Assets are not saved/reloaded correctly due to inconsistent default values between base class and derived classes.

The alternative solutions to a metadata class are:
1. Hide the base class property `public new Thickness Padding { get; set; }`
2. Change the property to `virtual` and override specifically to allow adding an attribute.

Not sure if there is an alternative solution for #111.

## Types of changes

## Checklist

### DIFF
--- a/sources/assets/Xenko.Core.Assets/Tracking/SourceHashesHelper.cs
+++ b/sources/assets/Xenko.Core.Assets/Tracking/SourceHashesHelper.cs
@@ -112,7 +112,14 @@ namespace Xenko.Core.Assets.Tracking
         {
             public const int DefaultOrder = int.MaxValue;
 
-            public static readonly SourceHashesDynamicMember Default = new SourceHashesDynamicMember { ShouldSerialize = x => { var asset = x as Asset; return asset != null && TryGet(asset, AbsoluteSourceHashesKey)?.Count > 0; } };
+            public static readonly SourceHashesDynamicMember Default = new SourceHashesDynamicMember
+            {
+                ShouldSerialize = (x, parentTypeMemberDesc) =>
+                {
+                    var asset = x as Asset;
+                    return asset != null && TryGet(asset, AbsoluteSourceHashesKey)?.Count > 0;
+                }
+            };
 
             static SourceHashesDynamicMember()
             {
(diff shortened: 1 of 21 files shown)


xen2:
I didn't know about `MetadataTypeAttribute` but I like the approach as it's easy to define stuff externally. Also this combines well in UI thanks to ability to add custom `DefaultValueAttribute` for any type.

Thanks, great work as usual and bonus points for the unit tests!

xen2:
I was thinking, wouldn't it be good to generate some warning (at build time, or runtime in debug mode) if some properties in the metadata type don't exist in the actual type anymore? (or if property type differ)
This would probably avoid some headache when user forget to update metadata properties after a property rename.
No need to do this right now as it's not used much (yet), just writing it so that we don't forget.
If you agree it's useful, maybe we could simply create a github issue to keep it in mind for later.

Basewq:
Yes, that's a good idea considering it's not strongly bound. I'll add the issue/enhancement.

PR #644, opened 2020-04-02
Minor code fixes, remove unsafe method

Basewq:
# PR Details

Minor fixes.

## Description

Use the correct variable for the for-loop bounds check. The two arrays already had the same sizes so it never caused any issues.

Removed `ParameterCollection.GetValuePointer` since the code leaves the `fixed` scope, the pointer can potentially point to an invalid location if the garbage collector moves the array.
The two methods that used it now implement the pointer code correctly.
Any external users that have been using it, 1. shouldn't, 2. have access to `DataValues` and can create a fixed pointer themselves.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Xenko/Rendering/ParameterCollection.cs
+++ b/sources/engine/Xenko/Rendering/ParameterCollection.cs
@@ -176,18 +176,6 @@ namespace Xenko.Rendering
         }
 
         /// <summary>
-        /// Gets pointer to directly copy blittable values.
-        /// </summary>
-        /// <typeparam name="T"></typeparam>
-        /// <param name="parameter"></param>
-        /// <returns></returns>
-        public unsafe IntPtr GetValuePointer<T>(ValueParameter<T> parameter) where T : struct
-        {
-            fixed (byte* dataValues = DataValues)
-                return (IntPtr)dataValues + parameter.Offset;
-        }
-
-        /// <summary>
         /// Sets an object.
         /// </summary>
         /// <typeparam name="T"></typeparam>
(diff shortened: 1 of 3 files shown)


Kryptos-FR:
Wondering if [`Span<T>`](https://docs.microsoft.com/en-us/dotnet/api/system.span-1) and/or [`Unsafe`](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.unsafe) could be used here, so that the `unsafe` keyword is not needed on the method signature.

But maybe it should be the subject of another issue or pull-request.

Eideren:
> 
> 
> Wondering if [`Span<T>`](https://docs.microsoft.com/en-us/dotnet/api/system.span-1) and/or [`Unsafe`](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.unsafe) could be used here, so that the `unsafe` keyword is not needed on the method signature.
> 
> But maybe it should be the subject of another issue or pull-request.

Framework doesn't support those types though

Basewq:
If you want that method reinstated properly, then I think you can use ref return.
That said, this method (even in it's existing state) doesn't seem usable for users if they're wanting to access an array of values since it doesn't return a 'count'.

sources/engine/Xenko/Rendering/ParameterCollection.cs
```diff
@@ -175,18 +175,6 @@ private unsafe Accessor GetValueAccessorHelper(ParameterKey parameterKey, int el
...
-        /// <summary>
-        /// Gets pointer to directly copy blittable values.
-        /// </summary>
-        /// <typeparam name="T"></typeparam>
-        /// <param name="parameter"></param>
-        /// <returns></returns>
-        public unsafe IntPtr GetValuePointer<T>(ValueParameter<T> parameter) where T : struct
-        {
-            fixed (byte* dataValues = DataValues)
-                return (IntPtr)dataValues + parameter.Offset;
```
xen2, line 187: Good catch! I think it was a remnant of the time when `DataValues` was either native or pinned.

xen2:
> EDIT: Oops, you guys are talking about something else.
> Actually Span & Unsafe are usable in Framework. You need to reference the library `System.Memory`. Not sure about the performance, though.
I will merge the PR as is for now as it fixes a potential bad memory access.
We can see about Span/Unsafe separately (might be especially worth doing once .NET Core 3 switch, and to do it in many other places as well at the same time).

PR #649, opened 2020-04-11
Rename Xenko to Stride

xen2:
# PR Details

Rename Xenko to Stride

## Motivation and Context

Background:
https://forums.xenko.com/t/xenko-help-us-find-a-new-name-logo-for-the-engine-if-we-have-to-rename/1980
https://forums.xenko.com/t/vote-for-xenkos-new-name/2016

## Types of changes

## Checklist


sources/assets/Stride.Core.Assets/PackageSession.Dependencies.cs
```diff
@@ -132,6 +136,18 @@ private async Task PreLoadPackageDependencies(ILogger log, SolutionProject proje
                             if (referencedProject != null)
                             {
                                 await PreLoadPackageDependencies(log, referencedProject, loadParameters);
+
+                                // Get package upgrader from dependency (a project might depend on another project rather than referencing Stride directly)
+                                // A better system would be to evaluate nuget flattened dependencies WITHOUT doing the actual restore (dry-run).
+                                // However I am not sure it's easy/possible to do it (using API) without doing a full restore/download, which we don't want to do
+                                // with old version (it might be uninstalled already and we want to avoid re-downloading it again)
+                                pendingPackageUpgradesPerPackage.TryGetValue(referencedProject.Package, out var dependencyPackageUpgraders);
+                                foreach (var dependencyPackageUpgrader in dependencyPackageUpgraders)
```
Basewq, line 145: Sorry if I'm jumping the gun and this is still a work in progress.
I've created a new SpriteStudio sample project and it crashed because `dependencyPackageUpgraders` is null, because the TryGetValue directly above it failed. Wrapping it in an if statement allowed it to go through and the project is created without any issues.

Kryptos-FR:
I was able to compile and create/run a new project. I did have a few errors when compiling tests but maybe that's already the case before the renaming (I need to check that).

sources/assets/Stride.Core.Assets/PackageSession.Dependencies.cs
xen2: Good point, something I overlooked when doing recent changes (I guess I was only testing upgrades, and didn't try new projects again yet).
Fixed with latest commit, thanks!

Basewq:
Another issue found: The custom shader (`.xksl` file) isn't renamed. I'm assuming `.xkfx` will probably have the same issue.
The autogenerated `.xksl.cs` also isn't renamed. The `using` statements are renamed in the `.xksl.cs` file, except for a line declared as an aliasing:
`using Buffer = Xenko.Graphics.Buffer;`

xen2:
@Basewq thanks for the extensive testing, it helps a lot!
I have just pushed a new version which should hopefully behave much better for xksl/xkfx files.

Also, Game Studio in new version doesn't add Navigation/Physics to GameSettings anymore (which might not be referenced, causing game to crash at startup).
This helps for some specific samples which wouldn't work once upgraded to Stride.

Basewq:
Yep, the xksl alias issue is fixed.

One minor issue: the auto-generated cs shader file correctly changes from `.cs` to `.sdsl.cs` (ie. official 3.1 -> 4.0), no problems here.
However, it doesn't rename any files that were already as `.xksl.cs` (ie. some fairly recent commit version -> 4.0)
Not sure if you want to tailor for the bleeding edge users (doesn't matter to me, just making sure you're aware of this).

xen2:
@Basewq I initially thought it was OK to ignore since it wasn't released officially, but thinking about it again, I think vvvv (@tebjan) uses a version containing it already so I can add support for it (esp. since it's quite easy).

Basewq:
Yes, your new changes have fixed the `xksl.cs` -> `sdsl.cs` issue.

I'm not actively trying to test this (honest!), but...
I created a new FPS sample template in the Stride editor. There are a few Asset warnings, eg.
```
Verbose:The model [Models/StrideCrate] from import [C:/FirstPersonShooter2/Resources/Models/StrideCrate.fbx] has been successfully imported.
Warning:Model references material 'XenkoLogo', but it was not defined in the ModelAsset.
```

If you try re-importing eg. StrideCrate.fbx, it'll import a material called XenkoLogo. I believe a few other fbx files will also be similar with this issue.
Nothing's crashing, so probably more of an FYI.

xen2:
@Basewq good catch, thanks!

Basewq:
Another coincidence, I swear...
The Launcher can't open any existing projects.
`RecentProjectViewModel.OpenWith` does a "name" search, but `StrideVersionName` is purely numeric because it's `Version.ToString()`, but `StrideVersionViewModel.Name` has now prepended `PackageSimpleName`, so the search always fails.

xen2:
@Basewq Fixed!
Thanks for the "coincidence testing", very helpful :)

PR #646, opened 2020-04-04
[Editor] Prevent multi-dimensional array from crashing the editor

Basewq:
## PR Details

Fix multi-dimensional array crashing the editor.
Annotate 'Not supported' text to property window for unsupported object types.

## Description

Treat multi-dimensional array as an object type to prevent the editor from crashing.
Append 'Not supported' text to unsupported object types in the property window.
Note that I don't know how localization really works, so would like someone to help in this area.

## Related Issue

Fixes #431 & #645

## Motivation and Context

Don't want the editor to crash.

## Types of changes

- [ ] Docs change / refactoring / dependency upgrade
- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [ ] My change requires a change to the documentation.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #654, opened 2020-04-19
[Editor] Fix property value showing '(Not supported)' for null values

### LINKED ISSUE #650: 'Not supported' property message showing on known types
**Release Type**: GitHub

**Version**: Commit 54a7a59c52eca744afb34d34887b2db27bec7863

**Platform(s)**: All

**Describe the bug**
I recently created pull #646, but the '(Not supported)' message is showing on null properties of supported types and some objects.

**Expected behavior**
Not append the (Not supported) message.

**Screenshots**
Example properties:
Untitled

**Additional context**
Either the change to the specific line can be reverted:
https://github.com/stride3d/xenko/blob/54a7a59c52eca744afb34d34887b2db27bec7863/sources/editor/Xenko.Core.Assets.Editor/View/DefaultPropertyTemplateProviders.xaml#L1346

or I will have a pull ready (after the rename is merged), and append the message on the specific types (ie. only explicitly for the multi dimensional array types).

Basewq:
## PR Details

Fixes bug introduced by #646

## Description

Can't just append the '(Not supported)' message to all objects.
A specific `ObjectDescriptor` type is needed to mark these types.

## Related Issue

Fixes #650 

## Motivation and Context

Unfortunately, this means only multi-array types can be marked.
The problem is that some properties accept null as a valid object type (this outputs 'None'). There are also properties that accept an actual object, eg. the 'Sort Mode' on the render stage.

A minor issue is that things like `DateTime` will not say '(Not supported)', however it does already just say DateTime and isn't editable.

## Types of changes

## Checklist

### DIFF
--- a/sources/core/Stride.Core.Reflection/TypeDescriptorFactory.cs
+++ b/sources/core/Stride.Core.Reflection/TypeDescriptorFactory.cs
@@ -96,7 +96,7 @@ namespace Stride.Core.Reflection
             }
             else if (type.IsArray)
             {
-                if (type.GetArrayRank() == 1)
+                if (type.GetArrayRank() == 1 && !type.GetElementType().IsArray)
                 {
                     // array[] - only single dimension array is supported
                     descriptor = new ArrayDescriptor(this, type, emitDefaultValues, namingConvention);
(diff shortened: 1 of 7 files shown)


tebjan:
hah, congrats to the first Stride PR :)

PR #656, opened 2020-04-20
[Editor] Add ability to 'Show source file in Explorer' for an asset

Basewq:
## Description

Add new right-click Asset context menu to 'Show source file in Explorer'
If the file no longer exists, then try to browse to the folder it belonged to.
If the folder no longer exists, then show an error message.
Menu is always disabled for Assets that don't have `MainSource` set.

## Motivation and Context

I want to browse to the actual file that the asset is pointing to because I want to do something with it.

## Types of changes

## Checklist


Kryptos-FR:
LGTM

PR #662, opened 2020-04-22
Change to open source AvalonDock & remove Xceed library

### LINKED ISSUE #659: Game Studio Editor docking layout resize bug
**Release Type**: Github

**Version**:  Commit 04bb068ca9eafe3b98dc412e86aba049c2d64cbd onwards (probably)

**Platform(s)**: Windows

**Describe the bug**
Extended.Wpf.Toolkit was updated, but the docking control has a bug.
Resizing a docking panels is buggy, it doesn't always resize to where your last mouse location is when you release the mouse.
It is easier to just see the image as per the open issue https://github.com/xceedsoftware/wpftoolkit/issues/1498

**To Reproduce**
1. Try to resize a layout panel.

**Expected behavior**
Should resize to where the mouse is released.

**Additional context**
As per the open github issue, it is fixed in their paid version and supposedly not going to be released to the open source version until some undetermined time.
Maybe use https://github.com/Dirkster99/AvalonDock instead as per the thread?
Digging through the rabbit hole, perhaps also use https://github.com/dotnetprojects/WpfExtendedToolkit as a whole instead (if any other of the toolkit controls are used)?

### LINKED ISSUE #661: Xceed licensing issue
**Release Type**: Github

**Version**:  Commit 04bb068ca9eafe3b98dc412e86aba049c2d64cbd onwards

**Platform(s)**: Windows

**Describe the bug**
As per [Xceed's readme file](https://github.com/xceedsoftware/wpftoolkit), Extended.Wpf.Toolkit changed its license from v3.7.0 onwards, replacing the MS-PL license with the Xceed Community License agreement (for non-commercial use).
The commit changed the Xceed library to 3.8.1

**Additional context**
Possible solutions:
1. The third party license file needs to be updated to reflect this.
2. Revert to an older version prior to this license change.
3. Change to open source forks which are still licensed under MS-PL, eg. https://github.com/Dirkster99/AvalonDock and https://github.com/dotnetprojects/WpfExtendedToolkit

Looking through the code, it seems only AvalonDock is used from the Xceed library, and the alternative AvalonDock could be a simple transfer since it's based on the same control.
The datagrid used to be used from the older version of the Xceed library, however it has already been replaced with the standard WPF datagrid. I believe some minor functionalities were removed due to the the standard WPF datagrid not supporting whatever the xceed datagrid did.
The WpfExtendedToolkit actually contains a datagrid so we could also reinstate the "xceed" datagrid using this fork, as well?

Basewq:
## Description

Change to open source MS-PL AvalonDock fork.
Removed Xceed library which doesn't seem to be used anymore (DataGrid removed prior, and AvalonDock changed in this pull).
Note that the `THIRD PARTY.md` still references `WPFToolKit` which is the old (expired?) link to the Xceed, which is not removed because we have a few xaml pages that were copied over from them (this was done before their license change so remains under the MS-PL license). Not a lawyer so leaving it unless advised otherwise.

## Related Issue

Fixes #661
Fixes #659

## Motivation and Context

The forked version fixes the docking layout resize bug, which we don't know when Xceed will release for the open source public.
Xceed license was changed, so using proper open-source is probably more welcome (in my opinion...)

## Types of changes

## Checklist

### DIFF
--- a/sources/editor/Stride.Core.Assets.Editor/View/AssetViewUserControl.xaml.cs
+++ b/sources/editor/Stride.Core.Assets.Editor/View/AssetViewUserControl.xaml.cs
@@ -5,7 +5,6 @@ using System.Linq;
 using System.Windows;
 using System.Windows.Controls;
 using System.Windows.Input;
-using Xceed.Wpf.Toolkit.Core;
 using Stride.Core.Assets.Editor.View.Controls;
 using Stride.Core.Assets.Editor.ViewModel;
 using Stride.Core.Presentation.Collections;
@@ -308,12 +307,6 @@ namespace Stride.Core.Assets.Editor.View
         {
             var control = (AssetViewUserControl)sender;
             e.CanExecute = control.CanBeginEdit();
-        }
-
-        private static void CanBeginEditEvent(object sender, CancelRoutedEventArgs e)
-        {
-            var control = (AssetViewUserControl)sender;
-            e.Cancel = !control.CanBeginEdit();
         }
 
         private static void BeginEdit(object sender, ExecutedRoutedEventArgs e)
--- a/THIRD PARTY.md
+++ b/THIRD PARTY.md
@@ -45,6 +45,7 @@ Stride uses the following open-source products.
 
 * [Assimp](http://assimp.sourceforge.net/) (BSD 3-clause License)
 * [AsyncEx](https://github.com/StephenCleary/AsyncEx) (MIT License)
+* [AvalonDock](https://github.com/Dirkster99/AvalonDock) (Ms-PL)
 * [DocFX](https://dotnet.github.io/docfx/) (MIT License)
 * [ffmpeg](https://ffmpeg.org/) (GPL 3)
 * [FreeImage](http://freeimage.sourceforge.net/) (FreeImage Public License - Version 1.0)
--- a/sources/editor/Stride.Core.Assets.Editor/Stride.Core.Assets.Editor.csproj
+++ b/sources/editor/Stride.Core.Assets.Editor/Stride.Core.Assets.Editor.csproj
@@ -10,7 +10,7 @@
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="Microsoft.CSharp" />
-    <PackageReference Include="Extended.Wpf.Toolkit" Version="3.8.1" />
+    <PackageReference Include="Dirkster.AvalonDock" Version="4.10.0" />
     <PackageReference Include="System.Collections.Immutable" Version="1.6.0" />
     <Reference Include="System.ComponentModel.Composition" />
     <PackageReference Include="System.Diagnostics.Tracing" Version="4.3.0" />
@@ -82,8 +82,5 @@
     <TargetsForTfmSpecificBuildOutput>$(TargetsForTfmSpecificBuildOutput);IncludeExtraAssemblies</TargetsForTfmSpecificBuildOutput>
   </PropertyGroup>
   <Target Name="IncludeExtraAssemblies">
-    <ItemGroup>
-      <BuildOutputInPackage Include="$(OutputPath)Xceed.Wpf.Toolkit.dll" />
-    </ItemGroup>
   </Target>
 </Project>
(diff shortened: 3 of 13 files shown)


Kryptos-FR:
I know we did use AvalonDock in the past, but we had some issues. Maybe the situation is reversed now.

Does switching and saving the editor layout still work with AvalonDock? I seem to remember the XML format was different.

sources/editor/Stride.Core.Assets.Editor/View/AssetViewUserControl.xaml.cs
```diff
@@ -5,7 +5,7 @@
 using System.Windows;
 using System.Windows.Controls;
 using System.Windows.Input;
-using Xceed.Wpf.Toolkit.Core;
+//using Xceed.Wpf.Toolkit.Core;
```
Kryptos-FR, line 8: Better remove code instead of commenting out.

sources/editor/Stride.Core.Assets.Editor/View/AssetViewUserControl.xaml.cs
```diff
@@ -310,11 +310,11 @@ private static void CanBeginEditCommand(object sender, CanExecuteRoutedEventArgs
             e.CanExecute = control.CanBeginEdit();
         }
 
-        private static void CanBeginEditEvent(object sender, CancelRoutedEventArgs e)
-        {
-            var control = (AssetViewUserControl)sender;
-            e.Cancel = !control.CanBeginEdit();
-        }
+        //private static void CanBeginEditEvent(object sender, CancelRoutedEventArgs e)
```
Kryptos-FR, line 313: Better remove code instead of commenting out.

sources/editor/Stride.Core.Assets.Editor/View/AssetViewUserControl.xaml.cs
Basewq: Ok, will update this and the other one.

sources/editor/Stride.Core.Assets.Editor/View/AssetViewUserControl.xaml.cs
Basewq: > 
> 
> I know we did use AvalonDock in the past, but we had some issues. Maybe the situation is reversed now.
> 
> Does switching and saving the editor layout still work with AvalonDock? I seem to remember the XML format was different.

Not quite sure what you mean in the past? You mean this specific fork's?
As far as I can tell, we've been using Xceed's AvalonDock (even in the official v3.1), and this fork is still based off Xceed's AvalonDock, but prior to their license change (as far as I can tell).

I've tested it running the editor, docking a panel somewhere else then reloading, and it seems to remember it properly.
Nothing crashing as far as I can tell... Probably needs more eyes to confirm it's working, though.

Basewq:
> 
> 
> I know we did use AvalonDock in the past, but we had some issues. Maybe the situation is reversed now.
> 
> Does switching and saving the editor layout still work with AvalonDock? I seem to remember the XML format was different.

Not quite sure what you mean in the past? You mean this specific fork's?
As far as I can tell, we've been using Xceed's AvalonDock (even in the official v3.1), and this fork is still based off Xceed's AvalonDock, but prior to their license change (as far as I can tell).

I've tested it running the editor, docking a panel somewhere else then reloading, and it seems to remember it properly.
Nothing crashing as far as I can tell... Probably needs more eyes to confirm it's working, though.

Kryptos-FR:
> Not quite sure what you mean in the past? You mean this specific fork's?

It is not visible in this repo as it was before Xenko went public. But actually we were using the Docking from Telerik. My memory is playing me tricks. So if this fork's AvalonDock is using the same XML serializer, then we should be ok.

xwellingtonx:
@Basewq @Kryptos-FR, I've tested it and seems that doesn't have any issue with XML, even if is a custom layout saved in GameStudioInternal.conf file.

xen2:
Bonus points for having chosen a solution compatible with .NET Core 3 :)
I remember we had some issues with open-sourced AvalonDock several years ago, it looks like it has come a long way.

PR #664, opened 2020-04-23
[Editor] Fix closing then reopening asset when undocked

Basewq:
## Description

Remove visibility logic check when closing event is raised so the asset window can actually close, otherwise it's hidden in limbo and the asset can't be reopened.
I have also removed the check in the CurveEditor code, but didn't test that which I'm not sure if it exists anymore.

## Motivation and Context

This bug is occurring to due to adding the new forked AvalonDock.
Not sure if this logic check was originally meant as some workaround for some bug in the original AvalonDock, eg. it raised close event when it shouldn't have, or if there's some condition where this is raised and we don't want it to close?

## Types of changes

## Checklist

### DIFF
--- a/sources/editor/Stride.GameStudio/AssetEditorsManager.cs
+++ b/sources/editor/Stride.GameStudio/AssetEditorsManager.cs
@@ -165,10 +165,6 @@ namespace Stride.GameStudio
 
         private void CurveEditorClosed(object sender, EventArgs eventArgs)
         {
-            var editorPane = (LayoutAnchorable)sender;
-            if (editorPane.IsVisible)
-                return;
-
             RemoveCurveEditor(true);
         }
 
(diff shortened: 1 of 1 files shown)


Kryptos-FR:
Ok, so I wrote that code 4 years ago. There were indeed some issues with the `CurveEditor`, but of course I don't remember most of it (unfortunately I don't seem to have kept any of my notes).

I'll look at the commit messages on my old fork and try to figure out what was happening at that time. I do remember that at the time the ÌsActive`, `IsHidden` and other visibility properties on a `RadPane` had some strange behaviors.

Basewq:
Thanks for investigating. I think the code I deleted was specifically for when the original AvalonDock was first introduced, so not sure if you know what happened to that.
I think you can ignore the `CurveEditor`. Looking at the code, it's placed inside a `LayoutAnchorable` which is the problematic control, not the `CurveEditor` itself (though, as said before I think it's not used anymore, so probably not too much of a concern)

Kryptos-FR:
@Basewq the curve editor is used to animate particles.

PR #666, opened 2020-04-23
[Editor] Add ability to make material deduping from Assimp importer optional

Basewq:
# PR Details

Not sure if anyone else wants this feature, but I was caught out with Assimp's importer deduping my materials on my model when I didn't want it to.

## Description

I've now added a checkbox so you can choose not to dedupe materials. It's ticked by default so it behaves the same as currently. However, note that FBX and Assimp importers actually behave inconsistently in this regard, since FBX importer does _not_ dedupe materials at all.

Here's a screenshot of what the option looks like:
image

Note that I've only added it for the Assimp importer since it's a flag option for the importer.
I did not do it for FBX importer, I couldn't find any option for the FBX SDK (or I didn't look hard enough), and I don't want to troll though the code to figure out what the correct logic is to manually deduped it.

Note that the warning only appears if a user drags&drops a mix of fbx and non-fbx models (ie. some using assimp importer and some using fbx importer).
If only assimp importer is used, then the warning doesn't appear.
If only fbx importer is used, then no option appears for it.

The property must also exist on the ModelAsset since reloading the asset goes though the importer during loading, so we must persist telling the importer to dedupe or not dedupe.
It's not made visible because ticking/unticking doesn't update the material slots properly, so it's better to just delete the model and re-import it again with the correct options.

## Motivation and Context

I exported a model from Blender as a Collada file and had multiple meshes with separate materials, but the materials shared the same texture file, so Assimp (rightly) detected they could be the same material. I still wanted the model to have all the material slots (so they could be changed at run-time), but since Assimp deduped the materials, the model ends up with less material slots.

A workaround is to force the Blender file to make each material distinct, but it's a bit of a hacky workflow.

For testing purposes (sorry, no unit tests!), attached is a Collada file (zipped so Github doesn't complain) with a model with three meshes, each with its own material but the materials are the same values (ie. can be deduped).
[material_test.zip](https://github.com/stride3d/stride/files/4532116/material_test.zip)

This is what the actual model looks like in Blender. Materials are just a simple blue color.
image

On import with dedupe ticked, the model only has one material slot (as per default).
On import with dedupe unticked, the model has all three material slots.
After placing in the scene, changing the material's color correctly updates the right mesh (or just selecting the material will highlight the affected mesh).
The dedupe flag still persists when you select the 'update from source' options.

## Types of changes

## Checklist

### DIFF
--- a/sources/assets/Stride.Core.Assets/AssetImporterParameters.cs
+++ b/sources/assets/Stride.Core.Assets/AssetImporterParameters.cs
@@ -24,6 +24,7 @@ namespace Stride.Core.Assets
         /// </summary>
         public AssetImporterParameters()
         {
+            InputParameters = new PropertyCollection();
             SelectedOutputTypes = new Dictionary<Type, bool>();
         }
 
@@ -53,6 +54,12 @@ namespace Stride.Core.Assets
                 SelectedOutputTypes[type] = true;
             }
         }
+
+        /// <summary>
+        /// Gets the import input parameters.
+        /// </summary>
+        /// <value>The import input parameters.</value>
+        public PropertyCollection InputParameters { get; private set; }
 
         /// <summary>
         /// Gets the selected output types.
--- a/sources/editor/Stride.Assets.Presentation/Templates/ModelAssetTemplateWindow.xaml
+++ b/sources/editor/Stride.Assets.Presentation/Templates/ModelAssetTemplateWindow.xaml
@@ -24,6 +24,8 @@
     <Border Margin="20,10,20,20">
       <StackPanel>
         <CheckBox Content="{sd:Localize Import materials, Context=Button}" IsChecked="{Binding ImportMaterials}" Margin="5"/>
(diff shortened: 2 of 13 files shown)


Basewq:
This pull is now back for review after fixing a bug (the dedupe setting wasn't persisting properly).
Hopefully there's no more issues (especially since I'm actually trying to use this!).

sources/assets/Stride.Core.Assets/AssetImporterParameters.cs
```diff
@@ -54,6 +55,12 @@ public AssetImporterParameters(IEnumerable<Type> supportedTypes) : this()
             }
         }
 
+        /// <summary>
+        /// Gets the import input parameters.
+        /// </summary>
+        /// <value>The import input parameters.</value>
+        public PropertyCollection InputParameters { get; private set; }
```
xen2, line 62: Small remark: no need to add `private set` anymore in recent C# versions if set only in ctor
(not a big deal, OK to keep as is -- other properties had it too anyway)

xen2:
Tested, looks good to me!
Thanks for the new feature and detailed explanation + test model!

PR #674, opened 2020-04-28
[Editor] Don't append track name to Audio asset name if there's only one track

Basewq:
##  Description

Imported audio assets won't append ' track ' + index to the name if there's only track.

## Motivation and Context

Majority of audio that I import are wav, mp3, etc, that don't have multiple tracks, so this naming appending feature is not needed for those cases.

## Types of changes

## Checklist


xen2:
I would be fine to simplify thing by removing the option and keeping your new approach as default (no `track` appended if only 1)

Basewq:
Done, removed the option and just only append when multiple tracks are detected.

Issue #403, opened 2019-03-07
GameBase.MinimizedMinimumUpateRate & ThreadThrottler appare to have issues/ Can't enforce minimum update rate on minimized games.


devjarmo:
**Release Type**: Official Release

**Version**: 3.0

**Platform(s)**: Windows

**Describe the bug**
When setting MinimizedMinimumUpdateRate and/or changing the throttle type to any option Standard, PreciseAuto or PreciseManual the game appears to sleep indefinitely when minimized. The throttle does not appear to be effective in enforcing a minimum rate. This is an issue for locally hosted multiplayer games when the server is minimized, as well as headless implementations when the server was minimized.

**To Reproduce**
1. Create a game
2. Try to modify ((GameBase)Game).MinimizedMinimumUpdateRate
3. Optionally change the type using the set methods.
4. Minimize the window

Note: You will likely need to put something into the game so that you are aware of game rate when minimized.

**Expected behavior**
I would expect setting the MinimizedMinimumUpdateRate would throttle or rate limit the game to enforce a minimum rate, in this case I don't want any limit. I'd request an option be added to the Throttler for Disabled and a SetDisabled() method be added. Optionally this could be added to GameBase to prevent the throttle from ever being checked in the if statement for if Minimized check.

**Screenshots**
If applicable, add screenshots to help explain your problem (use GitHub drag & drop)

**Log and callstacks**
If any, please attach here any log or callstack (preferably in a `.txt` file using GitHub drag and drop)

**Additional context**
I looked into the source and I'm not sure I'm seeing what I'm actually running locally; one type appears to not call the method in the throttler (SetToPreciseAuto). Also, I don't have the time at the moment to debug the sleep logic.

tebjan:
should be a 5 minute action for @Eideren :)

Eideren:
I cannot reproduce this issue: spawned a new animation sample, changed UIScript to 
```
using System;
using System.Threading.Tasks;
using Xenko.Engine;
using Xenko.Games;
using Xenko.Graphics;
using Xenko.UI;
using Xenko.UI.Controls;

namespace AnimatedModel3
{
    public class UIScript : AsyncScript
    {
        public Entity Knight;

        public SpriteFont Font;

        public override async Task Execute()
        {
            // Bind the buttons
            var page = Entity.Get<UIComponent>().Page;

            var btnIdle = page.RootElement.FindVisualChildOfType<Button>("ButtonIdle");
[...]

PR #679, opened 2020-04-29
[Game] Extend and fix property documentation around ThreadThrottler

Eideren:
# PR Details
See title.

## Description
Very small doc changes, see files

## Related Issue
#403 

## Motivation and Context
Trying to make that API less obtuse.

## Types of changes

## Checklist

### DIFF
--- a/sources/core/Stride.Core/ThreadThrottler.cs
+++ b/sources/core/Stride.Core/ThreadThrottler.cs
@@ -10,7 +10,7 @@ namespace Stride.Core
         /// <summary>
         /// Will block until this amount of time passed. 
         /// Set this to zero to disable throttling.
-        /// To control this value like a maximum FPS limiter, set this value to 10.
+        /// To control this value like an FPS limiter, set this value to TimeSpan.FromSeconds(1d / yourFPS).
         /// </summary>
         public TimeSpan MinimumElapsedTime { get; set; }
 
--- a/sources/core/Stride.Core/ThreadThrottler.cs
+++ b/sources/core/Stride.Core/ThreadThrottler.cs
@@ -8,12 +8,14 @@ namespace Stride.Core
     public class ThreadThrottler
     {
         /// <summary>
-        /// Minimum time allowed between each call
+        /// Will block until this amount of time passed. 
+        /// Set this to zero to disable throttling.
+        /// To control this value like a maximum FPS limiter, set this value to 10.
         /// </summary>
         public TimeSpan MinimumElapsedTime { get; set; }
 
         /// <summary>
-        /// The type of throttler used, call 
+        /// The type of throttler used
         /// </summary>
         public ThrottlerType Type { get; private set; }
 
--- a/sources/engine/Stride.Games/GameBase.cs
+++ b/sources/engine/Stride.Games/GameBase.cs
@@ -270,9 +270,7 @@ namespace Stride.Games
         public TimeSpan TargetElapsedTime { get; set; }
 
         /// <summary>
-        /// Access to the throttler used to set the minimum time allowed between each updates,
-        /// set it's <see cref="ThreadThrottler.MinimumElapsedTime"/> to TimeSpan.FromSeconds(1d / yourFramePerSeconds)
-        /// to control the maximum frames per second.
+        /// Access to the throttler used to set the minimum time allowed between each updates.
         /// </summary>
         public ThreadThrottler WindowMinimumUpdateRate { get; }
 


sources/core/Stride.Core/ThreadThrottler.cs
```diff
@@ -8,12 +8,14 @@ namespace Stride.Core
     public class ThreadThrottler
     {
         /// <summary>
-        /// Minimum time allowed between each call
+        /// Will block until this amount of time passed.
```
Kryptos-FR, line 11: As a personal preference, I would have kept the original description in the summary, and put all those additional lines under a `<remarks>` section.

The summary is what Visual Studio shows by default in a tooltip when hovering the mouse over it. So I like to keep it as short as possible. The remarks and other optional elements are usually shown only when displaying the full information.

I might be biased against the behavior that resharper shows since it can display that full information when pressing CTRL+SHIFT+F1. Visual Studio currently doesn't have the capacility but it is discussed in https://github.com/dotnet/roslyn/issues/31618.

Eideren: Issue is that people using this API more often than not want to throttle based on the amount of frames per second not the amount of seconds per frame, it most likely is hard for them to wrap their head around this so that comment is kind of essential.
We could also not require user to do that extra step in the first place and there's two way to go about this:
* Present users with a property that takes and shows that data in a 'fps' format.
This will lead to precision issues as converting between the two will be float-based and won't be obvious to the user which might in term lead to other user issues like running logic doing equality comparison with this property.
* Changing the entire system to use an fps format, doing so will make the system less precise than it is right now when users expect this to be quite precise in the first place.
Obviously those are fairly nitpick-y but I don't feel right loosing any kind of precision to the system to work around this specific issue.

Given what I said, do you still feel that this should go within remarks ?

Kryptos-FR: > set this value to 10

To 10 what? milliseconds, seconds, minutes, ticks?

I think, according to the previous comment on `GameBase.WindowMinimumUpdateRate` (now removed) that it is supposed to be `TimeSpan.FromSeconds(1.0/10)`. We should make that clear in the comment.

OK to keep it in the summary for now.

Kryptos-FR: Ok for me after commit 25b1b56.

Eideren: No reason for 10, probably mistyped that one while pushing to this pr.

tebjan:
@Eideren `TimeSpan.FromSeconds()` is not precise enough, for whether reason, the developer of this decided to round this to the closest millisecond. Here are some extensions that we use in our repo:

```csharp
    /// <summary>
    /// Creates TimeSpans in a precise way, normal "From" methods do weird rounding.
    /// </summary>
    public static class TimeSpanUtils
    {
        /// <summary>
        /// Creates a TimeSpan with precise tick count, normal "From" methods do weird rounding.
        /// </summary>
        public static TimeSpan FromMillisecondsPrecise(double milliSeconds)
            => TimeSpan.FromTicks((long)(milliSeconds * TimeSpan.TicksPerMillisecond));

        /// <summary>
        /// Creates a TimeSpan with precise tick count, normal "From" methods do weird rounding.
        /// </summary>
        public static TimeSpan FromSecondsPrecise(double seconds)
            => TimeSpan.FromTicks((long)(seconds * TimeSpan.TicksPerSecond));

        /// <summary>
        /// Creates a TimeSpan with precise tick count, normal "From" methods do weird rounding.
        /// </summary>
        public static TimeSpan FromMinutesPrecise(double minutes)
            => TimeSpan.FromTicks((long)(minutes * TimeSpan.TicksPerMinute));
    }
```

Kryptos-FR:
@tebjan I don't know why either. But does it matter here, a resolution of 1 ms still means an FPS of 1000 which a game is unlikely to get to.

And it looks like a lot of the internals in Windows don't even have such a high resolution. See [this article](https://aakinshin.net/posts/datetime/) for instance. `Thread.Sleep(1)` might actually sleep for 15 ms.

Kryptos-FR:
@tebjan I don't think you got my point.

So let's say you want 60 FPS. You certainly don't want to wait 16.66 ms, as otherwise you will have 0 ms remaining to execute one iteration of the game loop in order to keep that same 60 FPS rate. As one iteration of the game loop is very likely to take more than 1 ms to compute, there is no need to have that level of precision.

tebjan:
@Kryptos-FR if you read the code I linked above, you will see that the wait happens after the work is done, if course. However, since you know the time of the frame start before, you know how much time is left in the current frame to meet the throttle time.

You can never know who needs this kind of precision, if you don't need it in your project, the thread throttler has settings to turn high precision off.

Eideren:
@Kryptos-FR Just to extend on what @tebjan said:
The game runs for ``x`` ms, pushes the frame to screen then calls ``Throttle`` which forces the thread to sleep and spin for ``MinimumElapsedTime - x``. So if the game thread took 2ms to run, this will sleep and spin for 14.66...ms, sleeping for a certain amount depending on the type and OS response, and if the type allows, spinwait for the rest.

17ms is 58.8 fps, 16 is 62.5 fps, this will lead to uneven pacing when viewed on a 60hz monitors (we're not talking about vsync here as it would replace this system).
Some users might implement their logic relying on this system to do its job correctly, using frame counting instead of delta time like fighting games and games of the snes era, fixed delta time like physics engine. It should be precise enough for those kinds of usage so having an error as large as ~0.5 millisecond is too large compared to the real 16.6 for us to just ignore.

I'll make a PR to address those issues.

Kryptos-FR:
@Eideren If that's true, the name of the property is misleading and should be `MaximumElapsedTime`.

When I see `MinimumElapsedTime`, I expect the wait to be **at least** that amount of time, not the amount minus some variable.

Eideren:
@Kryptos-FR My point of view was minimum allowed time to be elapsed since that function for another call to the function not to block, perhaps I could name it periodicity or something. 
I don't like that name either, if you can think of another name for it be my guest !

Kryptos-FR:
@Eideren Naming is hard. Anyway, you both convinced that I was wrong in my assumptions. We can let the discussion here and continue on the next pull-request (#690).

PR #687, opened 2020-04-30
[Logger] Enhance logger and ExceptionInfo formatting

Eideren:
# PR Details
Hopefully this should make the logger less obtuse for users.

## Description
Adds a ``(...)`` at the end of exceptions to hint to users that they can use a similar button to show the rest. 
Changes ``ExceptionInfo``'s formatting to reflect how dot net does it by default, i.e.: 
``Type: Message\n{innerExceptions.ToString}\n{Stacktrace}``
instead of ours which was
``Message\n{Stacktrace}\n{innerException.ToString}``
this leads to the first stacktrace line being the line that actually created the exception instead of one of the many catch the exception climbed back through.

## Motivation and Context
See PR Details.

## Types of changes

## Checklist


xen2:
Good idea, thanks!

Issue #220, opened 2018-09-26
[Editor] Async method within ScriptComponent constructor crashes editor


Eideren:
**Release Type**: Official Release/GitHub (please choose appropriate option)

**Version**: 3.0.0.6 commit [ac4e7e0](https://github.com/xenko3d/xenko/commit/ac4e7e03b8e2e0e68005589bd52f7f2a2aba7b1a)

**Platform(s)**: Windows Editor

**Describe the bug**
Occurs when adding it to an entity, tested with an async method running a Task.Delay of one second,the crash handler then pops up.

**To Reproduce**
See description.

**Expected behavior**
Fix the issue or detect and prohibit those kinds of implementation.

**Screenshots**

**Log and callstacks**
[editor.txt](https://github.com/xenko3d/xenko/files/2419545/editor.txt)

**Additional context**

Kryptos-FR:
Unfortunately there is no way to "hide" the constructor in a script.
So the rule is: don't do anything in a constructor. Just don't.
And it's pretty hard to recover from an exception in a constructor. I'm not sure what we can do.

jonstelly:
This is somewhat similar to https://github.com/xenko3d/xenko/issues/209

While testing out my issue I noticed it happens also if you create any field/property initializers that fail when run inside the editor.

Guidance/rules to not put anything in the constructor are completely reasonable, but even documentation of those rules with big flashing warning signs will get missed by some people so are there ways we could catch these errors to give a meaningful message/feedback?

try/catch around the Activator.CreateInstance() call in ObjectFactoryRegistry.NewInstance() to throw a friendly/helpful message, whatever behavior you guys feel is reasonable for the Editor (message box or something else?).  Maybe even changing synchronization contexts during the Activator.CreateInstance() call to ensure that async errors in the constructor would get handled the same way?

I'm happy to work on this if you guys have a certain behavior you think makes sense.

Eideren:
@jonstelly Yep `CreateInstance()`, also initialize fields and auto properties, like a new() would.

I say let's not force users to adopt engine-specific behavior when programming, let them use their C# habits and notify them of issues instead of forcing them out of it. If we're going that way we would have to force them to avoid using field and auto prop initialization, 'doesn't seem like a good idea at all.

[...]

PR #692, opened 2020-05-02
[Microthread] ctor async throwing

Eideren:
# PR Details
A component whose ctor calls an async function crashes the editor, see #220.
While the changes within this PR works, those changes might not be safe, **I would like to hear from someone more familiar with the microthreading** implementation if this is alright **before setting this PR to ready**, is that exception essential, does it occur outside of case #220 ? If it's not safe at all investigate what other methods I/we could investigate.

## Related Issue
#220 

## Motivation and Context
Fix a bug, even if the context around this bug is questionable.

## Types of changes

## Checklist


xen2:

PR #694, opened 2020-05-03
[Compiler] Register the project's assemblies before loading the asset libraries

### LINKED ISSUE #683: Projects with Stride.Voxels as a dependency cannot be built
GitHub master(4.0.0.1-beta02), Windows

Firstly, sorry for never getting back after the merge, thanks so much for helping it through!

There's a problem though. Adding Stride.Voxels as a dependency to a project causes it to fail when building; everything works fine in the editor however. It seems that the asset compiler can't deserialize any of the classes in that module.

**To Reproduce**
1. Create a project (I usually use the FirstPersonShooter template)
2. Confirm it builds and runs
3. Add Stride.Voxels as a dependency
4. It no longer builds

**Expected behavior**
Probably should still build.

**Log and callstacks**
[VoxelsBuildFail.txt](https://github.com/stride3d/stride/files/4555937/VoxelsBuildFail.txt)

Thanks for any help!

Basewq:
## Description

**Warning:** This is not heavily tested, and I can't say I know what the full implications are of doing this, so probably needs someone more well versed into CompilerApp and probably make a better pull if necessary.
Basically the source of the issue for the Stride.Voxel library is that the asset's "definitions" are defined the optional library (ie. Stride.Voxels). The CompilerApp only registers assemblies that it knows about, ie. only mandatory Stride libraries.
What this does is load the assemblies of the project itself in order to register any missing definitions (eg. the Voxel classes, which are needed for the YAML deserializer).
The part I'm not understanding is the full implications of loading these assemblies. eg. will it conflict with other assemblies, especially if it's same assembly but different versions.

I think at the very least, this should provide pointers in which areas to look at.

## Related Issue

Fixes #683 

## Types of changes

## Checklist

### DIFF
--- a/sources/assets/Stride.Core.Assets.CompilerApp/PackageBuilder.cs
+++ b/sources/assets/Stride.Core.Assets.CompilerApp/PackageBuilder.cs
@@ -80,6 +80,7 @@ namespace Stride.Core.Assets.CompilerApp
                     AutoCompileProjects = !builderOptions.DisableAutoCompileProjects,
                     ExtraCompileProperties = builderOptions.ExtraCompileProperties,
                     RemoveUnloadableObjects = true,
+                    RegisterPackageAssemblies = true,
                     BuildConfiguration = builderOptions.ProjectConfiguration,
                 };
 
--- a/sources/assets/Stride.Core.Assets/PackageLoadParameters.cs
+++ b/sources/assets/Stride.Core.Assets/PackageLoadParameters.cs
@@ -42,6 +42,7 @@ namespace Stride.Core.Assets
         public PackageLoadParameters()
         {
             LoadMissingDependencies = true;
+            RegisterPackageAssemblies = false;
             LoadAssemblyReferences = true;
             AutoCompileProjects = true;
             AutoLoadTemporaryAssets = true;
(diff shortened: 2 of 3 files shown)


Kryptos-FR:
Unfortunately, this currently has no effect since it will return on one of the first two tests in `Package.LoadAssemblyReferencesForPackage`.

https://github.com/stride3d/stride/blob/273dfddd462fd3746569f833e1493700c070b14d/sources/assets/Stride.Core.Assets/Package.cs#L1020-L1035

Basewq:
Technically `LoadMissingDependencies` goes through `PreLoadPackageDependencies` first and that will throw an exception if the logger is null, however I will make the explicit check for that since it may potentially not call `PreLoadPackageDependencies` on some edge cases.
Though it should be pointed out that `LoadMissingAssets -> TryLoadAssets` assumes log is not null without checking it, so would crash the program too...

Kryptos-FR:
@Basewq I was not taking about the log check. Sorry should have made it more clear that I'm talking about the two "if" checks:
* `Container` is either a `StandalonePackage` so the first cast fails
* `LoadedAssemblies` already has the current assembly loaded, but maybe not its reference. Not sure why in my debugging session it also returned here, as I was tweaking some parameters so it could be my mistake.

Basewq:
Fixed by a8962a769ca0dc7d4ac17e1fe000b0a1d526c62a instead.

xen2:
Thanks for looking into it!

PR #705, opened 2020-05-06
[Rendering] Added NumOfTexCoords property to Procedural Models

tebjan:
# PR Details

Made the property to adjust the number of generated texture coordinates visible to the user.

## Motivation and Context

The vertex buffers generated by the procedural models are quite huge since they contain 10 channels of texture coordinates. This property allows the user to reduce them if needed.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Rendering/Rendering/ProceduralModels/PrimitiveProceduralModelBase.cs
+++ b/sources/engine/Stride.Rendering/Rendering/ProceduralModels/PrimitiveProceduralModelBase.cs
@@ -52,6 +52,16 @@ namespace Stride.Rendering.ProceduralModels
         public Vector3 LocalOffset { get; set; }
 
         /// <summary>
+        /// Gets or sets the number of texture coordinates to generate. A value between 1 and 10, inclusive.
+        /// </summary>
+        /// <value>
+        /// The number of texure coordinates.
+        /// </value>
+        [DataMember(540)]
+        [DataMemberRange(1, 10)]
+        public int NumOfTexCoords { get; set; } = 10;
+
+        /// <summary>
         /// Gets the material instance.
         /// </summary>
         /// <value>The material instance.</value>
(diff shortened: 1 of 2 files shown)


tebjan:
@xen2 does this need a version increment for the serializer because there is one more property to save?

xen2:
@tebjan in this case, it's not necessary:
- `Asset` itself is fine to upgrade (as long as it's a property addition, no need for upgrader)
- What matters is the build system `Command` it generates. In this specific case, the command hashes the full `Asset` (it inherits from `AssetCommand<ProceduralModelAsset>`).
As a result, any change to the `Asset` will make sure it doesn't reuse any of the previous build cache.
Of course, it would have been different is the command was more specific/optimized and not hashing the full `Asset`

sources/engine/Stride.Rendering/Rendering/ProceduralModels/PrimitiveProceduralModelBase.cs
```diff
@@ -51,6 +51,16 @@ public void SetMaterial(string name, Material material)
...
 
+        /// <summary>
+        /// Gets or sets the number of texture coordinates to generate. A value between 1 and 10, inclusive.
+        /// </summary>
+        /// <value>
+        /// The number of texure coordinates.
+        /// </value>
+        [DataMember(540)]
+        [DataMemberRange(1, 10)]
+        public int NumOfTexCoords { get; set; } = 10;
```
xen2, line 62: For API consistency, please use `NumberOfTexCoords`. Or maybe even `NumberOfTextureCoordinates` (it would matches `VertexElement.TextureCoordinate` and display better in property grid).

Note: if you want a more verbose version in property grid, feel free to add something like `[Display("Number of texture coordinate channels")]` or any description that you think is better if default one is not good enough.

Basewq:
Since you're fixing minor messages, you should also fix the one in the `<exception>` tag:
`/// maxTexcoord;maxTexcoord must be > 0`

sources/engine/Stride.Rendering/Rendering/ProceduralModels/PrimitiveProceduralModelBase.cs
tebjan: Sounds good, will improve on that. Which help text will be displayed in the help box below the property grid?

xen2: @tebjan for the detail in the property grid lower help area, the proper tag is `<userdoc>Some description</userdoc>` (you can put it after `</value>`)

PR #714, opened 2020-05-10
Removing entities improvements

Aggror:
# PR Details
Updated removing entities scene. renamed scene folder.

## Types of changes

## Checklist


_config.yml
```diff
@@ -0,0 +1 @@
+theme: jekyll-theme-slate
```
xen2, line 1: File added by mistake?

docs/_config.yml
```diff
@@ -0,0 +1 @@
+theme: jekyll-theme-slate
```
xen2, line 1: File added by mistake?

_config.yml
Aggror: Yeah I was toying around with Github pages, to see if we could generate the Documentation as github pages. That way we could include the docs in main repo. But it was a failed attempt. In the mean time my config file clutted the branch it seems.

Aggror:
I will have to check out why there are conflicts. Will do later this week.

Basewq:
> 
> 
> I will have to check out why there are conflicts. Will do later this week.

My bad, it's probably because of my pull where I fixed some typos in files you moving.
https://github.com/stride3d/stride/pull/716/files

PR #729, opened 2020-05-17
Fix FPS dependant scroll and cleanup JumpyJet sample

manio143:
# PR Details

JumpyJet used an `int` for the position offset of the background parallax. Every frame it was updated. But if the FPS count was big enough then position change was rounded down to 0 every frame, making the background seem static.

Chaning that offset to a float solved the issue.

At the same time I spotted two typos `GameStartedventKey` and `GetPipeRandoYPosition`.
Also the `PipeSet` class was no longer used in favor of prefabs.
And there was some lock file that I don't think was doing anything (correct me if I'm wrong).

## Motivation and Context

Those samples are visited by every newcomer, so they should be as clear to understand as possible.
I had come across the issue with parallax background not moving after adding GameProfiler script which disables VSync causing an increase in frames per second.

## Types of changes

## Checklist

I did a manual test - compile & run - to see if everything worked and it did.


manio143:
I have also stubbled upon another bug, but was unable to come up with a solution.
When JumpyJet dies after entering a collision with the PassedTrigger, then the collision is exited after restart which gives you a point at the start of the next round.

xen2:
Merged, thanks!

PR #740, opened 2020-05-28
[Threading] ThreadPool and Dispatcher efficiency

Eideren:
# PR Details
In general this PR reduces the amount of time wasted by threads calling and working with the dispatcher, also reduces spikes caused by lock contention and uneven work distribution.

"Changing the ThreadPool, again !" Yeah, I know, I just can't help myself ...

## Description
I replaced a dictionary in ``StreamingManager`` by a concurrent one, not the cleanest change but this was necessary as ``Get()`` has far too much lock contention, I can remove this if you feel like it should be its own PR / it requires more changes.

Swapped our pool implementation from **internal to public**, included a couple of properties that provide info on the pool's state and added documentation to public members. Our implementation should be mature enough by now to let users have access to it if they wish to profile it or use it directly.

Recursive dispatching should work better as calling threads will cooperate with the pool instead of just blocking until finished.

Sorting now schedules the right amount of threads and will prevent threads from exiting before the whole collection is sorted.

## Related Issue
N/A

## Motivation and Context
Performance

## Types of changes

## Checklist


xen2:
Thanks for improving it again!
I just took a quick glance and LGTM (I didn't check the finer details, let's see if any issues pop up)

Issue #340, opened 2019-01-23
Geometry Instancing


GutterLab:
**Is your feature request related to a problem? Please describe.**
Being able to draw a lot of the same thing while staying performant.

**Describe the solution you'd like**
Support for geometry instancing.

**Additional context**
https://en.wikipedia.org/wiki/Geometry_instancing

xen2:
@tebjan Fine with doing that at first.
I think the main other concern was how to gather instances (should we find duplicate draw call automatically or have an explicit way to describe those?)

yangguosdxl:
Maybe automatically gather instances not very emergency。this can do by game logic. Generally, may be we just want a method to commit geometry instancing :-)

Eideren:
Should we close this issue ? @tebjan

PR #741, opened 2020-05-28
[Rendering] Transformation instancing 

tebjan:
# PR Details

Implement transformation instancing per model.

## Description

Here are my thoughts on it so far, comments are welcome:
https://gitmind.com/app/doc/dc5bde667771d3fffea00a009eabd363

A working proof of concept can now be found here:
https://github.com/tebjan/StrideTransformationInstancing

The current issues to solve:
- [x] Where to upload the matrices to the dynamic buffer
- [x] Manage bounding box
- [x] Scan material shaders for World matrix usage and update accordingly

## Related Issue

#340 Geometry Instancing

## Motivation and Context

Efficiently batch draw the same model multiple times with different transformations.

## Types of changes

## Checklist


sources/engine/Stride.Rendering/Rendering/Transformation/TransformationWAndVPInstanced.sdsl
```diff
@@ -0,0 +1,41 @@
...
+/// </summary>
+shader TransformationWAndVPInstanced : TransformationBase, PositionStream4, PositionHStream4
+{
+    // Just for test, move to general transformation shader
+    struct InstanceTransform
+    {
+        float4x4 Matrix;
+    };
+
+    rgroup PerMaterial
```
xen2, line 14: Probably not in `PerMaterial`
This probably should go to `PerDraw`

Basewq:
Can I just say this is awesome?
Not really sure if I can provide any decent feedback, just some minor nitpicks.

Just wondering if the name needs to be clarified something like ModelInstancing, MeshInstancing or GpuInstancing, etc? I'm not a game dev so don't really know if this is the standard name or not.

sources/engine/Stride.Engine/Engine/Processors/InstanceProcessor.cs
```diff
@@ -0,0 +1,38 @@
...
+                component.Master = FindMasterInParents(component.Entity);
+            }
+        }
+
+        protected override void OnEntityComponentRemoved(Entity entity, [NotNull] InstanceComponent component, [NotNull] InstanceComponent data)
+        {
+            component.DisconnectInstancing();
+        }
+
+        private InstancingComponent FindMasterInParents(Entity entity)
```
Basewq, line 27: Unlikely to blow the stack, but could probably change this to non-recursive call for extra safety.

sources/engine/Stride.Engine/Engine/InstancingUserArray.cs
```diff
@@ -0,0 +1,84 @@
...
+        /// <summary>
+        /// A flag indicating whether the inverse matrices and bounding box should be calculated this frame.
+        /// </summary>
+        bool matricesUpdated;
+
+        /// <summary>
+        /// Updates the world matrices.
+        /// </summary>
+        /// <param name="matrices">The matrices.</param>
+        /// <param name="instanceCount">The instance count. When set to -1 the lenght if the matrices array is used</param>
```
Basewq, line 33: Typo in 'length'

sources/engine/Stride.Engine/Engine/InstancingUserArray.cs
```diff
@@ -0,0 +1,84 @@
...
+                {
+                    // Make sure inverse matrices array is big enough
+                    if (WorldInverseMatrices.Length < InstanceCount)
+                    {
+                        WorldInverseMatrices = new Matrix[InstanceCount];
+                    }
+
+                    // Invert matrices and update bounding box
+                    var bb = BoundingBox.Empty;
+                    for (int i = 0; i < InstanceCount; i++)
```
Basewq, line 67: I could be wrong, but my gut tells me you should hold a local copy of `InstanceCount` in this method because it's a virtual property otherwise it'll take a minor performance hit.

sources/engine/Stride.Engine/Engine/Processors/InstancingProcessor.cs
```diff
@@ -0,0 +1,206 @@
...
+                    renderInstancing.BuffersManagedByUser = false;
+                    renderInstancing.WorldMatrices = instancingUserArray.WorldMatrices;
+                    renderInstancing.WorldInverseMatrices = instancingUserArray.WorldInverseMatrices;
+
+                    if ((renderInstancing.InstanceWorldBuffer == null || renderInstancing.InstanceWorldBuffer.ElementCount < instancing.InstanceCount))
+                    {
+                        renderInstancing.InstanceWorldBuffer?.Dispose();
+                        renderInstancing.InstanceWorldInverseBuffer?.Dispose();
+
+                        renderInstancing.InstanceWorldBuffer = CreateMatrixBuffer(context.GraphicsDevice, instancing.InstanceCount);
```
Basewq, line 62: Could probably gain an extremely minor perf gain using `renderInstancing.InstanceCount` instead of `instancing.InstanceCount` since they should be the same value at this point.

sources/engine/Stride.Engine/Engine/InstancingUserArray.cs
```diff
@@ -0,0 +1,84 @@
...
+{
+    [DataContract("InstancingUserArray")]
+    [Display("UserArray")]
+    public class InstancingUserArray : InstancingBase
+    {
+        /// <summary>
+        /// The instance transformation matrices.
+        /// </summary>
+        [DataMemberIgnore]
+        public Matrix[] WorldMatrices = new Matrix[0];
```
Basewq, line 16: `Array.Empty<Matrix>()` to avoid needless additional allocation here and in `WorldInverseMatrices`.

sources/engine/Stride.Engine/Engine/InstancingUserArray.cs
tebjan: Good point!

sources/engine/Stride.Engine/Engine/InstancingUserArray.cs
tebjan: Good one!

sources/engine/Stride.Engine/Engine/Processors/InstancingProcessor.cs
tebjan: Thx!

tebjan:
@Basewq Thanks for taking the time to look through it! 
I also want to add that @xen2 helped me a lot with finding the correct implementation. He answered many questions I had over the course of implementing this.

> Just wondering if the name needs to be clarified something like ModelInstancing, MeshInstancing or GpuInstancing, etc? I'm not a game dev so don't really know if this is the standard name or not.

The official term is: [Geometry Instancing](https://en.wikipedia.org/wiki/Geometry_instancing)

Not 100% sure about the naming as well, it might be a bit simple, but every graphics developer should know what it is. However, the difference between `InstancingComponent` and `InstanceComponent` might be a bit slim... If someone knows something more distinguishable, I'd be happy to hear it.

Basewq:
Looking at the other engines,
Unity calls it [GPUInstancing](https://docs.unity3d.com/Manual/GPUInstancing.html) but they seem to do it on a material & shader level.
Unreal calls it [UInstancedStaticMeshComponent](https://docs.unrealengine.com/en-US/API/Runtime/Engine/Components/UInstancedStaticMeshComponent/index.html) or [Instanced Static Mesh for Blueprints](https://docs.unrealengine.com/en-US/BlueprintAPI/Components/InstancedStaticMesh/index.html)

I find it interesting Unreal calls it instance**d** but every method is 'instance' without the 'd'.

I'm a fan of prefixing or suffixing 'Model' or 'Mesh' to the names because it makes the name easier to understand in a list of entity components (since not all components will be graphics related), however this is just a personal preference (I'm no good at naming things anyway, lol).

tebjan:
@Basewq 
> I'm a fan of prefixing or suffixing 'Model' or 'Mesh' to the names because it makes the name easier to understand in a list of entity components (since not all components will be graphics related), however this is just a personal preference (I'm no good at naming things anyway, lol).

thanks for the thoughts, the reason I wasn't adding the "Model" prefix to the component name is that the components are already in the "Model" category of the component browser:

image

but I agree, in code, it could be more clear to add the prefix... maybe GPUInstancing isn't so bad. unity developers will feel at home and Godot users are not confused with the term.

xen2:
I just gave it a quick try (I didn't look at the code in detail yet).
Some remarks/improvements:
- Let's add `InstancingRenderFeature` to `DefaultGraphicsCompositorLevel10` so that it works out of the box. You can do it on a new game .sdgfxcomp and apply back the change manually to `DefaultGraphicsCompositorLevel10.sdgfxcomp`.
- Picking work, but not the selection wireframe gizmo is still tied to underlying model rather than the instance.
- Might be good to add an `Entity` template with an `InstanceComponent`:
image

Just so you know, we have the `Duplicate` command which is accessible by Ctrl+D (selected entity) or by moving an entity while pressing the Control key. This allows to easily create many entities with `InstanceComponent`.

xen2:
And nice job BTW, I like the final design you ended up choosing! (just need to figure out proper naming)

tebjan:
Thanks for having a look, good point with the graphics compositor. I'll also look into the other things.

tebjan:
> * Let's add `InstancingRenderFeature` to `DefaultGraphicsCompositorLevel10` so that it works out of the box. You can do it on a new game .sdgfxcomp and apply back the change manually to `DefaultGraphicsCompositorLevel10.sdgfxcomp`.

Ok, added the render feature to the compositor templates

> * Might be good to add an `Entity` template with an `InstanceComponent`:

I also added two entries there. Unfortunately, hierarchies don't work yet, so an entity with a child instance doesn't work yet. but it works well otherwise.

> * Picking work, but not the selection wireframe gizmo is still tied to the underlying model rather than the instance.

Yes, I thought it might be a bit too much if all instances are wireframed/highlighted if you pick only one. But if you select the model entity it makes sense to show all existing instances of it.

Drawing a single instance in wireframe would probably the best solution, but I doubt that I will have time to look into that.

sources/engine/Stride.Engine/Engine/Processors/InstancingProcessor.cs
```diff
@@ -0,0 +1,208 @@
...
+                    renderInstancing.WorldMatrices = instancingUserArray.WorldMatrices;
+                    renderInstancing.WorldInverseMatrices = instancingUserArray.WorldInverseMatrices;
+
+                    if (renderInstancing.InstanceWorldBuffer == null || renderInstancing.InstanceWorldBuffer.ElementCount < instanceCount)
+                    {
+                        renderInstancing.InstanceWorldBuffer?.Dispose();
+                        renderInstancing.InstanceWorldInverseBuffer?.Dispose();
+
+                        renderInstancing.InstanceWorldBuffer = CreateMatrixBuffer(context.GraphicsDevice, instanceCount);
+                        renderInstancing.InstanceWorldInverseBuffer = CreateMatrixBuffer(context.GraphicsDevice, instanceCount);
```
xen2, line 65: For the specific `InstancingUserArray` case, since we create them, shouldn't we destroy them on component remove too?

sources/engine/Stride.Engine/Engine/InstancingBase.cs
```diff
@@ -0,0 +1,31 @@
+using Stride.Core;
+using Stride.Core.Mathematics;
+using Stride.Engine.Processors;
+
+namespace Stride.Engine
+{
+    [DataContract]
+    public abstract class InstancingBase : IInstancing
```
xen2, line 8: I wonder if we want this class to implement `IInstancing`: it forces all `IInstancing` properties to have a setter (because we can't add one in inheriting class). However, those properties might not be something we want visible on user class.
For example, `InstancingUserArray.InstanceWorldBuffer` could be getter only (with internal set) or even totally hidden since it's not supposed to be visible to user.
Same goes for `InstanceCount` which might be `get` only or `get/set` depending on which implementation if `IInstancing` we are talking about.

sources/engine/Stride.Engine/Engine/Processors/InstancingProcessor.cs
tebjan: good point, yes!

sources/engine/Stride.Engine/Engine/InstancingBase.cs
tebjan: indeed, this can be modeled a bit clearer, I'll look into it.

xen2: OK, removed `InstancingBase` altogether with latest commit.
Let me know if everything still work as intended, thanks!

phr00t:
Good work guys.

Has any of this been tested or implemented on Vulkan, is this DirectX only, or should be API neutral?

sources/engine/Stride.Engine/Engine/Processors/InstancingProcessor.cs
tebjan: Disposing the buffers now on component remove and added file headers.

tebjan:
Ok, I think this is ready now (at least for beta builds).

There are only two things left to do:
The inverse matrix buffer is only required if the material uses normal map textures, so the calculation and upload of that data can be made dependent on whether the material uses normal map textures.
And an example of using `InstanceID` in a custom shader to, for example, change the color (roughness, metalness, etc) of individual instances.

najak3d:
@tebjan  - WOW, amazing.   I'm so glad you did this.   I was just googling to see if Stride supported instancing, and now, thanks to you, it does!

**One question -- does it support Animated Model instancing?** 

It's OK if all models share the same Animation State (i.e. move in perfect Unison), because variation can always be created by simply having a half-dozen or so InstanceGroups, each with a different Animation Speed, or at different intervals...  this eliminates the creepy "everyone is in unison" artifact. 

This is useful for creating simulations showing crowds of people walking through the town, and other things.

tebjan:
yes, it works with one animation for the whole batch. see this project for an example: https://github.com/tebjan/StrideTransformationInstancing

image

najak3d:
Amazing!  Thank you.   Urho.NET currently only supports instancing for non-Animated objects.

+1 (more) for Stride.

Issue #743, opened 2020-05-30
[Asset] Slightly complex prefabs and scenes are effectively unusable


Eideren:
**Release Type**: Github
**Version**: Latest
**Platform(s)**: Windows editor
**Describe the bug**
TL;DR: slightly complex scenes and prefabs are pretty much unusable because of the time the editor takes to save changes and add them to scenes.

- ***Details***
Modifying scenes and prefabs, drag&drop and other operations with those features takes an absurdly large amount of time and often ends in crashes in more complex scenarios, just dropping a prefab containing a couple of children in a scene can take more than a minute and will crash the game studio once the amount of children increases further.

- ***Profiling***
When profiling a drag and drop we can see that a lot of time is taken by reflection and fighting the GC with allocations but both of those doesn't take most of the time anyway (around 20%), the rest is spent all over the place in quantum so this is most likely a design issue.
**some of the functions where called more than 4.000.000 times** and it was barely a fourth of the way to completion.

- ***Question***
All of that work is to setup binding instance to the source prefab's for changes to cascade over, right ? I don't really understand why this needs to happen for a drag and drop operation, shouldn't this happen once a modification to the source has been made ?

- ***Potential Solution***
Loading from disk works fairly well even though that would require accessing files and parsing them which seems like it should take longer but clearly doesn't here, perhaps we should only write the changes to disk on save and propose to users to reload their opened scene if they contain changed prefabs ?

**To Reproduce**
1. Extract this zip: [MyGame7.zip](https://github.com/stride3d/stride/files/4705785/MyGame7.zip)
2. Load the solution in game studio
3. Drag and drop ``MediumPrefab`` in the scene view/hierarchy
4. Gamestudio will now block for the next couple of minutes or might just crash before successfully adding that prefab to the scene.

**Additional context**
I initially started investigating this because I had a prefab made out of around 2000 fairly simple items, the scene editor opened that prefab in a couple of second at most but drag and dropping it inside another scene led to a seemingly never ending load. I understand that 2000 items is perhaps too large for one prefab to hold but I don't think we should expect such an operation to either block for more than 20 minutes and/or crash afterward.

---
If you don't have time to @xen2 I can work on this but I only have a couple of days to do so and I need some pointers:
1. How is this whole system architectured ?
2. Where is it used ?
3. Why do we need all of this quantum reflection stuff when we actually generate serializers for binary serialization ?
4. Is what I proposed above in ``Potential Solution`` feasible

[...]

PR #745, opened 2020-06-02
[GameStudio] Prefabs and scene loading optimization

Eideren:
# PR Details
Reduce the time taken to add and load prefabs and scenes in the gamestudio.

## Description
Most of the changes made only target the hot path, as I have mentioned on #743 I think that a significant re-write is required before performances are at an acceptable level, this will have to do in the mean time.

The cost associated with allocating are far greater than attempting a couple of casts before hand as those methods are extremely frequently used, so:
- Explicitly tested IReadonlyCollection against common collection implementation to retrieve the alloc-less enum for cases where it made sense.
- Simplified or replaced a couple of linq operations by handwritten loops.
- ``FindSubEntity`` now uses the ``EntityManager.GetEnumerator()`` when it can instead of going through the hierarchy.
- Added another branch in ``CollectionDescriptor`` to handle a couple of functions for ``IDictionary`` and cleaned up unnecessary logic for ``IList``.
- And other minor changes.

## Related Issue
This PR only helps slightly #743, this is still an issue.

## Motivation and Context
Performance issues

## Types of changes

## Checklist

### DIFF
--- a/sources/assets/Stride.Core.Assets.Quantum/AssetToBaseNodeLinker.cs
+++ b/sources/assets/Stride.Core.Assets.Quantum/AssetToBaseNodeLinker.cs
@@ -48,7 +48,13 @@ namespace Stride.Core.Assets.Quantum
             var targetIds = CollectionItemIdHelper.GetCollectionItemIds(targetNode.Retrieve());
             var itemId = sourceIds[sourceReference.Index.Value];
             var targetKey = targetIds.GetKey(itemId);
-            return targetReference.FirstOrDefault(x => Equals(x.Index.Value, targetKey));
+            foreach (var targetRef in targetReference)
+            {
+                if (Equals(targetRef.Index.Value, targetKey))
+                    return targetRef;
+            }
+
+            return null;
         }
     }
 }
--- a/sources/assets/Stride.Core.Assets/Reflection/CollectionItemIdentifiers.cs
+++ b/sources/assets/Stride.Core.Assets/Reflection/CollectionItemIdentifiers.cs
@@ -134,7 +134,12 @@ namespace Stride.Core.Reflection
         public object GetKey(ItemId itemId)
         {
             // TODO: add indexing by guid to avoid O(n)
-            return keyToIdMap.SingleOrDefault(x => x.Value == itemId).Key;
+            foreach( var kvp in keyToIdMap )
+            {
+                if( kvp.Value == itemId )
+                    return kvp.Key;
+            }
+            return null;
         }
 
         public void CloneInto(CollectionItemIdentifiers target, IReadOnlyDictionary<object, object> referenceTypeClonedKeys)
--- a/sources/assets/Stride.Core.Assets/Reflection/CollectionItemIdentifiers.cs
+++ b/sources/assets/Stride.Core.Assets/Reflection/CollectionItemIdentifiers.cs
@@ -134,9 +134,9 @@ namespace Stride.Core.Reflection
         public object GetKey(ItemId itemId)
         {
             // TODO: add indexing by guid to avoid O(n)
-            foreach( var kvp in keyToIdMap )
+            foreach (var kvp in keyToIdMap)
             {
-                if( kvp.Value == itemId )
+                if (kvp.Value == itemId)
                     return kvp.Key;
             }
             return null;
--- a/sources/assets/Stride.Core.Assets/Reflection/CollectionItemIdentifiers.cs
+++ b/sources/assets/Stride.Core.Assets/Reflection/CollectionItemIdentifiers.cs
@@ -133,13 +133,18 @@ namespace Stride.Core.Reflection
 
         public object GetKey(ItemId itemId)
         {
+            object output = null;
             // TODO: add indexing by guid to avoid O(n)
             foreach (var kvp in keyToIdMap)
             {
                 if (kvp.Value == itemId)
-                    return kvp.Key;
+                {
+                    if(output != null)
+                        throw new InvalidOperationException("Two elements of the collection have the same id");
+                    output = kvp.Key;
+                }
             }
-            return null;
+            return output;
         }
 
         public void CloneInto(CollectionItemIdentifiers target, IReadOnlyDictionary<object, object> referenceTypeClonedKeys)
--- a/sources/editor/Stride.Assets.Presentation/AssetEditors/AssetCompositeGameEditor/ViewModels/AssetCompositeHierarchyEditorViewModel.cs
+++ b/sources/editor/Stride.Assets.Presentation/AssetEditors/AssetCompositeGameEditor/ViewModels/AssetCompositeHierarchyEditorViewModel.cs
@@ -137,11 +137,10 @@ namespace Stride.Assets.Presentation.AssetEditors.AssetCompositeGameEditor.ViewM
         [CanBeNull]
         public override IEditorGamePartViewModel FindPartViewModel(AbsoluteId id)
         {
-            var item = RootPart as IEditorGamePartViewModel;
-            if (item != null && id == item.Id)
+            if (RootPart is IEditorGamePartViewModel item && id == item.Id)
                 return item;
-
-            return RootPart?.EnumerateChildren().BreadthFirst(x => x.EnumerateChildren()).OfType<IEditorGamePartViewModel>().FirstOrDefault(part => part.Id == id);
+            
+            return RootPart?.EnumerateChildren().BreadthFirst(x => x.EnumerateChildren()).FirstOrDefault(part => part is IEditorGamePartViewModel viewModel && viewModel.Id == id) as IEditorGamePartViewModel;
         }
 
         /// <summary>
--- a/sources/editor/Stride.Editor/Engine/EntityExtensions.cs
+++ b/sources/editor/Stride.Editor/Engine/EntityExtensions.cs
@@ -20,12 +20,41 @@ namespace Stride.Editor.Engine
             if (entity.Id == subEntityId)
                 return entity;
 
-            return entity.GetChildren().DepthFirst(x => x.GetChildren()).FirstOrDefault(x => x.Id == subEntityId);
+            foreach (var child in entity.Transform.Children)
+            {
+                if (child.Entity.FindSubEntity(subEntityId) is Entity e)
+                    return e;
+            }
+
+            return null;
         }
 
         public static Entity FindSubEntity(this Scene scene, Guid subEntityId)
         {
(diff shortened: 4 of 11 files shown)


sources/assets/Stride.Core.Assets.Quantum/AssetToBaseNodeLinker.cs
```diff
@@ -48,7 +48,13 @@ public override ObjectReference FindTargetReference(IGraphNode sourceNode, IGrap
...
             var itemId = sourceIds[sourceReference.Index.Value];
             var targetKey = targetIds.GetKey(itemId);
-            return targetReference.FirstOrDefault(x => Equals(x.Index.Value, targetKey));
+            foreach (var targetRef in targetReference)
+            {
+                if (Equals(targetRef.Index.Value, targetKey))
+                    return targetRef;
+            }
+
+            return null;
```
Kryptos-FR, line 57: That's exactly what `FirstOrDefault` does and I find it easier to read.

sources/assets/Stride.Core.Assets/Reflection/CollectionItemIdentifiers.cs
```diff
@@ -134,7 +134,12 @@ public void Validate(bool isList)
         public object GetKey(ItemId itemId)
         {
             // TODO: add indexing by guid to avoid O(n)
-            return keyToIdMap.SingleOrDefault(x => x.Value == itemId).Key;
+            foreach (var kvp in keyToIdMap)
+            {
+                if (kvp.Value == itemId)
+                    return kvp.Key;
+            }
+            return null;
```
Kryptos-FR, line 142: Again why reinventing the wheel? That is the exact implementation of `SingleOrDefault`.

It does miss a null-check:

```csharp
return keyToIdMap.SingleOrDefault(x => x.Value == itemId)?.Key;
```

sources/editor/Stride.Assets.Presentation/AssetEditors/AssetCompositeGameEditor/ViewModels/AssetCompositeHierarchyEditorViewModel.cs
```diff
@@ -137,11 +137,10 @@ public override void Destroy()
...
         public override IEditorGamePartViewModel FindPartViewModel(AbsoluteId id)
         {
-            var item = RootPart as IEditorGamePartViewModel;
-            if (item != null && id == item.Id)
+            if (RootPart is IEditorGamePartViewModel item && id == item.Id)
                 return item;
-
-            return RootPart?.EnumerateChildren().BreadthFirst(x => x.EnumerateChildren()).OfType<IEditorGamePartViewModel>().FirstOrDefault(part => part.Id == id);
+            
+            return RootPart?.EnumerateChildren().BreadthFirst(x => x.EnumerateChildren()).FirstOrDefault(part => part is IEditorGamePartViewModel viewModel && viewModel.Id == id) as IEditorGamePartViewModel;
```
Kryptos-FR, line 143: I could be wrong, but it seems to me this version will do one more cast that the previous one.

Could be interesting to look at the generated IL to confirm.

sources/editor/Stride.Editor/Engine/EntityExtensions.cs
```diff
@@ -20,12 +20,41 @@ public static Entity FindSubEntity(this Entity entity, Guid subEntityId)
             if (entity.Id == subEntityId)
                 return entity;
 
-            return entity.GetChildren().DepthFirst(x => x.GetChildren()).FirstOrDefault(x => x.Id == subEntityId);
+            foreach (var child in entity.Transform.Children)
+            {
+                if (child.Entity.FindSubEntity(subEntityId) is Entity e)
```
Kryptos-FR, line 25: Incorrect. You replaced a depth-first search which is sequential with a recursive call which could potentially crash the stack. Deep entity hierarchies will crash the editor.

sources/editor/Stride.Editor/Engine/EntityExtensions.cs
```diff
@@ -20,12 +20,41 @@ public static Entity FindSubEntity(this Entity entity, Guid subEntityId)
...
         public static Entity FindSubEntity(this Scene scene, Guid subEntityId)
         {
-            return scene.Entities.DepthFirst(x => x.GetChildren()).FirstOrDefault(x => x.Id == subEntityId);
+            if (scene.Entities.Count == 0)
+                return null;
+            
+            if (scene.Entities[0].EntityManager is EntityManager manager)
+            {
+                // The entity manager contains all entities of all the scenes under the root scene,
+                // it should be faster in most cases to iterate over that instead of going through the entity tree
```
Kryptos-FR, line 40: That path could be faster, but recursive calls should be avoided.

sources/editor/Stride.Editor/Engine/EntityExtensions.cs
```diff
@@ -20,12 +20,41 @@ public static Entity FindSubEntity(this Entity entity, Guid subEntityId)
...
+                {
+                    if (e.Current.Id == subEntityId && ReferenceEquals(scene, e.Current.Scene))
+                        return e.Current;
+                }
+            }
+            else // Slow path, go through tree recursively
+            {
+                foreach (Entity entity in scene.Entities)
+                {
+                    if (entity.FindSubEntity(subEntityId) is Entity e)
```
Kryptos-FR, line 52: Recursive call = bad. Keep using `DepthFirst` or investigate if `BreadthFirst` is faster (could be since it doesn't need to do a reverse enumeration).

sources/assets/Stride.Core.Assets.Quantum/AssetToBaseNodeLinker.cs
Eideren: This is the hot path, those methods are called millions of times a second when loading assets. All of those FirstOrDefault allocation adds up and trashes the GC causing significant slowdown for any new allocation. All of those changes are not baseless, I profiled them thoroughly.

sources/assets/Stride.Core.Assets/Reflection/CollectionItemIdentifiers.cs
Eideren: Actually this is not quite right, single should throw if there's not exactly one match, I'll change this.

sources/editor/Stride.Assets.Presentation/AssetEditors/AssetCompositeGameEditor/ViewModels/AssetCompositeHierarchyEditorViewModel.cs
Eideren: Again, OfType creates unnecessary allocations and that's the main issue we're trying to solve here.

sources/editor/Stride.Editor/Engine/EntityExtensions.cs
Eideren: Threads have by default around one meg of stack space, let's say that half of it is already taken, somehow, and this recursion allocs around 50 bytes per call, you could recurse a hierarchy with a depth of 10000 items before crashing, of course all of that is napkin math but I'm being safe here, it's very unlikely that half of the stack is already taken and even more unlikely that each recursion takes that much of the stack. AFAIK Quantum has a couple of place were it has to recurse, that's likely the first place where it will crash because of deep hierarchies. If I end up rewriting quantum I'll solve them both

sources/editor/Stride.Editor/Engine/EntityExtensions.cs
Eideren: There aren't any recursive calls here, just the one below which you commented on ? Tell me if I missed something here

sources/editor/Stride.Editor/Engine/EntityExtensions.cs
Eideren: cfr. other replies

sources/editor/Stride.Editor/Engine/EntityExtensions.cs
Kryptos-FR: Not true at all. We had a test scene (from a real project) that could very easily reach that limit.

sources/editor/Stride.Editor/Engine/EntityExtensions.cs
Kryptos-FR: We are talking about a single allocation for the call to `DepthFirst`. I don't see how that could be an issue.

Kryptos-FR:
I think you have a big misconception of allocations. They are not always necessarily bad. Plus lots of them have been removed in .NET Core which will be the next version of .NET (in November).

Recursive calls on the other hand are a bigger concern. So as far as I am concerned I disapprove this pull-request. I will create more issues that it solves.

tebjan:
I'm intrigued, how much performance improvement did you see after these changes?

Eideren:
Allocations on the heap at high frequency is *definitely* one of the worst thing you could do for performance and core won't magically clear any alloc' logic related to linq's or our usage of enum'.
Again, those functions don't just run once or twice, they can run millions of times during the whole loading process. And again, I'm not talking out of my ass here as you seem to imply, those changes *have* been profiled and they are significantly faster:

One of the prefab I work on takes **51 seconds** to fully load before those changes and after them it's down to **13 seconds**.

Here's a snapshot, for this one I did a drag and drop of the prefab mentioned above onto an empty scene, probably takes 30 minutes or so to finish, here's the % of time taken by functions on the main thread over two minutes.
Before those changes:
Annotation 2020-06-03 191833

After those changes:
Annotation 2020-06-03 185749

As you can see the third and the tenth most costly calls were ``GetEnumerator()`` related, the first fell to 15th place and the other is too low to be on the screenshot.

Those snapshots weigh around 1.5GB each, I won't upload them here but if you really want to I can do so on google drive or somewhere else.

sources/editor/Stride.Editor/Engine/EntityExtensions.cs
Eideren: A real project that contains entities whose furthest ancestor is 'easily' 10000 generations deep ? Sorry but I don't believe that's likely, perhaps at runtime but within the game studio it's just not manageable and it doesn't make a lot of sense to have such a construct in a game either, what kind of ridiculous scenario could possibly require 10000 parents.
Anyway tested that inside a build of the gamestudio before those changes and the process silently crashes before it can even bring the main window up, gamestudio just can't load anything deeper than around 1250 on my end.

If you still don't think those recursions are safe I could conditionally swap back to the previous version through ``RuntimeHelpers.ProbeForSufficientStack``

xen2:
Pending a proper resolution of #743, a x4 improvement is nothing to sneeze at!
Thanks again for looking into it @Eideren

Issue #693, opened 2020-05-02
Documentation for Color[index] is incorrect


johang88:
**Release Type**: Github
**Version**: master

**Describe the bug**
The documentation states "The index of the component to access. Use 0 for the alpha component, 1 for the red component, 2 for the green component, and 3 for the blue component". 

The actual implementation uses:
```
switch (index)
{
	case 0: return R;
	case 1: return G;
	case 2: return B;
	case 3: return A;
}
```

**Expected behavior**
Documentation should match the implementation

Eideren:
Line in question: https://github.com/stride3d/stride/blob/f2aa3ed9d9077727a2bd1d152340c628932668b1/sources/core/Stride.Core.Mathematics/Color.cs#L222

raphaelschmitz00:
First time working on open source, this seems perfect to check if I can handle the git part :D
I'll have a go!

raphaelschmitz00:
...maybe _too_ easy, this was fixed 3 years ago.

If it's not about more than the single line mentioned by @Eideren, at least.

Eideren:
Yep, sorry about that @raphaelschmitz00 !

PR #746, opened 2020-06-02
Correction Color[index]

Greengreens3:
Corrected the param documentations for Color[index].

# PR Details

Params for color indexes were incorrect as described in [Issue #693](https://github.com/stride3d/stride/issues/693) and so were corrected to:
 - 0: Red
 - 1: Green
 - 2: Blue
 - 3: Alpha

## Related Issue

https://github.com/stride3d/stride/issues/693

## Motivation and Context

Change is good for readability and to avoid confusion.

## Types of changes

- [ x ] Docs change / refactoring / dependency upgrade

## Checklist

### DIFF
--- a/sources/core/Stride.Core.Mathematics/Color.cs
+++ b/sources/core/Stride.Core.Mathematics/Color.cs
@@ -199,7 +199,7 @@ namespace Stride.Core.Mathematics
         /// <summary>
         /// Initializes a new instance of the <see cref="Color"/> struct.
         /// </summary>
-        /// <param name="values">The values to assign to the alpha, red, green, and blue components of the color. This must be an array with four elements.</param>
+        /// <param name="values">The values to assign to the red, green, blue, or alpha components of the color. This must be an array with four elements.</param>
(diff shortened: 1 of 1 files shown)


sources/core/Stride.Core.Mathematics/Color.cs
```diff
@@ -219,7 +219,7 @@ public Color(byte[] values)
         /// Gets or sets the component at the specified index.
         /// </summary>
         /// <value>The value of the alpha, red, green, or blue component, depending on the index.</value>
```
Eideren, line 221: This could be re-ordered as well

Greengreens3: Indeed, 
https://github.com/stride3d/stride/blob/86ee808f19bf96a600d49039fb9eadbb1f77a20f/sources/core/Stride.Core.Mathematics/Color.cs#L200-L202

Should this also be corrected since its the only other occurrence?

Eideren: Yes for the first.
I'll raise an issue about the second one as it looks like a bug given how it is used.

Eideren:
Thanks !

PR #752, opened 2020-06-08
Fix CompilerApp to not load invalid assemblies

Basewq:
# PR Details

Fix some CompilerApp issues so it can build/load the game projects.

## Description

`builderOptions.ExtraCompileProperties` is nullable so needs to guarded before calling `TryGetValue`.

Sometimes nuget generates paths with "\_.\_", eg. my `[GAME].Windows\obj\project.assets.json` contains
```
{
  "version": 3,
  "targets": {
    ".NETFramework,Version=v4.7.2": {
// ...omitted
      "Stride.SpriteStudio.Offline/4.0.0.1-beta03": {
        "type": "package",
        "dependencies": {
          "Stride.Assets": "4.0.0.1-beta03",
          "Stride.SpriteStudio.Runtime": "4.0.0.1-beta03"
        },
        "compile": {
          "lib/net472/_._": {}
        },
        "runtime": {
          "lib/net472/_._": {}
        }
      },
// ...other stuff
```

According to [this thread](https://github.com/dotnet/aspnetcore/issues/744), nuget sometimes emits empty directories.
I have changed it so these paths are ignored, so it's not added into the `projectDependency.Assemblies` since later in the code it tries to load these paths which do not exist since they are not a real assembly paths, though I don't know if that's the proper fix or not.

@xen2 I know you're the expert in this area, so feel free to implement the proper fix :)

## Types of changes

## Checklist

### DIFF
--- a/sources/assets/Stride.Core.Assets.CompilerApp/PackageBuilder.cs
+++ b/sources/assets/Stride.Core.Assets.CompilerApp/PackageBuilder.cs
@@ -148,7 +148,7 @@ namespace Stride.Core.Assets.CompilerApp
                 // Add runtime identifier (if any) to avoid clash when building multiple at the same time (this happens when using ExtrasBuildEachRuntimeIdentifier feature of MSBuild.Sdk.Extras)
                 if (builderOptions.Properties.TryGetValue("RuntimeIdentifier", out var runtimeIdentifier))
                     indexName += $".{runtimeIdentifier}";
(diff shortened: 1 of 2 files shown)


xen2:
BTW, I also ran into the `builderOptions.ExtraCompileProperties` being `null` recently.
I think they shouldn't be nullable (we can fix that in a separate commit later though).

PR #755, opened 2020-06-11
Game Systems to use Services instead of Game object

Basewq:
# PR Details

Bullet2PhysicsSystem and DynamicNavigationMeshSystem to use services to retrieve GameSettings and game systems to remove dependency on Game object.

## Motivation and Context

Ideally Game object should have a 'headless server' mode, but this is a lot of effort.
In the mean time, this change makes it easier to make a "game" class that doesn't inherit the Game class.
eg. consider the following HeadlessGame, but we want to load the physics settings from the asset, or include DynamicNavigationMeshSystem.
https://github.com/xen2/Xenko.ClientServerSample/blob/master/Xenko.ClientServerSample.Server/HeadlessGame.cs

Note I haven't really tested if DynamicNavigationMeshSystem could work in a headless mode, so could be excluded from this commit if it's a bit iffy to change.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Navigation/DynamicNavigationMeshSystem.cs
+++ b/sources/engine/Stride.Navigation/DynamicNavigationMeshSystem.cs
@@ -11,6 +11,7 @@ using Stride.Core.Mathematics;
 using Stride.Core.Reflection;
 using Stride.Engine;
 using Stride.Engine.Design;
+using Stride.Engine.Processors;
 using Stride.Games;
 using Stride.Navigation.Processors;
 using Stride.Physics;
@@ -39,7 +40,7 @@ namespace Stride.Navigation
         /// </summary>
         [DataMember(20)]
         public NavigationMeshBuildSettings BuildSettings { get; set; }
-        
+
         /// <summary>
         /// Settings for agents used with the dynamic navigation mesh
         /// </summary>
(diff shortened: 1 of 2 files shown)


xen2:
LGTM!
Did you make sure everything was still working in the Game Studio as well, or should I give it a try?

Basewq:
Yep, opened Game Studio and even created a new project to see if the whole thing runs. Nothing's crashed.

slango0513:
Me too converted the Xenko.ClientServerSample project to stride, and it runs well.
In the current version, it is difficult to implement a headless server. I cannot find a way to add 'Features' like ScriptSystem, pathfinding without refactoring the engine. They always depend more or less on graphics.

devjarmo:
I’ve spent many cycles attempting this. It seems only possible If you implement a custom game base class and use/modify the game tick functionality and then implement a new type of script, script system and a few other items. While I’ve been busy working on other things, my branch has the idea laid out in only new classes if you want to take a look at the concept.

slango0513:
> This reply is re: slango’s headless comment.
> 
> I’ve spent many cycles attempting this. It seems only possible If you implement a custom game base class and use/modify the game tick functionality and then implement a new type of script, script system and a few other items. While I’ve been busy working on other things, my fork has the idea laid out in only new classes if you want to take a look at the concept.

Thank you, I will take a look

PR #753, opened 2020-06-08
Fix namespace generator for new project to use Namespace parameter in…

Greengreens3:
# PR Details

Fix issue while creating a new project with different namespace then name project that caused namespace starting with number if Project Name started with numer

## Description

Added variable `nameSpace` in `NewGameTemplateGenerator` to build namespace from instead of name if namespace isn't null

## Related Issue

[Issue #547 ](https://github.com/stride3d/stride/issues/547)

## Types of changes

- [ ] Docs change / refactoring / dependency upgrade
- [x ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [ ] My change requires a change to the documentation.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

Issue #547, opened 2019-11-02
Namespace start by number


cycpe950609:
**Release Type**: Official Release

**Version**: 3.1.0.1-beta02-0752

**Platform(s)**: I only test on Windows

**Describe the bug**
When I make a new project which name is start by Number (e.g 3DGame ),Xenko will generate source code file with name of project, not namespce in 'Create a game...' Dialog

**To Reproduce**
1. In "Project selection" ,select New project -> General -> New game , name project 3DGame 
2.Click Select
3.In "Create a game...", type ThreeDGame in namespace, and click OK

**Expected behavior**
namespace in 3DGame.Windows\3DGameApp.cs should be ThreeDGame.Windows

**Additional context**
```cs
using Xenko.Engine;

namespace 3DGame.Windows//Here is ERROR
{
    class 3DGameApp//Here is ERROR
    {
        static void Main(string[] args)
        {
            using (var game = new Game())
            {
                game.Run();
            }
        }
    }
}
```

Kryptos-FR:
I disagree that we need to do anything about it. That is a restriction of how C# works, not something we want to bother with. Just don't name your project starting with a number.

I do think there was a check at some point ot prevent creating such invalid names, that could be one area where we can do something. But transforming the name the user has typed is not a good solution and I would always be strongly against it.

xen2:
@Kryptos-FR I think you might have misread.
Issue is about game name being used as namespace instead of namespace specified by user.
I will see if it's still worth keeping them separate and test what visual studio does for new csproj projects.

Kryptos-FR:
@xen2 I definitely read to fast and misunderstood.

Eideren:
This is mostly fixed, the window where you specify platforms and such when creating a project blocks user if they don't have a valid namespace but the generated *.Windows (linux, osx, etc.) and the class hosting their main still carry the project name which may or may not have an invalid name.

herocrab:
@Eideren it looks as though this issue was resolved and the PR merged. Can this be closed?

PR #759, opened 2020-06-14
Template validation

Eideren:
# PR Details
Namespace changes within the namespace validation window weren't being carried over to generated project source files, this PR fixes that.

## Description
See above.

## Related Issue
#547 

## Motivation and Context
Fix a bug.

## Types of changes

## Checklist

This PR supersedes #753 , see my comment over there for more info.

Do note that the generated ``*App.cs`` class name is based on the project name and gets auto-corrected through ``BuildValidClassName()``, it is not based on the namespace, they won't match if users change the namespace, it works fine but I'm not sure that that's expected behavior for you @xen2, I'll leave that to you.


xen2:
I have also noticed when checking #753 PR that namespace wasn't passed through, and as you guessed it's the proper fix to do.
Thanks for getting ahead of me and making a new PR to fix that!

PR #765, opened 2020-06-21
Include default values in shader reflection

azeno:
## Description
The PR adds the default values as defined in a shader to the shader reflection. This allows to read them during runtime after they have been modified or new ones have been added.

I took the liberty and used the `EffectValueDescription.DefaultValue` property for it, as it wasn't used anywhere before.

I've also added tests to the PR which exposed a few other issues like invalid HLSL being generated when declaring a variable like `vector<float, 2> foo;`, constructors of `Half` vectors not being present or `ShaderClassString` not working.

## Related Issue
None that I know of.

## Motivation and Context
The default values of shader variables are being read at compile time by the shader key generator and stored in static C# fields. This works fine as long as one doesn't change those values during runtime or add new variables to a shader. Having to restart the application and regenerate the keys can be quite cumbersome when working on a shader.

## Types of changes

## Checklist

### DIFF
--- a/sources/core/Stride.Core.Mathematics/Half2.cs
+++ b/sources/core/Stride.Core.Mathematics/Half2.cs
@@ -36,6 +36,32 @@ namespace Stride.Core.Mathematics
     public struct Half2 : IEquatable<Half2>
     {
         /// <summary>
+        /// The size of the <see cref="Stride.Core.Mathematics.Half2"/> type, in bytes.
+        /// </summary>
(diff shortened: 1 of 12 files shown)


tebjan:
> or `ShaderClassString` not working.

great, you fixed that!? that was the missing part of my earlier PR #642. nice to see that working.

xen2:
LGTM, very appreciated changes!

PR #767, opened 2020-06-22
Little cleanup to previous PR regarding shader reflection

azeno:
## Description
Reverts change from previous commit that reflection wasn't cleaned up in debug.

## Related Issue
#765 

## Motivation and Context
The shader reflection should be the same in debug and release builds. My change in the last PR was a little too quick. It would've required user code to filter the reflection on its own.
It would still be nice to tell the compiler not to remove parameters, but that's for another day.

## Types of changes

## Checklist


xen2:
Thanks!
(Interestingly, I spotted it during review and had some doubts if it would be OK (we ran into this issue a lot), but assumed it was part of the hash and working if you didn't have any issues so far with it -- I will make sure next time!)

PR #769, opened 2020-06-24
[Engine] Scene = null also detaches parent

Eideren:
# PR Details
Users want a succinct way to remove entities from the game regardless of their place in the hierarchy, right now they have to set both parent and scene to null when they aren't sure that either of these is set.
Doing this operation should not require more than a single line, it might come more instinctively and will be easier to keep in mind.

In this PR when the value passed to ``Scene`` is null the engine detaches the entity from its parent if it has any or removes it from the scene.

I decided against making a specific function for it to keep it more in line with how we do other operations with transforms and entities and also to avoid having the same issues that unity has with the 'magic' Destroy() function.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Engine/Engine/Entity.cs
+++ b/sources/engine/Stride.Engine/Engine/Entity.cs
@@ -89,7 +89,8 @@ namespace Stride.Engine
         }
 
         /// <summary>
-        /// The parent scene.
+        /// The scene this entity is in. <br/>
+        /// Setting this to null will remove the entity from the scene and detach it from its parent if it has one.
         /// </summary>
         [DataMemberIgnore]
         public Scene Scene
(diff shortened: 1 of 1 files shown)


tebjan:
we could also simply have a method `Entity.Remove()` that does all the right things depending on the case.

Eideren:
I gave my two cents on this in the PR Details section, here's an excerpt
> I decided against making a specific function for it to keep it more in line with how we do other operations with transforms and entities and also to avoid having the same issues that unity has with the 'magic' Destroy() function.

Learning that ``Scene = null`` removes an entity from the game implies to the user that ``Scene = value`` adds it to one, also that if they do ``Parent = null`` and the value for ``Scene`` becomes null through that operation that the object would then be removed from the game.
``Remove()`` would probably be easier to find through intellisense but it would obfuscate all of the instinctive knowledge that you would gain from this specific pattern.

PR #772, opened 2020-06-30
Fix JumpyJet collision detection

aunpyz:
# PR Details

Refactor the code, fix some logic error (that I see during the walkthrough) and fix bugs from collider detection

## Description

I'll just focus on the bugs I fix, refactoring and fixing logic error are just trivial and does not break the game
- The original JumpyJet project has an overlapping collider on the PipeSet prefab and for some reason when the player goes through that overlapped area, the collision detection sometimes(?) does not broadcast `GameOverEventKey`.
And for the same reason, sometimes, too, the score will be increased to 1 right on game restarted when the player dies on the Pipe
**Overlapping collider**
collision-old
**Player goes through that overlapping area**
jumpyjet-old-collider
**Player gains a score on game restarted when it dies on the Pipe**
jumpyjet-old-score

- So what I do now is to move the `PassedTrigger` collider past the pipe and detect `NewCollision()` on `CountPassedPipes` task instead of `CollisionEnded()`
**New `PassedTrigger`**
collision-new
**Propose changes' result**
jumpyjet-new

## Motivation and Context

At first, I just wanted to walk through the code but accidentally found bugs in JumpyJet, so I just try to fix it and learn from it at the same time

## Types of changes

## Checklist

### DIFF
--- a/samples/Games/JumpyJet/Assets/Shared/Pipe Set.sdprefab
+++ b/samples/Games/JumpyJet/Assets/Shared/Pipe Set.sdprefab
@@ -36,8 +36,8 @@ Hierarchy:
                         ColliderShapes:
                             3fc7bd36a1bfa1290243340c6325db55: !BoxColliderShapeDesc
                                 Is2D: true
-                                Size: {X: 1.0, Y: 5.0, Z: 1.0}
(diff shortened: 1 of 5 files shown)


Eideren:
Hey, thanks a bunch !
Looks fine to me, I'll just merge this in.

PR #773, opened 2020-07-01
Increase UIRenderFeature SortKey to always render UI on top of Sprites

manio143:
# PR Details

This change prevents UI elements from being rendered underneath Sprites.
Previous value is `128` - the default, same as `SpriteRenderFeature`, which meant that during sorting the UI could potentially end up rendered before the sprites (or not).

The value `192` was chosen to match the `SortKey` of `BackgroundRenderFeature`.

## Motivation and Context

I was playing around with JumpyJet and noticed that for some combinations of entity order in the scene the UI was not rendered on top of everything else as it should be.

JumpyUIOverlayIssue

After I applied the change I played a bit with the order of entities as previously but now UI was always on top.

## Types of changes

## Checklist

_I'm not sure about changes to the documentation or in-code comments for this change._

### DIFF
--- a/sources/engine/Stride.UI/Rendering/UI/UIRenderFeature.cs
+++ b/sources/engine/Stride.UI/Rendering/UI/UIRenderFeature.cs
@@ -43,6 +43,11 @@ namespace Stride.Rendering.UI
         /// </summary>
         public UIElement UIElementUnderMouseCursor { get; private set; }
 
+        public UIRenderFeature()
+        {
+            SortKey = 192;
+        }
+
         protected override void InitializeCore()
         {
             base.InitializeCore();


manio143:
If you want to see the difference, other than in game, during debugging you can break at `sources\engine\Stride.Rendering\Rendering\RenderSystem.cs` [line 322](https://github.com/stride3d/stride/blob/1c01cd7a56217f7284735f132fa4cba88da2e2e8/sources/engine/Stride.Rendering/Rendering/RenderSystem.cs#L322) and view the `renderViewStage.SortedRenderNodes`. We want only the last items to have the `UIRenderFeature`.

aunpyz:
How can I set a breakpoint on source code when debugging the game? All source code is exposed as meta data
image

tebjan:
@aunpyz you need to check out the source code and open the Stride.sln and attach the debugger to the running game.

xen2:
Supposedly, even without checking out source code, it should download pdb from SourceLink and allow you to trace into Stride code out of the box.
Make sure SourceLink is enabled in VIsual Studio: 
image

aunpyz:
@xen2 Seems like it's enabled by default (at least on my end). But I can only trace up to source metadata just like the picture I posted above
@tebjan Thanks!! That's something I don't know up until now! Unfortunately, the breakpoint is not hit even I attach to a running game.
Here's what I got:
image

manio143:
If you build Stride yourself, and use that version, you can open the specific file from Stride in the game's solution and break there. This way VS can auto attach to your game when you run it and it manages to match the source from nuget's pdb.

aunpyz:
@manio143 I build all packages and switch nuget packages to the ones I built (I tried with JumpyJet solution), still getting the same result I mentioned
Would be nice if there's a tutorial for newcomers c# dev

manio143:
@xen2 I see there's a failed check, but cannot view details. Is there something more I should do for this PR?

xen2:
@manio143 I ran the test manually yesterday. It seems something didn't compile but it looks unrelated/spurious.
I am going to investigate.

xen2:
OK, sorted out the tests, all good!

Issue #749, opened 2020-06-06
GameTime needs more options


Basewq:
**Is your feature request related to a problem? Please describe.**
Maybe or maybe not the GameTime class itself, but there needs to be more features to control the 'timer' of a game.
A game needs options be able to pause a game, scale the timer (eg. for slow motion).
This also needs to be affect the physics engine, model animation, but NOT the UI (though not sure if UI has any built in animations?).
Not sure how Sound should be affected.

While you can just create your own "GameTime" object that your own scripts can use, this doesn't affect physics/animation since they specifically use GameTime.
Physics only has `Simulation.DisableSimulation`, but no time scale, though there is a `FixedTimeStep` field, but not sure if that is really the appropriate thing to change for time scaling.

**Describe the solution you'd like**
Honestly don't know what the solution is for this.
I know Unity uses a static field (bleh!) `Time.timeScale`.
Perhaps be able to set it directly on the GameTime object itself, or maybe via a service/game system?
One thing that needs to be clear if done is ensuring there's a distinction between the _current_ time's scale and the _requested change_ in time scale.

ykafia:
I'll have a go on this issue and see what i can do.

herocrab:
This looks to be merged, were we waiting on the docs to be updated before closing?

Basewq:
Closed, as per #785

PR #785, opened 2020-07-13
Time Warp in Stride

ykafia:
# PR Details

This PR is here to add a slow motion / time warp effect to Stride.

## Description

First is to add some public fields in `GameTime` :
* `Factor` is the factor by which the time will be slowed down from 0 to 1.
* `WarpedTotal` will represent the total game time accounting for slowed moments.
* `WarpedElapsed` will represent the elapsed time times the factor.

I'm yet unsure of next steps but from what I understood, slowing down the actual game time will slow the UI and everything related to time. So far the only thing that I believe needs to be slowed down in games are game logic, physics, animation and effects.

I'm still wondering about use cases, should we make every processors use the warped time by default or should we let the user decide which entities is slowed down?

EDIT : I have some answers for my own questions. Warped time is useful for physics, effects and logic. For animation and particles they use a render context for speed but there's already a time factor that can be set from the logic code.

## Related Issue

This PR is related to #749

## Types of changes

## Checklist

- [x] Rigid bodies are slowed down
- [x] Character component and Kinematic bodies are slowed down
- [x] Particles are slowed down
- [x] Animations are slowed down


sources/engine/Stride.Games/GameTime.cs
```diff
@@ -94,10 +97,51 @@ public GameTime(TimeSpan totalTime, TimeSpan elapsedTime)
...
+
+        /// <summary>
+        /// Gets the amount of time elapsed multiplied by the time factor.
+        /// </summary>
+        /// <value>The warped elapsed time</value>
+        public TimeSpan WarpElapsed{ get; private set; }
+
+
+        /// <summary>
+        /// Gets or sets the time factor for the time factor.
```
manio143, line 115: You should mention the `0..1` range here.

ykafia:
The slow motion works for particles, animations and physics.
For particles and animation, the update time is multiplied by the time factor in the GameTime class. As for physics, the simulation resolution is multiplied by the same value. 
Users can also use the WarpElapsed value in the GameTime class for logic.

I consider it finished but i'd be happy to receive any critics about it.

sources/engine/Stride.Games/GameTime.cs
```diff
@@ -94,10 +97,44 @@ public GameTime(TimeSpan totalTime, TimeSpan elapsedTime)
...
+        /// </summary>
+        /// <value>The multiply factor, a double between 0 and 1.</value>
+        public static double Factor
+        {
+            get => factor; 
+            set 
+            {
+                // TODO: Use pattern matching for a clearer code
+                if (value > 1)
+                    factor = 1;
```
salahchafai, line 124: why clamping the value between 0 and 1 ? i think it just needs to be 0 or higher, no need for preventing it from being higher than 1 cuz someone might set it to 2 (or more) for a fast forward effect

sources/engine/Stride.Games/GameTime.cs
```diff
@@ -94,10 +97,42 @@ public GameTime(TimeSpan totalTime, TimeSpan elapsedTime)
...
+        {
+            // TODO: When switching with .NET 5, use the multiply operator instead of this. elapsedGameTime * Factor
+            get => TimeSpan.FromSeconds(Elapsed.TotalSeconds * Factor);
+        }
+
+
+        /// <summary>
+        /// Gets or sets the time factor for the time factor.
+        /// </summary>
+        /// <value>The multiply factor, a double between 0 and 1.</value>
```
salahchafai, line 116: this should be changed, maybe "The multiply factor, higher or equal to 0"

ykafia: Ah right! i forgot the documentation!

sources/engine/Stride.Physics/Simulation.cs
```diff
@@ -687,7 +688,7 @@ internal void Simulate(float deltaTime)
 
             SimulationProfiler = Profiler.Begin(PhysicsProfilingKeys.SimulationProfilingKey);
 
-            if (discreteDynamicsWorld != null) discreteDynamicsWorld.StepSimulation(deltaTime, MaxSubSteps, FixedTimeStep);
+            if (discreteDynamicsWorld != null) discreteDynamicsWorld.StepSimulation(deltaTime, MaxSubSteps, FixedTimeStep * (float) GameTime.Factor);
```
salahchafai, line 691: here you need to multiply the deltaTime with the TimeFactor instead of the FixedTimeStep as multiplying the FixedTimeStep will mess up the simulation

this video shows how the physics act weirdly when slowing time https://www.youtube.com/watch?v=xj5wuY6jrWw&feature=youtu.be (the ball is supposed to keep bouncing forever)

salahchafai:
works fine for me now, thnx for the contribution, maybe tag xen2 to review it now ?

ykafia:
Thanks @salahchafai for reviewing!
And yes maybe @xen2 could review it as it's relatively simple?

sources/engine/Stride.Games/GameTime.cs
```diff
@@ -32,6 +32,7 @@ public class GameTime
     {
         private TimeSpan accumulatedElapsedTime;
         private int accumulatedFrameCountPerSecond;
+        private static double factor;
```
Eideren, line 35: Users can use stride to run multiple game instances concurrently in the same process, factor should not be static otherwise all game instances will share the same factor.

sources/engine/Stride.Games/GameTime.cs
```diff
@@ -94,10 +97,44 @@ public GameTime(TimeSpan totalTime, TimeSpan elapsedTime)
...
+        }
+
+
+        /// <summary>
+        /// Gets or sets the time factor.<br/>
+        /// This value controls how much the warped time flows, this includes physics, animations and particles.
+        /// A value between 0 and 1 will slow time, a value above 1 will make it faster.
+        /// </summary>
+        /// <value>The multiply factor, a double value higher or equal to 0</value>
+        public static double Factor
```
Eideren, line 119: Todo: You could write it like so:
```cs
set => factor = value > 0 ? value : 0;
```
what do you think ?

sources/engine/Stride.Games/GameTime.cs
```diff
@@ -94,10 +97,44 @@ public GameTime(TimeSpan totalTime, TimeSpan elapsedTime)
...
+
+
+        /// <summary>
+        /// Gets the amount of time elapsed multiplied by the time factor.
+        /// </summary>
+        /// <value>The warped elapsed time</value>
+        public TimeSpan WarpElapsed 
+        {
+            // TODO: When switching with .NET 5, use the multiply operator instead of this. elapsedGameTime * Factor
+            get => TimeSpan.FromSeconds(Elapsed.TotalSeconds * Factor);
```
Eideren, line 109: Use ``Elapsed.Ticks`` and ``TimeSpan.FromTicks`` instead, FromSeconds is less precise, see remarks over at https://docs.microsoft.com/en-us/dotnet/api/system.timespan.fromseconds?view=netframework-4.8

sources/engine/Stride.Particles/Components/ParticleSystemSimulationProcessor.cs
```diff
@@ -88,7 +88,7 @@ private void UpdateParticleSystem(ParticleSystemComponentState state, float delt
                 particleSystem.UniformScale = xScale;
             }
 
-            particleSystem.Update(deltaTime * speed);
+            particleSystem.Update(deltaTime * (float) GameTime.Factor * speed);
```
Eideren, line 91: Cast/convert to float the result of the three mult instead of ``factor`` itself, might as well use full precision since float mult isn't significantly faster than double mult on current architecture.

sources/engine/Stride.Physics/Simulation.cs
```diff
@@ -687,7 +688,7 @@ internal void Simulate(float deltaTime)
 
             SimulationProfiler = Profiler.Begin(PhysicsProfilingKeys.SimulationProfilingKey);
 
-            if (discreteDynamicsWorld != null) discreteDynamicsWorld.StepSimulation(deltaTime, MaxSubSteps, FixedTimeStep);
+            if (discreteDynamicsWorld != null) discreteDynamicsWorld.StepSimulation(deltaTime * (float) GameTime.Factor, MaxSubSteps, FixedTimeStep);
```
Eideren, line 691: See other remark about cast conversion

Eideren (commented):
Thanks a bunch @ykafia and sorry for the delay, I should have taken a look at this sooner but now that I'm here I'll take care of this, no need to bother xen :)

The audio system doesn't seem to be affected by this, unless I'm mistaken ? But it would probably warrant its own PR anyway and perhaps should be toggle-able.

sources/engine/Stride.Games/GameTime.cs
tebjan: 0 might be a bit risky because it would mean a 0-time increment per frame. this could mean that animation code that uses the delta time in as divisor will get a division by 0 error.

Eideren: I did think about that but I didn't mention it as I would expect engine code to already handle such cases: [PlayingAnimation](https://github.com/stride3d/stride/blob/master/sources/engine/Stride.Engine/Animations/PlayingAnimation.cs) already has a TimeFactor which itself can be any float value. If a system doesn't handle a delta of zero I do think it's his responsibility to work around that himself ?

ykafia: If we're thinking this way, this would make the game time factor a non-static field, but the user has to update the speed factors for each particle + animation ?

salahchafai: not exactly, from what i understood you just have to make the time factor a non-static field, what tebjan and eideren are talking about is like what's happening in the 3rd person template where the animation speed is divided by the delta time so if someone makes the time factor 0 then that code will return an error for dividing by 0

sources/engine/Stride.Games/GameTime.cs
ykafia: I've tried this first, one of my issue with it is that Particle and Animation systems use a RenderContext GameTime which is not a reference to the GameTime accessible in scripts.
My workaround was doing that in a sync script : 
```csharp
if(Input.IsKeyPressed(Keys.U))
{
        slowdown = !slowdown;
        Game.UpdateTime.Factor = slowdown ? 0.5f : 1;
        var rc = RenderContext.GetShared(SceneSystem.Services);
        rc.Time.Factor = slowdown ? 0.5f : 1;
}
```

Eideren: Alright, after taking a closer look you shouldn't have to change stuff in the animation and particle system; you change the update time's factor but not the draw time and that variable is passed onto the rendercontext/animation and particle system so if both the draw and update time factors are the same you shouldn't have to bother with including the factor mult in the animation and particle file.

But yeah, users shouldn't have to modify both factor for it to work correctly though, it doesn't make a lot of sense to have the game's logic slowdown but not the animation and particle so I'll create a PR after yours to merge update and draw time together, I don't see any reason to keep them separate.

So all you should do is replace Elapsed by WarpElapsed for animation, particle and physics (https://github.com/stride3d/stride/blob/aab1aa8fca05210fbfee0a0b0c33b17abb300419/sources/engine/Stride.Physics/Bullet2PhysicsSystem.cs#L95), replace static fields with instance fields, match the drawtime's factor with the update time's factor in gamebase (around here https://github.com/stride3d/stride/blob/aab1aa8fca05210fbfee0a0b0c33b17abb300419/sources/engine/Stride.Games/GameBase.cs#L635), and if that works you're good to go.

sources/engine/Stride.Games/GameTime.cs
ykafia: I favored Eideren's simple setter here.

sources/engine/Stride.Games/GameTime.cs
ykafia: Thanks for your insight and your time!

Eideren:
LGTM, thanks again @ykafia !

sources/engine/Stride.Games/GameTime.cs
herocrab: Did the docs receive an update?

PR #793, opened 2020-07-14
[Runtime] Move UI assets to Stride.UI + include StrideUILibrary as root

manio143:
# PR Details

## Description

Move UI assets from `Stride.Engine` to `Stride.UI` - this is non-breaking as assets are identified by GUID.
Include `StrideUILibrary.sduilib` as a root asset for the `Stride.Engine` package.

image

I've verified nothing is broken by running a few existing samples and creating a new UI page which references the UI library and assets being moved around.

## Motivation and Context
Currently when creating a `UIPage` in code one has to provide all the details for controls, like fonts, sprites, etc. Those assets are available, just not included. This change allows to do the following:

```csharp
var uiLibrary = Content.Load<UILibrary>("StrideUILibrary");
var button = uiLibrary.InstantiateElement<Button>("Button");
```
And get a fully displayable button.

I understand that not everyone might want to include the default UI elements in their project, but it's just additional 128KB (less than the default splash screen) and allows to easily create dev UI in code without having to link the font, button sprites, etc. in the asset.

## Types of changes

## Checklist

_Doesn't **require** documentation change, but:_
It would be nice to then mention this in https://doc.stride3d.net/latest/en/manual/ui/ui-libraries.html


xen2:
I need to take a quick look if OK.
In any case, the place it should be is `Stride.UI` package rather than `Stride.Engine`.

Eideren:
Did you end up taking a look at this one @xen2 ? 
@manio143 still needs to come back for that change request though

manio143:
Hmm, currently the Stride.UI doesn't contain any assets, but I think it should be possible to move the StrideUILibrary into that project while preserving the Guid identifier in order to not cause a breaking change for users.

Eideren:
Sure, I'll change this to a draft then

sources/engine/Stride.Assets/Stride.Assets.csproj
```diff
@@ -65,25 +65,6 @@
       <Link>Properties\SharedAssemblyInfo.cs</Link>
     </Compile>
   </ItemGroup>
-  <ItemGroup>
```
manio143, line 68: Removing those `<None Include` because it references files under Stride.Engine and Stride.UI - blame shows this is a remnant from the initial commit.
I'm guessing those were added to allow MSBuild to track changes in assets files and rebuild the project when one of them changes. Currently that is not in place elsewhere and changing assets requires manually touching any C# related files to trigger MSBuild to rebuild the project.

manio143:
I just noticed that for example GameProfilingSystem in Stride.Engine takes a runtime dependency on the default sprite font asset. With this PR I'd break it if someone imports Stride.Engine but doesn't import Stride.UI

Ethereal77:
Is `GameProfilingSystem` not using `DebugText` instead of the regular `SpriteFont`? If using `DebugText`, it is loading the Debug Font sprite sheet, which is not a "standard" sprite font asset. With this I mean that even if we move the default sprite font asset to other package, if we keep the debug font in `Stride.Engine` it should work. Shouldn't it?

manio143:
It uses the FastTextRender rather than DebugText. See here:
https://github.com/stride3d/stride/blob/ffd71acc8a8a26a80e5275f833b6e1442e98d01c/sources/engine/Stride.Engine/Profiling/GameProfilingSystem.cs#L354-L358

Ethereal77:
> It uses the FastTextRender rather than DebugText. See here:
> 
> https://github.com/stride3d/stride/blob/ffd71acc8a8a26a80e5275f833b6e1442e98d01c/sources/engine/Stride.Engine/Profiling/GameProfilingSystem.cs#L354-L358

Ok, that's what I meant. It is not a `SpriteFont` but a regular `Texture` that `FastTextRenderer` knows how to use. So that texture should stay in `Stride.Engine` as it is a "private" spritefont-like asset, not something the user can/should use.

xen2:
Sorry to revive old PR.
Agree with @Ethereal77 : it seems `StrideDebugSpriteFont` should stay in `Stride.Engine`
`Stride.UI` can reference it still and have UI-specific assets.

xen2:

PR #799, opened 2020-07-19
[Physics] Prevent invalid state when component was added right after removal

### LINKED ISSUE #798: Changing an entity's (with Rigidbody) parent in a single frame causes a crash
**Release Type**: GitHub

**Version**: master/beta04-1254

**Platform(s)**: Any

**Describe the bug**
Changing an entity's (with a Rigidbody) parent within a frame causes a game crash in the next frame. Specifically the physics system crashes.

**To Reproduce**
1. Create a new game
2. Add the following script

        public class Bug : SyncScript
        {
            // needs a rigid body
            public Entity EntityToRemove { get; set; }
            public Entity NewParent { get; set; }

            public override void Update()
            {
                //    EntityToRemove.Scene = null;
                EntityToRemove.GetParent().RemoveChild(EntityToRemove);
                NewParent.AddChild(EntityToRemove);
            }
        }
3. Plug two entities into the component in the Game Studio (if you use the first line (commented) they can be anywhere, otherwise `EntityToRemove` has to have a parent).
4. Run game, observe crash

**Expected behavior**
Fix the issue to allow changing a parent of an entity within one frame. Or, disallow doing so by throwing an exception when adding a child that was someone else's child within the same frame.

**Additional context**
Changing the parent across two frames works fine (first frame - remove parent, second frame - add child).

The crash is happening at https://github.com/stride3d/stride/blob/ad0d94f60445473b569a79b0da6178384aa0cb47/sources/engine/Stride.Physics/Simulation.cs#L690

If the Rigidbody is `Kinematic` then there is a `NullReferenceException` here (for `Data`): https://github.com/stride3d/stride/blob/ad0d94f60445473b569a79b0da6178384aa0cb47/sources/engine/Stride.Physics/Elements/RigidbodyComponent.cs#L406

manio143:
# PR Details

Fixes #798 by not removing a component that has been re-added in the same frame.

## Description

When an entity is removed from a scene it is removed from all the component processors. For physics, this removal is delayed till the next frame. So if the same entity is added back in the same frame, it's physics component should not be removed.

Note: This causes the associated data in the `ComponentDatas` to be a different object than in `component.Data`, however, they still have the same _value_ because they pass the `IsAssociatedDataValid` check.

## Related Issue
#798 

## Motivation and Context

Allows removing and adding entities, with a physics component, from the scene without the need to clone them.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Physics/Engine/PhysicsProcessor.cs
+++ b/sources/engine/Stride.Physics/Engine/PhysicsProcessor.cs
@@ -103,11 +103,9 @@ namespace Stride.Physics
 
         protected override void OnEntityComponentAdding(Entity entity, PhysicsComponent component, AssociatedData data)
         {
-            if (currentFrameRemovals.Contains(component))
-            {
-                currentFrameRemovals.Remove(component);
+            // Tagged for removal? If yes, cancel it
+            if (currentFrameRemovals.Remove(component))
                 return;
-            }
 
             component.Attach(data);
 
@@ -155,7 +153,7 @@ namespace Stride.Physics
             component.Detach();
         }
 
-        private readonly List<PhysicsComponent> currentFrameRemovals = new List<PhysicsComponent>();
+        private readonly HashSet<PhysicsComponent> currentFrameRemovals = new HashSet<PhysicsComponent>();
 
         protected override void OnEntityComponentRemoved(Entity entity, PhysicsComponent component, AssociatedData data)
         {
--- a/sources/engine/Stride.Physics/Engine/PhysicsProcessor.cs
+++ b/sources/engine/Stride.Physics/Engine/PhysicsProcessor.cs
@@ -103,6 +103,12 @@ namespace Stride.Physics
 
         protected override void OnEntityComponentAdding(Entity entity, PhysicsComponent component, AssociatedData data)
         {
+            if (currentFrameRemovals.Contains(component))
+            {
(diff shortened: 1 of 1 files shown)


Eideren:
Is there a particular reason why we would add components right away but delay removing to the next frame @xen2 ? 

This should work for now but I believe that it could introduce unexpected issues down the line by the fact that *remove, same frame add* would not behave like *remove, next frame add*, we could fix that by force-running the removing logic right before adding it back but it looks like there's a good reason we delay removing, perhaps related to async contact events, not sure.

sources/engine/Stride.Physics/Engine/PhysicsProcessor.cs
```diff
@@ -103,6 +103,12 @@ protected override bool IsAssociatedDataValid(Entity entity, PhysicsComponent ph
 
         protected override void OnEntityComponentAdding(Entity entity, PhysicsComponent component, AssociatedData data)
         {
+            if (currentFrameRemovals.Contains(component))
```
xen2, line 106: Contains/Remove are O(n) so we better avoid this (O(n^2) if we add n physics component, since it's called for every new component).
Let's turn `currentFrameRemovals` into a `HashSet<>`.
Then you can also simplify this code:
```
if (currentFrameRemovals.Remove(component))
    return;
```

PR #801, opened 2020-07-22
Allow components to have required members and report warning/error if they're null

### LINKED ISSUE #778: Allow components to have required members and warn when the asset value for such member is null
### Feature description
It is good practice to find bugs during the compilation steps rather than in runtime. When designing components it may be a good idea to put a `[NotNull]` attribute on a required field or property. However, currently there is no check in place that would actually validate that an asset fulfills such promise.

The Game Studio and more importantly the Asset Compiler should emit a warning when the asset for a required field is `null`.

The alternative is observing a runtime exception to find out an asset is missing.

### Additional context
I wanted to submit a PR, but then I read I should discuss the solution in an issue first.

In order to handle this situation I have created a step in the `YamlAssetSerializer` that checks each serialized/deserialized member for the `[NotNull]` attribute and if they have it whether they're equal to `null`. If so a warning is emitted.
You can see the changes here: [manio143/stride: asset_null_warning](https://github.com/manio143/stride/commit/6e236d6202e1132086c6f7f5b2c697788d2689c3)

The Asset Compiler reads assets and the Game Studio reads assets when opening project and writes assets when saving it.
My current solution seems fine for the Asset Compiler, but may be a bit too verbose for Game Studio as seen on the GIF below.

asset_compiler_null_warning

asset_null_warning

### Discussion topics

* Is `[NotNull]` a good attribute to be used here (there already are Stride components with members marked with it) or maybe a `[MemberRequired]` would be a better choice?
* Performance - is checking an attribute for every serialized/deserialized member going to have a performance impact? It would be interesting to test it on a bigger project.
* What is the best way to signal this warning in Game Studio?
    My current approach cannot give additional details about which Scene/Entity has the component with a missing asset.
* Should this validation process happen during serialization or be separately invoked?
* Similar approach for `[ItemNotNull]` in collections?

manio143:
# PR Details

## Description

This PR introduces a `[MemberRequired]` attribute that can be placed on fields and properties of components. If during the asset compilation a field/property with that attribute is `null`, a warning/error is issued.

Additionally, component checks in the `EntityHierarchyCompilerBase` have been abstracted into an interface `IEntityComponentCheck` to allow better maintenance and future expansion of such checks.

## Related Issue

Resolves #778 

## Motivation and Context

It's often very easy to forget to set a value for a component in Game Studio and you would only find that out with an exception. With this change, you can mark properties as required and get a warning during compilation which is more visible.

asset_compiler_null_warning_2

## Types of changes

## Checklist

The docs would need to be updated to include information about a possibility of marking properties with the added attribute.
All tests should be passing (I had some errors but they seemed to be completely unconnected).

### DIFF
--- a/sources/engine/Stride.Assets/Entities/ComponentChecks/RequiredMembersCheck.cs
+++ b/sources/engine/Stride.Assets/Entities/ComponentChecks/RequiredMembersCheck.cs
@@ -1,0 +1,71 @@
+using System;
+using System.Reflection;
+using Stride.Core.Annotations;
+using Stride.Core.Assets;
+using Stride.Core.Assets.Compiler;
+using Stride.Engine;
+
+namespace Stride.Assets.Entities.ComponentChecks
+{
+    /// <summary>
+    /// Checks if all members of a component with a <see cref="MemberRequiredAttribute"/> are assigned a value.
+    /// </summary>
+    public class RequiredMembersCheck : IEntityComponentCheck
+    {
+        /// <inheritdoc/>
+        public bool AppliesTo(Type componentType)
+        {
+            return true; // applies to any component
+        }
+
+        /// <inheritdoc/>
+        public void Check(EntityComponent component, Entity entity, AssetItem assetItem, string targetUrlInStorage, AssetCompilerResult result)
+        {
+            var type = component.GetType();
+            var componentName = type.Name; // QUESTION: Should we check attributes for another name?
+
+            var fields = type.GetFields(); // public fields, only those can be serialized?
+            var properties = type.GetRuntimeProperties(); // all properties may have a DataMember attribute
+
+            foreach (var field in fields)
+            {
+                if (field.FieldType.IsValueType)
+                    continue; // value types cannot be null, and must always have a proper default value
+
+                MemberRequiredAttribute memberRequired;
+                if ((memberRequired = field.GetCustomAttribute<MemberRequiredAttribute>()) != null)
+                {
+                    if (field.GetValue(component) == null)
+                        WriteResult(result, componentName, targetUrlInStorage, entity.Name, field.Name, memberRequired.ReportAs);
+                }
+            }
+
+            foreach (var prop in properties)
+            {
+                if (prop.PropertyType.IsValueType)
+                    continue; // value types cannot be null, and must always have a proper default value
+
+                MemberRequiredAttribute memberRequired;
+                if ((memberRequired = prop.GetCustomAttribute<MemberRequiredAttribute>()) != null)
+                {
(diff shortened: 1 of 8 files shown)


tebjan:
Wow, nice work there!

sources/engine/Stride.Assets/Entities/ComponentChecks/RequiredMembersCheck.cs
```diff
@@ -0,0 +1,71 @@
...
+            return true; // applies to any component
+        }
+
+        /// <inheritdoc/>
+        public void Check(EntityComponent component, Entity entity, AssetItem assetItem, string targetUrlInStorage, AssetCompilerResult result)
+        {
+            var type = component.GetType();
+            var componentName = type.Name; // QUESTION: Should we check attributes for another name?
+
+            var fields = type.GetFields(); // public fields, only those can be serialized?
```
xen2, line 27: I think we might want to use our existing reflection system to unify behavior with what's done for serialization. That way, it will skip/consider members the same way.
Here's how we can access them:
`TypeDescriptorFactory.Default.Find(typeof(XXX)).Members`

xen2:
Sorry for the delay (vacations).
Looks good to me, thanks for the nice change!
I have added a small comment. Hopefully, it's easy to adjust.

manio143:
Yes, it was indeed easy to adjust. I also had to fix the test where I forgot to add `[DataMember]` for non-public properties.

sources/engine/Stride.Assets/Entities/ComponentChecks/RequiredMembersCheck.cs
```diff
@@ -0,0 +1,59 @@
...
+
+            var members = typeDescriptor.Members;
+
+            foreach (var member in members)
+            {
+                if (member.Type.IsValueType)
+                    continue; // value types cannot be null, and must always have a proper default value
+
+                MemberRequiredAttribute memberRequired;
+                if ((memberRequired = member.GetCustomAttributes<MemberRequiredAttribute>(false).FirstOrDefault()) != null)
```
manio143, line 37: Here I used `inherit: false` - if I understand it correctly, this should be `true` when inspecting overloaded methods/properties and looking at attributes across the inheritance tree, should I change it?

xen2: Yes it should probably be true (in case it's declared on the `virtual` member but not on the `overridden` one)

manio143:
@xen2 Anything else?

xen2:
@manio143 LGTM, thanks!

PR #828, opened 2020-08-17
Use relative mode mouse mode when the mouse is locked

### LINKED ISSUE #345: Input.LockMousePosition exits window if mouse moved too quickly

**Version**: 3.1.0.1-beta01-0406

**Platform(s)**: Windows 10 Pro Build 17134

**Describe the bug**
When `Input.LockMousePosition` is used to lock the mouse cursor to the window, it exits the window despite its setting if the mouse is moved too quickly. This occurs on `forceCenter=true` and `forceCenter=false`.

**To Reproduce**
1. Use Third-Person Sample
2. Launch Game
3. Move mouse quick enough (Using a 6000 DPI Mouse, but not on Max. settings, so it should occur on slower ones, too.)

**Expected behavior**
Mouse-Cursor should never leave the window.

**Additional Information**
- Using a Tripple 2160p Monitor Setup
- Game launched via Xenko Editor AND Visual Studio 2017 Enterprise
- Tested in Window and Fullscreen

Zeroto:
# Use relative mouse input when the mouse is locked

## Description
By locking the cursor into position and using relative mouse input the cursor can't leave the window and accidentally cause input in another window. This fixes #345.

For UWP and SDL the changes are straightforward but for windows we need to use raw input. Seeing that SharpDX.RawInput doesn't have a .net core version and is end of life, I created my own raw input wrapper. In the future the existing usage of SharpDX.RawInput(which is fairly minimal) can be ported to that so that we can remove the dependency on SharpDX.RawInput.

## Related Issue
#345 

## Motivation and Context
This fixes a bug and adds raw input handling to remove the SharpDX.RawInput dependency in the future.

## Types of changes

## Checklist


xen2:
LGTM!
I'll just wait for @Eideren thumbs up and merge it!

Eideren (commented):
Thanks, just a small note

sources/engine/Stride.Input/Windows/RawInput/RawInput.cs
```diff
@@ -0,0 +1,144 @@
...
+
+        public static unsafe byte[] GetHIDRawData(ref RawHid rawHid)
+        {
+            var byteCount = rawHid.dwCount * rawHid.dwSizeHid;
+            var result = new byte[byteCount];
+            fixed (byte* dest = result)
+            {
+                fixed (int* src = &rawHid.bRawData)
+                {
+                    Unsafe.CopyBlockUnaligned(dest, src, byteCount);
```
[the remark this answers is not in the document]
Zeroto: Hmm, that is a weird one. It is most definitely available under netfw. I tested it both under netcore and netfw and it seems to work fine on both. Don't know why the docs say that it isn't available on netfw.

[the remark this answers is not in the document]
Eideren: Built a simple project under ``<TargetFramework>net472</TargetFramework>`` containing that signature
and it returns ``error CS0103: The name 'Unsafe' does not exist in the current context``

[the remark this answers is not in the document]
Zeroto: Huh, weird. Pretty sure it compiled here yesterday, but I might be mistaken. I will check again tonight. Worst case scenario that part of the code can be removed completely because it handles HID input which isn't used. I only added it to completely handle the WM_INPUT message.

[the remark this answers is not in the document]
Zeroto: After a bit of research I think I found the cause. `System.Runtime.CompilerServices.Unsafe` is being included as a transient package by another dependency(don't know which one yet) so that is why that class is available under netfw. I will add it as a direct dependency.

Eideren (approved):
Sorry, didn't see your request, looks good to me ! Your call @xen2 :)

sources/engine/Stride.Input/Windows/RawInput/RawInput.cs
```diff
@@ -0,0 +1,144 @@
...
+        {
+            uint cbSize = 0;
+            Win32.GetRawInputData(lParam, (uint)RawInputDataType.RID_INPUT, IntPtr.Zero, ref cbSize, (uint)sizeof(RawInputHeader));
+            if (cbSize == 0)
+            {
+                return default;
+            }
+            var buffer = stackalloc byte[(int)cbSize];
+            var count = Win32.GetRawInputData(lParam, (uint)RawInputDataType.RID_INPUT, (IntPtr)buffer, ref cbSize, (uint)sizeof(RawInputHeader));
+            return (RawInputData*)buffer;
```
Ethereal77, line 141: Wouldn't this cause problems? Basically you are allocating a buffer on the stack and returning a pointer to that buffer. After exiting the function that buffer is out of scope and can be potentially overwritten by other data.

Zeroto: ooh, yeah. That is a very good point. I had it different before but apparently screwed it up later. Good catch.

Zeroto: This is now fixed :)

sources/engine/Stride.Input/Windows/RawInput/RawInput.cs
```diff
@@ -138,7 +138,7 @@ public static bool RegisterDevice(UsagePage usagePage, UsageId usageId, ModeFlag
             }
             var buffer = stackalloc byte[(int)cbSize];
             var count = Win32.GetRawInputData(lParam, (uint)RawInputDataType.RID_INPUT, (IntPtr)buffer, ref cbSize, (uint)sizeof(RawInputHeader));
-            return (RawInputData*)buffer;
+            return *(RawInputData*)buffer;
```
Eideren, line 141: Nice catch @Ethereal77 
You're still returning something pointing to stackalloc'ed data, that data returns to the stack as soon as it exits the scope so if anything re-allocates on the stack it'll manipulate what's in that pointer. Allocate from the caller and pass in the pointer to that function instead to read it from the caller safely

Eideren (changes requested):

sources/engine/Stride.Input/Windows/RawInput/RawInput.cs
Zeroto: No. Because it is a struct it will make a copy(I checked the memory addresses to see if they were different). However, I just realized this will break for the HID inputs because that is a byte pointer in the provided struct. So I will need to change some more stuff.

Eideren: Ah nevermind, somehow swapped diffs around

Ethereal77:
As all of raw input handling is encapsulated inside Stride.Input maybe it'll be easier to pass the responsibility to the caller. That's the way many of the Win32 APIs are designed. By this I mean that in `PreFilterMessage()` where `GetRawInputData()` is called (or any other appropriate call site) is where you allocate the buffer and pass it to the function. The function then returns the data in the buffer instead of returning a copy of the structure or a pointer to the stack.

Zeroto:
Except that in this case that doesn't really work because the size needed for the buffer is unknown until I call it first with size 0.  This is because the HID structure has a byte array in it, but we can't use a managed byte array for that because of the need for unsafe calls and pointer magic. I either need to move all the logic into the message filter, or add a callback to the getrawinput.

Ethereal77:
Will not be sufficient to just call first (in the message filter) to `GetRawInputData` with data = `null` to retrieve the size of the needed buffer, then `stackalloc` a buffer of that size, then call again `GetRawInputData` with the now correctly-sized buffer and finally convert the `ptr` to a `ref RawInput` or a `RawInput*`?
The message filter will be a little more cluttered, but all the logic will be there and it is not much code I suppose.

What do you think?

Ethereal77:
This is an extract of my message filter:

```
private static void HandleInput(IntPtr hWnd, IntPtr ptrRawInputMessage, IntPtr rawInputMode)
{
    uint dataLength = GetRawInputDataSize(ptrRawInputMessage);
    if (dataLength == 0)
        return;

    Span<byte> inputData = stackalloc byte[(int) dataLength];
    ref RawInputData rawInput = ref GetRawInputData(ptrRawInputMessage, inputData);

    var device = rawInput.Header.Device;
    switch (rawInput.Header.Type)
    {
        case DeviceType.Mouse:
            ...
```

I was casually writing a RawInput library last week when I first saw this PR. It is inspired by SharpDX.RawInput but using .NET Core 3.1 and later .NET 5.

xen2:
LGTM.
I think this is ready for merge, right?

Zeroto:
Yeah. I think so. If @Eideren and @Ethereal77 can do their last check over it then it can be merged

Eideren:
I don't have anything to add, thanks again @Zeroto !

sources/engine/Stride.Input/Windows/RawInput/RawInput.cs
```diff
@@ -0,0 +1,131 @@
...
+        public struct RawInputData
+        {
+            public RawInputHeader header;
+            public RawInputDataInner data;
+        }
+
+        public static unsafe byte[] GetHIDRawData(ref RawHid rawHid)
+        {
+            var byteCount = rawHid.dwCount * rawHid.dwSizeHid;
+            var result = new byte[byteCount];
```
Ethereal77, line 99: This would result in an allocation (although a small one) potentially every frame, or at least everytime a message is received from a HID. An alternative would be to move the ownership of the buffer outside the function.

This is not very important though. Maybe a good point to keep in mind for a further optimization pass.

Ethereal77:
Apart from the comment above, it seems good to me.

sources/engine/Stride.Input/Windows/RawInput/RawInput.cs
xen2: Yes it would be good to fix.
Maybe it can be stackalloc on the caller side? (assuming it can't be too large to avoid stackoverflow)

Zeroto: IMO this can be merged for now. This allocation only happens when a HID device event occurs which is not the case with the current usage. Even in the future it will be rare that it will be used because the HID devices would either be accessed through their own API or through other input libraries such as directinput/xinput.

xen2: Sorry, I thought it was happening all the time.
Totally fine then :)

Issue #835, opened 2020-09-08
Alpha of color behaving strange when MSAA is on


azeno:
**Release Type**: Official Release

**Version**: 4.0.0.1 (latest)

**Platform(s)**: Windows

**Describe the bug**
The alpha of colors behaves rather weired if MSAA is turned on. In my example I have a pure emissive material where the alpha channel of its white color gets increased over time. One would expect a blending with the background in a linear fashion, but when turning on MSAA one gets a kind of step function based on the amount of MSAA levels.

**To Reproduce**
See attached project.
[MyGame5.zip](https://github.com/stride3d/stride/files/5190477/MyGame5.zip)

**Expected behavior**
MSAA should not affect alpha blending.

Does the blend state of the MSAAResolver need to be different? Or is it the shader itself - maybe it shouldn't take the alpha values into account at all - at least the referenced original shader doesn't, only loading rgb (https://github.com/TheRealMJP/MSAAFilter/blob/master/MSAAFilter/Resolve.hlsl#L378) and returning alpha = 1 for the final color value.

tebjan:
Close this now, since #836 is merged?

xen2:
Fixed by #836

PR #836, opened 2020-09-09
[Rendering] Disabled alpha-to-coverage for Blend and Additive transparency

azeno:
# PR Details

## Description

Added a flag to the material pass to explicitly control the enabled state of the alpha-to-coverage multisample technique. It needs to be disabled for the Blend and Additive transparency features as it uses the alpha channel for alpha testing and not blending.

## Related Issue

#835 

## Motivation and Context

Alpha blending works like expected when MSAA is on.

## Types of changes

## Checklist


tebjan:
Glad we found this one. But how about allowing the user to set it explicitly via a checkbox on the material, instead of doing an automagic. With proper defaults, as the code is doing it now.

xen2:
@tebjan that might be a great addition, but I will be merging that version for now.

PR #841, opened 2020-09-11
CameraComponent now can be updated by ignorant third party...

gregsn:
... without any harm.

Calling `Update(null)` had the effect that the Custom Aspect Ratio stored in property `AspectRatio` was used, even if `UseCustomAspectRatio` was turned off.

Calling `Update(null)` now will just use the last fed Aspect Ratio.

A helper visualizing the frustum can now just call `Update(null)` without any side effect and visualize the actual frustum with the actually used aspect ratio.

# PR Details

The change shouldn't be problematic, because the `CameraProcessor` was checking 
for the flag `UseCustomAspectRatio` and only called `Update(null)` when true. The behavior in this case will be the same.

## Description

Added a cache for the last used Aspect Ratio.

## Motivation and Context

We now can Update a currently unused Camera and visualize it from another perspective without falling back to the custom aspect ratio that actually isn't used by the `SceneCameraRenderer` when `UseCustomAspectRatio` is turned off (which is the default). 

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Engine/Engine/CameraComponent.cs
+++ b/sources/engine/Stride.Engine/Engine/CameraComponent.cs
@@ -218,8 +218,7 @@ namespace Stride.Engine
             // A helper visualizing the state of the camera will not change it's state.
             if (UseCustomAspectRatio)
                 ActuallyUsedAspectRatio = AspectRatio;
-            else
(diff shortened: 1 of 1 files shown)


xen2:
LGTM, thanks!

gregsn:
thank you!

PR #842, opened 2020-09-14
[Engine] Small fix for time warp

Eideren:
# PR Details
I mentioned on the previous PR regarding this topic (#785) that I would look at merging UpdateTime and DrawTime together into a single 'Time' class, it's actually not a good idea to do so for a whole lot of reasons.
So instead of that I'll just fix DrawTime to UpdateTime's factor, see #785 for why that's important. Also, factor should only affect elapsed time after the frame/update is processed otherwise you might have 'one half of the frame' in real time and the other in slow mo when that variable is modified in the middle of its execution.

## Description
See above

## Related Issue
None

## Motivation and Context
See PR details

## Types of changes

## Checklist


Eideren:
I'll just merge this in

PR #843, opened 2020-09-14
Explicitly pass database to effect compiler

azeno:
# PR Details

## Description

The effect compiler uses the virtual file provider to lookup shader source code and the database to cache compiled shaders. Allowing to pass both services explicitly (and not assuming they're implemented by the same instance) one can highly customize the shader source code lookup without affecting the caching mechanism.

## Motivation and Context

We need to be able to load shaders at runtime. These changes allow us to use a custom file provider for shader source code lookup.

## Types of changes

## Checklist


xen2:
Great PR, thanks!

PR #845, opened 2020-09-15
[WIP] Workaround in ScrollingText to avoid endless loop

CharlesWoodhill:
# PR Details

Workaround in ScrollingText to avoid endless loop (when measuring text)

## Description
added bailout for while loop when measuredText = 0

## Motivation and Context

gameStudio hangup when using scrollingtext

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.UI/Controls/ScrollingText.cs
+++ b/sources/engine/Stride.UI/Controls/ScrollingText.cs
@@ -168,6 +168,11 @@ namespace Stride.UI.Controls
         {
             if (string.IsNullOrEmpty(Text))
                 return;
+            // todo: better handling of while loop below,
+            // if for any reason CalculateTextToDisplayWidth() returns 0, its endlessLoop
+            // making entire gamestudio unresponsible
+            if (Font is null)
+                return;
 
             var elapsedSeconds = time != null ? (float)time.Elapsed.TotalSeconds : 0f;
 
@@ -190,6 +195,10 @@ namespace Stride.UI.Controls
                     nextLetterIndex = 0;
 
                 textToDisplayWidth += addedCharacterWidth;
+
+                // workaround to avoid endless loop
+                if (addedCharacterWidth <= 0.0f)
+                    break;
             }
 
             // Check if all the string has finished to scroll, if clear the message


sources/engine/Stride.UI/Controls/ScrollingText.cs
```diff
@@ -190,6 +195,10 @@ private void UpdateAndAdjustDisplayText(GameTime time = null)
                     nextLetterIndex = 0;
 
                 textToDisplayWidth += addedCharacterWidth;
+
+                // workaround to avoid endless loop
+                if (addedCharacterWidth <= 0.0f)
+                    break;
```
Eideren, line 201: So, from what I understand, this component displays an 'infinitely' repeating text horizontally. The issue happens when the whole text has a width of less than or equal to zero. It fills the component with characters until it reaches the max width but since the text's width is effectively zero it can't ever finish that loop.

If I did understand that correctly, a cleaner solution would be to just take the width of the whole text that would be included and if that width is below an epsilon avoid the whole loop/drawing the text, this can be done through ``CalculateTextSize(YourText).X``.
As, right now, from what I understood of your workaround, it seems to bail out as soon as *just one* of the characters has a zero width ? I think there are a couple of languages where zero-width characters are useful, not sure how applicable it is to our engine but might as well account for it ( [Zero-width_joiner](https://en.wikipedia.org/wiki/Zero-width_joiner) and [Zero-width_non-joiner](https://en.wikipedia.org/wiki/Zero-width_non-joiner) perhaps ? )

Eideren (commented):
Thanks !

Unrelated comment: Looking at this component, it's kind of a shame that it just uses string internally instead of a StringBuilder, it's as easy since we already support that type when computing font text sizes and would greatly really reduce allocs which might increase ui performances.

CharlesWoodhill:
yeah sorry for that, i currently had other problems, its just a workarounding quick fix dirty solution, thats why i labeled it with that,
didnt have the time to make it better... 
the problem i had wasnt with text length, this is catched at start of function ( if text.isnullorempty)
the problem i had was that i somehow managed to have no font, and without font it also cant calculate it.
but i couldnt set a font because as soon as i open the page => endless loop :)

CharlesWoodhill:
maybe leave it open, when i have time i try to make it better

Zeroto:
> the problem i had wasnt with text length, this is catched at start of function ( if text.isnullorempty)

I don't think @Eideren was talking about the length of the text, but of the width.

`if (addedCharacterWidth <= 0.0f)`
This is not a good solution because there are valid zero width characters in the unicode codeset. Most likely the proper solution would be to check at the repeat to see if the text before the repeat is zero width or not.

CharlesWoodhill:
jep, i understood, when i have time i will make it better.

Eideren:
Continuing through #846, closing this one off

Issue #663, opened 2020-04-22
Folders added in VS don't add in Stride


bdphil86:
**Release Type**: Official Release/GitHub (please choose appropriate option)

Official Release

**Version**: Version number and/or git branch

3.1.0.1

**Platform(s)**: Does the problem occur on Windows, Android...?

Windows

**Describe the bug**

When new folders are added in VS, they don't get added in Stride. This leads to disorganization within Stride. All of the files are created and added to the Assets folder, but they are just there disorganized.

**To Reproduce**
1. add new folders and files in VS.
2. Open solution in Stride.

The new files exist but not the folder structures.

**Expected behavior**

The new folders AND files should exist in VS to maintain organization. I shouldn't have to recreate the folders in both tools.

**Screenshots**
If applicable, add screenshots to help explain your problem (use GitHub drag & drop)

**Log and callstacks**
If any, please attach here any log or callstack (preferably in a `.txt` file using GitHub drag and drop)

**Additional context**

manio143:
Ok, so I looked into it. Here's my observations:

- directories show up in the game studio only if they are not empty
    That's because directories are added into the viewmodel by enumerating assets and checking their URLs.
- code has a separate viewmodel (`ProjectCodeViewModel` and `CodeViewModel`) from other assets (`AssetMountpointViewModel` and `AssetViewModel`)
- subfolders of "Assets" are showing up correctly, whereas subfolders for code are skipping one level (for unknown reasons)
    So if I have code in "subfolder" it will show up straight in "Code", but code in "subfolder/puppy" will show up in "Code/puppy".

I'm going to submit a PR for that last one as it is indeed making it hard to find the right class among all scripts in first level subfolders.

Adding non-code assets to new subfolders requires "Reload project" in GameStudio for them to be loaded correctly.

@bdphil86 Could you confirm the behaviour I outlined above? Is my PR for code assets the solution to this bug report?

manio143:
@Eideren I think this can be closed.

PR #848, opened 2020-09-17
[Editor] Don't skip first level directories for code assets

manio143:
# PR Details

## Description

If you create folders to organize your code, all scripts are visible on the same level in the "Code" folder in GameStudio. That is because the first level folders are skipped in the viewmodel. This PR removes that skip, which means all this organized code is organized in GameStudio as well.

## Related Issue
#663 (I've written more details in the comment there)

## Types of changes

## Checklist


manio143:
I just noticed that the CSharpBegginer tutorial has a `Code` folder in its game project and likely this is the reason why the skip was introduced. However, this is not an expected practice in .NET projects. What we can do is skip the first level only if it matches the name `Code`.

Eideren:
I seem to remember that issue being there since before those tutorials were introduced, perhaps older projects had a similar setup to that one. I don't think it makes much sense to have an exception just for that anyway.

Looks good to me, I'll let xen chime in just in case he has more context on this.

Issue #851, opened 2020-09-19
Memory leak when no instancing is used


tebjan:
**Release Type**: Official Release/GitHub (please choose appropriate option)

Release

**Version**: Version number and/or git branch

4.0.0.1-beta04-1265

**Platform(s)**: Does the problem occur on Windows, Android...?

Windows

**Describe the bug**
Running the third-person platformer template shows a memory leak. see: 
https://forums.stride3d.net/t/inefficient-memory-management-garbage-collection/2210

**To Reproduce**
1. Open Third-Person platformer and run

**Expected behavior**
No allocations, no matter whether instancing is used or not.

**Screenshots**
If applicable, add screenshots to help explain your problem (use GitHub drag & drop)

![image](https://user-images.githubusercontent.com/1094716/93666134-dd5cf680-fa7b-11ea-8e45-5e142e5167ad.png)
![image](https://user-images.githubusercontent.com/1094716/93666111-bdc5ce00-fa7b-11ea-9cbb-8fdb96202ff2.png)

**Log and callstacks**
```
Stride.Core!Stride.Core.Collections.FastListStruct<Stride.Rendering.EffectValidator.EffectParameterEntry>.EnsureCapacity( int )
Stride.Core!Stride.Core.Collections.FastListStruct<Stride.Rendering.EffectValidator.EffectParameterEntry>.Add( EffectParameterEntry )
Stride.Rendering!Stride.Rendering.EffectValidator.ValidateParameter( PermutationParameterKey<int>,int )
Stride.Rendering!Stride.Rendering.InstancingRenderFeature.PrepareEffectPermutations().Closure_0.<PrepareEffectPermutations>b__0( RenderObject )
Stride.Core!Stride.Core.Threading.Dispatcher.ForEach().Closure_0<T>.<ForEach>b__0( int )
Stride.Core!Stride.Core.Threading.Dispatcher.ExecuteBatch( int,int,Action<int> )
Stride.Core!Stride.Core.Threading.Dispatcher.Fork( int,int,int,Action<int>,BatchState )
Stride.Core!Stride.Core.Threading.Dispatcher.Fork().Closure_0.<Fork>b__0()
Stride.Core!Stride.Core.Threading.ThreadPool.TryCooperate()
Stride.Core!Stride.Core.Threading.ThreadPool.WorkerThreadScope()
mscorlib!System.Threading.ExecutionContext.RunInternal( ExecutionContext,ContextCallback,object,bool )
[...]

PR #854, opened 2020-09-19
[Rendering] Instancing: Skip effect validation

tebjan:
If instancing count <= 0

# PR Details

Memory leak bug fix.

## Description

The effect validation doesn't run if instancing count is <= 0. This fixes the bug, but I am not sure whether this is the correct way of doing it.

Questions:

* Does this work if an effect has instancing and then the instancing component is removed?
* Is there a way to not even create the `InstancingData` and also don't run the `InstancingRenderFeature` for a mesh, if no `InstancingComponent` is attached? Here is also the question of whether it works if an effect has instancing and in the next frame not anymore.

## Related Issue

#851 

## Motivation and Context

Reported memory leak

## Types of changes


tebjan:
This works and is good enough for as a hot fix, I think.

PR #858, opened 2020-09-23
Remove <RestorePackages> from project files to improve build time

manio143:
# PR Details

## Description

This PR removes `<RestorePackages>true</RestorePackages>` from all project files. I believe it is the reason why the projects are rebuilt everytime - Stride.Core.Tasks among others did a force restore which touched its `project.assets.json` and forced all dependants to recompile as well.

The removed tag is from an older version of MSBuild that didn't do automatic NuGet restores (see [link](https://docs.microsoft.com/pl-pl/nuget/consume-packages/package-restore)).

**Please verify that my observations are reproduced.**

## Related Issue

N/A (we talked about it on Discord ([msg](https://discordapp.com/channels/500285081265635328/500290856532574209/743858738561155093)))

## Motivation and Context

Well, after the change I don't have to wait 3 minutes to start GameStudio (event without any changes), but more like 20 seconds.

## Types of changes

## Checklist


manio143:
Retested, seems there's no difference to previous behavior 😢
Doing 'Clean Solution' and then building twice - the second time is fast but after reloading the solution we're back with the slow builds.

PR #860, opened 2020-09-23
Fix AsyncScript priority handling

### LINKED ISSUE #859: AsyncScript priority isn't respected

**Version**: 1264 (since initial commit)

**Platform(s)**: N/A

**Describe the bug**
Setting Priority on an AsyncScript has no practical effect. The scheduler will get the Execute method with priority 0.

**To Reproduce**
1. Create a script like this:
 
        using System.Threading.Tasks;
        using Stride.Engine;
        using Stride.Core.Diagnostics;

        namespace AsyncPriorityTest
        {
            public class PrintPriority : AsyncScript
            {
                private static Logger logger = GlobalLogger.GetLogger(nameof(PrintPriority));
                public override async Task Execute()
                {
                    logger.Info($"Priority: {Priority}");
                }
            }
        }

2. Create two entities with that script
3. Give the first one a higher value for `Priority`
4. See higher priority printed first

**Expected behavior**
Lower priority should be printed first.

**Additional context**
This is due to an incorrect binary operation in `ScriptSystem` (an `&` instead of `|`).

manio143:
# PR Details

## Description

Fix binary operation for setting `AsyncScript`'s priority in the scheduler.

## Related Issue
Fixes #859 

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Engine/Engine/Processors/ScriptSystem.cs
+++ b/sources/engine/Stride.Engine/Engine/Processors/ScriptSystem.cs
@@ -85,7 +85,7 @@ namespace Stride.Engine.Processors
                     var asyncScript = script as AsyncScript;
                     if (asyncScript != null)
                     {
-                        asyncScript.MicroThread = AddTask(asyncScript.Execute, asyncScript.Priority & UpdateBit);
+                        asyncScript.MicroThread = AddTask(asyncScript.Execute, asyncScript.Priority | UpdateBit);
                         asyncScript.MicroThread.ProfilingKey = asyncScript.ProfilingKey;
                     }
                 }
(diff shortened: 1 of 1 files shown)


sources/engine/Stride.Engine/Engine/Processors/ScriptSystem.cs
```diff
@@ -85,7 +85,7 @@ public override void Update(GameTime gameTime)
                     var asyncScript = script as AsyncScript;
                     if (asyncScript != null)
                     {
-                        asyncScript.MicroThread = AddTask(asyncScript.Execute, asyncScript.Priority & UpdateBit);
+                        asyncScript.MicroThread = AddTask(asyncScript.Execute, asyncScript.Priority | UpdateBit);
```
xen2, line 88: There's another case line 174.
I will fix it and merge

PR #861, opened 2020-09-23
Couple of fixes to the shader compiler

azeno:
# PR Details

## Description

This PR addresses two issues we ran into when loading a broken shader:
1) It was impossible to reload it due to some internal dependency not being tracked and therefor a cache not being invalidated by ResetCache calls.
2) The compiler result was not deterministic when invoked a second time - a mixin was marked with error flag but the result log had no error which lead to subsequent code paths crashing as they only checked the result log.

## Related Issue

https://github.com/vvvv/VL.Stride/issues/221

## Types of changes

## Checklist


azeno:
Sorry, ran into an issue, will need to do another one.

PR #871, opened 2020-10-07
[Rendering] Instancing: fix instance back to scene

Eideren:
# PR Details
Small fix to keep state valid for an instance that was removed from the scene and added right back. (previously didn't end up rendering the instance after this operation)

## Related Issue
None reported

## Motivation and Context
Fix bug

## Types of changes

## Checklist

@tebjan note the modifier changes from ``public``->``internal``, seemed more fitting, was there a specific reason to keep it public ? 
On that note, ``OnEntityComponentAdding`` calling ``FindMasterInParents`` seems to me like unexpected or unwanted behavior, should we keep it ?

### DIFF
--- a/sources/engine/Stride.Engine/Engine/InstanceComponent.cs
+++ b/sources/engine/Stride.Engine/Engine/InstanceComponent.cs
@@ -58,16 +58,16 @@ namespace Stride.Engine
             ConnectInstancing();
         }
 
-        private void ConnectInstancing()
+        internal void ConnectInstancing()
         {
-            if (master != null && master.Type is InstancingEntityTransform instancing)
+            if (connectedInstancing == null && master != null && master.Type is InstancingEntityTransform instancing)
             {
                 instancing.AddInstance(this);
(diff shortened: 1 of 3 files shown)


Eideren:
Included another small fix but that's it, waiting on tebjan to look at this :)

tebjan:
> the modifier changes from `public`->`internal`, seemed more fitting, was there a specific reason to keep it public ?

that looks correct, I think it was just an oversight.



> On that note, `OnEntityComponentAdding` calling `FindMasterInParents` seems to me like unexpected or unwanted behavior, should we keep it ?

this is a wanted feature that automatically connects to the next InstancingComponent of the parent. i would even recommend to use this as the default pattern to create instances:

ParentEntity
 - Child1
 - Child2
 - ...

if you build it like that, you don't have to connect every instance with the parent component. it also makes it easier to create instance dynamically, because you don't have to connect then either. so it's a convenience feature...

Eideren:
Sure, I'll leave it as is then

PR #878, opened 2020-10-14
[Core] Customize minimum log level for GlobalLogger by introducing AppSettings

### LINKED ISSUE #876: Allow setting minimum log level for Logger (currently hardcoded)
Currently the minimum log level has to be set explicitly for every instance of `GlobalLogger` by calling inherited `ActivateLog`. There is a method that allows calling `ActivateLog` for every registered `GlobalLogger` (with overloads for module name filtering), however it cannot be used to setup a default log level for new instances and would have to be rerun periodically.

Provide a mechanism for setting the default minimum log level
a) from code,
b) by automatically loading a settings file.
If the user sets the minimum log level in code we should ignore the settings file. If there's no file provided, then a hardcoded default value is used, but still it's possible for the user to change it.

Moreover, add a secondary constructor to `Logger` and `GlobalLogger.GetLogger` that takes a minimum level and sets it for this particular instance. This is so that the user may test a single logger with a more verbose setting while having the default set higher.

**Additional context**
I'm going to implement this feature. The implementation for the second part (secondary constructors) is simple. For the first part, to support loading `.config` files I would have to add a dependency on the [System.Configuration.ConfigurationManager](https://www.nuget.org/packages/System.Configuration.ConfigurationManager/) package for `Stride.Core`. But this isn't a great solution - we want to keep Core small. A better option would be to implement a small configuration reader from what is already in Stride - I will outline a proposed design in the comment below.

manio143:
# PR Details

## Description

This PR introduces a new way to define application level settings and uses it to allow user to override the minimum log level, which is currently hardcoded to Info.

* `Stride.Core.Settings.AppSettings` is the settings container - a collection of objects which should differ by type (otherwise the first object of the given type is returned).
* `Stride.Core.Settings.IAppSettingsProvider` is an interface that abstracts how we obtain the AppSettings. This is because we don't want to add any dependencies to `Stride.Core`.
* `Stride.Core.Settings.AppSettingsManager` is a static class that uses `AssemblyRegistry` to find a type which implements the above interface, loads the settings and caches them.
* `Stride.Core.Settings.AppSettingsProvider` (in `Stride.Core.Design`) provides a way to load Yaml based `.appsettings` file. Note: `.settings` is special for Visual Studio, so couldn't be used.
* `Stride.Core.Diagnostics.Logger` has been modified to try loading the user provided override from `AppSettings` to determine the `MinimumLevelEnabled`. Moreover, the property can be overridden from code.
* `Stride.Core.Diagnostics.GlobalLogger` now uses the above property when creating new loggers. It also got an overload which can specify the minimum log level per instance (when created).
* `AssemblyRegistry` got a static constructor in which we register `Stride.Core` assembly - this is so that we can use the default `YamlSerializer` which only allows deserializing registered types. Moreover, the static logger is initialized lazily because otherwise there's a static initializer dependency cycle.
* I also included a small change to the `YamlSerializer` making `Load<T>` static as it just called a static field.

### Usage
To override the default log level create a file called "_YourExecutable_.appsettings" and mark it in VS to copy to output directory, or include the following in your project file:

```xml
  <ItemGroup>
    <None Update="MyProject.appsettings">
      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
    </None>
  </ItemGroup>
```

Edit the file contents to look like this:

```yaml
!AppSettings
Settings:
    - !GlobalLoggerConfig
        Level: Debug
```
You can set the log level to one of those: `Debug`, `Verbose`, `Info`, `Warning`, `Error`, `Fatal` (but usually only the first three make sense).

## Related Issue

Closes #876 
I haven't waited long for comments under my issue, so I'll gladly take any criticism here.

## Motivation and Context

When creating `static Logger Logger = GlobalLogger.GetLogger(nameof(Class));` for many classes I want to be able to change one setting to have them all logging more or less information. I could not use the `GlobalLogger.ActivateLog` because I would have to call it after my loggers are created and that's not easy.

## Types of changes

**Note:** This is a binary interface breaking change, but a non-breaking change with regard to usage and behavior.

## Checklist

### DIFF
--- a/sources/core/Stride.Core/Settings/AppSettingsManager.cs
+++ b/sources/core/Stride.Core/Settings/AppSettingsManager.cs
@@ -22,17 +22,21 @@ namespace Stride.Core.Settings
                     return settings;
 
                 foreach (var assembly in AssemblyRegistry.FindAll())
-                    foreach (var type in assembly.GetTypes())
-                        if (typeof(IAppSettingsProvider).IsAssignableFrom(type))
-                        {
-                            var ctor = type.GetConstructor(new Type[0]);
-                            if (ctor != null)
+                {
+                    var scanTypes = AssemblyRegistry.GetScanTypes(assembly);
+                    if (scanTypes != null &&
+                        scanTypes.Types.TryGetValue(typeof(IAppSettingsProvider), out var providerTypes))
+                    {
+                        foreach (var type in providerTypes)
+                            if (!type.IsAbstract &&
+                                type.GetConstructor(Type.EmptyTypes) != null)
                             {
-                                var instance = (IAppSettingsProvider)ctor.Invoke(new object[0]);
+                                var instance = (IAppSettingsProvider)Activator.CreateInstance(type);
                                 settings = instance.LoadAppSettings();
                                 return settings;
                             }
-                        }
+                    }
+                }
 
                 settings = new AppSettings();
                 return settings;
--- a/sources/core/Stride.Core/Settings/AppSettingsManager.cs
+++ b/sources/core/Stride.Core/Settings/AppSettingsManager.cs
@@ -1,0 +1,44 @@
+using System;
+using Stride.Core.Reflection;
+
+namespace Stride.Core.Settings
+{
+    /// <summary>
+    /// Manages the loading of application settings with <see cref="IAppSettingsProvider"/>.
+    /// </summary>
+    public static class AppSettingsManager
+    {
+        /// <summary>
+        /// <see cref="AppSettings"/> instance for the application.
+        /// </summary>
+        /// <note>
+        /// Loaded with an <see cref="IAppSettingsProvider"/>. If no provider implementation is found, returns empty <see cref="AppSettings"/> instance.
+        /// </note>
+        public static AppSettings Settings
+        {
+            get
+            {
+                if (settings != null)
+                    return settings;
+
+                foreach (var assembly in AssemblyRegistry.FindAll())
+                    foreach (var type in assembly.GetTypes())
+                        if (typeof(IAppSettingsProvider).IsAssignableFrom(type))
+                        {
+                            var ctor = type.GetConstructor(new Type[0]);
+                            if (ctor != null)
+                            {
+                                var instance = (IAppSettingsProvider)ctor.Invoke(new object[0]);
+                                settings = instance.LoadAppSettings();
+                                return settings;
+                            }
+                        }
+
+                settings = new AppSettings();
+                return settings;
+            }
+        }
+
+        private static AppSettings settings;
+    }
+}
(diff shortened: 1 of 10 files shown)


sources/core/Stride.Core/Settings/AppSettingsManager.cs
```diff
@@ -0,0 +1,44 @@
...
+        /// </note>
+        public static AppSettings Settings
+        {
+            get
+            {
+                if (settings != null)
+                    return settings;
+
+                foreach (var assembly in AssemblyRegistry.FindAll())
+                    foreach (var type in assembly.GetTypes())
```
xen2, line 25: Listing all types in all assemblies is something we are trying to avoid in the runtime (as opposed to design tools).
Reason is it forces lot of initialization (JIT, cctor, etc.). I remember some cases were it was throwing exceptions on some platforms (maybe not if using `FindAll()`, but at least when listing all referenced assemblies).

The recommended method is to use the assembly processor to list and register those types inheriting from `IAppSettingsProvider`, that way the work is done once at compile time only.
In order to do that, please add `[AssemblyScan]` to `IAppSettingsProvider`. Then there's some logic in `AssemblyRegistry.RegisterEngineAssembly` to register in specific collections (but ideally this should be external for easier extension later)

sources/core/Stride.Core/Settings/AppSettingsManager.cs
```diff
@@ -0,0 +1,44 @@
...
+                        if (typeof(IAppSettingsProvider).IsAssignableFrom(type))
+                        {
+                            var ctor = type.GetConstructor(new Type[0]);
```
xen2, line 28: Maybe using `Activator.CreateInstance` is more explicit, and possibly even faster (this point doesn't matter much if called once)?

manio143: Looked up online - `Activator.CreateInstance` can be faster for the default constructor. I used this here because iterating over all types also gave the interface itself (which I could have eliminated another way). I'll change it with the fix for the above.

sources/core/Stride.Core/Settings/AppSettingsManager.cs
manio143: Thanks for the info! I didn't know about `AssemblyScan` and I will definitely also use it in my project. Pushed a fix.

manio143:
The game engine does not reference `Stride.Core.Design` by default, which means it either has to be referenced by the user, or maybe I should move the `AppSettingsProvider` to another assembly?

xen2:
@manio143 the whole Yaml concept doesn't exist at typical runtime either.
It could always be moved to a specific assembly later but it doesn't need to be part of the core runtime due to all the dependencies it needs (yaml and such).

PR #883, opened 2020-10-21
[SDL] Drop files and clipboard

Eideren:
# PR Details
Provides delegate subscription to listen for dropped files and present clipboard IO through SDL Application

## Description
See above.

## Related Issue
None

## Motivation and Context
Those things are useful

## Types of changes

## Checklist

If @tebjan could take a quick look at this since you have worked around this namespace before, do we want the clipboard to sit within Application.cs ?

Note that I haven't implemented the ``SDL_DROPTEXT`` variant as it doesn't look like it's supported on windows.

### DIFF
--- a/sources/engine/Stride.Graphics/SDL/Application.cs
+++ b/sources/engine/Stride.Graphics/SDL/Application.cs
@@ -190,6 +190,11 @@ namespace Stride.Graphics.SDL
                     // Send these events to all the windows
                     Windows.ForEach(x => x.ProcessEvent(e));
                     break;
+                
+                case SDL.SDL_EventType.SDL_DROPTEXT:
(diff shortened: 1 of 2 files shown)


xen2:
@tebjan any remark? otherwise, I will merge it.

tebjan:
looks good, it is confined to SDL implementations. one could think about adding this to GameWindow. but that can be done in another step.

tebjan:
just checked the SDL docs, they say:

> If these events are enabled you must free the filename in the events using SDL_free().

@Eideren did you check whether this is done somewhere? or whether this is even relevant in our managed context?

Eideren:
Yes, passing ``true`` to ``UTF8_ToManaged`` frees it

Issue #477, opened 2019-06-24
PBR issues


nemesis567:
**Release Type**: Official Release/GitHub (please choose appropriate option)

**Version**: 3.1.0.1-beta2-0615

**Platform(s)**: Windows

**PROJECT LINK for DEBUG**: http://198.23.255.26/test.rar
**Describe the bug**
1. Cannot use normal maps alongside displacement maps at the domain stage, the material disappears.
2. Displacement maps, after tesselation with edge fixing so the object doesn't split, mess up the textures in the uv set.
3. Displacement maps after tesselation, make faces disappear from the camera view as it gets closer/further away to the edges of the rendered screen.
**To Reproduce**

1.  Open the project, check the different objects seen in the screenshots, you'll see the issues I mention.

**ScreenShots**
Tesselation+Displacement working fine except for the face issue, mentioned in #3. It's without normal, if I add normal I get across issue 1.
![WOWOWOW](https://user-images.githubusercontent.com/5153909/60042821-815bc480-96b6-11e9-81c4-aaf306b12e47.png)

Issue #2, the texture is messed up, if I remove displacement and tesselation, the texture looks ok.
![facefixing](https://user-images.githubusercontent.com/5153909/60042822-815bc480-96b6-11e9-9170-147d34f1baa9.png)

#3, with camera away it seems ok.
![FullWithCameraAway](https://user-images.githubusercontent.com/5153909/60042823-81f45b00-96b6-11e9-8c72-97fe3f7953fc.png)

#3, with camera close by faces are removed.
![cutasclose](https://user-images.githubusercontent.com/5153909/60042996-f29b7780-96b6-11e9-9111-989b07596547.png)

With vertex stage tesselation, I can add the normal map, but the object breaks apart.
![wall0](https://user-images.githubusercontent.com/5153909/60042826-81f45b00-96b6-11e9-8a13-a06f40736b4b.png)

![rock0](https://user-images.githubusercontent.com/5153909/60042825-81f45b00-96b6-11e9-883c-aa7e54f0e4aa.png)

But if I set it to domain stage #1 happens.
![wall1](https://user-images.githubusercontent.com/5153909/60042827-828cf180-96b6-11e9-8a09-2b047bf3db98.png)

Kryptos-FR:
Consider not attaching files from an external IP address, it is unsafe. I personally would never click on such links.

nemesis567:
edited

johang88:
[...]

PR #888, opened 2020-10-22
[#477] normalWS exposed when tessellation and normal mapping is used

johang88:
# PR Details

Tessellation (with displacment) requires normalWS to be available, this is only the case if normal mapping is not used. This change exposes normalWS in the various normal mapping scenarios. 

The `GenerateNormal_VS` implementation from `NormalFromMesh` is used in all cases.

## Related Issue

#477 This solves one of the reported issues

## Types of changes

## Checklist


sources/engine/Stride.Rendering/Rendering/Core/NormalFromNormalMappingTessellation.sdsl
```diff
@@ -0,0 +1,14 @@
...
+// Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
+/// <summary>
+/// Calculates the normal  the normals from a normal map.
+/// </summary>
+shader NormalFromNormalMappingTessellation : NormalFromNormalMapping
+{
+    override stage void GenerateNormal_VS()
+    {
+        // Perform normal generation at the end in case meshNormal is modified
+        streams.meshNormalWS = mul(streams.meshNormal, (float3x3)WorldInverseTranspose);    // TODO: PERFORMANCE: Normalization required?
```
Eideren, line 11: Is that todo still valid ?

sources/engine/Stride.Rendering/Rendering/Core/NormalFromNormalMappingTessellationInstanced.sdsl
```diff
@@ -0,0 +1,14 @@
...
+// Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
+/// <summary>
+/// Calculates the normal  the normals from a normal map.
+/// </summary>
+shader NormalFromNormalMappingTessellationInstanced : NormalFromNormalMappingInstanced
+{
+    override stage void GenerateNormal_VS()
+    {
+        // Perform normal generation at the end in case meshNormal is modified
+        streams.meshNormalWS = mul(streams.meshNormal, (float3x3)WorldInverseTranspose);    // TODO: PERFORMANCE: Normalization required?
```
Eideren, line 11: It's here as well

sources/engine/Stride.Rendering/Rendering/Skinning/NormalVSSkinningNormalMappingTessellation.sdsl
```diff
@@ -0,0 +1,14 @@
...
+// Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
+/// <summary>
+/// Calculates the normal  the normals from a normal map.
+/// </summary>
+shader NormalVSSkinningNormalMappingTessellation : NormalVSSkinningNormalMapping
+{
+    override stage void GenerateNormal_VS()
+    {
+        // Perform normal generation at the end in case meshNormal is modified
+        streams.meshNormalWS = mul(streams.meshNormal, (float3x3)WorldInverseTranspose);    // TODO: PERFORMANCE: Normalization required?
```
Eideren, line 11: yep

Eideren (commented):
Looks pretty good, just one small thing to note

sources/engine/Stride.Rendering/Rendering/Core/NormalFromNormalMappingTessellation.sdsl
johang88: Not sure if the todo is valid, its present in `NormalFromMesh` which is why I kept it.

Eideren:
@johang88 Give me a heads up when you're ready to merge this in :)

johang88:
WIll do, found some issues with instancing (it crashes) that I have to sort out first :)

johang88:
There we go, instanced tessellation works without crashing now and looks correct. Had to change the `GetInstanceWorld` and `GetInstanceWorldInverse` methods as I was unable to find a way to get the shader compiler to generate distinct versions for each shader stage otherwise, which resulted in it generating a single function taking stream inputs from both VS and PS stages.

If this is possible to do somehow then we can change it back and get slightly cleaner shader code :)

Eideren:
No clue, what do you think @tebjan ?

tebjan:
@johang88 that looks correct. 

In general, I wonder why the effect file needs all this if/else constructs. my feeling is that this could be solved more "general" with permutation keys. @xen2 what do you think? my concern is, that for the next feature there would be another if/else, which kind of defies the purpose of dynamic composition...

Anyway, this doesn't influence the functionality, at the moment. So is there a test project somewhere that demonstrates the working features?

johang88:
I agree, it' getting a bit unwieldy, especially in this case where it does not really add much. Some kind of inter-stage dependency analysis would work quite well in this case, ie always calculate `normalWS` in the vertex stage, the optimizer then removes it if not in use in subsequent shader stages.

johang88:
Test project available here here https://github.com/johang88/Stride-Tessellation

tebjan:
@johang88 I had some time to test the changes. it does indeed fix the cobblestone material used in your test project.

then I tried it with the instancing test project and added tesselation and displacement to a material but it didn't work there. as far as I remember, it wasn't working before your PR either, but I think I'll let you know, maybe the problem is related:
image

johang88:
@tebjan I will check it out, not sure if I tested skinned instanced meshes so there might be an issue there. You are correct in that it did not work before but would be nice to fix all edge cases before merging.

tebjan:
right, this could be related to skinning. if you find the time, it would be great to see a fix there, but this could also be another PR. so from my side, this PR is approved.

johang88:
I did some test with the instancing project and it seems to work fine. So I think we are ready to merge.

image

tebjan:
indeed, it works correctly here too with the latest beta build. great, thanks for the PR!

PR #892, opened 2020-10-25
[Physics] Send CollisionEnded on entity/component removal

### LINKED ISSUE #788: [Physics] Removing an entity will not signal CollisionEnd()
**Release Type**: GitHub

**Version**: master

**Platform(s)**: Any

**Describe the bug**
I have two entities colliding and one of them has a script that does `await physics.CollisionEnd()`. I remove the other entity by setting `otherEntity.Scene = null`. The script never continues.

**To Reproduce**
1. Create two entities in the same spot
2. Add a rigidbody to each one, such that they collide.
3. Create an async script that awaits `CollisionEnd()` and then prints to debug log.
4. Create another script on the second entity that waits a second and then removes the entity

**Expected behavior**
Removing an entity should signal the collision end.

**Additional context**
Why is this happening (or rather not happening)?
As far as I can tell the situation looks as follows:

1. Entity is removed, so its physics component is marked for removal.
2. `Bullet2PhysicsSystem` L89 calls `PhysicsProcessor.UpdateRemovals()` which goes through the list of contacts and removes them (`Simulation.CleanContacts(component)`), finally removing the collision from both components in `Simulation.ContactRemoval` L768 and adding it to the `removedCollisionsCache` in L770
3. `Simulation.BeginContactTesting` clears the `removedCollisionsCache`
4. `removedCollisionsCache` is iterated over to send CollisionEnd event in `Simulation.SendEvents()`

Possible fix: https://github.com/manio143/stride/commit/ed0ae4329cf2127244bcf48755530f6ff026059a

manio143:
# PR Details

## Description

Send a message to channels waiting on `physicsComponent.CollisionEnded()` when the collision is removed due to the removal of a physics component. The collision object received by the user will have been _destroyed_, so both `collision.ColliderA` and `collision.ColliderB` will be `null`.

> Note: every collision received from `CollisionEnded` or `NewCollision` is destroyed before the next frame. Here I'm introducing sending of the events in between the last frame scripts and the destroying before the next frame.

## Related Issue

Fixes #788 

## Motivation and Context

Removing an entity/physics component should be valid means of ending a collision (rather than sending into some far away location for a frame). Rather than checking the collisions collection on the physics component we should encourage users to rely on the `CollisionEnded()` awaitable event.

## Types of changes

Currently we could assume that each collider of the collision will not be `null` (however, there is no `NotNull` contract on the `Collision` class, so this assumption is purely by convention).

## Checklist

For some strange reason my test is not being run by default in VS Test explorer.

### DIFF
--- a/sources/engine/Stride.Physics.Tests/GameAssets/SendCollisionEndedWhenEntityIsRemovedTest.sdscene
+++ b/sources/engine/Stride.Physics.Tests/GameAssets/SendCollisionEndedWhenEntityIsRemovedTest.sdscene
@@ -1,0 +1,89 @@
+!SceneAsset
+Id: ad1acc09-3637-40e7-b185-3bfdfb77b885
+SerializedVersion: {Stride: 2.1.0.1}
+Tags: []
+ChildrenIds: []
+Offset: {X: 0.0, Y: 0.0, Z: 0.0}
+Hierarchy:
+    RootParts:
+        - ref!! b2be5192-5dfb-4fc1-9cb3-0690ea5171dd
+        - ref!! b5866a46-8ea6-49a9-a46e-a534ce3db87d
+        - ref!! fa2ec94d-e743-4b7f-9b0c-731af5efeac1
+    Parts:
+        -   Entity:
+                Id: b2be5192-5dfb-4fc1-9cb3-0690ea5171dd
+                Name: Camera
+                Components:
+                    54d2b97cfa75fa43958a4db8e83d054b: !TransformComponent
+                        Id: 7cb9d254-75fa-43fa-958a-4db8e83d054b
+                        Position: {X: 0.0, Y: 0.0, Z: 2.215951}
+                        Rotation: {X: 0.0, Y: 0.0, Z: 0.0, W: 1.0}
+                        Scale: {X: 1.0, Y: 1.0, Z: 1.0}
+                        Children: {}
+                    ba901191469d00419a686d5db1a0d8c4: !CameraComponent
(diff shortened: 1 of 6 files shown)


Eideren:
> The collision object received by the user will have been destroyed, so both collision.ColliderA and collision.ColliderB will be null
If I understand this right I'm not sure that's a good idea, users logic might still need to know about what actually exit-ed their collision, perhaps mark those collisions as being caused by a collider removal inside the collision data

manio143:
Well, this would be a larger issue - given that the entity has been removed from the scene, or its physics component has been removed, returning a partially valid collision would be worse, than returning a collision that has no information (_destruction_ clears it).
Otherwise, the user might try to invoke further actions upon the entity/component which would result in an exception, due to it being detached.

If the user is interested in checking what entity has exited the collision they should manage this information themselves.
The main goal of this PR is not to trap scripts waiting forever in `await physicsComponent.CollisionEnded()` or `await collision.Ended()` if the other colliding party is removed.

Eideren:
That's why I proposed:
> perhaps mark those collisions as being caused by a collider removal inside the collision data but keep collider references
Adding a simple bool to it and users would naturally end up seeing it through intellisense when manipulating the collision, you could even include in the summary that it returns partial data when caused by component removal

manio143:
Right, thanks for clarifying, I will add that.

Eideren:
Looks good to me, I'll merge this in.
Note to @xen2 : "For some strange reason my test is not being run by default in VS Test explorer."

PR #893, opened 2020-10-25
[Physics] Move libbulletc.dll preloading to module initializer

### LINKED ISSUE #368: Creating any ColliderShape before a PhysicsComponent has been created results in an exception

**Version**: 3.1.01-beta01-0430

**Platform(s)**: Windows

**Describe the bug**
Creating any kind of ColliderShape, like a ConeColliderShape will currently result in an exception related to libbulletc not being loaded (if done before a PhysicsComponent has been created).

**To Reproduce**
1.  In any project before a **PhysicsComponent** has been created, create a shape like so: 
```cs
var newShape = new ConeColliderShape(32, 32, ShapeOrientation.UpX);
```

**Screenshots**
image

**Additional context**
Currently, you can work around the problem by making sure some kind of PhysicsComponent is created before any ColliderShape, because of this:

> https://github.com/xenko3d/xenko/blob/5f596657e006bfea9ebbf0f04302df5e3f3aed85/sources/engine/Xenko.Physics/Engine/PhysicsComponent.cs#L27-L31

manio143:
# PR Details

## Description

Moves native library preloading to module initializer.

**Question:** Should I also remove the preload call from `Bullet2PhysicsSystem`?

## Related Issue

Fixes #368 

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Physics/Engine/PhysicsComponent.cs
+++ b/sources/engine/Stride.Physics/Engine/PhysicsComponent.cs
@@ -33,12 +33,6 @@ namespace Stride.Engine
     {
         protected static Logger logger = GlobalLogger.GetLogger("PhysicsComponent");
 
-        static PhysicsComponent()
-        {
-            // Preload proper libbulletc native library (depending on CPU type)
-            NativeLibrary.PreloadLibrary("libbulletc.dll", typeof(PhysicsComponent));
-        }
-
         protected PhysicsComponent()
         {
             CanScaleShape = true;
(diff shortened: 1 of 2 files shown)


Eideren:
I seem to remember it loading only when a physics component is used, which seems smart given that not all games will use physics but should probably trigger on any physics related thing, is that behavior retained with this PR ?

manio143:
The native library will be loaded if your project references `Stride.Physics` package (even if you don't use any physics components).
So if your game won't be using physics, you need to remove that reference (which is included by default).

Eideren:
Alright that's good enough, thanks !

tebjan:
Great, that's a really good one, this will get rid of many null exceptions. Thanks @manio14

xen2:
I can confirm this is perfect as module initializer since Physics became a plugin (I think it was even mentioned on discord as TODO at some point)

PR #896, opened 2020-10-29
[Graphics] Fixed loading an image as SRgb

tebjan:
will now also convert the format of the pixel buffers.

(cherry picked from commit 659f6eef93e715558ed784eb7870468f32e0b557)

# PR Details

Fixes format to SRgb conversion on image load.

## Related Issue

Copying of pixel buffers failed because the formats did not match.

## Motivation and Context

Loading many images as texture array.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride/Graphics/Image.cs
+++ b/sources/engine/Stride/Graphics/Image.cs
@@ -137,25 +137,25 @@ namespace Stride.Graphics
         public ImageDescription Description;
 
         /// <summary>
-        /// Converts the format of the description and the pixel buffers to SRgb.
+        /// Converts the format of the description and the pixel buffers to sRGB.
         /// </summary>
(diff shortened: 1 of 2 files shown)


sources/engine/Stride/Graphics/Image.cs
```diff
@@ -136,6 +136,28 @@ public sealed class Image : IDisposable
         /// </summary>
         public ImageDescription Description;
 
+        /// <summary>
+        /// Converts the format of the description and the pixel buffers to SRgb.
+        /// </summary>
+        public void ToSRgb()
```
xen2, line 142: It sounds like it returns a SRGB version of this image. Maybe this should be renamed `ConvertToSRgb` to show it's a mutable action?

Issue #653, opened 2020-04-16
Xenko cant install bec of visual studio ERROR 1


MuktadaCreations:
when I try to install xenko, just after downloading all the needed files xenko ask for an administrator request access for visual studio it does that a couple of times intel it starts it after starting it immediately goes to the modify tap and after that it completely shuts down VS the  and says something about error 1 in a VS massage there is a video to see https://youtu.be/hUsp1EyGASE

P.S. I have visual studio 2017, 2019 and it does opens the modify window and I don't have .NET thing in it bec it's huge, if I can manually install what xenko need's please tell me

Kryptos-FR:
Please look at the documentation:
* https://doc.xenko.com/latest/en/manual/get-started/install-xenko.html
* https://doc.xenko.com/latest/en/manual/troubleshooting/xenko-doesnt-run.html

Also, in the video ti is clear that you are preventing the installer to complete its work. MSBuild Tools is part of the requirements for Stride (new name of Xenko), so you need to let it install or update necessary files.

arcazrael:
Before installing xenko.
-Checks if visual studio install has any updates.
-Open visual studio install, check if it opens without a problem
-Install visual studio tool 2017

I'm using xenko 3.1 and visual studio 2019 community

RayKoopa:
This isn't working for me too since a _long_ time - I tried installing Xenko 3.1 throughout the last year and today, and it _never_ worked on multiple PC installs here.

I'm using VS2019 Enterprise 16.6 Preview 3 with Workloads ".NET desktop development", "Desktop development with C++" and ".NET Core cross-platform development", and I also tried it with older stable 2019 Community versions back in the days.

Whenever the Xenko launcher tries to install _something_ with the VS installer, the VS installer aborts with the log message
```
[7980:000a][2020-04-20T18:29:45] Running pre-check verification.
[7980:000a][2020-04-20T18:29:45] Warning: Pre-check verification: Your machine is busy installing a program other than Visual Studio.

Select Continue to install Visual Studio while the other install is running. This might cause problems with other parts of the installation.

Select Retry to continue with the Visual Studio install once the other install has completed.

Select Cancel to cancel the Visual Studio install.
```
and error log message
```
[7980:000a][2020-04-20T18:29:45] Pre-check verification failed with warning(s) :  AnotherInstallationRunning.
```
so there seems to be something wrong in the installation process of Xenko 3.1.

[...]

PR #898, opened 2020-11-03
Fix to stop the visual studio installer exiting with code -1 during prerequisites install

Zeroto:
# PR Details

Moved vs prerequisite install before the other prerequisite installs.

## Description

The prerequisites installer somehow keeps the msiexec service in an installing state even though the installer is already finished and closed. This causes the visual studio installer to refuse to install because it thinks there is still an other installer running.

By installing the visual studio prerequisite before installing the other prequisites this problem is avoided. This is a hack because the windows installer state will still be wrong after this, but this will at least allow people to install stride properly without encountering an error that they can't easily solve. The windows installer being in a wrong state will most likely not cause any other problems for the user, because the next windows installer will detect the wrong state and fix it, or in the case of the visual studio installer it will fix it after hitting retry or restarting the visual studio installer(this will not happen when it is run in passive mode such as during the prerequisite install).

The proper solution for not ending up in this state would be in the advanced installer script/version. However, I could not find any online solutions for this or other people with the same problem.

## Related Issue

#653

## Types of changes

## Checklist

### DIFF
--- a/sources/tools/Stride.PackageInstall/Program.cs
+++ b/sources/tools/Stride.PackageInstall/Program.cs
@@ -36,15 +36,15 @@ namespace Stride.PackageInstall
                     case "/install":
                     case "/repair":
                     {
+                        // Make sure we have the proper VS2019/BuildTools prerequisites
+                        CheckVisualStudioAndBuildTools();
+
                         // Run prerequisites installer (if it exists)
                         var prerequisitesInstallerPath = @"install-prerequisites.exe";
                         if (File.Exists(prerequisitesInstallerPath))
                         {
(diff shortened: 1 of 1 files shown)


tebjan:
oh wow, that looks like a nice one!

xen2:
Interesting, so it was all due to our own prerequisite installer!

xen2:
It might be worth asking the Advanced Installer team/support about it.

xen2:
@Zeroto How did you notice/check this msiexec state? Is there any tool or command line I can use to check the current state?

Zeroto:
Sorry for the late reply, busy with my moving

No, as far as I know there is not anything available that will tell you the state of the service. My process was more a deduction of seeing what tasks were running and experimenting with running various installers in a different order and seeing the effects. I do remember encountering a few forum posts about people have similar problems (unrelated to stride)

PR #906, opened 2020-11-08
[Editor] Allow adding default configurations to GameSettings

### LINKED ISSUE #844: [Editor] Add a button to extend the GameSettings with new settings sections
The GameSettings asset cannot be extended from the Game Studio. If I create a custom configuration (see [this](https://www.youtube.com/watch?v=QKizq1xC4vk)), or include `Stride.Navigation` in my project, I need to manually edit the GameSettings asset file (and I need to know how to do this properly).

Similarly to components on an entity, there should be a button in the GameSettings asset view which says "Add settings section" and upon pressing displays a combo box with available, not already added, sections.

Those sections would be discovered similarly to how the editor finds instantiatable implementations for an interface.

manio143:
# PR Details

## Description

This PR modifies the presentation of `GameSettingsAsset.Defaults` in such way, that the Quantum nodes with default configurations are no longer spilled into the root node for the asset, but kept as a collection which allows to use the existing templates to display an add button that provides a way to add custom configurations.

Comparison between old and new look:
stride_gamesettings_add_configuration_change

I tried to add a button without modifying the way it looks now, but it seems that might not be possible. In particular, after the items in `Defaults` are spilled into the root, any modifications to the collection won't properly update the view without opening a different asset in the property grid and forcing a full refresh.

## Related Issue

Closes #844 

## Motivation and Context

We need a quick way to add settings for non-default packages (e.g. Physics or Navigation). It can also be used for adding custom configurations created by the user. 

## Types of changes

## Checklist

### DIFF
--- a/sources/editor/Stride.Assets.Presentation/NodePresenters/Updaters/GameSettingsAssetNodeUpdater.cs
+++ b/sources/editor/Stride.Assets.Presentation/NodePresenters/Updaters/GameSettingsAssetNodeUpdater.cs
@@ -27,7 +27,7 @@ namespace Stride.Assets.Presentation.NodePresenters.Updaters
                     item.Order = node.Order + item.Index.Int;
                     item.AttachedProperties.Add(CategoryData.Key, true);
                 }
-                node.BypassNode();
+                node.Commands.Add(new AddNewItemCommand());
             }
 
             if (typeof(Configuration).IsAssignableFrom(node.Type))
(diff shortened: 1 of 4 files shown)


manio143:
In order to allow removing configurations in the editor I had to change the attribute on the `GameSettingsAsset.Defaults` to say `ReadOnly = false`. This does not influence the actual use of the GameSettings during game runtime.

PR #922, opened 2020-11-30
[Input] Fixes compile error introduced in previous commit

azeno:
## Description

The previous PR introduced a compiler error. Not sure why I didn't catch it when running the tests locally :/


azeno:
Would it be possible to compile the PRs on teamcity so we can avoid such errors in the future?

xen2:
I did run some on teamcity but not everything.
It's not run automatically for safety reasons but I'll check if I can allow regular contributors.
Locally it builds only D3D by default.

xen2:
@azeno OK, from now on, it should auto-build PR from outside contributors (which you are now tagged with).

PR #923, opened 2020-12-01
Decoupling GraphicsDevice and GraphicsPresenter

azeno:
## Description

This PR decouples `GraphicsDevice` and `GraphicsPresenter` by removing the `GraphicsDevice.Presenter` property. A device can have many presenters.

- The central change is the remove of the `GraphicsDevice.Presenter` property. For code paths which are directly tied to the game instance the presenter of the main window gets retrieved through the game itself now (new abstract property `GameBase.Presenter`). Rendering code paths retrieve it through the `RenderDrawContext` (new key `GraphicsPresenter.Current` which gets set during `Game.Draw`).
- Much code moved from `GraphicsDeviceManager` to a new `GameWindowManager` class - managing a game window and its graphics presenter. However there're still code paths which are doubled in those two classes due to the rather complex device/screen mode queries. Would be eager to hear how to clean that up. Also having to add those `CreatePresenter` and `FindBestScreenModes` methods to the `IGraphicsDeviceService` didn't feel right (see comments in code).
- The `GraphicsDeviceManager` and the `GameWindowRenderer` make use of that new `GameWindowManager`. However the former probably shouldn't - but wanted to keep the API changes small so I forwarded most of the properties for now.
- I've made the `GraphicsPresenter` a `GraphicsResourceBase` to automatically getting picked up by the `ResumeManager` - how to test this?
- `CommandList.ClearStates` sets the render targets to null - should it take the render targets as parameters instead?
- The change in the Direct3D12 command list looks like a bug fix anyways?
- In `UIRenderFeature` I used the viewport to query for width/height - that correct? Otherwise the presenter or currently bound render target would need to be passed through those methods, the viewport was already there.

## Motivation and Context

When rendering to multiple windows we had to copy a lot of code from the `GraphicsDeviceManager` and also had to ensure to set and reset the `Presenter` property accordingly.

## Types of changes

## Checklist


azeno:
I just realized that under OpenGL the `GraphicsPresenter` is tied to the `GraphicsDevice` through internal properties like `WindowProvidedFrameBuffer` and `WindowProvidedRenderTexture` - what are the plans with OpenGL anyways?

xen2:
I was hoping we could drop it in the long term (Vulkan for Linux/Android and Metal for iOS through Vulkan bindings).
That's probably not going to happen right away...

Maybe there's some other way to disconnect them and just handle this behind the scene?
Even if you remove `GraphicsDevice.Presenter`, there's still this "default presenter" often created the same time as the device (this one for OpenGL, Android or some other windowing API might provide you something else with a specific ID, etc.).
Maybe this fits better as a return value for the function creating the `GraphicsDevice`?

xen2:

PR #925, opened 2020-12-02
[Build] Fixes Stride.Assimp.dll not found during package build

azeno:
# PR Details
## Description

Seems like something changed in msbuild with the net5 update that the dll isn't copied anymore. I guess in the past it came through the `Stride.Importer.Assimp.vcxproj` reference even though there it's marked as `CopyLocal = false`

Referencing the project directly ensures that the assembly gets copied to the build output folder of `Stride.Assets.Models` - it's needed by the `IncludeExtraAssemblies` target (`<BuildOutputInPackage Include="$(OutputPath)Stride.Assimp.dll" />`).

For reference, this is my build invocation:
`msbuild Stride.build /t:Package /m /v:m /restore /nr:false /p:StridePlatforms="Windows" /p:StrideGraphicsApiDependentBuildAll=false /p:StrideBuildPrerequisitesInstaller=false /p:StrideSign=false /p:StrideOfficialBuild=false`

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Assets.Models/Stride.Assets.Models.csproj
+++ b/sources/engine/Stride.Assets.Models/Stride.Assets.Models.csproj
@@ -14,6 +14,8 @@
     </Compile>
   </ItemGroup>
   <ItemGroup>
+    <ProjectReference Include="..\..\tools\Stride.Assimp\Stride.Assimp.csproj" PrivateAssets="All" />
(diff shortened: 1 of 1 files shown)


xen2:
I've already had similar issues in the past and it always took me a while to find the proper MSBuild fix (rather than adding a direct reference). This is due to `.vcxproj` behaving slightly differently.

But considering it happened once again, I guess using direct reference (rather than relying on indirect one) is probably the safer choice for now.

PR #927, opened 2020-12-03
[Rendering] Allows to use in-memory generated materials as blend layers

azeno:
# PR Details

## Description

By querying the `Material.Descriptor` property first an in-memory generated material can be used in blend layers.

## Types of changes

## Checklist


xen2:
Do you directly invoke the `Material.New` yourself?
I wonder if we want to make this behavior default (we probably want to get rid of unused data, and also you never know if user might mutate this descriptor to create multiple materials).

azeno:
I could live without the `Material.New` - we could simply assign the descriptor on our own after calling it. However the lookup was important I remember - to not rely on having the material as an asset.

azeno:
I just updated the PR and added a comment in code.

Issue #178, opened 2018-09-02
GameStudio: async asset loading


xen2:
We don't need to load all projects/assets before the editor is usable.
This can happen asynchronously in the background.
Two options:
(1) Still load all assets automatically, but in the background
(2) Don't load asset automatically, but on demand (asset selected/edited by user)

Note: At some point we changed our file format to use ref!! (which should be !!ref btw), so that we can easily scan files for references without deserializing full Yaml. The idea is to be able to quickly figure out dependencies if not all assets are loaded yet but we want to load a specific asset asap for editing.

PR #933, opened 2020-12-05
Check for XML documentation file and document dependencies. Fixes #917

### LINKED ISSUE #917: Game Studio doesn't show tooltip for public properties that have <userdoc> or <summary>

**Version**: 4.0.0.1-beta04-1317

**Platform(s)**: Windows

**Describe the bug**
When `<userdoc>` is added to a public property, game studio should show the text when the mouse is over that property in the Property Grid. But it remains blank. Intellisense in Visual Studio shows the text as expected when using `<summary>`.

**To Reproduce**
1. Attach script to Entity.
2. Add public property in script.
3. Add `<userdoc>` in XML comment above the property with text.

**Expected behavior**
Game studio shows text when hovering over property.

**Screenshots**
Screenshot 2020-11-20 143720
Screenshot 2020-11-20 143807

**Log and callstacks**
None

**Additional context**
Maybe game studio could also show the text in `<summary>` so that using both is unnecessary.

AmbulantRex:
# PR Details

This changes the way that documentation is cached for the editor to consider the standard XML documentation if available as a fallback when usrdoc is not provided. It will also now look at dependencies of solution project assemblies.

***Recreated due to invalid GitHub link with commit user.***

## Description

When first loading a project, the UserDocumentationService caches docs it finds in the `.usrdoc` files. This is a simple key-value format and is not generated automatically by Stride projects. The XML documentation file can be produced easily by including the appropriate directives in a csproj file as laid out in https://docs.microsoft.com/en-us/dotnet/csharp/codedoc. With this change, we will first look for a `.usrdoc` file before falling back to the `.xml` path. A future enhancement could be made to respect the project's specified documentation path. When looking at XML, it only considers the `<userdoc>` element

## Related Issue

https://github.com/stride3d/stride/issues/917

#1102 

## Motivation and Context

When <userdoc> is added to a public property, game studio should show the text when the mouse is over that property in the Property Grid. The only supported file is a custom key-value doc with the `.userdoc` extension and XML documentation is ignored. The `<userdoc>` element is currently what is being used in the `.usrdoc` file so the convention is continued here. 

## Types of changes

## Checklist


xen2:
Thanks, good change!
I am a little bit worried about the large XML parsing in the main thread. I wonder if it could be moved outside of UI thread (i.e. using async, Task or Lazy).
Maybe not trivial since it seems needed while building the nodes in `DocumentationNodeUpdater`.  Possibly `DocumentationData.Key` needs to change from `string` to `Task<string>`?

AmbulantRex:
XML Parsing is slow, but I'm not sure what an average size of this file would be. My project is small/medium that I was testing against with a 132 KB docs file and it wasn't perceptible. Using async should be okay. When you say DocumentationData.Key, are you referring to the cachedDocumentations dictionary in here? Since I will have parsed by that point should I attempt to wrap the parsing of the files in an async task instead?

AmbulantRex:
@xen2

AmbulantRex:
@xen2 I'm just coming back to this now. It seems like the async changes you discussed are related to #178. I'm not familiar enough with the engine to make a change that drastic. 

This information doesn't seem like something we should be caching on-demand as it generates with the build. Should we instead pre-cache while loading the projects into the session?

For example, the following area:
https://github.com/stride3d/stride/blob/a8977da1e215e301ed5f9af6d1ca44cc6a853004/sources/editor/Stride.Core.Assets.Editor/ViewModel/SessionViewModel.cs#L1260-L1270

Could be modified to perform this work like this:
```csharp
internal void ProcessAddedProjects(LoggerResult loggerResult, WorkProgressViewModel workProgress, bool packageAlreadyInSession)
{
    var newPackages = new List<PackageViewModel>();
    var documentationService = ServiceProvider.Get<UserDocumentationService>();
    foreach (var package in session.Projects.Where(x => !packageMap.ContainsValue(x)))
    {
        var viewModel = CreateProjectViewModel(package, packageAlreadyInSession);
        viewModel.LoadPackageInformation(loggerResult, workProgress);
        newPackages.Add(viewModel);

        // Add documentation at load time.
        foreach (var packageAssembly in viewModel.LoadedAssemblies)
        {                    
            documentationService.CacheAssemblyDocumentation(packageAssembly.Assembly);
        }
    }
    ProcessAddedPackages(newPackages);
}
```
(Is there a better place?)

These `.xml` and `.usrdoc` files are generally quite small for most projects and we already require a reload on code change. What are your thoughts?

sources/editor/Stride.Core.Assets.Editor/Services/UserDocumentationService.cs
```diff
@@ -95,59 +104,142 @@ private bool CacheAssemblyDocumentation([NotNull] Assembly assembly)
...
-                                location = asm.Path;
-                                break;
+                                var name = asm.Assembly.GetName();
+                                if (name.Name == assemblyName)
+                                {
+                                    location = asm.Path;
+                                    break;
+                                }
                             }
                         }
```
Eideren, line 129: A nitpick if you have time: this function works fine but it's kind of a mess right now; 
Those scopes don't really need to be as deep as they are, ex: you can merge two ``if`` together, return from one other ...
You could swap some of your logic around to avoid calling ``CacheAssemblyDocumentationFromPath(..., location)`` below when ``location`` has not changed, like calling it within the if scope above this instead and move the one below this to an else of the ``if (string.IsNullOrEmpty(location))`` test.

Eideren (commented):
Thanks for looking into this, looks pretty good !
I agree with xen's comment though, this could be done on a background thread/task while the thread it was started on continues loading the rest of the project. Those documentation files can get quite heavy with bigger projects and there's no reason to block the project from loading while you load tooltips in.

Edit: The systems you are interacting with are self contained, #178 shouldn't matter in your case. You can just spawn a new task like you would in c# and do your work in there while making sure that the other functions of this type are thread safe.

sources/editor/Stride.Core.Assets.Editor/Services/UserDocumentationService.cs
AmbulantRex: I did some superficial work to flatten this logic. I didn't go crazy since it's late, but let me know if you'd like to see more than that.

AmbulantRex:
@Eideren mind giving this one another look? I _think_ this should take care of the concurrency concerns, but my fire-and-forget per assembly vs. one `Task` with the foreach loop within has me second guessing myself.

manio143:
@AmbulantRex while you're working on this service can you take a look at fixing #1102? Should be quick, as it just needs a replace of `+` with `.` wherever the Type.FullName is used.

sources/editor/Stride.Core.Assets.Editor/Services/UserDocumentationService.cs
```diff
@@ -23,57 +27,57 @@ public class UserDocumentationService
...
-                var suffix = propertyKey.Name.Split('.').Last();
-                var key = $"F:{declaringType}.{suffix}";
-                return cachedDocumentations.TryGetValue(key, out string result) ? result : null;
+        public void ClearCachedAssemblyDocumentation([NotNull] Assembly assembly)
+        {
+            var assemblyName = assembly.GetName().Name;
+
+            lock (lockObj)
+            {
+                documentedAssemblies.Remove(assemblyName);
```
Eideren, line 76: You need to clear from ``undocumentedAssemblies`` as well, right ?

sources/editor/Stride.Core.Assets.Editor/Services/UserDocumentationService.cs
```diff
@@ -89,65 +93,161 @@ private bool CacheAssemblyDocumentation([NotNull] Assembly assembly)
...
-                            if (separator < 0 || separator >= line.Length - 1)
-                                continue;
-
-                            var key = line.Substring(0, separator);
-                            var documentation = line.Substring(separator + 1);
-                            cachedDocumentations[key] = documentation;
+                            continue;
                         }
+
+                        cachedDocumentations.TryAdd(key, documentation);
```
Eideren, line 192: ``cachedDocumentations[key] = documentation;`` otherwise documentation changes will keep the initial value

Eideren (commented):
Almost done, two small issues ! 

Do note that right now documentation removed while systems are running are not removed from this cache, I don't think it's a big deal as it would increase complexity quite a bit for something that has no side effect. If someone has a good argument to make sure it's cleaned up feel free to step in.

As for @manio143 's suggestion, if you really want to go the extra mile you could look at the source for those formats to see how/why it formats inner class with a ``.`` instead of a ``+`` and replicate that formatting on our end to make sure that it's always accurate. Feel free to just do a string replace ``+``->``.`` if you don't have time.

AmbulantRex:
> 
> 
> Almost done, two small issues !
> 
> Do note that right now documentation removed while systems are running are not removed from this cache, I don't think it's a big deal as it would increase complexity quite a bit for something that has no side effect. If someone has a good argument to make sure it's cleaned up feel free to step in.
> 
> As for @manio143 's suggestion, if you really want to go the extra mile you could look at the source for those formats to see how/why it formats inner class with a `.` instead of a `+` and replicate that formatting on our end to make sure that it's always accurate. Feel free to just do a string replace `+`->`.` if you don't have time.

I was able to throw in the simple replacement as part of this one. I dug around to see if there was a more elegant way to solve it, but .NET under the hood populates the qualified name using `+` for subclasses. The classes in the engine that dealt with this were very generic so I was afraid of unintended consequences modifying it any further up the chain, but I could come back to it if I come across a better way down the road.

sources/editor/Stride.Core.Assets.Editor/Services/UserDocumentationService.cs
```diff
@@ -23,57 +27,69 @@ public class UserDocumentationService
...
         {
             if (propertyKey == null) throw new ArgumentNullException(nameof(propertyKey));
+            var ownerType = propertyKey.OwnerType;
+            if (ownerType == null)
+                return null;
+
+            if (!CacheAssemblyDocumentation(ownerType.Assembly))
+                return null;
+
+            var declaringType = ownerType.FullName;
```
manio143, line 71: Can you also do the replace `+`, `.` here? Thanks

AmbulantRex: I'll give it a look!

Issue #952, opened 2020-12-10
Build fails when BaseIntermediateOutputPath is absolute path (Stride.Core.Assets.CompilerApp.targets)


jaredthirsk:
**Release Type**: Official Release

**Version**: 4.0.0.1-beta04-1347

**Platform(s)**: Windows

**Describe the bug**
Build fails when $(IntermediateOutputPath) is absolute path

**To Reproduce**
1. Create Directory.Build.props file in parent directory with this:

``` xml
<Project>
  <PropertyGroup>
    <BaseIntermediateOutputPath>z:\mydir\obj\$(MSBuildProjectName)\</BaseIntermediateOutputPath>	  </PropertyGroup>
</Project>
```

**Expected behavior**
Build from VS2019 succeeeds

**Log and callstacks**
```
4>C:\Users\ja\.nuget\packages\stride.core.assets.compilerapp\4.0.0.1-beta04-1347\buildTransitive\Stride.Core.Assets.CompilerApp.targets(81,5): error MSB3491: Could not write lines to file "C:\mydir\Stride\myapp.Stride.Windows\z:\mydir\obj\myapp.Stride.Windows\Debug\win-x64\stride\assetcompiler-uptodatecheck.inputs.expanded". The given path's format is not supported.
```

**Idea for fix:**

In Stride.Core.Assets.CompilerApp.targets, change this:

``` xml
<StrideCompileAssetUpToDateCheckFileBase>$(ProjectDir)$(IntermediateOutputPath)stride\assetcompiler-uptodatecheck</StrideCompileAssetUpToDateCheckFileBase>
```

to this:

``` xml
<StrideCompileAssetUpToDateCheckFileBase>$(IntermediateOutputPath)stride\assetcompiler-uptodatecheck</StrideCompileAssetUpToDateCheckFileBase>
```

jaredthirsk:
There may be a couple more places in this file that also need the chagne.

line 214:
```
<Exec Command="&quot;$(StrideCompileAssetCommand)&quot; --pack --package-file=&quot;$(MSBuildProjectFullPath)&quot; --build-path=&quot;$(BaseIntermediateOutputPath)stride\pack&quot;" ConsoleToMsBuild="true">
```

line 43:
```
<StrideCompileAssetOutputPath Condition="'$(StrideCompileAssetOutputPath)' == '' and '$(StridePlatform)' == 'Android'">$(IntermediateOutputPath)assets\data</StrideCompileAssetOutputPath>
```

[...]

PR #956, opened 2020-12-11
Allow absolute paths in $(IntermediateOutputPath)

jaredthirsk:
# PR Details

## Description

Remove $(ProjectDir), only rely on naked $(IntermediateOutputPath) which is typically relative, but this change allows IntermediateOutputPath to be an absolute path.
May require current working directory is set correctly for the typical case. 

## Related Issue

#952  

## Motivation and Context

I moved my IntermediateOutputPath to a ramdisk on a different drive, and so $(ProjectDir)$(IntermediateOutputPath) results in two absolute paths crammed into one which breaks the build.

## Types of changes

## Checklist


jaredthirsk:
I believe this breaks the samples due to incorrect working directory.  See issue for discussion of more potential fixes.

jaredthirsk:
See https://github.com/jaredthirsk/stride/blob/master/sources/assets/Stride.Core.Assets.CompilerApp/build/Stride.Core.Assets.CompilerApp.targets for my latest fixes (working so far with both relative and absolute $(BaseIntermediateOutputPath)

PR #965, opened 2020-12-14
Catch type load exceptions when trying to load asset compilers from an assembly

azeno:
# PR Details

## Description

This PR simply turns a compiler error into a warning. I had the asset compiler app crashing while it was processing an unrelated assembly. So instead of crashing completely a warning that it skipped the assembly was a nice way out.

## Types of changes

## Checklist

### DIFF
--- a/sources/assets/Stride.Core.Assets/Compiler/AssetCompilerRegistry.cs
+++ b/sources/assets/Stride.Core.Assets/Compiler/AssetCompilerRegistry.cs
@@ -181,26 +181,33 @@ namespace Stride.Core.Assets.Compiler
         private void RegisterCompilersFromAssembly(Assembly assembly)
         {
             // Process Asset types.
-            foreach (var type in assembly.GetTypes())
+            try
             {
-                // Only process Asset types
-                if (!typeof(IAssetCompiler).IsAssignableFrom(type) || !type.IsClass)
-                    continue;
-
-                // Asset compiler
-                var compilerAttribute = type.GetCustomAttribute<AssetCompilerAttribute>();
-
-                if (compilerAttribute == null) // no compiler attribute in this asset
-                    continue;
-
-                try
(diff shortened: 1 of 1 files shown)


xen2:
From a quick check, assemblies going through this should have been explicitly tagged by user with a call such as `AssemblyRegistry.Register(typeof(Module).Assembly, AssemblyCommonCategories.Assets);`.

I wonder how an unrelated assembly went through this code? (or maybe I missed some part of the logic)

(If registered explicitly by user, I think it makes more sense to make it an `Log.Error`, don't you think?)

azeno:
Let me debug this again to better understand what the actual issue was.

azeno:
So here is the initial exception I get while debugging
image
It complains that it can't find VL.Core

If I now go one up in the call stack, hover `assembliesToRegisterCopy` and look at the next assemblies which would be in the list with the debugger, I suddenly see in the debug output "Loaded VL.Core" and the assembly which caused the exception initially can be inspected.
The difference between them is that the former (causing the exception) is coming from a nuget package, while the latter is the build result of a project. It seems to me the assembly resolve event depends on the assembly load order.

azeno:
Having breakpoints in the resolve handlers, when looking for VL.Core, we first end up here:
image

It finds the requesting assembly (VL.Stride.Runtime) but using its path (pointing to the nuget package lib folder) it can't find VL.Core.

So now we get to the second resolve handler:
image

Here the `assembliesResolved` is already set, but `assemblies` doesn't list VL.Core because the restore was running on Stride.Core.Assets.CompilerApp:
image

azeno:
Here I see all assemblies (including VL.Core):
image

So maybe one solution would be to somehow simply register them to the current assembly container?

azeno:
Close me if https://github.com/stride3d/stride/pull/972 is the better option.

PR #970, opened 2020-12-15
[Graphics] Back buffer format adjustments based on the device color space

azeno:
# PR Details

## Description

Back buffer format adjustments based on the device color space will also be applied on resize. 
It will now also work for all pixel formats by using ToSrgb and ToNonSrgb extension methods.
Also fixes format adjustment not being applied if color space was set to Gamma but device supported Linear space.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Graphics/GraphicsPresenter.cs
+++ b/sources/engine/Stride.Graphics/GraphicsPresenter.cs
@@ -55,31 +55,7 @@ namespace Stride.Graphics
             GraphicsDevice = device;
             var description = presentationParameters.Clone();
 
-            // If we are creating a GraphicsPresenter with 
-            if (device.Features.HasSRgb && device.ColorSpace == ColorSpace.Linear)
-            {
-                // If the device support SRgb and ColorSpace is linear, we use automatically a SRgb backbuffer
-                if (description.BackBufferFormat == PixelFormat.R8G8B8A8_UNorm)
-                {
-                    description.BackBufferFormat = PixelFormat.R8G8B8A8_UNorm_SRgb;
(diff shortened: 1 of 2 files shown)


sources/engine/Stride.Graphics/GraphicsPresenter.cs
```diff
@@ -175,14 +151,29 @@ public void Resize(int width, int height, PixelFormat format)
...
 
+        private PixelFormat NormalizeBackBufferFormat(PixelFormat backBufferFormat)
+        {
+            // If we are creating a GraphicsPresenter with 
+            if (GraphicsDevice.Features.HasSRgb && GraphicsDevice.ColorSpace == ColorSpace.Linear)
+            {
+                // If the device support SRgb and ColorSpace is linear, we use automatically a SRgb backbuffer
+                return backBufferFormat.ToSRgb();
+            }
+            else
```
xen2, line 170: Previous code has a condition for this case: `else if (!device.Features.HasSRgb)`
Then the default situation would be to keep `backBufferFormat` as is.

azeno: We removed this condition on purpose because as we understood it, one either 
- has a device which is not capable -> ToNonSrgb (that was the !device.Featues.HasSRGB)
- asked for Gamma space -> ToNonSrgb
So in both cases one wants to end up with a NonSrgb format.

xen2: OK, it makes sense, thanks (I wasn't sure if it was intentional)

PR #972, opened 2020-12-16
Register assemblies in the assembly container to ensure later assembly loads succeed

azeno:
# PR Details

## Description

Different approach to what I proposed in https://github.com/stride3d/stride/pull/965. This time all the dependencies as given by the project are registered in the used assembly container to ensure that a later assembly load will work.

## Related Issue

https://github.com/stride3d/stride/pull/965

## Types of changes

## Checklist

### DIFF
--- a/sources/assets/Stride.Core.Assets/PackageSession.Dependencies.cs
+++ b/sources/assets/Stride.Core.Assets/PackageSession.Dependencies.cs
@@ -303,6 +303,10 @@ namespace Stride.Core.Assets
             // 1. Load store package
             foreach (var projectDependency in project.FlattenedDependencies)
             {
+                // Make all the assemblies known to the container to ensure that later assembly loads succeed
+                foreach (var assembly in projectDependency.Assemblies)
+                    AssemblyContainer.RegisterDependency(assembly);
+
                 var loadedPackage = packages.Find(projectDependency);
                 if (loadedPackage == null)
                 {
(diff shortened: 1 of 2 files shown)


xen2:
Does it run `AssemblyContainer.CurrentDomain_AssemblyResolve` before throwing the exception? (with a not-null `container`)
If yes, I think it might be better to do the few required assembly-loads lazily in there rather than forcing all of them to load.

However, I am not sure why we don't do that for the `.deps.json` case already. Maybe it doesn't work in practice.

Do you happen to have a repro project so that I could quickly test?

azeno:
1) Checkout `dev/azeno/assembly-load-repro` of https://github.com/vvvv/VL.Stride/tree/dev/azeno/assembly-load-repro and build the `packages/VL.Stride.Runtime` project (if you build the whole solution it will try to install vvvv which you don't need). You should end up with VL.Stride.* packages in `LocalAppData\Stride\Nuget Dev`
2) Download and build this console app: 
[ConsoleApp1.zip](https://github.com/stride3d/stride/files/5729454/ConsoleApp1.zip)

The build step should crash with a type load exception.

azeno:
Oh and the assemblies are not force loaded - just their paths get registered in case they need to be loaded.

xen2:
Sorry, I misread the code and thought `LoadAssemblyByName` was called by another top-level function rather than `AssemblyResolve`.
So it's already doing what I was describing, thanks 👍

Issue #803, opened 2020-07-26
Change how scripts are handled


salahchafai:
**Is your feature request related to a problem? Please describe.**
Currently there are 3 different types of scripts (startup, sync and async), i think it'll be better if startup and sync scripts get unified (and maybe async too) and have more functions like FixedUpdate, Awake, LateUpdate, OnEnable, OnDisable (virtual so you can override only the ones you need), the current system isn't scalabe as new methods require new classes (which is not ideal), so having a single class (maybe next to async) that has all the functions and then the user can override the ones he needs would be super helpful, also using virtual methods, or interfaces is standard with a lot of component based frameworks. As such, this is how most users would expect it to function. finally i think it'll be way better to have scripts activable so you can easily enable and disable scripts (which is very useful)

**Describe the solution you'd like**
Have a single class (let's say ScriptBehavior or something) that contains all the virtual functions, and make the scriptComponent inherit from ActivableEntityComponent instead of EntityComponent

**Describe alternatives you've considered**
The user can add these functions by himself (in async script for example) but that's not a great way of doing things and it's unnecessary extra work imo

tebjan:
Yes having a few more convenient methods to override would be nice.

This could be done in a new script type that takes care of that logic and provides these as virtual methods to override by the user.

I'd not remove the previous script types, that would make all existing projects invalid...

salahchafai:
@tebjan actually no, we can have some kind of a class responsible for upgrading the scripts so if it's a sync script it gets changed to the new type, if it's a startup script it gets changed to the new type and only override the start function, also i'll most likely keep the async script so that will work with no extra work), having many script types only causes confusion imo

Kryptos-FR:
> FixedUpdate, Awake, LateUpdate, OnEnable, OnDisable

Looks like you want to mimic Unity. But Stride doesn't work like Unity.

salahchafai:
@Kryptos-FR i don't want to mimic Unity, many people expect these functions (and many other engines have them like Unigine), and these are just functions i suggested and are not necessarily (the ones i see important are FixedUpdate, OnEnable, OnDisable)

manio143:
Would you consider providing those methods in the async context first?
While I understand your motivation it seems to me that creating a ScriptBehaviour with optionally overridable methods and using reflection to check those would be bad architecture.
On the other hand, extending the ScriptSystem to support `await Script.FixedUpdate()` would achieve the goal and get more people to use the async features of Stride.
[...]

PR #980, opened 2020-12-26
[POC] Async entity processors

AmbulantRex:
# PR Details

This PR provides a base processor for async entity processing and modifies the entity manager to support it in the same way that the `ScriptSystem` does for `AsyncScript`. 

## Description

Moving towards an asynchronous, ECS type system, we need to be able to support asynchronous component processing.  The `EntityProcessor` class was built to support synchronous operations only. Specifically, it was doing Update and Draw operations which do need to be performed on time. Stride supports asynchronous scripting through it's `ScriptSystem` and the `AsyncScript` base type, but the `EntityManager` had no concept of microthreads or async entity processors. There were bits of code in the `EntityManager` that suggested it may have been intended at some point (see `Scheduler` property) and was never fully implemented.  

## Related Issue

https://github.com/stride3d/stride/issues/20
https://github.com/stride3d/stride/issues/803

## Motivation and Context

Entity component processing can only happen synchronously. We allow asynchronous scripting through the AsyncScript, but no async entity processor if the developer would like to follow a pattern closer to ECS. 

## Types of changes

## Checklist

### DIFF
--- a/sources/core/Stride.Core/Collections/OrderedCollection.cs
+++ b/sources/core/Stride.Core/Collections/OrderedCollection.cs
@@ -146,7 +146,7 @@ namespace Stride.Core.Collections
         public void RemoveAt(int index)
         {
             if (index < 0 || index >= size) throw new ArgumentOutOfRangeException(nameof(index));
-            RemoteItem(index);
+            RemoveItem(index);
         }
 
         /// <summary>
@@ -184,7 +184,7 @@ namespace Stride.Core.Collections
             size++;
         }
 
-        protected virtual void RemoteItem(int index)
+        protected virtual void RemoveItem(int index)
         {
             size--;
             if (index < size)
--- a/sources/engine/Stride.Engine.Tests/TestEntityManager.cs
+++ b/sources/engine/Stride.Engine.Tests/TestEntityManager.cs
@@ -10,6 +10,9 @@ using Stride.Core;
 using Stride.Core.Collections;
 using Stride.Engine.Design;
 using Stride.Engine.Processors;
+using System.Threading.Tasks;
+using System.Collections.Concurrent;
+using Stride.Games;
 
 namespace Stride.Engine.Tests
 {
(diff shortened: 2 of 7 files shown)


manio143:
What is the use case for an asynchronous component processor?
I believe component processors need to be running each frame to properly observe component changes. If we're waiting in the processor for a long (across frames) action to happen, it cannot process new entities that have been spawned.
Also remember that running things on the scheduler is not parallel, so in practice the processors are going to run just as synchronously, except for cases where you await an asynchronous action (like an `EventReceiver`) to take place, which would mean you're "dropping" some frames in your processor.

I read through the changes and your design seems to have a flaw - it schedules the `Execute` method of the processor to be executed on each frame. If you want a behavior similar to that of `AsyncScript` which relies on `await Script.NextFrame` for scheduling continuations, you would need to schedule the `Execute` method just once per processor.

AmbulantRex:
> I read through the changes and your design seems to have a flaw - it schedules the Execute method of the processor to be executed on each frame. If you want a behavior similar to that of AsyncScript which relies on await Script.NextFrame for scheduling continuations, you would need to schedule the Execute method just once per processor.

Ah, I think I see what you mean. I misread this believing that it executed every update. I also neglected the importance of the `NextFrame` continuation stuff and failed to include a method of cancellation in the event that all entities have disappeared. 

https://github.com/stride3d/stride/blob/9219cbbb9de7a238060d97cf1cfbbf759120aa23/sources/engine/Stride.Engine/Engine/Processors/ScriptSystem.cs#L65-L133

As for use case. I would imagine anything that the `AsyncScript` would be valid for. The goal was to produce a processor which is responsible for executing some async behavior based on the data provided by the component. 

Taken from the Wikipedia article on Entity-Component-System:
> Each System runs continuously (as though each System had its own private thread) and performs global actions on every Entity that possesses a Component of the same aspect as that System

It's possible I misunderstand the intent of the `EntityProcessor`, and I also know that Stride is not "ECS", but it has been offered in the past that it could be a stand-in for the concept of a System as described in that pattern. My change was an effort to introduce a convenient solution for asynchronous processing, but if this doesn't feel "Core" enough it could be implemented easily enough by the developer in their own project through extension of the `EntityProcessor` class and managing a `Scheduler` from within it.

manio143:
Well, I think we could say that Stride works in discrete time space, therefore a continuous system wouldn't need to be running more than once per update frame (nor should it be expected to). Right now I think the best solution is to have the synchronous processor use the script system to schedule asynchronous tasks and monitor their execution in some way. The only issue would be if you wanted to inject asynchronous tasks to be executed in between some processors - as the script system usually runs after other processors.

PR #984, opened 2021-01-03
[Threading] Allow switching default pool + small performance improvement

Eideren:
# PR Details
``ThreadPool.Instance`` can now be assigned to, giving users the ability to change the object, and so, amount of threads the pool uses. Users should dispose of a pool if they intend to replace it.
Also imported/ported the latest changes from dotnet source, benchmarking shows a fairly small but significant enough improvement, should also waste slightly less resources on top of that improvement.

## Related Issue
Request from @tebjan related to https://github.com/vvvv/VL.Stride/issues/294

## Motivation and Context
I'm sure you can figure that one out :)

## Types of changes

## Checklist

### DIFF
--- a/sources/core/Stride.Core/Threading/ThreadPool.SemaphoreW.cs
+++ b/sources/core/Stride.Core/Threading/ThreadPool.SemaphoreW.cs
@@ -27,7 +27,8 @@ namespace Stride.Core.Threading
             /// </summary>
             private readonly Semaphore lifoSemaphore;
             private readonly int spinCount;
-            public Internals internals;
+            private Internals internals;
+            public uint SignalCount => internals.SignalCount;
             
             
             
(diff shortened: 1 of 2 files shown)


sources/core/Stride.Core/Threading/ThreadPool.SemaphoreW.cs
```diff
@@ -17,14 +17,17 @@ public sealed partial class ThreadPool
...
             private static readonly int OptimalMaxSpinWaitsPerSpinIteration;
+            private static readonly int SpinMult;
             
             /// <summary>
             /// Eideren: Is not actually lifo, standard 2.0 doesn't have such constructs right now
             /// </summary>
             private readonly Semaphore lifoSemaphore;
             private readonly int spinCount;
-            private Internals internals;
+            public Internals internals;
```
xen2, line 30: First letter should be uppercase since it's now public.

Issue #985, opened 2021-01-03
Wrong description in Stride installer when missing Visual Studio component


Basewq:
**Release Type**: Official Installer

**Version**: (?)

**Platform(s)**: Windows

**Describe the bug**
When installing Stride, if you are missing the Development Tools component, the installer says ".NET Core 2.1 Runtime (LTS)" is missing and needs to be installed, however this is incorrect.

As per this line
https://github.com/stride3d/stride/blob/b73b2a956422171c4173cecc7d5433d2d9c67d36/sources/tools/Stride.PackageInstall/Program.cs#L25
the installer is looking for `Microsoft.NetCore.ComponentGroup.DevelopmentTools.2.1`, but this is called `Development Tools plus .NET Core 2.1` as seen in this link https://docs.microsoft.com/en-us/visualstudio/install/workload-component-id-vs-community?view=vs-2019

The closest match to the description `.NET Core 2.1 Runtime (LTS)` is the one with ID `Microsoft.Net.Core.Component.SDK.2.1` as per this link https://docs.microsoft.com/en-us/visualstudio/install/workload-component-id-vs-build-tools?view=vs-2019

**Expected behavior**
Need the installer to state the correct missing/required component.

frionx:
I noticed this too, I got really confused while installing because of this. I installed like 10gb of .net runtimes and sdks and it finally worked so.

cstdbool:
Corrected and merged with PR #988 (commit 065b78b61219b41651eb813f8b29e1767305cc09). The issue can be closed.

Basewq:
> 
> 
> Corrected and merged with PR #988 (commit [065b78b](https://github.com/stride3d/stride/commit/065b78b61219b41651eb813f8b29e1767305cc09)). The issue can be closed.

Thanks.
FYI, if you write something like `Fixes #985` in the PR, Github can usually detect the link and auto-close the issue.

PR #988, opened 2021-01-08
Correct VS workload name and fix VS installer arguments

cstdbool:
# PR Details

This fixes a problem where Stride installation always fails for the machines which do not have the `Development Tools plus .NET Core 2.1` workload installed and then shows an incorrect error message telling the user to install the incorrect workload `.NET Core 2.1 Runtime (LTS) (inside .NET core cross-platform development)` manually.

## Description

This problem was introduced with b73b2a956422171c4173cecc7d5433d2d9c67d36 - the VS installer is invoked incorrectly due to invalid generated arguments. The commit changes the array type containing necessary VS workloads from string to tuple. However, not all places that referred to the tuple were updated to access the appropriate tuple item, which in turn resulted in the tuple being consumed directly by its `ToString()` method instead of the actual workload ID. This PR changes the code in the remaining places.

Furthermore, VS installer workload `Microsoft.NetCore.ComponentGroup.DevelopmentTools.2.1` was incorrectly labeled as `.NET Core 2.1 Runtime (LTS) (inside .NET core cross-platform development)`. This PR changes it to `Development Tools plus .NET Core 2.1` (as seen in the official documentation [here](https://docs.microsoft.com/en-us/visualstudio/install/workload-component-id-vs-community?view=vs-2019)).

## Related Issue

#653, #985

## Motivation and Context

The motivation is to fix the installation problem and the incorrect error message, so that the newcomers do not have to waste their time trying to figure out why the installation is constantly failing for their machine.

## Types of changes

## Checklist

### DIFF
--- a/sources/tools/Stride.PackageInstall/Program.cs
+++ b/sources/tools/Stride.PackageInstall/Program.cs
@@ -22,7 +22,7 @@ namespace Stride.PackageInstall
         {
             ("Microsoft.VisualStudio.Workload.ManagedDesktop", ".NET desktop development"),
             ("Microsoft.VisualStudio.Workload.NetCoreTools", ".NET core cross-platform development"),
-            ("Microsoft.NetCore.ComponentGroup.DevelopmentTools.2.1", ".NET Core 2.1 Runtime (LTS) (inside .NET core cross-platform development)"),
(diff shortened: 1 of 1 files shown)


xen2:
Good catch, thanks for the fixes!

tebjan:
Thanks a lot, this one was bugging new users since a while.

cstdbool:
I'm glad I could help. Thanks for merging!

PR #991, opened 2021-01-11
Fix InstancingProcessor dependency on ModelRenderProcessor

### LINKED ISSUE #813: Instancing crashes if instancing added before model component
**Release Type**: Github
**Describe the bug**
Adding an instancing component and then a model component crashes the instancing processor.

**To Reproduce**
1. Add instancing component to an entity,
2. Add model component to the same entity
3. Exception time

**Expected behavior**
It should not crash.

**Screenshots**
If applicable, add screenshots to help explain your problem (use GitHub drag & drop)
image

**Log and callstacks**
[stacktrace.txt](https://github.com/stride3d/stride/files/5003587/stacktrace.txt)

Basewq:
# PR Details

## Description

Create and add `ModelRenderProcessor` if it doesn't exist in `InstancingProcessor`.

## Related Issue
Fixes #813

## Motivation and Context
`InstancingProcessor` has a hard dependency on `ModelRenderProcessor` which it stores in a field.
A subtle bug can occur (usually in the Game Studio) where a user adds an `InstancingComponent` before adding any `ModelComponent`s in a scene. In this scenario, `InstancingProcessor` will be created first, on its own, and null will be stored in the field, so the fix is to create the processor immediately.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Engine/Engine/Processors/InstancingProcessor.cs
+++ b/sources/engine/Stride.Engine/Engine/Processors/InstancingProcessor.cs
@@ -23,8 +23,8 @@ namespace Stride.Engine.Processors
             public RenderInstancing RenderInstancing = new RenderInstancing();
         }
 
-        public InstancingProcessor() 
-            : base (typeof(TransformComponent), typeof(ModelComponent)) // Requires TransformComponent and ModelComponent
+        public InstancingProcessor()
+            : base(typeof(TransformComponent), typeof(ModelComponent)) // Requires TransformComponent and ModelComponent
         {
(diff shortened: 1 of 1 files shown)


tebjan:
Good one, so this checks whether there is a ModelProcessor and if not, it will add one, right?

Basewq:
> Good one, so this checks whether there is a ModelProcessor and if not, it will add one, right?

Yes, that's all it does.
The only other 'change' is removing a bunch of trailing whitespaces in the file, which doesn't actually impact the engine itself.

tebjan:
This is approved. Can someone merge it?

PR #993, opened 2021-01-12
Remove GameStudio dependencies when a build is uninstalled

### LINKED ISSUE #992: Uninstalling a build only removes Stride.GameStudio package

**Version**:
* Stride GameStudio: `4.0.1.1381-beta`
* Stride Launcher: `v5.0.3`

**Platform(s)**: Windows 10

**Describe the bug**
After uninstalling a Stride build, only `Stride.GameStudio` package is removed. All of the other `Stride.*` packages remain.

**To Reproduce**
1. Open Stride Launcher
2. Download any Stride build
3. Uninstall that build
4. Notice that the `C:\Users\<USERNAME>\.nuget\packages\stride.*` directories contain the files for the removed build version

**Expected behavior**
All `Stride.*` package dependencies from `Stride.GameStudio` should be removed for the uninstalled build version to save disk space.

**Screenshots**
None.

**Log and callstacks**
None.

**Additional context**
None.

cstdbool:
# PR Details

Uninstalling a build should remove all of the `Stride.*` packages for that build version, not just `Stride.GameStudio`.

## Description

Added additional logic to `/sources/launcher/Stride.Launcher/ViewModels/PackageVersionViewModel.cs`. When the deleted package ID is one of the main package IDs, then remove all of its dependencies (where IDs are like `Stride.*` or `Xenko.*`) for that particular build version.

## Related Issue

Fixes #992.

## Motivation and Context

This will remove the unneeded files from `C:\Users\<USERNAME>\.nuget\packages\` and it will save significant amounts of disk space for people who frequently update Stride builds.

## Types of changes

## Checklist

### DIFF
--- a/sources/launcher/Stride.Launcher/ViewModels/PackageVersionViewModel.cs
+++ b/sources/launcher/Stride.Launcher/ViewModels/PackageVersionViewModel.cs
@@ -10,6 +10,7 @@ using Stride.LauncherApp.Services;
 using Stride.Core.Presentation.Commands;
 using Stride.Core.Presentation.Services;
 using Stride.Core.Presentation.ViewModel;
+using System.Linq;
 
 namespace Stride.LauncherApp.ViewModels
 {
@@ -189,7 +190,7 @@ namespace Stride.LauncherApp.ViewModels
                         {
                             progressReport.ProgressChanged += (action, progress) => { Dispatcher.InvokeAsync(() => { UpdateProgress(action, progress); }).Forget(); };
                             progressReport.UpdateProgress(ProgressAction.Delete, -1);
-                            await Store.UninstallPackage(LocalPackage, progressReport);
+                            await UninstallPackage(Store, LocalPackage, progressReport);
                             CurrentProcessStatus = null;
                         }
                     }
@@ -236,7 +237,7 @@ namespace Stride.LauncherApp.ViewModels
                         var localPackage = Store.FindLocalPackage(ServerPackage.Id, ServerPackage.Version);
                         if (localPackage != null)
                         {
-                            await Store.UninstallPackage(localPackage, null);
+                            await UninstallPackage(Store, localPackage, null);
                         }
                     }
                     catch
@@ -295,7 +296,7 @@ namespace Stride.LauncherApp.ViewModels
                     progressReport.ProgressChanged += (action, progress) => { Dispatcher.InvokeAsync(() => { UpdateProgress(action, progress); }).Forget(); };
                     progressReport.UpdateProgress(ProgressAction.Delete, -1);
                     CurrentProcessStatus = string.Format(Strings.ReportDeletingVersion, FullName);
-                    await Store.UninstallPackage(LocalPackage, progressReport);
+                    await UninstallPackage(Store, LocalPackage, progressReport);
                     CurrentProcessStatus = null;
                 }
             }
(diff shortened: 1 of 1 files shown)


sources/launcher/Stride.Launcher/ViewModels/PackageVersionViewModel.cs
```diff
@@ -316,6 +317,38 @@ private async Task DeleteInternal(bool displayErrorMessage)
...
+
+        private static async Task UninstallDependencies(NugetStore store, NugetPackage package)
+        {
+            foreach (var dependency in package.Dependencies)
+            {
+                string dependencyId = dependency.Item1;
+                string dependencyIdPrefix = dependencyId.Split('.').First();
+                PackageVersionRange dependencyVersionRange = dependency.Item2;
+
+                // Dependency must be from Stride/Xenko and package version must match exactly
```
xen2, line 339: Good idea, it should prevent it from uninstalling the shared `Stride.Samples.Templates` (which is versioned separately and usually shared amongst multiple versions).

xen2:
LGTM, thanks!

sources/launcher/Stride.Launcher/ViewModels/PackageVersionViewModel.cs
xen2: Thinking about it, I am not sure how this will behave if at some point we have versions that overlap and evolve at different speeds, because the `dependencyVersionRange` is usually a "at least" constraint, not a "equal" (i.e. if both 4.1 and 4.2 references samples templates 4.1, deleting 4.1 will delete the templates required by 4.2).
I wonder if we should hardcode this specific package id, but it's not great...

Worst case it's not a blocking/critical issue because GameStudio and/or Visual Studio will restore nuget packages on next launch.

cstdbool: If this is the only package like this, then it makes sense to make it an exception. I've created a PR #994 that addresses this problem.

xen2: Sorry, I was thinking about it again and think it's better to not hardcode.
Reasons:
- This `Samples.Templates` package name might change, this would make details of Stride package spill even more into the launcher (and force an upgrade)
- Also I realized there are other packages like this, such as `Stride.OpenTK`

Anyway, I was thinking that a better way to fix that is that versions will (almost) never collide, since the semver now contains the git height (i.e. `4.0.1.1431` with the 1431 being auto-generated).
As a result, we can simply rewrite the test like `dependencyVersionRange.Contains(package.Version)` to `dependencyVersionRange.MinVersion == package.Version && dependencyVersionRange.IsMinInclusive && dependencyVersionRange.MaxVersion == null` ?

cstdbool: But we would still end up with the same problem. Imagine the hypothetical scenario like this:
* `MainPackage.1.0` depends on `DependencyPackage.1.0` (minimum version)
* `MainPackage.2.0` depends on `DependencyPackage.1.0` (minimum version)

In this case, the removal of `MainPackage.1.0` would end up deleting `DependencyPackage.1.0`, but it is still needed by `MainPackage.2.0`.

cstdbool: Better solution #995 - On launcher start or on GameStudio version install/update/uninstall, all unique dependency versions are enumerated for every installed GameStudio package version. The set is compared to the previous set content, and if previous set contains a version that is no longer present in the new set, only then it is removed.

PR #994, opened 2021-01-13
Uninstalling GameStudio should not remove Stride.Samples.Templates package

cstdbool:
# PR Details

Uninstalling `Stride.GameStudio`/`Xenko.GameStudio` should not remove `Stride.Samples.Templates`/`Xenko.Samples.Templates` package.

## Description

PR #993 added a functionality where removing `Stride.GameStudio`/`Xenko.GameStudio` via the launcher would also remove all of its dependencies with package IDs like `Stride.*`/`Xenko.*`. However, it was recently brought to my attention by @xen2 that there exists a shared package (`Stride.Samples.Templates`/`Xenko.Samples.Templates`) which does not get released for every released build, but rather some of the newer builds reference the past builds of this package. This package should not be removed because there is a chance that it is also referenced by a build newer than the build that is being uninstalled.

## Related Issue

#992

## Motivation and Context

Preventing a possible bug in the future. See [comment](https://github.com/stride3d/stride/pull/993#discussion_r556267309) for more details.

## Types of changes

## Checklist

### DIFF
--- a/sources/launcher/Stride.Launcher/ViewModels/PackageVersionViewModel.cs
+++ b/sources/launcher/Stride.Launcher/ViewModels/PackageVersionViewModel.cs
@@ -328,16 +328,21 @@ namespace Stride.LauncherApp.ViewModels
             }
         }
 
+        // These packages will not be uninstalled with GameStudio
+        private static readonly string[] fixedDependencies = { "Samples.Templates" };
+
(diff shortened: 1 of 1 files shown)


xen2:
Sorry, after some more thinking I thought it would be better to do it another way:
https://github.com/stride3d/stride/pull/993/files#r556958469

PR #995, opened 2021-01-14
Remove packages which are no longer referenced by any GameStudio package after uninstall or update

cstdbool:
# PR Details

Remove packages which are no longer referenced by any GameStudio package after uninstall or update.

## Description

When a GameStudio package is uninstalled via launcher, a lot of other packages will remain on disk which are no longer referenced by any of the remaining GameStudio packages. This change makes sure that **only** the packages which are not referenced by any GameStudio package are removed.

## Related Issue

#992, #993, #994

## Motivation and Context

Previous solution from #993 was error-prone.

## Types of changes

## Checklist

### DIFF
--- a/sources/launcher/Stride.Launcher/ViewModels/LauncherViewModel.cs
+++ b/sources/launcher/Stride.Launcher/ViewModels/LauncherViewModel.cs
@@ -13,7 +13,6 @@ using System.Threading.Tasks;
 using Microsoft.Win32;
 
 using Stride.Core.Extensions;
-using Stride.PrivacyPolicy;
 using Stride.LauncherApp.Resources;
 using Stride.LauncherApp.Services;
 using Stride.Core.Packages;
@@ -23,6 +22,7 @@ using Stride.Core.Presentation.Services;
 using Stride.Core.Presentation.ViewModel;
 using Stride.Metrics;
 using Stride.Core.VisualStudio;
+using System.Diagnostics.CodeAnalysis;
 
 namespace Stride.LauncherApp.ViewModels
 {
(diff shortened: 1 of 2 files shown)


sources/launcher/Stride.Launcher/ViewModels/LauncherViewModel.cs
```diff
@@ -218,6 +264,22 @@ public async Task RetrieveLocalStrideVersions()
                 var localPackages = await RunLockTask(() => store.GetPackagesInstalled(store.MainPackageIds).FilterStrideMainPackages().OrderByDescending(p => p.Version).ToList());
                 lock (objectLock)
                 {
+                    // Try to remove unused Stride/Xenko packages after uninstall or update
+                    try
+                    {
+                        Task.WaitAll(RemoveUnusedPackages(localPackages));
```
xen2, line 270: Curious:
I assume here it would slow down Stride initial listing, and 99% it should not be necessary.
Any reason it's not done during uninstall/update action? (it might need to be done after the install of the new version though)

Also, in the unlikely case the user explicitely want one of those separately versioned package for his project (i.e. directly as nuget ref), it will end up being deleted **every** Stride launcher start, rather than just when a Stride version is uninstalled.
It seems little bit harsh/intrusive (esp. if there's some unexpected scenario or bug, it might end up deleting a lot of packages every start).

What do you think?

cstdbool: I didn't experience any noticable slowdown at all, most likely because it's only checking for locally installed packages (no web requests).

Actually, packages are **not** uninstalled when the launcher starts, it just checks which packages are referenced by currently installed GameStudio versions. The actual uninstalling occurs **only** when a version is installed/updated/removed.

In an unlikely event where the user wants to keep a package, there are two possible scenarios:
* package is on disk and no GameStudio is referencing it - in this case any install/update/uninstall will **not** remove the package (unless the GameStudio version which references that exact package is installed and then uninstalled afterwards)
* package is on disk and one GameStudio is referencing it - in this case, when this GameStudio is uninstalled, the package will be deleted as well, however, after the package is restored (manually or whatever), it will remain on the disk indefinitely

Issue #999, opened 2021-01-19
[WIP] Introduce Interface IGameBase, change IGame


CharlesWoodhill:
I created this issue to track my changes to IGame and the new IGameBase interface.
The rest here assumes stride accepts my changes:

What the soon coming PullRequest would change:

Renamed and enriched Interface: from IGame to IBaseGame
New Interface: IGameBaseInternal (to cover internal marked props/methods)
New Interface: IGame
Let Game implement IGame
Let GameBase implement IGameBase (implicit) and IGameBaseInternal (explicit)
Replaced most uses of GameBase to IGameBase (where appropriate)
Changed internal uses of GameBase to use IGameBaseInternal
Replaced most uses of Game to IGame (where appropriate)

ATTENTION:
i only tested my changes with windows and there with PlatformDesktop
but the change needed to edit source encapsulated in compiler directives
so i can only hope i did everthing right.

Can anybody compile and test this for me on other platform(s)?

CharlesWoodhill:
Forgot to reason why i need this.
Short answer: 
Get more control over the engine when using it as library/codeonly
Make it possible to create own Game (including GameBase)

CharlesWoodhill:
no rushing

PR #1000, opened 2021-01-19
Interfaces IGame, IGameBase

CharlesWoodhill:
# PR Details
Issue: #999

Renamed and enriched Interface from IGame to IBaseGame
Created Interface: IGameBaseInternal (to cover internal marked props/methods)
Created Interface: IGame
Let Game implement IGame
Let GameBase implement IGameBase (implicit) and IGameBaseInternal (explicit)
Replaced most uses of GameBase to IGameBase (where appropriate)
Replaced most uses of Game to IGame (where appropriate)

## ATTENTION

i only tested my changes with windows and there with PlatformDesktop
but the change needed to edit source encapsulated in compiler directives
so i can only hope i did everthing right.

Can anybody compile and test this for me on other platform(s)?

## Motivation and Context

More control from "Outside" when using Stride as library.

## Types of changes

## Checklist

### DIFF
--- a/sources/editor/Stride.Assets.Presentation/AssetEditors/AssetHighlighters/MaterialAssetHighlighter.cs
+++ b/sources/editor/Stride.Assets.Presentation/AssetEditors/AssetHighlighters/MaterialAssetHighlighter.cs
@@ -81,7 +81,7 @@ namespace Stride.Assets.Presentation.AssetEditors.AssetHighlighters
             }
         }
         
-        private async Task HighlightMaterial(Game game, Material material, float duration)
+        private async Task HighlightMaterial(IGame game, Material material, float duration)
         {
             // TODO: this could be factorized between the different highlighters
             if (material == null)
--- a/sources/editor/Stride.Assets.Presentation/AssetEditors/AssetHighlighters/ModelAssetHighlighter.cs
+++ b/sources/editor/Stride.Assets.Presentation/AssetEditors/AssetHighlighters/ModelAssetHighlighter.cs
@@ -60,7 +60,7 @@ namespace Stride.Assets.Presentation.AssetEditors.AssetHighlighters
             }
         }
 
-        private async Task HighlightMesh(Game game, Mesh mesh, float duration)
+        private async Task HighlightMesh(IGame game, Mesh mesh, float duration)
         {
             // TODO: this could be factorized between the different highlighters
             if (mesh == null)
--- a/sources/editor/Stride.Assets.Presentation/AssetEditors/EntityHierarchyEditor/Game/EditorGameNavigationMeshService.cs
+++ b/sources/editor/Stride.Assets.Presentation/AssetEditors/EntityHierarchyEditor/Game/EditorGameNavigationMeshService.cs
@@ -671,7 +671,7 @@ namespace Stride.Assets.Presentation.AssetEditors.EntityHierarchyEditor.Game
             /// <summary>
             /// Updates the highlight material animation
             /// </summary>
-            public void UpdateHighlighting(GameBase game)
+            public void UpdateHighlighting(IGameBase game)
             {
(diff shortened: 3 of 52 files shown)


manio143:
Can you check if this change has any impact on performance? Virtual calls for interfaces are ever so slightly [more expensive](https://stackoverflow.com/questions/7225205/performance-of-direct-virtual-call-vs-interface-call-in-c-sharp) but I'm not sure if `Game` methods are called often enough for it to matter.

Eideren:
@manio143 the large majority of functions within 'game' are called once at most per frame, perf hit won't be meaningful here

@CharlesWoodhill 
Could you give an example where this is necessary ? 
With this PR we'll have ``IGameBase``, ``GameBase``, ``IGame`` and ``Game``, it's kind of getting ridiculous, I would rather we had a single class which just contains virtual properties and functions to be overridden by users instead of this mess of obtuse types.

Could you verify that all of the changes you included are actually necessary ?
``sources/engine/Stride.Engine/Engine/Sim.cs `` is commented out but still included with this PR, 
``sources/targets/Stride.Core.TargetFrameworks.Editor.props``'s changes lay outside the scope of this PR.

Should ``sources/engine/Stride.Games/SimBase.cs`` really be included in this PR if, according to the comment, it is still experimental, probably needs a better name if that's the case, like ``GameCodeOnly`` / ``MinimalGame``, something more obvious.

CharlesWoodhill:
I did this to be able to create a more "open" Game object (working tiltle Sim object)
without tampering with exitsting Game object.
The Sim object should be targeting users who use Stride from their existing project
things it should be able to do:
provide a feature to use "external" log,
provide a feature to run on other thread than mainthread  (the main thread is owned by the user not the engine)
provide a feature to use "external" window as "mainWindow"
more to come :)

Eideren:
I don't think it's smart to just go ahead and implement another ``Game`` every time we have a specific setup to cover, that's one more class we'll have to maintain every time we make changes to games and you'll never have an implementation that would satisfy all users, so the smartest way to go about this is:
- Either you have to make it incredibly basic and provide documentation for users to build it up however they want, i.e. ``game`` now is just an interface, users will deal with the implementation on their side but you expose functions to help them build it up.
- Or you make it complete and extensible enough so that most users don't have to fiddle with it but those who do, can change it however they want.

I would rather you go with the later, see what you need to change to make ``Game`` or ``GameBase`` work with your setup and expose/branch/decouple them.

---

> provide a feature to use "external" log,

Afaik you can already do that, not sure what you mean otherwise.

> provide a feature to run on other thread than mainthread (the main thread is owned by the user not the engine)

``Game`` uses the thread you ``Run()`` it on, perhaps under certain windowing system that's not the case ? Feel free to correct me.

> provide a feature to use "external" window as "mainWindow"

Yes, that would require some changes to the engine but can still be used from a standard ``Game`` instance, like when we use ``Run(someWindowingSystem)``, this parameter could be changed to something more abstract.

---

I really think that you need to make changes to this pr before we can merge it, Xen can do so if he wants to but I don't agree with a lot of those changes.
- Rollback changes to ``sources/targets/Stride.Core.TargetFrameworks.Editor.props``, it's out of scope and I would rather Xen deal with these kinds of changes.
- Remove unnecessary / logic commented out.
- ``IGame.AutoLoadDefaultSettings`` this member should be specified through a constructor, not an interface member, why should all implementations of ``IGame`` implement this ?
- Change how you implement this, see the first section of this message.

CharlesWoodhill:
maybe it was too fast, better close it unmerged

Eideren:
@CharlesWoodhill please, don't be discouraged by what I told you. You've got the right idea, 'just needs a little more work on your side before I merge this in.

CharlesWoodhill:
nono you are right, it was too fast

PR #1006, opened 2021-01-25
[DO NOT MERGE] Move to Vortice.Direct3D11 and Vortice.Direct3D12

amerkoleci:
Move to latest DirectX bindings for Direct3D11 and Direct3D12.


xen2:
Thanks for the contribution!
It's a good idea to move on from the deprecated SharpDX.
However, it might be good to discuss plans before you go further with the port (the latest discussion consensus was to switch to Silk.NET for D3D11/12 to have struct-like lower-level bindings).

amerkoleci:
Ok, will close PR, ping me in case.

Thanks

PR #1014, opened 2021-02-06
[Editor] Add DateTime property editor and improve TimeSpan property editor

### LINKED ISSUE #961: Datetime cannot be serialized
**Release Type**: Official Release/GitHub 

**Version**: 4.0.0.1-beta04-1347  with .NET 5.0.01

**Platform(s)**: For now I am only testing it on windows.

**Describe the bug**
When adding a DateTime field into any custom struct/class or even in the scripts directly as a variable it will only be serialized as the pure string "DateTime" and not the actually value or even the possibility to change the value in the description:

**To Reproduce**
1. Open existing .cs script (i. e CharachterScript)
2. Add a public DateTime field
3. Save solution
4. Now u should see it only serializes the word "DateTime"

**Expected behavior**
A proper value serialization (if one was setup in the Start or constructor) or the possibility to sett it via a DateTime-Picker within the inspector

**Screenshots**
see attached image!

**Log and callstacks**
Successful Log:
_[(0,0)]: Info: Patch for assembly [SpaceEscape.Game, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null]
[(0,0)]: Info: SpaceEscape.Game -> C:\Users\shho3\OneDrive\Dokumente\Stride Projects\SpaceEscape\SpaceEscape.Game\bin\Debug\netstandard2.0\SpaceEscape.Game.dll
Info: Unloading assembly SpaceEscape.Game, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null
Info: Loading assembly SpaceEscape.Game, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null
Info: Updating components with newly loaded assemblies_

**Additional context**
Nope.
error

cstdbool:
# PR Details

The `DateTime` property editor was added. The `TimeSpan` property editor was improved.

## Description

The `DateTime` property editor was requested in #961, so I've implemented it with this pull request. The date and time value is displayed in the read-only text box, and next to it there is a calendar button (icon taken from `Fugue Icons` pack, the same pack from where the other Stride icons come from). When clicked, a popup is shown with each of the `DateTime` components editable. Originally, I intended to use WPF `DatePicker`, however, it seemed that it wouldn't work in well with the Stride's UI theme (keep in mind that I'm not a WPF expert).

The `TimeSpan` property editor was improved as well. Previously it was just a single text box (implicitly in seconds). Now it separates the components of the time span into `Days`, `Hours`, `Minutes` and `Seconds`, making it more user friendly.

DateTime_TimeSpan_Updates

## Related Issue

Closes #961.

## Motivation and Context

Extending the GameStudio functionality. Making the GameStudio more user friendly.

## Types of changes

## Checklist


sources/editor/Stride.Core.Assets.Editor/View/DefaultPropertyTemplateProviders.xaml
```diff
@@ -807,29 +807,35 @@
...
+  <!-- Provider for date time editor -->
+  <view:TypeMatchTemplateProvider x:Key="DateTimePropertyTemplateProvider" Type="{x:Type s:DateTime}" edvw:PropertyViewHelper.TemplateCategory="PropertyEditor">
+    <DataTemplate DataType="qvm:NodeViewModel">
+      <sd:DateTimeEditor x:Name="DateTimeEditor" Margin="2"
+                         Value="{Binding NodeValue, Converter={cvt:DifferentValuesToNull}}"
+                         WatermarkContentTemplate="{StaticResource DifferentValuesWatermarkTemplate}">
+      </sd:DateTimeEditor>
+      <DataTemplate.Triggers>
+        <DataTrigger Binding="{Binding NodeValue}" Value="{x:Static qvm:NodeViewModel.DifferentValues}">
+          <Setter TargetName="DateTimeEditor" Property="WatermarkContent" Value="{sd:Localize (Different values)}"/>
```
manio143, line 832: What do you mean here by `Different values`? When would this show up?

sources/presentation/Stride.Core.Presentation/Themes/ThemeSelector.xaml
```diff
@@ -1801,6 +1802,122 @@
...
+    <Setter Property="Focusable" Value="False"/>
+    <Setter Property="Template">
+      <Setter.Value>
+        <ControlTemplate TargetType="ctrl:DateTimeEditor">
+          <Grid>
+            <Grid.ColumnDefinitions>
+              <ColumnDefinition Width="*"/>
+              <ColumnDefinition Width="Auto"/>
+            </Grid.ColumnDefinitions>
+            <ctrl:TextBox Grid.Column="0" IsEnabled="False" SelectAllOnFocus="True" Margin="-1" TextAlignment="Left"
```
manio143, line 1857: Conside using `IsReadOnly` instead of `IsEnabled` so that the text isn't grayed out.

sources/editor/Stride.Core.Assets.Editor/View/DefaultPropertyTemplateProviders.xaml
cstdbool: When you select multiple objects containing the same script. If the values differ, this string will be shown.

sources/presentation/Stride.Core.Presentation/Themes/ThemeSelector.xaml
cstdbool: Thanks for the hint! I amended the last commit.

tebjan:
Thanks, nice one!
Did you consider smaller time spans than seconds? Sometimes ms or even ticks are required...

cstdbool:
> Thanks, nice one!
> Did you consider smaller time spans than seconds? Sometimes ms or even ticks are required...

Yes, of course. Look at the pic above, the seconds text box accepts real numbers, so anything smaller than a second can be specified as well.

tebjan:
Very good, didn't see that one. Did you check whether the parsing respects sub ms or does it round like the FromSeconds method? We had another PR with a similar topic. For some reason the standard .NET methods round to the nearest ms and one has to write an extension to get the actual precision down to ticks.

cstdbool:
Initially I wasn't thinking about the need for precision to go below the millisecond level. But it wouldn't be a bad idea to support that as well. I changed the way the component properties are updated in my latest commit on this PR (bd19288583d36951a5a45964112a6cc01db84bf5) so that it now supports tick-level precision. Thanks for your suggestion @tebjan, and please feel free to double-check that it works.

tebjan:
Looks great, thanks for taking care of the precision. It's important sometimes, for example, 60fps is 16.66... ms per frame.

xen2:
Thanks, great addition!
LGTM, one quick question though:
where does the `calendar--pencil.png` come from? Are we allow to use/distribute it? Any license to include?

cstdbool:
@xen2 It comes from `Fugue Icons` pack from [here](https://p.yusukekamiyamane.com/). It is already credited in the `About` -> `Third Party Software` section, since the other icons came from there as well.

PR #1015, opened 2021-02-08
[Editor] Show property type on hover in property grid

### LINKED ISSUE #963: Show for every Inspector-field "OnHover" the specified datatype of that particular Inspector-field
Not a problem directly but sometimes with alot of serialized datatypes its hard to keep track of what field is what, its not always 
directly known or remembered so it could help to visually show the underlying type of the inspector-field.

Title!

None.

**Additional context**
Nope.

cstdbool:
# PR Details

Property type should be shown when mouse is hovered over the property in property grid (inspector).

## Description

Property type is now shown when the mouse hovers over the property in property grid (inspector). See example below:

TypeInfo

## Related Issue

Closes #963.

## Motivation and Context

User should be able to tell the data type of the property without having to read the code of the script.

## Types of changes

## Checklist

### DIFF
--- a/sources/editor/Stride.GameStudio/GameStudioWindow.xaml
+++ b/sources/editor/Stride.GameStudio/GameStudioWindow.xaml
@@ -876,7 +876,16 @@
 
                     <Border Grid.Row="6" Background="{StaticResource ToolBarBackgroundBrush}">
                       <DockPanel>
-                        <TextBlock Margin="4" DockPanel.Dock="Top" FontWeight="Bold" Text="{Binding HoveredItem.DataContext.DisplayName, ElementName=AssetPropertyView}" TextTrimming="CharacterEllipsis"/>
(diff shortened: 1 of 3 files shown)


xen2:
Thanks!

I tested to make the property grid narrower and it doesn't behave so nicely:
image

We should keep a small forced space between Name and Type (20 pixel+) and use ellipsis if typename is too long (maybe on the namespace part first? not sure)

cstdbool:
@xen2 I implemented your suggestions. It looks a bit better now. See the pic below for various widths of property grid.
property_grid_resize

For the last two, maybe it is a bit strange that the value is cut-off like this, but nevertheless these are very narrow property grid widths, and the actual property values above would be cut-off as well.

Anyway, this control is usable for reasonable widths.

xen2:
Thanks for the change!
I think the way you split/prioritize is perfect for narrow situations (I had the same in my mind).

PR #1023, opened 2021-02-18
[Launcher] Distinguish between dev updates and official updates

### LINKED ISSUE #719: launcher confuses "official" vs "locally built"
**Version**: Launcher v4.0.6
**Platform(s)**: Windows

**Describe the bug**

* The launcher says "(update available) if you locally-build a dev version, if you "update" it updates to locally built version
* there is no obvious difference between locally built and official versions in launcher

**To Reproduce**
1. install stride
2. locally build stride
3. run launcher, WITNESS "Stride 4.0 (update available)

**Expected behavior**
I expect the launcher to better separate "official releases" and "locally built" dev builds. I expect the launcher to only say "(update available)" when there is an official release update available.

**Screenshots**

This is what the launcher looks like when I'm on the current release (v4.0.0.1-vbeta02-0926) and I build a new local head-of-tree-version. Notice it says "(update available)" but that is just my locally built version.

image

There are also some more subtle bugs in the transition... watch this video...

[xenko launcher version confusion - YouTube](https://www.youtube.com/watch?v=7KJGdClGosI)

**Proposed Solution**
I think most of these problems would go away if locally built dev-versions got a different top-level item in the listing... As in "Stride 4.0-dev (locally built)":

image

cstdbool:
# PR Details

When the Launcher shows that a GameStudio package update is available, it will now show "(local update available)" in case if it is a local build (from disk, with location such as `%LocalAppData%\Stride\NugetDev`).

## Description

Previously it was not possible to distinguish between locally built dev packages and official packages from NuGet. This PR changes that by showing "(local update available)" instead of "(update available)" for locally built packages. See the screenshot below (`4.0.*.*` versions are official, whereas `4.1.*.*` versions were built locally).

launcher_example

## Related Issue

Closes #719 (proposed solution not implemented exactly as suggested, but instead just the local build indication was implemented - in my opinion, if the package was supposed to get its own separate version section entry, the actual version identifier should be changed in `/sources/shared/SharedAssemblyInfo.cs` file - for example, by appending `-dev` suffix).

## Motivation and Context

First time that I built Stride locally I was confused by this and actually thought that a new version was released in the meantime. It took me a bit to realize that building the solution will make the packages available to the officially installed launcher as well. It would be a good idea to merge this change to remove further confusion for newcomers.

## Types of changes

## Checklist

### DIFF
--- a/sources/launcher/Stride.Launcher/Resources/Strings.Designer.cs
+++ b/sources/launcher/Stride.Launcher/Resources/Strings.Designer.cs
@@ -722,6 +722,15 @@ namespace Stride.LauncherApp.Resources {
         }
         
         /// <summary>
+        ///   Looks up a localized string similar to (local update available).
+        /// </summary>
+        public static string VersionButtonLocalUpdateAvailable {
+            get {
+                return ResourceManager.GetString("VersionButtonLocalUpdateAvailable", resourceCulture);
+            }
+        }
+        
+        /// <summary>
         ///   Looks up a localized string similar to Visual Studio plugin.
         /// </summary>
         public static string VisualStudioPlugin {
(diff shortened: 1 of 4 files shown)


tebjan:
Wow thanks, big improvement for engine contributors.

PR #1057, opened 2021-04-27
Motion blur post process

ykafia:
# PR Details

This PR adds 2 motion blur techniques as post processing effects

## Description

1. [Chapman](http://john-chapman-graphics.blogspot.com/2013/01/per-object-motion-blur.html)/[NVIDIA](https://developer.nvidia.com/gpugems/gpugems3/part-iv-image-effects/chapter-27-motion-blur-post-processing-effect)
2. [Guertin, McGuire et al](https://casual-effects.com/research/Guertin2014MotionBlur/index.html)

The first technique is a simple and straight forward technique, it does wonders without adding a lot of computation complexity but has many caveats for those who aims for a very photorealistic result namely :
* Assumes all motion is linear
* Static object in the front can bleed on moving objects in the back
* The silouhette of moving objects is maintained, the blur is only inside objects
* Since the computation of screen space velocity is pixel precise, it does not compute realistic blur

The second technique tries to solve some of the previous issues but still have some caveats. The good point is that it's currently used in Unreal Engine (albeit with some modifications of their own), I'd like to see how Stride would handle this technique.

## Related Issue

#1058

## Motivation and Context

Motion blurs add to the realism of a scene and they look super cool!

## Types of changes

## Checklist

Change in the documentation include adding a page describing the effect and the different parameters.


ykafia:
Just for information the first motion blur technique is finished, it works, parameters works. I want to finish implementing the second one but i could move it into another PR ?

sources/engine/Stride.Rendering/Rendering/Images/MotionBlur/IMotionBlur.cs
```diff
@@ -0,0 +1,11 @@
+using Stride.Rendering.Images;
+
+namespace Stride.Rendering.Rendering.Images.MotionBlur
+{
+    public interface IMotionBlur : IImageEffect
+    {
+        bool NeedRangeDecompress { get; }
```
manio143, line 7: Seems you're not referencing this property anywhere - also naming wise, should it follow the `Requires*` naming convention or is `Need` something else?

ykafia: I've copy pasted the fields from the anti aliasing interface but you're right this property is not used, i removed it in my last commit!

ykafia:
Just something i noticed, and maybe due to my lack of math knowledge but the MeshVelocity computes velocity of animated meshes (skinning) but not when they are translated/rotated. Shader wise, the math seems to check out, i'm not sure if it's the intended behavior but i'll try to see if something can be different on the TransformRenderFeature

xen2:

PR #1071, opened 2021-06-08
[Reflection] Give the deps.json file a higher priority when loading assemblies by name

azeno:
# PR Details

## Description

The deps.json file preserves the compilation context - if it exists we can be sure that the assembly it points to by simple name is the correct one. Therefor it should get a higher priority when a dependency needs to be loaded by name.

## Motivation and Context

We ran into an issue where the same library was loaded multiple times from different locations (`.nuget\VL.Stride\...\VL.Stride.dll` as well as `src\PROJECT_X\bin\Release\VL.Stride.dll`) because a nuget dependency of `PROJECT_X` referenced it in a different version than the rest of the projects did. Such conflicts get resolved just fine by MSBuild and can easily be preserved using the `PreserveCompilationContext` switch.

## Types of changes

## Checklist


azeno:
Note that I target `master-4.0` on purpose because we can't switch to `4.1` yet. Would it be possible to make another `4.0` release?

xen2:
Sure, making a new 4.0 release shouldn't be a problem.

PR #1074, opened 2021-06-11
Dev/joreg/system pointer

joreg:
# PR Details

Introduces a SystemPointerId property which is different from the already existing PointerId in that it passes through the original, unmodified pointer Id provided by the OS.

## Description

Adds a SystemPointerId to PointerEvent and a SystemId to PointerPoint allowing the different framework implementations to pass through the original, unmodified Ids as provided by the OS. 

# Questions
* For now, I only implemented this for SDL. How would I do this for the other backends which I cannot test?
* Looking at the implementations of how the existing PointerId is generated in the different backends [iOS](https://github.com/stride3d/stride/blob/7e836297cb5930c01e6dfa0183e7f3cc64748fb6/sources/engine/Stride.Input/iOS/PointeriOS.cs#L120), [Android](https://github.com/stride3d/stride/blob/7e836297cb5930c01e6dfa0183e7f3cc64748fb6/sources/engine/Stride.Input/Android/PointerAndroid.cs#L89), [SDL](https://github.com/stride3d/stride/blob/7e836297cb5930c01e6dfa0183e7f3cc64748fb6/sources/engine/Stride.Input/SDL/PointerSDL.cs#L97) and [UWP](https://github.com/stride3d/stride/blob/7e836297cb5930c01e6dfa0183e7f3cc64748fb6/sources/engine/Stride.Input/UWP/PointerUWP.cs#L82), it is obvious that iOS and SDL work the same. Android might be the same, hard to understand without testing (the Android docs are not really vocal about the nature of the Id). And UWP, doesn't seem to modify the Id, ie may already currently be working differently than the others.. So how to deal with that?

## Motivation and Context

Under Windows, this solves the issue that we can now use an API like [InteractionContext](https://docs.microsoft.com/en-us/windows/win32/api/_input_intcontext/), which relies on those original Ids.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Input/PointerDeviceState.cs
+++ b/sources/engine/Stride.Input/PointerDeviceState.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Stride contributors (https://stride3d.net) and Silicon Studio Corp. (https://www.siliconstudio.co.jp)
+// Copyright (c) Stride contributors (https://stride3d.net) and Silicon Studio Corp. (https://www.siliconstudio.co.jp)
 // Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
 
 using System;
@@ -91,6 +91,7 @@ namespace Stride.Input
         public void UpdatePointerState(PointerEvent evt, bool updateDelta = true)
         {
             var data = GetPointerData(evt.PointerId);
+            data.Id = evt.PointerId;
 
             if (updateDelta)
             {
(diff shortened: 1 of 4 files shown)


xen2:
> For now, I only implemented this for SDL. How would I do this for the other backends which I cannot test?

Good question. Maybe we can have a default fallback behavior, as long as it's explained in the member documentation (default to 0 or random or `PointerId` or something like that)

joreg:
it now defaults to 0 on all implementations but SDL. so you mean just adding to the remarks of SystemPointerId: "NOTE: If this returns 0, then it is not implemented yet!"?

and how about the UWP case where I suspect the PointerId currently working like SystemPointerId. there, changing the behavior of PointerId to conform to the other implementations would be a breaking change. but one could argue it is a bugfix?!

Eideren:
I'm not very familiar with this so bare with me. For ``InteractionContext`` to work in your case you need a 'unique' value for each new ``PointerEvent`` ? Should those value match those reported by sdl exactly or can you emulate the behavior sdl uses to generate them and use those ?
If the latter is true, you could emulate that behavior for all other platforms and rename that variable to something that would reflect that behavior, that way we don't have any platform differences where there is no need to.

PR #1086, opened 2021-07-02
[Build] Couple of fixes to get it building with latest Visual Studio

azeno:
## Description

Not sure I did something completely wrong but with the following changes I could successfully compile again with latest VS (16.10.3). I was working on the `master-4.0` branch.

## Types of changes

## Checklist

### DIFF
--- a/sources/assets/Stride.Core.Assets/Stride.Core.Assets.csproj
+++ b/sources/assets/Stride.Core.Assets/Stride.Core.Assets.csproj
@@ -13,7 +13,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.Build" Version="16.0.461" ExcludeAssets="runtime" />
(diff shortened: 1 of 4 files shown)


xen2:
Sorry I wanted to fix the build server first but this will take a few more days.
This is probably fine so I will merge it.

xen2:
@azeno Curious, you remember what was the issue with `net5.0-windows` => `net5.0-windows7.0`?
I wonder if it should be ported to `master`.

azeno:
See my commit message - I put a link there. Didn't really look too deep into it.

xen2:
Thanks, I didn't see it.
Let me check if we need to update other places then (I could see some other references to `net5.0-windows` without `7.0`.

azeno:
Well as long as there's no need don't. Once we're on 4.0 and these build issues persist I'll come back here :)

PR #1089, opened 2021-07-13
Added a Goto method to Vectors

VexedUXR:
# PR Details

Added a Goto method that moves an object from its current position to another position.

## Motivation and Context

Usefull if you want to move an object to another object/point at a certain speed.

## Types of changes

### DIFF
--- a/sources/core/Stride.Core.Mathematics/Vector2.cs
+++ b/sources/core/Stride.Core.Mathematics/Vector2.cs
@@ -208,7 +208,7 @@ namespace Stride.Core.Mathematics
         }
 
         /// <summary>
-        /// Moves the first vector3 to the second one in a straight line.
+        /// Moves the first vector2 to the second one in a straight line.
         /// </summary>
         /// <param name="First">The first point.</param>
         /// <param name="Second">The second point.</param>
--- a/sources/core/Stride.Core.Mathematics/Vector2.cs
+++ b/sources/core/Stride.Core.Mathematics/Vector2.cs
@@ -208,6 +208,24 @@ namespace Stride.Core.Mathematics
         }
 
         /// <summary>
+        /// Moves the first vector3 to the second one in a straight line.
+        /// </summary>
+        /// <param name="First">The first point.</param>
+        /// <param name="Second">The second point.</param>
+        /// <param name="Speed">The rate at which the first point is going to move towards the second point.</param>
+        public static Vector2 Goto(Vector2 First, Vector2 Second, float Speed)
+        {
+            Vector2 Distance = Subtract(Second, First);
+
+            float length = (float)Math.Sqrt(Distance.X * Distance.X + Distance.Y * Distance.Y);
+
+            if (length == 0 || Speed >= length)
+                return Second;
+
+            return new Vector2(First.X + Distance.X / length * Speed, First.Y + Distance.Y / length * Speed);
+        }
+
+        /// <summary>
         /// Adds two vectors.
         /// </summary>
         /// <param name="left">The first vector to add.</param>
--- a/sources/core/Stride.Core.Mathematics/Vector2.cs
+++ b/sources/core/Stride.Core.Mathematics/Vector2.cs
@@ -210,19 +210,34 @@ namespace Stride.Core.Mathematics
         /// <summary>
         /// Moves the first vector2 to the second one in a straight line.
         /// </summary>
-        /// <param name="First">The first point.</param>
-        /// <param name="Second">The second point.</param>
-        /// <param name="MaxDistancePerCall">The rate at which the first point is going to move towards the second point.</param>
-        public static Vector2 Goto(Vector2 First, Vector2 Second, float MaxDistancePerCall)
+        /// <param name="from">The first point.</param>
+        /// <param name="to">The second point.</param>
+        /// <param name="maxDistancePerCall">The rate at which the first point is going to move towards the second point.</param>
+        /// <param name="result">When the method completes, contains a version of the Vector that is closer to the second vector.</param>
+        public static void Goto(ref Vector2 from, ref Vector2 to, ref float maxDistancePerCall, out Vector2 result)
         {
-            Vector2 Distance = Subtract(Second, First);
+            Vector2 distance;
+            Subtract(ref to, ref from, out distance);
 
-            float length = (float)Math.Sqrt(Distance.X * Distance.X + Distance.Y * Distance.Y);
+            float length = distance.Length();
 
-            if (length == 0 || MaxDistancePerCall >= length)
-                return Second;
+            if (maxDistancePerCall >= length || length == 0)
+                result = to;
+            else
+                result = new Vector2(from.X + distance.X / length * maxDistancePerCall, from.Y + distance.Y / length * maxDistancePerCall);
+        }
 
-            return new Vector2(First.X + Distance.X / length * MaxDistancePerCall, First.Y + Distance.Y / length * MaxDistancePerCall);
+        /// <summary>
+        /// Moves the first vector2 to the second one in a straight line.
+        /// </summary>
+        /// <param name="from">The first point.</param>
+        /// <param name="to">The second point.</param>
+        /// <param name="maxDistancePerCall">The rate at which the first point is going to move towards the second point.</param>
+        public static Vector2 Goto(Vector2 from, Vector2 to, float maxDistancePerCall)
+        {
+            Vector2 result;
+            Goto(ref from, ref to, ref maxDistancePerCall, out result);
+            return result;
         }
 
         /// <summary>
--- a/sources/core/Stride.Core.Mathematics/Vector2.cs
+++ b/sources/core/Stride.Core.Mathematics/Vector2.cs
@@ -212,19 +212,19 @@ namespace Stride.Core.Mathematics
         /// </summary>
         /// <param name="from">The first point.</param>
         /// <param name="to">The second point.</param>
-        /// <param name="maxDistancePerCall">The rate at which the first point is going to move towards the second point.</param>
+        /// <param name="maxTravelDistance">The rate at which the first point is going to move towards the second point.</param>
         /// <param name="result">When the method completes, contains a version of the Vector that is closer to the second vector.</param>
-        public static void Goto(ref Vector2 from, ref Vector2 to, ref float maxDistancePerCall, out Vector2 result)
+        public static void MoveTo(ref Vector2 from, ref Vector2 to, float maxTravelDistance, out Vector2 result)
         {
             Vector2 distance;
             Subtract(ref to, ref from, out distance);
 
             float length = distance.Length();
 
-            if (maxDistancePerCall >= length || length == 0)
+            if (maxTravelDistance >= length || length == 0)
                 result = to;
             else
-                result = new Vector2(from.X + distance.X / length * maxDistancePerCall, from.Y + distance.Y / length * maxDistancePerCall);
+                result = new Vector2(from.X + distance.X / length * maxTravelDistance, from.Y + distance.Y / length * maxTravelDistance);
         }
 
         /// <summary>
@@ -232,11 +232,11 @@ namespace Stride.Core.Mathematics
         /// </summary>
         /// <param name="from">The first point.</param>
         /// <param name="to">The second point.</param>
-        /// <param name="maxDistancePerCall">The rate at which the first point is going to move towards the second point.</param>
-        public static Vector2 Goto(Vector2 from, Vector2 to, float maxDistancePerCall)
+        /// <param name="maxTravelDistance">The rate at which the first point is going to move towards the second point.</param>
+        public static Vector2 MoveTo(Vector2 from, Vector2 to, float maxTravelDistance)
         {
             Vector2 result;
-            Goto(ref from, ref to, ref maxDistancePerCall, out result);
+            MoveTo(ref from, ref to, maxTravelDistance, out result);
             return result;
         }
 
--- a/sources/core/Stride.Core.Mathematics/Vector2.cs
+++ b/sources/core/Stride.Core.Mathematics/Vector2.cs
@@ -212,17 +212,17 @@ namespace Stride.Core.Mathematics
         /// </summary>
         /// <param name="First">The first point.</param>
         /// <param name="Second">The second point.</param>
-        /// <param name="Speed">The rate at which the first point is going to move towards the second point.</param>
-        public static Vector2 Goto(Vector2 First, Vector2 Second, float Speed)
+        /// <param name="MaxDistancePerCall">The rate at which the first point is going to move towards the second point.</param>
+        public static Vector2 Goto(Vector2 First, Vector2 Second, float MaxDistancePerCall)
         {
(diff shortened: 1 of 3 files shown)


sources/core/Stride.Core.Mathematics/Vector2.cs
```diff
@@ -207,6 +207,24 @@ public float[] ToArray()
             return new float[] { X, Y };
         }
 
+        /// <summary>
+        /// Moves the first vector2 to the second one in a straight line.
+        /// </summary>
+        /// <param name="First">The first point.</param>
+        /// <param name="Second">The second point.</param>
+        /// <param name="MaxDistancePerCall">The rate at which the first point is going to move towards the second point.</param>
+        public static Vector2 Goto(Vector2 First, Vector2 Second, float MaxDistancePerCall)
```
tebjan, line 216: method input parameters should be lower case to match the convention in the stride source code.

tebjan, line 216: the naming could be `from` and `to` which would be more expressive.

sources/core/Stride.Core.Mathematics/Vector2.cs
```diff
@@ -207,6 +207,24 @@ public float[] ToArray()
...
+        {
+            Vector2 Distance = Subtract(Second, First);
```
tebjan, line 218: here you could use the subtract with `ref/out` parameters to solve some copies.

sources/core/Stride.Core.Mathematics/Vector2.cs
```diff
@@ -207,6 +207,24 @@ public float[] ToArray()
...
+
+            float length = (float)Math.Sqrt(Distance.X * Distance.X + Distance.Y * Distance.Y);
+
+            if (length == 0 || MaxDistancePerCall >= length)
```
tebjan, line 222: swap cases, the second one is more likely which will exit the check early.

sources/core/Stride.Core.Mathematics/Vector2.cs
tebjan, line 216: there could be a version with `ref/out` parameters to save copies. and the version without them will just call the other, like it is done with most math functions.

sources/core/Stride.Core.Mathematics/Vector2.cs
```diff
@@ -207,6 +207,24 @@ public float[] ToArray()
...
+        /// Moves the first vector2 to the second one in a straight line.
+        /// </summary>
+        /// <param name="First">The first point.</param>
+        /// <param name="Second">The second point.</param>
+        /// <param name="MaxDistancePerCall">The rate at which the first point is going to move towards the second point.</param>
+        public static Vector2 Goto(Vector2 First, Vector2 Second, float MaxDistancePerCall)
+        {
+            Vector2 Distance = Subtract(Second, First);
+
+            float length = (float)Math.Sqrt(Distance.X * Distance.X + Distance.Y * Distance.Y);
```
tebjan, line 220: here you can use the `Length` method with `ref/out` params.

tebjan (changes requested):
Hello and thanks a lot for the addition. I've added a few improvements for naming and performance.

it could also be discussed whether this should go into the https://github.com/dfkeenan/StrideToolkit package because the methods are non-essential....

sources/core/Stride.Core.Mathematics/Vector2.cs
VexedUXR: alright but should I make all of them lower case or just the M for MaxDistancePerCall?

tebjan: The first letter of all parameters should be lower case.

VexedUXR:
> it could also be discussed whether this should go into the https://github.com/dfkeenan/StrideToolkit package because the methods are non-essential....

Should I make a PR there?

Eideren:
@tebjan I don't think it would make sense to move it to the toolkit; most of the math functions are not essential (``CatmullRom`` is used once maybe, ``Lerp`` is trivial to write inline, etc.), but they are still part of the API. Those are not shortcuts to another part of the API either and it doesn't require any maintenance since math stuff is fairly static. What do you think ?

sources/core/Stride.Core.Mathematics/Vector2.cs
```diff
@@ -210,19 +210,34 @@ public float[] ToArray()
...
         /// </summary>
-        /// <param name="First">The first point.</param>
-        /// <param name="Second">The second point.</param>
-        /// <param name="MaxDistancePerCall">The rate at which the first point is going to move towards the second point.</param>
-        public static Vector2 Goto(Vector2 First, Vector2 Second, float MaxDistancePerCall)
+        /// <param name="from">The first point.</param>
+        /// <param name="to">The second point.</param>
+        /// <param name="maxDistancePerCall">The rate at which the first point is going to move towards the second point.</param>
+        /// <param name="result">When the method completes, contains a version of the Vector that is closer to the second vector.</param>
+        public static void Goto(ref Vector2 from, ref Vector2 to, ref float maxDistancePerCall, out Vector2 result)
```
tebjan, line 217: great, just one minor thing, single float values don't need to be passed by ref, only vectors. also, the name could be `maxTravelDistance`. as the "perCall" already implies a certain usage of the method.

sources/core/Stride.Core.Mathematics/Vector2.cs
```diff
@@ -210,19 +210,34 @@ public float[] ToArray()
         /// <summary>
         /// Moves the first vector2 to the second one in a straight line.
         /// </summary>
-        /// <param name="First">The first point.</param>
-        /// <param name="Second">The second point.</param>
-        /// <param name="MaxDistancePerCall">The rate at which the first point is going to move towards the second point.</param>
-        public static Vector2 Goto(Vector2 First, Vector2 Second, float MaxDistancePerCall)
```
tebjan: In general, i think `MoveTo` is a better name. `Goto` might be confused with the `goto` language statement.

tebjan (changes requested):
Good, found two other small things. Otherwise, looking good!

tebjan:
> @tebjan I don't think it would make sense to move it to the toolkit; most of the math functions are not essential (`CatmullRom` is used once maybe, `Lerp` is trivial to write inline, etc.), but they are still part of the API. Those are not shortcuts to another part of the API either and it doesn't require any maintenance since math stuff is fairly static. What do you think ?

yes, you're right. it is totally fine to add it to the math lib.

sources/core/Stride.Core.Mathematics/Vector2.cs
VexedUXR: alright 👍

sources/core/Stride.Core.Mathematics/Vector4.cs
```diff
@@ -282,6 +282,39 @@ public float[] ToArray()
...
+        {
+            Vector4 distance;
+            Subtract(ref to, ref from, out distance);
+
+            float length = distance.Length();
+
+            if (maxTravelDistance >= length || length == 0)
+                result = to;
+            else
+                result = new Vector4(from.X + distance.X / length * maxTravelDistance, from.Y + distance.Y / length * maxTravelDistance, from.Z + distance.Z / length * maxTravelDistance, from.W + distance.W / length * maxTravelDistance);
```
Eideren, line 302:
```
var v = 1f / length * maxTravelDistance;
result = new Vector4(from.X + distance.X * v, from.Y + distance.Y * v, from.Z + distance.Z * v, from.W + distance.W * v);
```
Would be faster, three less divs for one more mult, but it doesn't really matter at this point.

tebjan: Right, its common practice in the stride repo to do that, so indeed recommended.

sources/core/Stride.Core.Mathematics/Vector2.cs
xen2: Shouldn't we use the `in` keyword and have only a single version of this function?

xen2: From my understanding, it should be fine if we add `readonly` to `Length()`, which is the only non-field (property) accessed by the method.

Eideren: > Not sure about if compiler would make defensive copies for such a struct or not though, anybody knows?

It shouldn't trigger a defensive copy as long as we don't use struct properties, or at least that's how it should work, maybe they changed it in newer versions but I couldn't repro that quirk.
Benchmark with net6 showed no perf' difference between passing by ``in`` or by ``ref``, by value was significantly slower of course.
```cs
[Benchmark]
public void PassedByValue()
{
    Vector3 a = new Vector3( 10, 5, 9 ), b = new Vector3( 8, 6, -10 );
    for( int i = 0; i < 100000; i++ )
    {
        a = Vector3.MoveTo( a, b, 0.15f );
        a.X += 10f;
    }
}

[Benchmark]
public void PassedByRef()
{
    Vector3 a = new Vector3( 10, 5, 9 ), b = new Vector3( 8, 6, -10 );
    for( int i = 0; i < 100000; i++ )
    {
        Vector3.MoveTo( ref a, ref b, 0.15f, out a );
        a.X += 10f;
    }
}

[Benchmark]
public void PassedByIn()
{
    Vector3 a = new Vector3( 10, 5, 9 ), b = new Vector3( 8, 6, -10 );
    for( int i = 0; i < 100000; i++ )
    {
        a = Vector3.MoveToIn( a, b, 0.15f );
        a.X += 10f;
    }
}
```

**That said though, we have been going back and forth with @VexedUXR for a while now, it would be kind of rude to ask of him to change stuff again, one of us three should incorporate the rest of what's required and merge this in. I'll get to that tomorrow if neither of you had the time to do it by then** :)

If you want to take a look at the IL: 
https://sharplab.io/#v2:C4LglgNgPgAgTABgLACh4EZWpugnACgCJAeDcBB9wgSgG4sVsBmAAgGdgAnAVwGNhGA1AKY8A9m3qoA3qkYzZMmExwA2RmwEBDACbCAdhACejMDt4BlMAC8BASR0AhfcAHNGAXkb0AVM0sDhAM3x/CGF1YGpaOTkFRmVVDW09Q0ERMUYALQE2YTdGHQEAd34hYFF6fAi6FCjZGLi1LV0DYtSmAFUdMGAADVz8opTSsXx0ADoEfwAaRgRxqZm5yuka2MV0FQbE5sGyxg6ugE0+wpah8tmJ6bHLhYml6pW69fjGpNPd/eB044GSsvwLvNAVdFjQqitVrFnpsmsk/mkAPL5H7vYbXebokF3MHLGoAegJjAAPMwOABbMnqNj6AB8uKiBLxjAAKgALASMXpcYRkgAOugEJkYAUYwHZjAAbvDRvS5IziXjSRSqbTZbUmMFQrxujiHvjCSTyZTqXS9QzCWyOUduXyBUKRWKOVLUjKzXKDYqjSrTRCYpqwowDrqIfLDcqTWqZPLLRlGDb+fl7f5ReLnUNXSGPUrjaq3erGP7eOlg48mDthoXJeoIBwBBRI4wpHmar13BLq7WwRC5Ec2x2BF3uzJvn2awOGwBfSKl1HlSsAD2mlf0S5CAYs9ebjYbclbjHng6HvcY+kP3ZHjAsZ7kU7NDZiACNhMIIIxrMwAHKiSkQXyaBtNkOMgAOYCLwEixAA7IwACyYSsqMACCD7MPg+C9J4nIUIwADUjD4EcmEHNheH4N8mHpNhAC0jDoP42FEjM/hUIwt4QmxM6VmKYDMAA2sYvDGJoAjzgAugBO6yKBwCSTIgFAXIzAFF0XCsvhQkiZuClyPJ2lRFw6jMByCAgFBnLXnpcaGRy6CmTA0FBrJ3YGUZjBwHZ0HFk5N7TnpYrZEU/SMIhbDAeSgrAAiHCRf4ABK6g6KBACi85cAIvLAGAuhEBp86ENMhC2JoYBpS4/iiLOqgcDoBbZGSMyijkcDTMYXA1j4UqjJQFk1LeTlGTJW46d5shKSpan4LlWmWbplkyC5xmmXu7ZjixD4NAA1j12kLbRpnHitnaMOtGhbSNNS7e5sajkdJ3qGdQ1AcJ/jqBwECgCmAV5CcIVhWSEVRTF8WJQIKVpRlWU6DlOjCXlBVFSVzgFhV5ZMJwNX+HVDWlG5LU6G1HAdQIXWVHNHFDuTUR9VufprrwAAygrAWKFQSY98jQUEdMUHBYqjKYACObDAGhGFYbh+GEYGJH4eRGQUKTKzU76Gp04wjOJWKgscFSAiaKzW6zb6nNi90MsEYwRHm3LlHbROjD0veawqDAAAssHCFKzLCPgajJqjtU8tMfuVaUq5aowlLzsybDqFKEAACI8cACVpdMwjRZVaiku9001EbKwB8VrCp+O7OMKYHAPuw6g8L7AjJmH8TJpjQfCpnxcp/jdYlkOlYQEzYq5J3pejBrzOshUvfdmAyb4FHMdxwIifJ6XjA0u4A+a2pUBQIwW8T247gIHn2nZ29vDuKU21yMvRnnfEOeX99vytEEdWjL0eEj93n+MEyB8h6YQXrHeOScS7dyXB/I439V6/yOAAwealgHqGjqA5e4Cu5p0DmSUY3xYEQLSng/++8kGW0jqgxeYC4FpUVr1XyUQngqADjBT2Ahvb4ADq3Mk0wA5N0rCApeK9CF1jZgpAO593o31kKwr2PsQ7cODg3Rq0xBHUJEenTOkjwjSI5o/C+dsGHRGdpVWR7DhC2EmjVLhdU8ah2EOHAMaiME0NEYbJyRdXG6IrlXGuPBLH8Nse3XgP9aHTwhP3Mh7hQnE3Hize4ClZ74WccIrBHIN6kO3owXemTD6uGPqfBS9lGreLvgIB+xSgqo3fjyP+BC0l/0QVklBaChGYNLlA2pMDGAxNGAg3JQCKGtPUWkzpuD8E9NccQpph8WlUJcSIuhVMjH5jiG7Hx1dY51xDgHAe/hgBKP9vCNGYBgKsgOcErOzgL6FNkAXGo2iUTVL2cAP+NE2CnPOZ/aYLy+mMHeZ815BwfkN1ed8AFZywVLJ8uCGcaz3aV02bXYAljjCVReXYiRgLNG8Akdc3OYigKPPcFU45+Bfm9AhV87oIL9l/KpUC2lYL/mqEBXg6FshbwTiAA==

xen2: Interesting, I though calling `Length()` would make it do defensive copies.
Totally agree with one of us doing it, sorry for chiming in late.

tebjan: Ah didn't know about this improvement, i thought it's nice to have it in sync with the other method. So we can also keep it as it's now and clean up the ref methods with all the others at a later point.

@VexedUXR so basically do with this info as you please :-)

Eideren:
Continued in PR #1090 as I don't have write access to this one.

PR #1096, opened 2021-07-25
Setting an entity's parent to null makes it a root entity

### LINKED ISSUE #1095: Setting entity parent to null removes it from the scene
**Release Type**: Official Release/GitHub

**Describe the bug**
If you set an entity's parent to null (`Entity.Transform.Parent = null`) the entity gets removed from the scene.

**Expected behavior**
I expected the entity to become a root entity in the scene.

salahchafai:
# PR Details
If you set the parent of an entity to null it won't be removed from the scene

## Related Issue
Closes #1095 

## Types of changes

### DIFF
--- a/sources/engine/Stride.Engine/Engine/TransformComponent.cs
+++ b/sources/engine/Stride.Engine/Engine/TransformComponent.cs
@@ -204,12 +204,21 @@ namespace Stride.Engine
                 var oldParent = Parent;
                 if (oldParent == value)
                     return;
-                
-                // SceneValue must be null if we have a parent
-                if( Entity.SceneValue != null )
-                    Entity.Scene = null;
 
                 var previousScene = oldParent?.Entity?.Scene;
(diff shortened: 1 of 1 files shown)


Eideren:
This is a **major** breaking change, I'm not sure how we're supposed to communicate to our users that they should fix their stuff to align with this PR, and I'm not convinced that users expect this behavior more than the current one or that it communicates better how the scene/hierarchy system works under the hood.

tebjan:
I understand why it might be odd that setting the parent to null will also set the scene to null, but I'd also say that a change could break too much existing code. I'd rather improve documentation to inform the programmer what happens and what options there are. Also, we could add an extension method that makes the entity a root entity in the scene...

salahchafai:
yep right, changing this might break many projects, would be better to just improve the docs so users know what happens when setting the parent to null

PR #1099, opened 2021-07-28
[Math] Math -> MathF

Eideren:
# PR Details
The majority of cases of ``(float)Math.SomeFunctionThatTakesAndReturnsDouble(aFloatInput)`` uses ``MathF`` now, skipping unnecessary double precision operation, the cast to double and back to float.

A couple of minor cases using doubles as input were converted as well; ``LightSpot.cs`` at line 156: ``LightRadiusAtTarget = (float)Math.Abs(Range * Math.Sin(AngleOuterInRadians * 0.5f));``, converting both of those to ``MathF`` would skip the improved precision of ``Abs(Range*Sin)``. For those cases, I didn't think the improved precision was significant enough to leave them in.

Other cases might benefit from these changes but were left untouched as it would require proper testing, see ``SphericalHarmonics.cs``.

``Abs()``, ``Min()``, ``Max()``, and a couple of other minor functions have float32 version within ``Math`` but were converted for consistency with other usages within the file.

## Related Issue
None

## Motivation and Context
Cleaner and more efficient code base.

## Types of changes

## Checklist

Sorry, there's a ton of changes, but I triple-checked all of them. If you still want to go through and review all of this, godspeed to you.

### DIFF
--- a/sources/engine/Stride.Engine/Engine/TransformComponent.cs
+++ b/sources/engine/Stride.Engine/Engine/TransformComponent.cs
@@ -146,8 +146,7 @@ namespace Stride.Engine
                 float xw = rotation.X * rotation.W;
 
                 rotationEuler.Y = MathF.Asin(2.0f * (yw - zx));
-                double test = Math.Cos(rotationEuler.Y);
-                if (test > 1e-6f)
+                if (MathF.Cos(rotationEuler.Y) > MathUtil.ZeroTolerance)
                 {
                     rotationEuler.Z = MathF.Atan2(2.0f * (xy + zw), 1.0f - (2.0f * (yy + zz)));
                     rotationEuler.X = MathF.Atan2(2.0f * (yz + xw), 1.0f - (2.0f * (yy + xx)));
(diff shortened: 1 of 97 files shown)


sources/engine/Stride.Engine/Engine/TransformComponent.cs
```diff
@@ -145,16 +145,16 @@ public Vector3 RotationEulerXYZ
                 float yz = rotation.Y * rotation.Z;
                 float xw = rotation.X * rotation.W;
 
-                rotationEuler.Y = (float)Math.Asin(2.0f * (yw - zx));
+                rotationEuler.Y = MathF.Asin(2.0f * (yw - zx));
                 double test = Math.Cos(rotationEuler.Y);
```
xen2, line 149: Found this one leftover

xen2:
LGTM, thanks!

sources/engine/Stride.Engine/Engine/TransformComponent.cs
Eideren: If you're referring to the ``double test = ...`` yes, I didn't end up changing stuff that used the double precision just to stay safe but this one is fine.

PR #1100, opened 2021-08-04
[Physics] Improve collision and contact harvesting

Eideren:
# PR Details
Retrieving collision and contact information was previously done by re-testing all components for collisions ... which obviously led to awful performance for physics heavy scenes (could take up to and above 90% of the frame).
This PR instead just reads from bullet's collisions collection.

Contacts are now lazily evaluated to reduce overhead when nothing actually ends up reading them.
Users can now read and iterate over all collisions through ``Simulation.CurrentCollisions``.

I introduced some breaking changes though, ``NewContact``, ``ContactUpdate`` and ``ContactEnded`` channels have been removed and Collision is now a struct instead of a pooled class.

On the removal of ``NewContact``, ``ContactUpdate`` and ``ContactEnded``:
I don't think that's something we should leave in; when some logic awaits new contacts or contact updates and that collision is removed the logic will never resume nor throw by itself, workarounds for this issue aren't straightforward to setup.
``ContactChanged()`` though will always resume those awaiting when collision exits since the collision got from ``x`` contacts to no contacts, the user can then test if there aren't any contacts or call ``AreColliding`` to continue await if it is still colliding.

On changing Collision from class to struct:
We don't have to worry about sharing this anymore, users can now safely store those if they need to.

## Related Issue
None reported in the GitHub issues.

## Motivation and Context
Performance, safety and usability.

## Types of changes

## Checklist


sources/engine/Stride.Physics/Collision.cs
```diff
@@ -67,70 +28,56 @@ internal void Destroy()
...
 
-        internal bool InternalEquals(PhysicsComponent a, PhysicsComponent b)
+        public override int GetHashCode()
         {
-            return (ColliderA == a && ColliderB == b) || (ColliderB == a && ColliderA == b);
+            int aH = ColliderA.GetHashCode();
+            int bH = ColliderB.GetHashCode();
+            // This ensures that a pair of components will return the same hash regardless
+            // of if they are setup as [b,a] or [a,b]
+            return aH > bH ? HashCode.Combine(aH, bH) : HashCode.Combine(bH, aH);
```
xen2, line 80: Are we sure `aH != bH`?
I suppose `ColliderA` and `ColliderB` can't be the same?
Still, there could be a collision hash (especially if poorly implemented).

Eideren: I don't think I understand what you meant here ?
If they have the same hash values the order you combine them in doesn't matter;
Let's say ``aH`` is ``10``, now if ``aH == bH`` then ``bH`` must be ``10`` as well, so the call would look like ``HashCode.Combine(10, 10)``, swapping the tens around wont change the resulting hash generated since it's the same value ... ?

xen2: Yes obviously it's fine, please disregard my comment, I must have been too tired when I reviewed :)

PR #1101, opened 2021-08-07
Fix the dictionary editor

D3ZAX:
- Fix the problem dictionary with int type key can't change key in editor.
- Add support to dictionary editor with key type as char, bool, or other number type, and can modify key now.


D3ZAX:
It appears that the CLA show wrong user for this pull request and I can't sign CLA with that user. I will try another pull request.

PR #1105, opened 2021-08-08
[Physics] Raycast and ShapeSweep can filter out triggers, true by default

Eideren:
# PR Details
Users most likely wouldn't expect triggers to be returned when looking for collisions through those one shot tests given that triggers don't collide with anything.
If they still would like to test for those as well, they can do so by setting the appropriate boolean when calling those methods.

## Related Issue
None reported.

## Motivation and Context
Bug fix, kind of.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Physics/Simulation.cs
+++ b/sources/engine/Stride.Physics/Simulation.cs
@@ -513,10 +513,11 @@ namespace Stride.Physics
         /// <param name="to">To.</param>
         /// <param name="filterGroup">The collision group of this raycast</param>
         /// <param name="filterFlags">The collision group that this raycast can collide with</param>
-        /// <returns>The list with hit results.</returns>
-        public HitResult Raycast(Vector3 from, Vector3 to, CollisionFilterGroups filterGroup = DefaultGroup, CollisionFilterGroupFlags filterFlags = DefaultFlags)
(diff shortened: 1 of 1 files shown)


sources/engine/Stride.Physics/Simulation.cs
```diff
@@ -1170,11 +1176,26 @@ public virtual void Recycle(ref Vector3 from, ref Vector3 to, CollisionFilterGro
...
+                    return false;
+                }
+
+                return base.NeedsCollision(proxy0);
             }
         }
 
         private abstract class StrideReusableConvexResultCallback : BulletSharp.ConvexResultCallback
         {
+            public bool HitNoContResp;
```
xen2, line 1197: Can we use a non-abbreviated name so that it's easier to understand?
If I understood correctly, this is the value of `hitTriggers` whose doc is "Whether this test should collide with <see cref="PhysicsTriggerComponentBase"/>". Maybe we can copy-paste this doc here too?

Eideren:
@xen2 I added a summary to that field and refactored the callbacks a bit, let me know what you think.

xen2:
Perfect, thanks!
Ready to merge?

Eideren:
Sure, go ahead :)

Issue #957, opened 2020-12-11
OpenXR Support


tebjan:
**Is your feature request related to a problem? Please describe.**
Add a VR implementation based on OpenXR, similar to the current OpenVR integration.

This also came up in #722 

**Additional context**
This could be attractive because Facebook might sponsor OpenXR development, as they did for the Godot engine.

phr00t:
FYI, I just finished implementing and testing OpenXR support in Focus Engine with success:

https://github.com/phr00t/FocusEngine/tree/master/sources/engine/Xenko.VirtualReality/OpenXR

You might be able to use some of this code -- although it is targeting Vulkan.

tebjan:
That's very cool, thanks for letting us know! That will definitely become handy as soon as someone works on it.

Eideren:
While a PR regarding this issue was implemented (#1483), it is not yet supported on all graphics API

Eideren:
Current state of the implementation running on Quest3 from @Aggror :
> Turned on dev mode the quest 3
> Started a new project with the VR template.
> Connected both through cable as well as Airlink
> Checked the existing documentation but everything was already setup correctly.
> ... but unfortunately not working right away. The split view in the editor works, but I was not seeing any updates regarding movement nor updates on the quest 3 itself.
> interestingly when I start the application, MetaVr en SteamVR start automatically as well. 
> launching Airlink on the quest 3, you can actually see the latest build from stride in the Overlay menu.
> I haven't gotten it to work natively/standalone on quest 3. Through virtual desktop, no problem though

MelvinG24:
> Current state of the implementation running on Quest3 from @Aggror :
> 
> > Turned on dev mode the quest 3
> > Started a new project with the VR template.
> > Connected both through cable as well as Airlink
> > Checked the existing documentation but everything was already setup correctly.
> > ... but unfortunately not working right away. The split view in the editor works, but I was not seeing any updates regarding movement nor updates on the quest 3 itself.
> > interestingly when I start the application, MetaVr en SteamVR start automatically as well.
> > launching Airlink on the quest 3, you can actually see the latest build from stride in the Overlay menu.
> > I haven't gotten it to work natively/standalone on quest 3. Through virtual desktop, no problem though

[...]

PR #1107, opened 2021-08-10
[WIP] OpenXR for Stride

anotherdevguy:
# PR Details

This will add support for OpenXR, at least for Vulkan. May not support all features provided in Focus Engine, as I'm trying not to modify the overall Stride system as much as possible. I never implemented it for DirectX and have no real experience with that API, so I only feel confident submitting the Vulkan implementation. It shouldn't be too difficult to convert (or add support for) DirectX.

## Description

Adds OpenXR as a selectable VR API.

## Related Issue

https://github.com/stride3d/stride/issues/957

## Motivation and Context

OpenXR is the future of VR support. Also, you guys offered help getting Focus Engine to target .NET Core 5 😄 

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.VirtualReality/OpenXR/OpenXrTouchController.cs
+++ b/sources/engine/Stride.VirtualReality/OpenXR/OpenXrTouchController.cs
@@ -1,0 +1,167 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+using Silk.NET.OpenXR;
+using Stride.Core.Mathematics;
+using Stride.Games;
+
+namespace Stride.VirtualReality
+{
+    class OpenXrTouchController : TouchController
+    {
+        private string baseHandPath;
+        private OpenXRHmd baseHMD;
+        private SpaceLocation handLocation;
+        private TouchControllerHand myHand;
+
+        public ulong[] hand_paths = new ulong[12];
+
+        public Space myHandSpace;
+        public Silk.NET.OpenXR.Action myHandAction;
+
+        public OpenXrTouchController(OpenXRHmd hmd, TouchControllerHand whichHand)
+        {
+            baseHMD = hmd;
+            handLocation.Type = StructureType.TypeSpaceLocation;
+            myHand = whichHand;
+            myHandAction = OpenXRInput.MappedActions[(int)myHand, (int)OpenXRInput.HAND_PATHS.Position];
+
+            ActionSpaceCreateInfo action_space_info = new ActionSpaceCreateInfo()
+            {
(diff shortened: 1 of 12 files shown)


anotherdevguy:
> [CLA assistant check](https://cla-assistant.io/stride3d/stride?pullRequest=1107)
> Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/stride3d/stride?pullRequest=1107) before we can accept your contribution.
> You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.io/check/stride3d/stride?pullRequest=1107) it.

I replied to you, silly CLAssistant! recheck doesn't seem to help 😢

sources/engine/Stride.VirtualReality/OpenXR/OpenXrTouchController.cs
```diff
@@ -0,0 +1,167 @@
+using System;
```
xen2, line 1: Need copyright header (applies to other files)
(might need to check copyright permission with original author).

phr00t:
Where this pull request stands:

It builds, and "should work" -- problem is, this dynamic call:

https://github.com/anotherdevguy/stride/blob/openxr/sources/engine/Stride.VirtualReality/OpenXR/OpenXRHmd.cs#L406

... should be telling OpenXR which Vulkan device is in-use in addition to setting a handle to that device in physicalDevice. This code works in my fork, but fails for some reason on Stride (physicalDevice comes back as null). If I try using the Vulkan's physical device from GraphicsDevice.Vulkan.cs, you'll get an OpenXR validation error saying you have to use xrGetVulkanGraphicsDeviceKHR to get the handle.

However, this could all be bypassed by using DirectX instead, which seems more prevalent on Stride. You could use https://www.khronos.org/registry/OpenXR/specs/1.0/html/xrspec.html#XrGraphicsRequirementsD3D11KHR instead, and use the XR_KHR_D3D11_enable extension (instead of "XR_KHR_vulkan_enable") up on line 89 of OpenXRHmd.cs (to get started on using DirectX instead of Vulkan).

Eideren has passed on his NET 5 work to me, so I'm going to pick up where he left off (thank you!). As I'm switching back to that, I won't be driving further development on this pull request for now -- but will definitely answer questions and help out where I can.

tebjan:
Depends on #1425

Eideren:
It sounds like #1483 supersedes this one ? Should we close this PR ?
@arturoc @tebjan

tebjan:
Yes!

PR #1108, opened 2021-08-12
 Fix the dictionary editor

D3ZAX:
- Fix the problem dictionary with int type key can't change key in editor.
- Add support to dictionary editor with key type as char, bool, or other number type, and can modify key now.


JackPilley:
It looks like it's this commit https://github.com/stride3d/stride/pull/1108/commits/4aa5c9caee5fb6fdfb79b40fbe546a79fc13670c

PR #1111, opened 2021-08-14
Solve editor problem with dictionary

D3ZAX:
Fix the problem dictionary with int type key can't change key in editor.
Add support to dictionary editor with key type as char, bool, or other number type, and can modify key now.

### DIFF
--- a/sources/editor/Stride.Core.Assets.Editor/Quantum/NodePresenters/Commands/AddPrimitiveKeyCommand.cs
+++ b/sources/editor/Stride.Core.Assets.Editor/Quantum/NodePresenters/Commands/AddPrimitiveKeyCommand.cs
@@ -55,10 +55,8 @@ namespace Stride.Core.Assets.Editor.Quantum.NodePresenters.Commands
             var value = nodePresenter.Value;
 
             NodeIndex? newKey;
-			if (dictionaryDescriptor.KeyType == typeof(string))
-                newKey = GenerateStringKey(value, dictionaryDescriptor, parameter as string);
-            else if (TypeDescriptor.GetConverter(dictionaryDescriptor.KeyType).CanConvertFrom(typeof(string)))
-                newKey = GenerateGenericKey(value, dictionaryDescriptor, parameter as string);
+			if (parameter != null && TypeDescriptor.GetConverter(dictionaryDescriptor.KeyType).CanConvertFrom(parameter.GetType()))
+                newKey = GenerateGenericKey(value, dictionaryDescriptor, parameter);
             else if (dictionaryDescriptor.KeyType.IsEnum)
                 newKey = new NodeIndex(parameter);
             else
@@ -66,12 +64,15 @@ namespace Stride.Core.Assets.Editor.Quantum.NodePresenters.Commands
 
             if (newKey != null)
             {
-                var newItem = dictionaryDescriptor.ValueType.Default();
-                var instance = CreateInstance(dictionaryDescriptor.ValueType);
-                if (!AddNewItemCommand.IsReferenceType(dictionaryDescriptor.ValueType) && (assetNodePresenter == null || !assetNodePresenter.IsObjectReference(instance)))
-                    newItem = instance;
+                if (!dictionaryDescriptor.ContainsKey(nodePresenter.Value, newKey.Value.Value))
+                {
+                    var newItem = dictionaryDescriptor.ValueType.Default();
+                    var instance = CreateInstance(dictionaryDescriptor.ValueType);
+                    if (!AddNewItemCommand.IsReferenceType(dictionaryDescriptor.ValueType) && (assetNodePresenter == null || !assetNodePresenter.IsObjectReference(instance)))
+                        newItem = instance;
 
-                nodePresenter.AddItem(newItem, newKey.Value);
+                    nodePresenter.AddItem(newItem, newKey.Value);
+                }
             }
         }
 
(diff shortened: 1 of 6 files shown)


sources/editor/Stride.Core.Assets.Editor/View/TemplateProviders/DictionaryNumberKeyTemplateProvider.cs
```diff
@@ -0,0 +1,51 @@
+// Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net) and Silicon Studio Corp. (https://www.siliconstudio.co.jp)
```
xen2, line 1: No need for `and Silicon Studio Corp. (https://www.siliconstudio.co.jp)` in new files.

sources/editor/Stride.Core.Assets.Editor/Quantum/NodePresenters/Commands/AddPrimitiveKeyCommand.cs
```diff
@@ -94,7 +99,25 @@ private static object CreateInstance(Type type)
...
-        internal static NodeIndex GenerateStringKey(object dictionary, ITypeDescriptor descriptor, string baseValue)
+        internal static NodeIndex? GenerateGenericKey(object dictionary, ITypeDescriptor descriptor, string baseValue)
+        {
+            // TODO: use a dialog service and popup a message when the given key is invalid
+            DictionaryDescriptor dictionaryDescriptor = descriptor as DictionaryDescriptor;
+            Type keyType = dictionaryDescriptor.KeyType;
+            object key = keyType.Default();
+
+            if (!string.IsNullOrWhiteSpace(baseValue))
+                key = keyType.GetMethod("Parse", new Type[] { typeof(string) }).Invoke(null, new string[] { baseValue });
```
Eideren, line 110: The ``GetMethod("Parse", ...`` here and in other cases can be replaced by ``System.ComponentModel.TypeDescriptor.GetConverter(keyType).ConvertTo(baseValue , typeof(string))`` (and other functions of that type like ``CanConvertTo``, etc.) 
**BUT** ``NodeIndex`` supports ``object`` and so does ``DictionaryDescriptor`` for its keys. Is there any reason to use strings instead of the actual object ?

sources/editor/Stride.Core.Assets.Editor/View/TemplateProviders/DictionaryNumberKeyTemplateProvider.cs
```diff
@@ -0,0 +1,51 @@
...
+// Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
+using System;
+using System.Collections.Generic;
+using Stride.Core.Assets.Editor.Quantum.NodePresenters.Updaters;
+using Stride.Core.Presentation.Quantum;
+using Stride.Core.Presentation.Quantum.ViewModels;
+
+namespace Stride.Core.Assets.Editor.View.TemplateProviders
+{
+    public class DictionaryNumberKeyTemplateProvider : DictionaryTemplateProvider
```
Eideren, line 11: Can this be made more generic, to perhaps support any other value type ?

Eideren (commented):
Nice you were able to fix the commit issue ! Thanks for the PR !

sources/editor/Stride.Core.Assets.Editor/Quantum/NodePresenters/Commands/AddPrimitiveKeyCommand.cs
D3ZAX: Yes, I solve the invalid user problem by deleting my old repository and forking new one.

This command here I just find it's used in dictionary with string key and enum key. My purpose is to fix the number key dictionary, and support other value types as key as much as posible, so I just consider string as key here.
And as a text input ui triggered command, I'd like see it like a string based command function to construct dictionary key. Maybe it can extend to much more types, but not in this merge request.

sources/editor/Stride.Core.Assets.Editor/View/TemplateProviders/DictionaryNumberKeyTemplateProvider.cs
D3ZAX: This template provider here is just to support number value key dictionary to create new key shows default value as 0, I just reference the DictionaryEnmuKeyTemplateProvider to write this.
As my purpose is to fix the dictionary editor with value types key, so I decide to keep the old style of the code. So a better way with much modification is not in my consideration.
And also, other value type key can use default dictionary template with the default value, such as '' for char, false for bool and so on, I think it's not nessary to support so generic for the default has done this work.

sources/editor/Stride.Core.Assets.Editor/View/TemplateProviders/DictionaryNumberKeyTemplateProvider.cs
D3ZAX: OK, I have removed it!

sources/editor/Stride.Core.Assets.Editor/Quantum/NodePresenters/Commands/AddPrimitiveKeyCommand.cs
Eideren: > Maybe it can extend to much more types, but not in this merge request.

Why not, the changes required to introduce this don't seem significantly harder to implement than what you already have to do with replacing reflection ``Parse`` with the ``TypeDescriptor.GetConverter`` stuff. 

If you don't want to dedicate any more time on this PR I understand but I don't think this PR should be merged in this state with it being so close to potentially support any key type.

D3ZAX: Yeah, I have tried it and it's really a great idea! Thanks for the advice!
I have changed my code to support more generic type keys, with some annoying can't edit bug fixed.

Eideren:
LGTM, thanks again @D3ZAX

PR #1114, opened 2021-08-15
PhysicsConstraintComponent with a gizmo and minor fixes

manio143:
# PR Details

This PR aims to add a component for establishing physics constraint in the editor, rather than doing that just in code.

## Description

Following how collider shapes have classes that describe them, I created classes that describe basic properties required to create physics constraints. The constraint component takes two rigid bodies and a description of the constraint and its processor will invoke the creation of those constraints in the Simulation.

I have put a lot of effort into documenting all the public properties of constraints so that they may also be used as a guide for people creating constraints without those components.

The gizmo for constraints is disabled by default just like the gizmo for physics components. The size of the gizmo is controlled with the size slider as usual. I wanted to make it constant in size in relation to the view, like the transform gizmo is, but couldn't figure it out, and it can be amended later if people feel the need.

Note: I did not create a component for 6DoF constraint as I believe it's not used as often and would require a really big description - if someone decides they need it, it shouldn't be very hard to add it following my work.

### Minor fixes
I had to add another method to Simulation to create the Hinge constraint - it's frame is very complicated to create and everyone is using the secondary constructor which takes an axis vector. I also added the `SetAxis` method into the constraint.

In SceneInstance I have moved assignment of the `rootScene` field in the property setter to before processors are initialized - this way if a scene starts with a PhysicsProcessor it can access a required dependency and create debug shapes.

### GeometricPrimitive.Disc
I needed a disc/sector for previewing angular limits, so I'm adding that in, but I know little about texturing so it might need some fixing up to be properly reusable - I just had a uniform color.

### Screenshots
image
image

### Notes
Angular limits are in radians, because they need to be usable in code. I'm not sure if there's any way to do an adhoc converter from angles in the editor.

The small cone at the pivot point perpendicular to the constraint axis (X) is the Z axis.

## Related Issue

N/A

## Motivation and Context

The current API for creating constraints is limited to just code (`Simulation.CreateConstraint`) and requires a lot of experimental adjustments when used with non-trivial objects (esp. when creating a ragdoll). I needed to see the constraints in the editor.

Moreover, the documentation around Bullet constraints is lacking - I've spent hours searching on the internet how to set up the constraints without luck. E.g. how do you create a frame for a constraint, what does swingSpan1 mean for a ConeTwist constraint? While experimenting and reading documentation of other game engines that use bullet I was able to figure this stuff out.

## Types of changes

## Checklist

- [ ] I have added tests to cover my changes. _I have a test project here: [manio143/StrideConstraintsDemo](https://github.com/manio143/StrideConstraintsDemo)_


xen2 (approved):

sources/engine/Stride.Physics/Engine/PhysicsConstraintProcessor.cs
```diff
@@ -0,0 +1,92 @@
...
+            if (component.Enabled)
+            {
+                if (component.Description == null || component.BodyA == null)
+                {
+                    logger.Warning("ConstraintComponent with an empty description or missing required body. Skipping constraint creation.");
+                    return;
+                }
+
+                if (component.BodyB != null && component.BodyB.Simulation != component.BodyA.Simulation)
+                    return; // simulation mismatch - may happen when first loading the scene
```
Eideren, line 79: Shouldn't this case and the one below throw then since the function could not recreate it ? Or is it going to be recreated automatically later in the frame ?

sources/engine/Stride.Physics/Simulation.cs
```diff
@@ -756,11 +795,39 @@ static Constraint CreateConstraintInternal(ConstraintTypes type, RigidbodyCompon
...
             }
+            constraintBase.RigidBodyA = rigidBodyA;
             rigidBodyA.LinkedConstraints.Add(constraintBase);
 
             return constraintBase;
         }
 
+        static HingeConstraint CreateHingeConstraintInternal(RigidbodyComponent rigidBodyA, RigidbodyComponent rigidBodyB, Vector3 pivotInA, Vector3 pivotInB, Vector3 axisInA, Vector3 axisInB, bool useReferenceFrameA = false)
+        {
+            if (rigidBodyB != null && rigidBodyB.Simulation != rigidBodyA.Simulation) throw new Exception("Both RigidBodies must be on the same simulation");
```
Eideren, line 806: @xen2 what do you think about this specific coding style, I know it's fairly common for the rest of the code base to use such a style for exception throwing but not sure if we should avoid it since it doesn't match with the base style for ``if`` statement, should I comment on it when doing reviews ?

Eideren:
Thanks a lot for this PR @manio143 !

sources/editor/Stride.Assets.Presentation/AssetEditors/Gizmos/PhysicsConstraintGizmo.cs
```diff
@@ -0,0 +1,564 @@
...
+        {
+            if (!MaterialCache.TryGetValue(device, out var cache))
+            {
+                cache = new Dictionary<Color, Material>();
+                MaterialCache.Add(device, cache);
+            }
+
+            if (!cache.TryGetValue(color, out var material))
+            {
+                material = GizmoUniformColorMaterial.Create(device, color);
```
salahchafai, line 203: the transformation gizmos use `GizmoEmissiveColorMaterial`, I think this one should too? (in order to have the same style)

sources/engine/Stride.Physics/Engine/PhysicsConstraintProcessor.cs
manio143: I need to check again what was up with that - I can't remember the exact reason why this processor was running before the physics one (I remember there was some something, but not the details) - it's possible that it should be the other way around - thus ensuring rigidbodies are properly initialized by the time this runs.

sources/engine/Stride.Physics/Simulation.cs
xen2: I prefer two lines, but not so picky about one line version if it's short.
This one is quite long so probably better as 2 lines?

sources/editor/Stride.Assets.Presentation/AssetEditors/Gizmos/PhysicsConstraintGizmo.cs
xen2: Good point, otherwise it might disappear without lighting.

manio143: Looks kind of harsh to me and doesn't seem to support transparency ( I could use the uniform color for just the limit disc )
I mean, not sure what the preference here is from people, visually - for me the only important thing is that it's functional.
image

xen2: @manio143 sure, please use whatever you feel is better UX wise.

sources/engine/Stride.Physics/Engine/PhysicsConstraintProcessor.cs
manio143: So apparently the ordering of adding entities' components is a little different than I expected (e.g. child entity is added in the transform component processor, so their components will be processed before any other parent components). I think I did a good job mitigating it (see new commit).

The one thing that bugs me a little is - if you remove a rigidbody before you remove the constraint component, on the next frame the constraint component will try to recreate itself and throw an exception because the rigidbody is no longer valid. Do you think I should be handling this in some way - i.e. set some flag on the component not to recreate the constraint unless user explicitly calls `Recreate()` or removes and re-adds the component, or just have them ensure that they remove the constraint component on the same frame as they remove the rigid body?

sources/editor/Stride.Assets.Presentation/AssetEditors/Gizmos/PhysicsConstraintGizmo.cs
manio143: I added transparency support to the emissive material (copied from the uniform one) and used slightly lower intensity.
image

sources/engine/Stride.Physics/Engine/PhysicsConstraintProcessor.cs
Eideren: I see, perhaps the ``PhysicsConstraintComponent`` should be a property of the ``RigidbodyComponent`` then if the former always requires the later and creates unexpected issues when it is removed ?

Then you would just need to release the constraint when ``BodyB`` was set but is no longer in the scene ?

manio143: Well, my approach to constraints has always been a little to the side of rigidbodies - which is why I didn't tie `BodyA` to the entity holding the component - this allowed me to put all constraints into a single entity when doing a ragdoll, which made it a bit easier to place constraints for left/right side, rather than switching between entities and trying to memorize the placements of constraint pivots. Plus, I feel like the `RigidbodyComponent` already has a lot going on.

I see two options:
- one is to provide an event on the constraint component `RigidbodyDetach`, which would be called when one of the rigidbodies gets detached - this can be then wired up by the users with a callback, but would likely require people to set up those components from code
- add an internal flag `Detached`, which would prevent exceptions when a rigidbody referenced in the constraint was previously successfully in a constraint but then was detached, but should the rigidbody be reattached the constraint would be re-established.

And document the behaviour so that users aren't surprised.

Eideren: Sure, that makes sense.
I don't think I would expect constraints to throw, and in the small number of cases where I need to ensure that those connections will not break, a callback seems like the best tool for the job so, I would rather go for the first option.
Release the connection as soon as either body is no longer in the scene or if they are set to null by the user and invoke a method to let them know, yeah.

manio143: @Eideren have you had a chance to look at the last update?

sources/engine/Stride.Physics/Engine/PhysicsConstraintProcessor.cs
```diff
@@ -0,0 +1,132 @@
...
+                else if (component.Constraint == null && component.Enabled)
+                {
+                    Recreate(component, component.Detached);
+                }
+            }
+
+            foreach (var component in detachedComponents)
+            {
+                component.OnDetach();
+            }
```
Eideren, line 78: ``detachedComponents.Clear`` right here instead to avoid holding refs within that list for longer than necessary (or 'forever' for cases where things have been detached in the previous frame but the current has ``Simulation.DisableSimulation`` true which would prevent clear and the rest of update to execute)

sources/engine/Stride.Physics/Engine/PhysicsConstraintComponent.cs
```diff
@@ -0,0 +1,95 @@
...
+        /// Simulation to which this constraint was added.
+        /// </summary>
+        [DataMemberIgnore]
+        public Simulation Simulation { get; internal set; }
+
+        /// <summary>
+        /// Invoked when a constraint is broken because one of the rigidbodies
+        /// involved in the constraint is removed from the scene.
+        /// </summary>
+        public event Action<PhysicsConstraintComponent> Detach;
```
Eideren, line 83: ``Detach`` would mislead users going through intellisense into thinking that this is the event to *call* to detach the object, which it isn't and granted can't be called anyway. ``OnDetach`` would be a better fit, you could rename the internal function below to ``InvokeOnDetach`` / ``InvokeDetachEvent`` or something similar instead ?

Eideren (commented):
Sorry, I was waiting on your approval before looking at the changes

manio143:
I'm still finding some ambiguity with the hinge constraint for certain scenarios - for example when dealing with a bone whose Up is downwards I'm not displaying the limits quite right - this might need some more debugging to figure out

Issue #1113, opened 2021-08-14
Could not restore Stride.VisualStudio.Commands 4.1.0.1459-beta


AmbulantRex:
**Release Type**: Official Release

**Version**: 4.1.0.1459-beta

**Platform(s)**: Windows

**Describe the bug**
Visual Studio extension attempts to restore the Stride.VisualStudio.Commands nuget package necessary to function and is unable to do so. Latest plugin has been installed and reinstalled with no change in behavior.

**To Reproduce**
1. Load solution in VS that contains `.sdsl` files.
2. Go to Output -> Show output from: General

**Expected behavior**
Would expect the extension to be able to restore the appropriate version that has been installed.

**Log and callstacks**
Error initializing Stride command proxy: System.InvalidOperationException: Could not restore Stride.VisualStudio.Commands 4.1.0.1459-beta, this visual studio extension may fail to work properly without it. To fix this you can either build Stride.VisualStudio.Commands or pull the right version from nugget manually
   at Stride.VisualStudio.Commands.StrideCommandsProxy.<FindStrideSdkDir>d__16.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at Stride.VisualStudio.StridePackage.<InitializeCommandProxy>d__19.MoveNext()

**Additional context**
- Visual Studio Community 2019, Version 16.10.4
- Windows 10 Pro, Version 10.0.19042
- .NET SDK 5.0.400

AmbulantRex:
Seems like this is an issue with compatibility between Net5.0 and Net5.0-windows7.0.

```
Package Stride.VisualStudio.Commands 4.1.0.1459-beta is not compatible with net50 (.NETFramework,Version=v5.0). Package Stride.VisualStudio.Commands 4.1.0.1459-beta supports: net5.0-windows7.0 (.NETCoreApp,Version=v5.0)
```

AmbulantRex:
This issue was resolved with the above commit. Closing.

PR #1117, opened 2021-08-20
Changed NuGet restore to use friendly names

AmbulantRex:
# PR Details

Switches the ambiguous and error prone long names to the friendly names instead.

## Related Issue

#1113 

## Motivation and Context

In some newer versions of .NET 5 SDK, restoration of the visual studio extension was failing. It was unable to resolve `.NET,Version=v5.0` to the version with the windows binding (and specifying other attributes like `Profile` did not work either). 

The recommendation for Target Framework Names is to standardize around the friendly name as it already encodes the platform and version. https://github.com/dotnet/designs/blob/main/accepted/2020/net5/net5.md#mapping-to-long-names

## Types of changes

## Checklist


xen2:
Nice!
Another positive effect is that grepping the code for TFMs will yield this result, so we won't forget it if it needs any kind of update.

Issue #1097, opened 2021-07-28
[Math] Quaternion API improvement


ykafia:
**Is your feature request related to a problem? Please describe.**
It's frustrating to have to go through matrix everytime i need a LookAt function

**Describe the solution you'd like**
Add a Quaternion static function that returns a look at rotation

**Describe alternatives you've considered**
Keep using the matrix function

Eideren:
Changed this issue to include one more requests from discord: ``Quaternion * Vector -> Vector`` shorthand for ``Quaternion.Rotate()``

AmbulantRex:
I've looked at this one in the past. I'm also interested in a "RotateAround" type function which is a common use case for a LookAt type function.

AmbulantRex:
> 
> 
> Changed this issue to include one more requests from discord: `Quaternion * Vector -> Vector` shorthand for `Quaternion.Rotate()`

I'm worried about the implications of this. Multiplying a quaternion by a vector isn't a rotation and would require conjugation like we do in the `Quaternion.Rotate()`. I think adding this as a `*` operator would be mathematically incorrect unless I misunderstood what you meant.

ykafia:
I think we should have a "LookAt" function for TransformTRS since Transform Component already has it by extension

Eideren:
@AmbulantRex 
> I think adding this as a `*` operator would be mathematically incorrect unless I misunderstood what you meant.

Quaternion operators have never been element-wise operation, ``quat * quat`` is not returning a new quat with the product of each element of the quats multiplied together. 
So given that, if ``quat * quat`` rotates a given rotation by another, it is expected that ``quat * vector`` would return the result of rotating the Vector by the Quaternion. 
This is common to most of the engines out there now and even before like it was in UE3.

Eideren:
This request has been implemented and merged in, I'll close this off

PR #1121, opened 2021-08-22
[Math] Quaternion LookRotation, quat*vec and AngleBetween

Eideren:
# PR Details
Added two methods, an operator and more coverage for ``in``s.
See related issue for more information.

## Related Issue
#1097 

## Motivation and Context
User had issues writing logic that required such functions.

## Types of changes

## Checklist

### DIFF
--- a/sources/core/Stride.Core.Mathematics/Quaternion.cs
+++ b/sources/core/Stride.Core.Mathematics/Quaternion.cs
@@ -29,6 +29,7 @@
 using System;
 using System.Globalization;
 using System.Runtime.InteropServices;
+using static System.MathF;
 
 namespace Stride.Core.Mathematics
 {
@@ -300,7 +301,7 @@ namespace Stride.Core.Mathematics
         /// <see cref="Stride.Core.Mathematics.Quaternion.LengthSquared"/> may be preferred when only the relative length is needed
         /// and speed is of the essence.
         /// </remarks>
-        public float Length()
+        public readonly float Length()
         {
             return MathF.Sqrt((X * X) + (Y * Y) + (Z * Z) + (W * W));
         }
@@ -313,7 +314,7 @@ namespace Stride.Core.Mathematics
         /// This method may be preferred to <see cref="Stride.Core.Mathematics.Quaternion.Length"/> when only a relative length is needed
         /// and speed is of the essence.
         /// </remarks>
-        public float LengthSquared()
+        public readonly float LengthSquared()
         {
             return (X * X) + (Y * Y) + (Z * Z) + (W * W);
         }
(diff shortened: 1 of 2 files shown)


Eideren:
@tebjan Sure, please share if you do find one, I have thought about this quite a bit and looked around a bunch but the other solutions aren't any faster than the ``RotationMatrix``, if you can somehow replace the orthogonal stuff with it you might have a chance though.

AmbulantRex:
> 
> 
> @tebjan Sure, please share if you do find one, I have thought about this quite a bit and looked around a bunch but the other solutions aren't any faster than the `RotationMatrix`, if you can somehow replace the orthogonal stuff with it you might have a chance though.

I think the Matrix solution in this case is computationally less expensive since it doesn't require any trig functions and relies on a few `MathF.Sqrt()` when constructing the result quaternion. That's probably only interesting from an academic perspective though. I've likely wasted more time typing this than it might ever save me in the long run.

sources/core/Stride.Core.Mathematics/Quaternion.cs
```diff
@@ -677,6 +695,24 @@ public static Quaternion Lerp(Quaternion start, Quaternion end, float amount)
             Lerp(ref start, ref end, amount, out result);
             return result;
         }
+        
+        /// <summary>
+        /// Returns a rotation whose facing direction points towards <paramref name="forward"/>
+        /// and whose up direction points as close as possible to <paramref name="up"/>.
+        /// </summary>
+        public static Quaternion LookRotation(in Vector3 forward, in Vector3 up)
```
AmbulantRex, line 703: To a novice like myself, the concept of `forward` in this context is a bit hard to grasp. Could this have an overload to do the simple `Vector3.Normalize(target - source)` but takes a `source` and `target` parameter instead of `forward`?

tebjan:
> > @tebjan Sure, please share if you do find one, I have thought about this quite a bit and looked around a bunch but the other solutions aren't any faster than the `RotationMatrix`, if you can somehow replace the orthogonal stuff with it you might have a chance though.
> 
> I think the Matrix solution in this case is computationally less expensive since it doesn't require any trig functions and relies on a few `MathF.Sqrt()` when constructing the result quaternion. That's probably only interesting from an academic perspective though. I've likely wasted more time typing this than it might ever save me in the long run.

Right, let's not waste any time on this. Unlikely that it will cause problems for anyone. If someone comes across a faster algorithm, er can exchange them later.

tebjan:
> There doesn't seem to be any specific convention but unity and udk does quaternion first, cryengine does both sides, unreal I have no clue.

Ok, maybe having it both ways is an option?

AmbulantRex:
> 
> 
> > There doesn't seem to be any specific convention but unity and udk does quaternion first, cryengine does both sides, unreal I have no clue.
> 
> Ok, maybe having it both ways is an option?

I created the flip side in a separate PR so we'd have either. If we decide not to go that route (or not use my PR), I can rip it out,

Eideren:
> 
> 
> > There doesn't seem to be any specific convention but unity and udk does quaternion first, cryengine does both sides, unreal I have no clue.
> 
> Ok, maybe having it both ways is an option?

@tebjan 
Actually, now that I think about it, it makes sense to have only one and it be the ``Quaternion`` first given that ``quatA * quatB`` != ``quatB * quatA``, the mult operator can be thought of as ``LocalRotToWorldRot(quat parentWorld, quat local) => parentWorld * local`` so it would make sense to have it behave in a similar way for vectors, that might help user build an intuition, I would still need to include this in the remarks for that operator's docs though, what do you think ?

AmbulantRex:
> 
> 
> > > There doesn't seem to be any specific convention but unity and udk does quaternion first, cryengine does both sides, unreal I have no clue.
> > 
> > 
> > Ok, maybe having it both ways is an option?
> 
> @tebjan
> Actually, now that I think about it, it makes sense to have only one and it be the `Quaternion` first given that `quatA * quatB` != `quatB * quatA`, the mult operator can be thought of as `LocalRotToWorldRot(quat parentWorld, quat local) => parentWorld * local` so it would make sense to have it behave in a similar way for vectors, that might help user build an intuition, I would still need to include this in the remarks for that operator's docs though, what do you think ?

Quaternions being noncommutative is expressly important to the concept so I do agree that intuition might dictate _not_ having it both ways. I wondered something similar when I wrote the flip side in my other PR which has since been retracted.

PR #1122, opened 2021-08-22
[Math] Vector RotateAround & Vector3 * Quaternion rotation shorthand

AmbulantRex:
# PR Details

Adds a convenience method for performing rotation around a target for a given axis and angle of rotation. Includes a rotation shorthand on the Vector3 to match up with @Eideren's Quaternion shorthand in PR https://github.com/stride3d/stride/pull/1121. 

## Related Issue

#1097 

## Motivation and Context

Rotation around objects is a common task and it would be nice to have a single method to handle it.

The rotation shorthand is common in many game engines either as Quaternion * Vector3 or Vector3 * Quaternion. PR #1121 addresses the Quaternion side shorthand.

## Types of changes

## Checklist

### DIFF
--- a/sources/core/Stride.Core.Mathematics/Vector3.cs
+++ b/sources/core/Stride.Core.Mathematics/Vector3.cs
@@ -240,6 +240,34 @@ namespace Stride.Core.Mathematics
         }
 
         /// <summary>
+        /// Rotates the vector around the origin by the angle using the given axis of rotation.
+        /// </summary>
+        /// <param name="axis">The axis of rotation.</param>
+        /// <param name="angle">The angle of rotation.</param>
+        public void RotateAround(Vector3 axis, float angle)
+        {
+            RotateAround(Zero, axis, angle);
+        }
+
+        /// <summary>
+        /// Rotates the vector around the target by the angle using the given axis of rotation.
+        /// </summary>
+        /// <param name="target">The position vector of the target to rotate around.</param>
+        /// <param name="axis">The axis of rotation.</param>
+        /// <param name="angle">The angle of rotation.</param>
+        public void RotateAround(Vector3 target, Vector3 axis, float angle)
+        {
+            Vector3 local = this - target;
+            Quaternion q = Quaternion.RotationAxis(axis, angle);
+            q.Rotate(ref local);
+            var result = target + local;
+
+            X = result.X;
+            Y = result.Y;
+            Z = result.Z;
+        }
+
+        /// <summary>
         /// Creates an array containing the elements of the vector.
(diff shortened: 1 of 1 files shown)


sources/core/Stride.Core.Mathematics/Vector3.cs
```diff
@@ -1507,6 +1552,19 @@ public static void RotationYawPitchRoll(ref Quaternion quaternion, out Vector3 y
             return new Vector3(left.X * right.X, left.Y * right.Y, left.Z * right.Z);
         }
 
+        /// <summary>
+        /// Return the vector rotated by the quaternion.
+        /// </summary>
+        /// <remarks>
+        /// Shorthand for <see cref="Quaternion.Rotate(ref Vector3)"/>
+        /// </remarks>
+        public static Vector3 operator *(in Vector3 left, in Quaternion right)
```
Eideren, line 1561: It would be confusing to have two separate PRs/commits for the same thing, I'll cover this with the other one I have in my PR.

sources/core/Stride.Core.Mathematics/Vector3.cs
```diff
@@ -239,6 +239,34 @@ public void Pow(float exponent)
...
+            RotateAround(Zero, axis, angle);
+        }
+
+        /// <summary>
+        /// Rotates the vector around the target by the angle using the given axis of rotation.
+        /// </summary>
+        /// <param name="target">The position vector of the target to rotate around.</param>
+        /// <param name="axis">The axis of rotation.</param>
+        /// <param name="angle">The angle of rotation.</param>
+        public void RotateAround(Vector3 target, Vector3 axis, float angle)
```
Eideren, line 258: I think we should be as consistent as possible and avoid code duplication; since most of the functions here are statics let's keep the static one below and get rid of those two.

I understand that you might want instance methods for succinctness, if that's the case you can work around this by adding a ``using static Stride.Core.Mathematics.Vector3;`` to your class which would give you the ability to omit the type when calling static methods of that type. So you would be able to write ``Vector3.Lerp(x, y, z)`` as just ``Lerp(x, y, z)``.

Eideren (commented):
Nice !
If you could get rid of the unnecessary white-space changes while you're at it that would be great :)

sources/core/Stride.Core.Mathematics/Vector3.cs
AmbulantRex: Done!

Eideren:
Thanks ! I'll merge this one as soon as #1121 is merged in

PR #1140, opened 2021-09-01
[Engine] Fixes null exception in InstancingProcessor

tebjan:
Fixes null exception in InstancingProcessor when no ModelComponent is present

# PR Details

Just a null check.

- [ ] Needs a cherry-pick to master

## Types of changes

### DIFF
--- a/sources/engine/Stride.Engine/Engine/Processors/InstancingProcessor.cs
+++ b/sources/engine/Stride.Engine/Engine/Processors/InstancingProcessor.cs
@@ -87,18 +87,19 @@ namespace Stride.Engine.Processors
             if (instancingComponent.Enabled && instancingComponent.Type != null)
             {
                 var instancing = instancingComponent.Type;
+                var modelComponent = instancingData.ModelComponent;
 
                 // Calculate inverse world and bounding box
                 instancing.Update();
 
-                if (instancingData.ModelComponent != null && instancing.InstanceCount > 0)
+                if (modelComponent != null && modelComponent.Model != null && instancing.InstanceCount > 0)
                 {
                     // Bounding box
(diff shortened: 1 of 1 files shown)


sources/engine/Stride.Engine/Engine/Processors/InstancingProcessor.cs
```diff
@@ -87,18 +87,19 @@ private void UpdateInstancing(InstancingComponent instancingComponent, Instancin
             if (instancingComponent.Enabled && instancingComponent.Type != null)
             {
                 var instancing = instancingComponent.Type;
+                var modelComponent = instancingData.ModelComponent;
 
                 // Calculate inverse world and bounding box
                 instancing.Update();
 
-                if (instancingData.ModelComponent != null && instancing.InstanceCount > 0)
+                if (modelComponent != null && modelComponent.Model != null && instancing.InstanceCount > 0)
```
xen2, line 95: `modelComponent?.Model != null`?

tebjan: yes, that is indeed much shorter in terms of changes. But some lines got a little shorter with the local variable. just a matter of taste. if you prefer the `?` operator I'll change it.

vdvman1: There's no local variable being assigned in that comparison
xen2 was saying that `modelComponent != null && modelComponent.Model != null` can be written as `modelComponent?.Model != null`
No behaviour change there, no change to what variables get assigned

tebjan: Oh, i see it now... Thanks! :-)

Issue #1141, opened 2021-09-01
Error when starting from local build


tebjan:
**Release Type**: Official Release/GitHub (please choose appropriate option)

Local build master4.0 branch.

**Version**: Version number and/or git branch

**Platform(s)**: Does the problem occur on Windows, Android...?

Windows

**Describe the bug**
Game studio doesn't start due to a restore error related to a file access violation.

**To Reproduce**
1. Start game studio from visual studio in master-4.0 branch

**Expected behavior**
Normal start of game studio after nuget restore.

**Screenshots**
![image](https://user-images.githubusercontent.com/1094716/131680331-422a6ea0-cef1-474a-8900-97c7aa0bbdbf.png)

**Log and callstacks**

```
[Verbose] Acquired lock for the installation of Stride.Core.Assets 4.0.1.1
[Information] Installing Stride.Core.Assets 4.0.1.1.
[Error] Error restoring NuGet packages: System.AggregateException: Mindestens ein Fehler ist aufgetreten. ---> System.UnauthorizedAccessException: Der Zugriff auf den Pfad "Stride.Irony.dll" wurde verweigert.
   bei System.IO.Directory.DeleteHelper(String fullPath, String userPath, Boolean recursive, Boolean throwOnTopLevelDirectoryNotFound, WIN32_FIND_DATA& data)
   bei System.IO.Directory.Delete(String fullPath, String userPath, Boolean recursive, Boolean checkHost)
   bei NuGet.Packaging.PackageExtractor.<>c__DisplayClass5_0.<<InstallFromSourceAsync>b__0>d.MoveNext()
--- Ende der Stapelüberwachung vom vorhergehenden Ort, an dem die Ausnahme ausgelöst wurde ---
   bei System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   bei System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   bei NuGet.Common.ConcurrencyUtilities.<ExecuteWithFileLockedAsync>d__4`1.MoveNext()
--- Ende der Stapelüberwachung vom vorhergehenden Ort, an dem die Ausnahme ausgelöst wurde ---
   bei System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   bei NuGet.Common.ConcurrencyUtilities.<ExecuteWithFileLockedAsync>d__4`1.MoveNext()
...
[...]

PR #1142, opened 2021-09-01
[Rendering] Adds IsOrthographic case for AmbientOcclusion 

tebjan:
Adds IsOrtho case for AmbientOcclusion that will work with orthographic cameras.

# PR Details

Makes ambient occlusion work with orthographic cameras.

## Description
* The projection matrix of the current render view gets checked for orthographic projection
* The shaders have a new generics variable `IsOrtho` that selects the right method to restore view depth and position
* The `Draw` method of AmbientOcclusion sets the variable and adapts the projection parameters accordingly
* Ambient occlusion only modulates RGB, not alpha
- [ ] Needs cherry-pick to master branch

Good links that helped to make this PR:

Stride AO algorithm paper:
https://research.nvidia.com/publication/alchemy-screen-space-ambient-obscurance-algorithm

SSAO comparison, claiming stride one has the best quality to performance ratio:
http://frederikaalund.com/a-comparative-study-of-screen-space-ambient-occlusion-methods/

Reconstruct view position for ortho:
https://www.derschmale.com/2014/03/19/reconstructing-positions-from-the-depth-buffer-pt-2-perspective-and-orthographic-general-case/

## Related Issue

Works well in our repo, see https://github.com/vvvv/VL.Stride/pull/505
Not tested in Stride Game Studio due to #1141

## Motivation and Context
We had to use an orthographic camera for a large scale media installation.

## Types of changes

## Checklist


xen2:
LGTM.
Two things:
- We use `IsOrthographic` in various locations. Maybe better to switch to that? (however I agree even Stride is not consistent with itself because we also have methods such as `Matrix.OrthoLH`.
- One extra thing to do is to find all `.sdeffectlog` files and add `AmbientOcclusionBlur.IsOrthographic: false` in the `AmbientOcclusionBlurEffect` parameters (just search/grep in the repo).

Eideren:
Xen2's suggestions have been implemented so I'll just merge this in then. Thanks !

tebjan:
Thanks!

Issue #827, opened 2020-08-15
Cannot create android project


makemefeelgr8:
**Release Type**: Official Release

**Version**: 4.0.0.1-beta04-1265

**Platform(s)**: Windows

**Describe the bug**
New project cannot be created.

**To Reproduce**
1. Create a new empty project for .net core 3.1

**Expected behavior**
I expect the project to be created.

**Log and callstacks**
Verbose: Restore NuGet packages for MyGame4.Android...
[\MyGame4\MyGame4.Android\MyGame4.Android.csproj(93,3)]: Error: The imported project "C:\Program Files\dotnet\sdk\3.1.302\Xamarin\Android\Xamarin.Android.CSharp.targets" was not found. Confirm that the expression in the Import declaration "C:\Program Files\dotnet\sdk\3.1.302\\Xamarin\Android\Xamarin.Android.CSharp.targets" is correct, and that the file exists on disk.

xen2:
Does it work better with .NET Framework version of Stride ?

makemefeelgr8:
I cannot create .NET Framework android project. The error looks just like this one:
https://github.com/stride3d/stride/issues/780
Unfortunately, there is no working way to create an android project. So, the engine is a no-go for me.

laske185:
I fixed it temporary for me by removing the `MSBuildExtensionsPath` variable in the project import of the `Xamarin.Android.CSharp.targets`.

So I replaced in the *.Android.csproj file 
`<Import Project="$(MSBuildExtensionsPath)\Xamarin\Android\Xamarin.Android.CSharp.targets" />`
by
`<Import Project="C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Xamarin\Android\Xamarin.Android.CSharp.targets" />`.

bmello4688:
![image](https://user-images.githubusercontent.com/7825858/99675309-0f231700-2a45-11eb-9c6b-3f14edf76c37.png)

bmello4688:
From issue #780

Agredo:
This is still not working? This needs to be fixed. I would love to use it but without this platforms there is no reason to use it.

MeharDT:
I'm looking into both issues. The Resource.Designer.cs file isn't a requirement for Android projects in Stride/Xenko 3.1 and it actually doesn't generate one for new projects either (only at build). Something changed with 4.0 that now makes this a requirement at project creation.

[...]

PR #1146, opened 2021-09-06
Missing Resource.Designer.cs fix

MeharDT:
# PR Details
The Resource.Designer.cs file is required to compile Android apps with Xamarin. By default the file is not included in either Stride/Xenko 3.x or 4.x and is instead generated when the Android project is first compiled. A change in 4.x makes this file a requirement for Stride's Project Generator to create a new Android project, when the file isn't found an error is thrown and the project fails to create.

## Description
An empty Resource.Designer.cs.t4 file has been added to the Android template. This file is regenerated in a project specific context when the user first compiles their Android app via Stride or Visual Studio.

## Related Issue
#827 

## Motivation and Context
While workarounds are possible (documented in the above issue), this bug prevents the Android platform from working "out of the box" on impacted versions of Stride.

## Types of changes

## Checklist


MeharDT:
If the fix is accepted I'll create a separate PR for master-4.0

tebjan:
great, thank you! this was a really annoying bug and the fix looks surprisingly easy.

PR #1150, opened 2021-09-17
Linux Platform Build Fix

MeharDT:
# PR Details
The Linux template no longer generates with TargetFramework incorrectly set to net5.0-windows and instead uses .net5.0

## Description
TargetFramework has been manually set to net5.0, this is inline with the macOS template and also the Linux template in 4.0, as a result of this change Linux builds will now compile.

## Types of changes

## Checklist


xen2:
I ended up doing it slightly differently (cf 60f1b9f on `master-4.0` and ebb3bee on `master`).
Thanks for the idea/initial PR!

Issue #1155, opened 2021-09-30
DebugText.Print(string msg, Int2 pos, Color4 col) doesn't display given color


DrBellubins:
**Release Type**: Official Release

**Version**: 4.1.0.1459-beta

**Platform(s)**: Windows

**Describe the bug**
DebugText.Print() has an overload which takes a Color4, yet isn't reflected in-game.

**To Reproduce**
1. Use DebugText.Print(string message, Int2 position, Color4 color) overload.

**Expected behavior**
Given color be reflected in-game.

MeharDT:
Line 103 in DebugTextSystems.cs is always false even when the colours are different.

`msg.TextColor` != currentColor`

Will have a PR ready shortly.

TheKeyblader:
maybe close the issue ?

PR #1156, opened 2021-09-30
DebugText Color Fix

MeharDT:
# PR Details
RenderText.Print() will now work as intended when a custom text color (or colour) is set.

## Description
If a custom color is set in RenderText.Print() the text will remain the default Light Green. This is because msg.TextColor != currentColor was essentially comparing against itself and as a result always returning a value of False. With this change, msg.TextColor is now being compared to the current/default TextColor value.

## Related Issue
#1155

## Types of changes

## Checklist


MeharDT:
As always, I'll backport and create a new 4.0 PR if this is accepted.

PR #1158, opened 2021-10-02
Changed Assimp binding to Silk.Net.Assimp

johang88:
# PR Details

Changed the Assimp native binding to Silk.Net.Assimp.

## Description

Changed the Assimp native binding to Silk.Net.Assimp and converted the importer to a native C# library. I have also converted the `Stride.Importer.Common` library to a native C# library as it contained no native code.

This is a WIP PR and it currently lacks some of the material import code, otherwise it's close to a 1 to 1 conversion of the previous Assimp importer.

## Motivation and Context

This change allows us to remove much of the C++/CLR code used by the asset compiler and brings us one step closer to running the asset compiler on non windows systems.

## Types of changes

## Checklist


johang88:
Things left:
* Complete material importer
* Test skeleton / animation importing
* Clean up code / "csharpify" it where appropriate
* Maybe fix some todo's from the old importer, but might be better to push these in other PR's later on

sources/tools/Stride.Importer.Assimp/MeshConverter.cs
```diff
@@ -0,0 +1,1364 @@
...
+                            {
+                                if (AllowUnsignedBlendIndices)
+                                    ((ushort*)(vbPointer + blendIndicesOffset))[bone] = (ushort)vertexIndexToBoneIdWeight[(int)i][bone].Item1;
+                                else
+                                    ((short*)(vbPointer + blendIndicesOffset))[bone] = (short)vertexIndexToBoneIdWeight[(int)i][bone].Item1;
+                            }
+                            else
+                            {
+                                if (AllowUnsignedBlendIndices)
+                                    ((byte*)(vbPointer + blendIndicesOffset))[bone] = (byte)vertexIndexToBoneIdWeight[(int)i][bone].Item1;
```
ykafia, line 785: This one seems to need a check for the bone index since the resulting list might end up having a size of less than 4

sources/tools/Stride.Importer.Assimp/MeshConverter.cs
```diff
@@ -0,0 +1,1364 @@
...
+                    Logger.Warning(
+                        "The input file contains vertices that are associated to more than {0} bones. In current version of the system, a single vertex can only be associated to {0} bones. Extra bones will be ignored",
+                        new ArgumentOutOfRangeException("To much bones influencing a single vertex"));
+                }
+
+                curVertexWeights.Capacity = nbBoneByVertex;
+
+                var totalWeight = 0.0f;
+                for (var boneId = 0; boneId < nbBoneByVertex; ++boneId)
+                    totalWeight += curVertexWeights[boneId].Item2;
```
ykafia, line 876: Same size issue here and on line 882.
Maybe can be corrected on line 872?

sources/tools/Stride.Importer.Assimp/MeshConverter.cs
```diff
@@ -0,0 +1,1364 @@
...
+            var compositionFathers = new Stack<IComputeColor>();
+
+            var sets = new Stack<int>();
+            sets.Push(0);
+
+            var nbTextures = assimp.GetMaterialTextureCount(pMat, textureType);
+
+            IComputeColor curComposition = null, newCompositionFather = null;
+
+            var isRootElement = true;
```
ykafia, line 1062: So if i understood well, a material element is root if it's not currently referencing the previous element of the stack, or it's not being referenced by the next element of the stack.
```suggestion
            //var isRootElement = true; Should go in the loop
            
            bool isRootElement = top.type != Material.StackType.Operation && stack.Peek().type != Material.StackType.Operation;
            
```

sources/tools/Stride.Importer.Assimp/MeshConverter.cs
```diff
@@ -0,0 +1,1364 @@
...
+                IComputeColor curCompositionFather = null;
+                if (!isRootElement)
+                {
+                    if (compositionFathers.Count == 0)
+                        Logger.Error("Texture Stack Invalid : Operand without Operation.");
+
+                    curCompositionFather = compositionFathers.Pop();
+                }
+
+                var set = sets.Pop();
```
ykafia, line 1078: This one is removed too early, if the current material element is not result of an operation there's no need to pop this one here

sources/tools/Stride.Importer.Assimp/MeshConverter.cs
```diff
@@ -0,0 +1,1364 @@
...
+
+                var type = top.type;
+                var strength = top.blend;
+                var alpha = top.alpha;
+
+                if (type == Material.StackType.Operation)
+                {
+                    var realTop = (Material.StackOperation)top;
```
ykafia, line 1086:
```suggestion
                    var set = sets.Pop();
                    var realTop = (Material.StackOperation)top;
```

ykafia:
I got to a point where assimp loads a gltf mesh, albeit very broken, will continue to work on it.

tebjan:
> looks like a vertex declaration issue

Did you try the assimp loading flags, for example FlipWinding? It's important to check the intended coordinate system for a mesh and convert it to strides coordinate system... If the handedness changes, the winding has to be flipped, and/or the normals...

johang88:
Will check this out, it should all match the existing code base just moved to C# from C++, but quite possible that I made some errors when converting : ) If it matches the existing code base then I would prefer to fix stuff in a later pass.

ykafia:
Sure! Those are only suggestions, I've only tested gltf.
And yeah you're right, I'll try the other file formats before focusing on gltf!

johang88:
Tested gltf models from https://kaylousberg.itch.io/kaykit-dungeon seems to be working, skeletal meshes import but the skeleton itself is correct, bone weights looks fine though as mesh is correct if I switch to skeleton imported from FBX instead. Will have to check if the issue exists in the old assimp importer or not : )

Perksey:
FYI A lot of this code doesn't seem Stride-specific. It's an ongoing issue that the C API that we bind to is pretty thin, and isn't easy to work with. As such, if you'd like feel free to contribute some of the wrappers you've made back to Silk.NET! 

Issue: dotnet/Silk.NET#249

sources/tools/Stride.Importer.Assimp/MeshConverter.cs
johang88: Good catch, had a translation error on the line above where i just set `Capacity` on the list instead of adding empty elements like the original version does with `curVertexWeights.resize(nbBoneByVertex, std::pair<short, float>(0, 0.0f));`

phr00t:
FYI I had to make 2 commits to make this work on my fork:

https://github.com/phr00t/FocusEngine/commit/f8d2683e2b566f23b79d6ab19eb3999ce7f4b555 <-- this fixes counting UV channels

https://github.com/phr00t/FocusEngine/commit/3111c241d41d4150cc17a8241749703cc1c956b7 <-- this allowed my game assets to be built (patches Assimp64.dll missing errors).

Unsure this is the best way to fix these things, but now this appears usable.

Jklawreszuk:
@Eideren Please close this since my PR based on Johan's work was merged 😊

Eideren:
Thank you @johang88 for the initial work

PR #1161, opened 2021-10-04
System.Numerics conversions

ykafia:
# PR Details

Adds implicit conversions between System.Numerics and Stride.Maths vectors
Some casts use `Unsafe.As` for better performance, haven't done any tests though.

## Motivation and Context

A lot nugets use System.Numerics, it would be nice to be able to cast implicitely between Sys Nums and Stride Maths

## Types of changes

## Checklist

### DIFF
--- a/sources/core/Stride.Core.Mathematics/UInt4.cs
+++ b/sources/core/Stride.Core.Mathematics/UInt4.cs
@@ -184,6 +184,12 @@ namespace Stride.Core.Mathematics
         }
 
         /// <summary>
+        /// Casts from System.Numerics to Stride.Maths vectors
+        /// </summary>
+        /// <param name="v">Value to cast</param>
+        public static implicit operator UInt4(System.Numerics.Vector4 v) => new((uint)v.X, (uint)v.Y,(uint)v.Z,(uint)v.W);
+
+        /// <summary>
         ///   Creates an array containing the elements of the vector.
         /// </summary>
         /// <returns>A four-element array containing the components of the vector.</returns>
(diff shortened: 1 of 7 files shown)


sources/core/Stride.Core.Mathematics/UInt4.cs
```diff
@@ -183,6 +183,12 @@ public UInt4(uint[] values)
             }
         }
 
+        /// <summary>
+        /// Casts from System.Numerics to Stride.Maths vectors
+        /// </summary>
+        /// <param name="v">Value to cast</param>
+        public static implicit operator UInt4(System.Numerics.Vector4 v) => new((uint)v.X, (uint)v.Y,(uint)v.Z,(uint)v.W);
```
Eideren, line 190: Conversion must be explicit when data is lost in the process, similar to how c# does it between the base value types

Eideren:
Thanks, not sure if you're done but you might want to cover matrix as well. Will have to verify whether they are row major or column major though.

sources/core/Stride.Core.Mathematics/UInt4.cs
ykafia: I understand! will change it on the next commit

ykafia:
Of course, there are other conversions i will add too that can be useful

manio143:
Can you also do the conversions the other way from Stride vectors into System.Numerics?

ykafia:
Didn't think of it but will do, shouldn't be much of a trouble!

Eideren:
Are you sure that stride and dotnet's matrix have the same memory layout ?

PR #1163, opened 2021-10-04
Bind opaque output during transparent render stage

### LINKED ISSUE #1162: Bind opaque output during transparent render stage
Refractive effects currently require the scene to be rendered twice, once in the Opaque render pass and once to a render texture. This has a larger performance impact than needed.

Expose the output from the opaque render pass in the transparent stage, just like we can currently do for the depth buffer by enabling a settings in the forward renderer. 

Continue using a render target and render the scene twice.

**Additional context**
Actual implementation would differ from the binding of the depth buffer as its not possible to bind the opaque render target as a read only texture, due to the fact the transparent render pass will be writing it's output to the same target. I suggest creating a temporary texture with the same dimensions and format as the active render target, then blitting the output from the opaque stage to it before binding it to the active resource sets just like the Depth buffer.

johang88:
# PR Details

Exposes the output of the opaque render stage in the transparent render stage

## Description

Output of opaque render stage is bound during transparent rendering if `BindOpaqueAsResourceDuringTransparentRendering`is enabled on the `ForwardRenderer`. Added shader `OpaqueBase.sdsl` to give access to it.

## Related Issue

Fixes #1162 

## Motivation and Context

Improves performance for common scenarios like refraction where access to the opaque output is needed by removing the need for an extra rendering of the entire scene.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Engine/Rendering/Compositing/ForwardRenderer.cs
+++ b/sources/engine/Stride.Engine/Rendering/Compositing/ForwardRenderer.cs
@@ -118,6 +118,12 @@ namespace Stride.Rendering.Compositing
         [DefaultValue(true)]
         public bool BindDepthAsResourceDuringTransparentRendering { get; set; } = true;
 
+        /// <summary>
+        /// If true, render target generated during <see cref="OpaqueRenderStage"/> will be available as a shader resource named OpaqueBase.OpaqueRenderTarget during <see cref="TransparentRenderStage"/>.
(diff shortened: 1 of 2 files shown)


Eideren (commented):
Thanks ! Looks pretty solid, just two small notes

johang88:
Moved the resource group and updated the summary comment

Eideren:
Thanks a bunch !

PR #1165, opened 2021-10-10
Fixed shader compiler parsing of ByteAddressBuffer and RWByteAddressByffer

### LINKED ISSUE #248: [Shaders] ByteAddressBuffer not parsed correctly
official

3.0.0.6

**Platform(s)**: 
windows d3d11

**Describe the bug**
the generated c# shader keys are wrong 

**To Reproduce**
add a ByteAddressBuffer shader variable in xksl

**Expected behavior**
no error in visual studio and a c# shader object key of type buffer

**Screenshots**
image
image

**Log and callstacks**
Error	CS0246	The type or namespace name 'ByteAddressBuffer' could not be found (are you missing a using directive or an assembly reference?)	CoNature.Game	C:\vvvv\PROJECTS\2018\11_MLF_CoN\CoNature\CoNature\CoNature.Game\Shaders\Utils_CalcDispatchArgs.cs	23	N/A

johang88:
# PR Details

Fixes both parsing and shader key generation for ByteAddressBuffer and RWByteAddressBuffer

## Related Issue
Fixes #248 

## Motivation and Context

The current system only works properly for generic types and textures. This change modifies that to also handle the ByteAddressBufferRWByteAddressBuffer types.

## Types of changes

## Checklist

### DIFF
--- a/sources/shaders/Stride.Core.Shaders/Ast/Hlsl/ByteAddressBufferType.cs
+++ b/sources/shaders/Stride.Core.Shaders/Ast/Hlsl/ByteAddressBufferType.cs
@@ -1,0 +1,39 @@
+// Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net) and Silicon Studio Corp. (https://www.siliconstudio.co.jp)
+// Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Stride.Core.Shaders.Ast.Hlsl
+{
+    public static class ByteAddressBufferType
+    {
+        public static readonly ObjectType ByteAddressBuffer = new ObjectType("ByteAddressBuffer");
+
+        public static readonly ObjectType RWByteAddressBuffer = new ObjectType("RWByteAddressBuffer");
+
+        private static readonly ObjectType[] ObjectTypes = new[] { ByteAddressBuffer, RWByteAddressBuffer };
+
+        public static bool IsByteAddressBufferType(this TypeBase type)
+        {
+            return Parse(type.Name) != null;
+        }
+
+        /// <summary>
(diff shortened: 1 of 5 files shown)


sources/shaders/Stride.Core.Shaders/Ast/Hlsl/ByteAddressBufferType.cs
```diff
@@ -0,0 +1,39 @@
...
+        /// <returns></returns>
+        public static ObjectType Parse(string name)
+        {
+            foreach (var objectType in ObjectTypes)
+            {
+                if (string.Compare(name, objectType.Name.Text, StringComparison.OrdinalIgnoreCase) == 0)
+                    return objectType;
+            }
+            return null;
+        }
```
Eideren, line 37: This function operates like a .Net ``TryParse``, ``Parse``'s common operating behavior in .Net is to throw when the input not valid.
Maybe change the function signature to ``public static bool TryParse(string name, out ObjectType type)``

Eideren (commented):
Just a small note, otherwise everything looks fine, thanks !

sources/shaders/Stride.Core.Shaders/Ast/Hlsl/ByteAddressBufferType.cs
johang88: Yes I made it like this as it matches the other `Parse(string)` methods in the shader, see `SamplerType`. I prefer to keep it consistent but could change if needed, if so should I update the other places as well or might that be fit for a separate PR?

tebjan: Makes sense, I'd also go with consistency. Such a fix could then be provided by a separate PR for all `Parse` methods in the same way.

Eideren: I see, yes, let's keep it that way for now

Eideren:
Thanks !

PR #1173, opened 2021-10-25
Partial workaround to compile native libs for Linux and MacOS

manio143:
# PR Details

## Description

This is a minimal workaround to compile native libs for Linux and MacOS while `StridePlatform` is equal to `Windows`. Basically I added special outputs for linux and mac and modified the `EXEC`s a bit for them to not depend on props for things like file extension.

If there's nicer ways to go about it, I'm welcome to suggestions and learning opportunities, as currently I view MSBuild as an overly complex system.

## Related Issue

Since 8228690ffe27efeb73aaf22191c19dbebb47fc06 native libs for non-Windows desktop platforms have not been compiled - we've been skipping their targets.

The `StridePlatform` depends on the RID for platforms that target the same TFM (`net5.0`) and we are no longer setting it.

## Motivation and Context

This workaround allows you to compile Linux/MacOS native libs if those are listed in the `StridePlatforms` (note the plural) property. They get packaged properly.

## Types of changes

### DIFF
--- a/sources/core/Stride.Core/Stride.Core.csproj
+++ b/sources/core/Stride.Core/Stride.Core.csproj
@@ -113,7 +113,7 @@
   </ItemGroup>
   
   <ItemGroup>
-    <StrideContent Condition="'$(StridePlatform)' == 'Linux'" Include="Stride.Core.dll.config" />
+    <StrideContent Condition="$(StridePlatforms.Contains('Linux'))" Include="Stride.Core.dll.config" />
   </ItemGroup>
   
   <ItemGroup>
--- a/sources/engine/Stride.Audio/Stride.Audio.csproj
+++ b/sources/engine/Stride.Audio/Stride.Audio.csproj
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <Project>
   <PropertyGroup>
     <StrideNativeOutputName>libstrideaudio</StrideNativeOutputName>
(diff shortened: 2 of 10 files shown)


xen2:
I looked into it quickly.
There are several needs:
- Stride itself:
  - Compile-time: it makes sense to merge Windows/macOS/Linux platforms.
  - Runtime: We might still want to differentiate at runtime though (using `Stride.Core.Platform.Type`).
- User projects:
  - Compile-time: right now we need to know platform because it is sent to AssetCompiler (but use is quite limited, and mostly for GraphicsAPI determination so we could just pass the GraphicsAPI separately and possibly merge Windows/macOS/Linux? I need to check further all uses in the AssetCompiler).
  - Runtime: some code still change depending on platform so `Stride.Core.Platform.Type` need to differentiate

As a result, we might want to separate Compile-time from Runtime.
Several options:
(1) Like you did: stick to a single enum `PlatformType` for both compile and runtime, and add `Desktop = Windows`. This might be misleading because we always have to remember which "context" we are in (is this a compile time or runtime setting?)
(2) Add another enum for compile-time.
(3) Stop using enum for compile-time and rely directly on TFM. Advantage is that targets might be easier to understand (usual `net6.0-ios` and such). We still want a place where we map Platform from their names to TFM so that in TeamCity or Stride.build we can keep checkboxes such as `Windows`  or `iOS` (even if we bump from `net6.0-ios` to `net7.0-ios`) so that we don't have to update CI everytime we bump TFMs, so it might end up being almost same as (2) anyway.

xen2:
I did a bunch of commits in `master` to try to improve the situation.
I removed most `StridePlatform` usage in Stride internal build system. It's still not completely removed yet, I decided to defer complete removal in a future commit for further testing (longer term, we could keep `StridePlatform` meaning only for "executables" such as tests projects calling AssetCompiler, in sync with user project).

Instead, the build system now relies on TFMs which are clearly defined in a single location.

Note: native currently compiles Windows, Linux and macOS on `net6.0`. The optimization to compile only Windows with certain flags is not done yet. But since we properly cache and check C++ source updates, it's probably not urgent.

manio143:
I believe xen2 has implemented some of this PR and I'm closing it.

PR #1180, opened 2021-11-06
[WIP] Silk Vulkan

ykafia:
# PR Details

Switching Vortice.Vulkan to Silk.NET.Vulkan

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Graphics/Vulkan/DescriptorSetLayout.Vulkan.cs
+++ b/sources/engine/Stride.Graphics/Vulkan/DescriptorSetLayout.Vulkan.cs
@@ -7,8 +7,8 @@ using Stride.Core;
 using Stride.Core.Collections;
 using Stride.Shaders;
 #if STRIDE_GRAPHICS_API_VULKAN
-using Vortice.Vulkan;
-using static Vortice.Vulkan.Vulkan;
+using Silk.NET.Vulkan;
+using static Silk.NET.Vulkan.Vk;
 
 namespace Stride.Graphics
 {
@@ -51,16 +51,16 @@ namespace Stride.Graphics
         }
 #endif
 
-        internal static unsafe VkDescriptorSetLayout CreateNativeDescriptorSetLayout(GraphicsDevice device, IList<DescriptorSetLayoutBuilder.Entry> entries, out uint[] typeCounts)
+        internal static unsafe Silk.NET.Vulkan.DescriptorSetLayout CreateNativeDescriptorSetLayout(GraphicsDevice device, IList<DescriptorSetLayoutBuilder.Entry> entries, out uint[] typeCounts)
         {
-            var bindings = new VkDescriptorSetLayoutBinding[entries.Count];
-            var immutableSamplers = new VkSampler[entries.Count];
+            var bindings = new Silk.NET.Vulkan.DescriptorSetLayoutBinding[entries.Count];
+            var immutableSamplers = new Sampler[entries.Count];
 
             int usedBindingCount = 0;
 
             typeCounts = new uint[DescriptorTypeCount];
 
-            fixed (VkSampler* immutableSamplersPointer = &immutableSamplers[0])
+            fixed (Sampler* immutableSamplersPointer = &immutableSamplers[0])
             {
                 for (int i = 0; i < entries.Count; i++)
                 {
@@ -70,12 +70,12 @@ namespace Stride.Graphics
                     if (entry.ArraySize == 0)
                         continue;
 
-                    bindings[usedBindingCount] = new VkDescriptorSetLayoutBinding
+                    bindings[usedBindingCount] = new Silk.NET.Vulkan.DescriptorSetLayoutBinding
                     {
-                        descriptorType = VulkanConvertExtensions.ConvertDescriptorType(entry.Class, entry.Type),
-                        stageFlags = VkShaderStageFlags.All, // TODO VULKAN: Filter?
-                        binding = (uint)i,
-                        descriptorCount = (uint)entry.ArraySize
+                        DescriptorType = VulkanConvertExtensions.ConvertDescriptorType(entry.Class, entry.Type),
+                        StageFlags = ShaderStageFlags.ShaderStageAll, // TODO VULKAN: Filter?
+                        Binding = (uint)i,
+                        DescriptorCount = (uint)entry.ArraySize
                     };
 
                     if (entry.ImmutableSampler != null)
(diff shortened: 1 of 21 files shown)


ykafia:
Debug time for this one, it should be fairly straight forward to debug since 99% of the code is the same, Vortice did very little in abstracting Vulkan.

amerkoleci:
Why this? Vortice.Vukan is maintained and always updated (by me), I can understand SharpDX replacement but don't actually see the need of this changes. @xen2?

ykafia:
This PR was mostly my decision and there's no real reason for it to happen besides seeing what it would look like if we were using one library for all apis (plus Assimp and windowing).

amerkoleci:
> This PR was mostly my decision and there's no real reason for it to happen besides seeing what it would look like if we were using one library for all apis (plus Assimp and windowing).

Doesn't make sense but whatever, go with it, assimp and windowing are all separate bindings.

tebjan:
From what I understand, the Silk binding has a lower overhead because it is generated differently (structs instead of classes). Is this true, or do I remember it incorrectly?

MingHuiLiu:
> From what I understand, the Silk binding has a lower overhead because it is generated differently (structs instead of classes). Is this true, or do I remember it incorrectly?

The way of loading dll and calling api is also different。and it binding is updated from upstream

MingHuiLiu:
> > From what I understand, the Silk binding has a lower overhead because it is generated differently (structs instead of classes). Is this true, or do I remember it incorrectly?
> 
> The way of loading dll and calling api is also different。and it binding is updated from upstream

Sorry, I haven’t seen the implementation of Vortice.Vulkan😓

amerkoleci:
> From what I understand, the Silk binding has a lower overhead because it is generated differently (structs instead of classes). Is this true, or do I remember it incorrectly?

That's the case for d3d bindings, vulkan bindings uses struct, just take look at the generated commmand and structs:
https://github.com/amerkoleci/Vortice.Vulkan/blob/main/src/Vortice.Vulkan/Generated/Structures.cs
https://github.com/amerkoleci/Vortice.Vulkan/blob/main/src/Vortice.Vulkan/Generated/Commands.cs

Perksey:
From my perspective (my views do not represent Silk.NET, Stride, or anyone else's views)

There is no tangible reason to use Silk.NET over Vortice, or Vortice over Silk.NET. The codegen is basically identical, unlike other cases like DirectX where Silk.NET may have an edge. (**but read on!**)

<sub>I guess if you wanted to get pernickety you could use "it looks more at home with C# code" as a tangible reason given our overloading system and our naming conventions</sub>

---

Silk.NET has always meant to be the "one library" you need for anything graphics related, and is designed so that you can fully buy into the Silk.NET stack if you so please, and not need any other bindings library. Having less libraries reduces maintenance hassle as well as having greater predictability of breaking changes, releases, intellectual property, and so on (there are more niche benefits as well like reduced risk of supply-chain attacks).

For instance, you can always count on Silk.NET having an update on the first Friday of every month (or within 1 or 2 days of that date as let's face it this is OSS) and our generator always fetches the most up-to-date specifications for all of our bindings when releasing them. This could, in theory, let Stride control the version of Silk.NET they're using from a common properties file and have that traverse across the entire engine.

And the final reason I guess could be that you can count on us not dying tomorrow. There are multiple people on the team with full control of the project and, if one of us went AWOL, the project would continue unharmed. The same cannot be said for Vortice. Moreover, if all of us were to go AWOL the Foundation has a copy of the keys to everything so they could find a new maintainer (or even let Stride engulf the project) and once again the project would continue unharmed.

---

What excites me about this line of work is the relationship that our projects are able to build. .NET Foundation projects working side-by-side and helping each-other is a really good image: Stride depending on Silk.NET, Silk.NET depending on ClangSharp - all .NET Foundation projects working in tandem.

Especially in architecting our upcoming rewrite, having the much-needed input of larger consumers like Stride in addition to the "hobbyist" input we get today will allow us to make a much more fleshed-out library that meets more users' needs.

But again, this is my view and not anyone else's; and while I did reach out to Stride a loooooooong time ago all recent work has been done at their and their contributors' accord - the Silk.NET team themselves have had no influence on Stride's 
decisions here.

Regardless, we're happy to have Stride on board and are excited to hopefully work closely with them in the future :)

---

TL;DR: consistency, dependency consolidation, predictability, maintainability, reliability, perception, and teamwork.

amerkoleci:
@Perksey, no problem for me, go with whatever works for stride and others :)

sources/engine/Stride.Graphics/Vulkan/DescriptorSetLayout.Vulkan.cs
```diff
@@ -89,21 +89,21 @@ internal static unsafe VkDescriptorSetLayout CreateNativeDescriptorSetLayout(Gra
...
                 {
-                    sType = VkStructureType.DescriptorSetLayoutCreateInfo,
-                    bindingCount = (uint)usedBindingCount,
-                    pBindings = usedBindingCount > 0 ? (VkDescriptorSetLayoutBinding*)Core.Interop.Fixed(bindings) : null,
+                    SType = StructureType.DescriptorSetLayoutCreateInfo,
+                    BindingCount = (uint)usedBindingCount,
+                    PBindings = usedBindingCount > 0 ? (DescriptorSetLayoutBinding*)Core.Interop.Fixed(bindings) : null,
                 };
-                vkCreateDescriptorSetLayout(device.NativeDevice, &createInfo, null, out var descriptorSetLayout);
+                GetApi().CreateDescriptorSetLayout(device.NativeDevice, &createInfo, null, out var descriptorSetLayout);
```
Perksey, line 106: Note: GetApi is an extremely expensive call, please don't use it every time you use an API function!

The `Vk` object stores all function pointers, which are lazily initialized (i.e. the first time you use a function on that object, there's a small penalty in which we call `vkGetInstanceProcAddr`, `vkGetDeviceProcAddr`, or the kernel `GetProcAddress` method). If you make a new object for every call, we call this every time and incur this penalty every time. Moreover, there's a few allocations that happen under-the-hood that are intended to be long-lived when instantiating the object, and once again are pretty expensive.

As a general rule of thumb, you should have one Vk instance for each instance-device combo. This ensures you load the correct function pointers for each device, and don't encounter any indirection in the Vulkan driver in multi-GPU scenarios.

Usually this means one object for a whole program, but with Stride it doesn't look to be as simple. It looks like:
- create an object in [GraphicsAdapterFactory.InitializeInternal](https://github.com/stride3d/stride/blob/7e836297cb5930c01e6dfa0183e7f3cc64748fb6/sources/engine/Stride.Graphics/Vulkan/GraphicsAdapterFactory.Vulkan.cs#L24)
- add an argument to the `GraphicsAdapterFactoryInstance` constructor for the Vk object. make it nullable and default to null, and add logic such that if not provided it just creates a new API object (this is primarily for the `debugInstance`)
- when constructing the `defaultInstance`, pass this Vk object into the constructor. `defaultInstance` can now be thought of as "owning" that API object, and when `vkCreateInstance` is called it will capture your instance handle implicitly and rebuild the function table such that it calls `vkGetInstanceProcAddr` with that handle.
- `GraphicsDevice` should also hold a Vk object. In [GraphicsDevice.InitializePlatformDevice](https://github.com/stride3d/stride/blob/5bb8915ccc596c5bd0a404b5f59ed61031a9251e/sources/engine/Stride.Graphics/Vulkan/GraphicsDevice.Vulkan.cs#L248), do something like:
  ```cs
  // TODO clean this up when https://github.com/dotnet/Silk.NET/issues/697 is a thing
  if (_vk is null)
  {
      _vk = Vk.GetApi();
      _vk.CurrentInstance = GraphicsAdapterFactory.GetInstance(false).NativeInstance;
  }
  ```
- perhaps find some way to have a `Vk` property in [GraphicsResourceBase](https://github.com/stride3d/stride/blob/5bb8915ccc596c5bd0a404b5f59ed61031a9251e/sources/engine/Stride.Graphics/Vulkan/GraphicsResourceBase.Vulkan.cs) to make code look nicer.

This is just my finger-in-the-air estimate of how best to work this into the engine, and is by no means "the best way" (as I'm not that familiar with Stride's internals), but basically it boils down to:
- have one Vk object for each instance-device combo.
- pass it around to the objects that need it.

There are hackier ways to work this in, for example have a static class containing the Vk object, but the above is the best way to account for multiple logical devices.

sources/engine/Stride.Graphics/Vulkan/PipelineState.Vulkan.cs
```diff
@@ -229,8 +229,8 @@ private unsafe void CreateRenderPass(PipelineStateDescription pipelineStateDescr
             if (hasDepthStencilAttachment)
                 attachmentCount++;
 
-            var attachments = new VkAttachmentDescription[attachmentCount];
-            var colorAttachmentReferences = new VkAttachmentReference[renderTargetCount];
+            var attachments = new AttachmentDescription[attachmentCount];
```
Perksey, line 232: One optimization you could consider in general (potentially for the future) is to conditionally stackalloc:
```cs
Span<AttachmentDescription> attachments = (attachmentCount * sizeof(AttachmentDescription)) < 512 ? stackalloc AttachmentDescription[attachmentCount] : new AttachmentDescription[attachmentCount];
```

Might be a bit verbose for this codebase, but it's just a little thing to not invoke the GC if you don't need it. The `512` is so we don't risk stack overflows and don't allocate too much on the stack - `512` is just a number I've pulled from my backside and I usually scale it according to how much code like this I have i.e. if I have one of these "conditional stackallocs" then I use 512, if I have two I use 256, etc...

ykafia:
Needs to start again

PR #1254, opened 2021-12-20
Improve native library search on Linux

manio143:
# PR Details

## Description

For the Stride native libs (`libcore`, `libstride`, `libstrideaudio`) the existing way of locating libraries was mostly fine, but there may be libraries used like `freetype` (and previously SDL) which are assumed to be installed in the system. In order to find those, we need to pass in the correct name of the library file and let the CLR search for it in global library locations.

In order to make the code easier to read, all calls to `PreloadLibrary` should not include the extension (e.g. `.dll`).

## Related Issue

This is a follow up to the work of #1173 as I found myself having to manually create links for some libraries the current approach wasn't finding.

## Types of changes

## Checklist

### DIFF
--- a/sources/core/Stride.Core/Native/NativeLibraryHelper.cs
+++ b/sources/core/Stride.Core/Native/NativeLibraryHelper.cs
@@ -11,6 +11,7 @@ namespace Stride.Core
 {
     public static class NativeLibraryHelper
     {
+        private const string UNIX_LIB_PREFIX = "lib";
         private static readonly Dictionary<string, IntPtr> LoadedLibraries = new Dictionary<string, IntPtr>();
 
         /// <summary>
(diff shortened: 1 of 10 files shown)


sources/core/Stride.Core/Native/NativeLibraryHelper.cs
```diff
@@ -65,19 +67,54 @@ public static void PreloadLibrary(string libraryName, Type owner)
...
+                        extension = ".dll";
+                        break;
+                    case PlatformType.Linux:
+                        extension = ".so";
+                        break;
+                    case PlatformType.macOS:
+                        extension = ".dylib";
+                        break;
+                    default:
+                        throw new PlatformNotSupportedException();
```
tebjan, line 82: won't this throw on mobile?

manio143: This method starts with `#if STRIDE_PLATFORM_DESKTOP`, so I'm assuming mobile will not invoke this code. I don't think we load native libs on mobile (and use some platform specific APIs)

Issue #1069, opened 2021-06-07
`ViewportRestore` incorrectly saving current state


vdvman1:
**Release Type**: Official Release

**Version**: 4.1.0.1459-beta (also affects 4.0.1.1401)

**Platform(s)**: Tested on windows, almost certainly affects all platforms

**Describe the bug**
When calling `RenderContext.SaveViewportAndRestore()` to save the viewport the `ViewportRestore` class only stores a reference to the current `ViewportState` class. As a result any changes to the current viewport state will affect the "saved copy", and when it is restored nothing is changed. Overall this means the viewport state changes will bleed outside the `using` block.

This can be seen affecting the `RenderTextureSceneRenderer`, any renderers invoked after this renderer finishes will see the viewport state set by the `RenderTextureSceneRenderer` rather than the original, leading to incorrect aspect ratios

**To Reproduce**
1. Create a new project using the `Sample: Animation` sample
2. Run and observe the correct aspect ratio for the main model in the middle of the screen
3. Change the width or height of the `RenderTexture` asset
4. Run again and notice the main model in the middle of the screen stretches the same way as the render textures

Alternative reproduction steps:
1. Create a new game project
2. Follow the official [manual](https://doc.stride3d.net/latest/en/manual/graphics/graphics-compositor/render-textures.html) for render textures, except place the `Camera Renderer` for the render texture prior to the main camera renderer
3. Run and observe the incorrect aspect ratio of the sphere

**Expected behavior**
`ViewportRestore` should either make a deep copy of the `ViewportState` or replace the current state with a new instance so that the original viewport is not modified
The aspect ratio of a render texture should not affect independent renderers in the parent scene collection

**Screenshots**
With the default `RenderTextureSceneRenderer` which calls  `RenderContext.SaveViewportAndRestore()` which creates an instance of the default `ViewportRestore` class:
![default-rendertexture](https://user-images.githubusercontent.com/2172023/120950071-4eb71380-c789-11eb-9f18-f5369747a85c.png)

With a custom `ViewportRestore` class which copies the `Viewport0` value rather than the `ViewportStatus` reference:
![fixed-rendertexture](https://user-images.githubusercontent.com/2172023/120950165-8de56480-c789-11eb-9072-8dec3bf65b8d.png)

These are in the exact same scene with the same graphics compositor other than the change to the custom render texture scene renderer script

**Log and callstacks**
N/A

**Additional context**
[...]

PR #1259, opened 2022-01-03
Added Clone method to ViewportState and copy parameter to ViewportRestore constructor

Artromskiy:
# PR Details

Added Clone method to ViewportState and copy parameter to ViewportRestore constructor

## Description

Done to fix issue of not correctly working render texture

## Related Issue

https://github.com/stride3d/stride/issues/1069

## Motivation and Context

https://github.com/stride3d/stride/issues/1069

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Games/GraphicsDeviceManager.cs
+++ b/sources/engine/Stride.Games/GraphicsDeviceManager.cs
@@ -1039,7 +1039,6 @@ namespace Stride.Games
                             {
                                 try
                                 {
-                                    GraphicsDevice.ColorSpace = graphicsDeviceInformation.PresentationParameters.ColorSpace;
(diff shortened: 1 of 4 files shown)


Eideren:
Maybe you could make ``ViewportState`` into a struct ? I haven't looked thoroughly at the source but it sounds like that would be preferable given that it is cloned fairly often and so generates a bit of unnecessary garbage for the GC to churn through ? Pinging @vdvman1 in case he has more insight into this.

Artromskiy:
Changing ViewportState to struct also works

Eideren:
Great, thanks ! It seems safe so I'll just merge this in, we'll revert and come back to this if it ever becomes an issue

tebjan:
Very good one, we had a few bugs because of this... Thanks!

Issue #1035, opened 2021-03-18
"Color Space" setting change to "Linear" will get wrong display


D3ZAX:
Get wrong display when changing "Color Space" in "Game settings" from "gamma" to "Linear"!
Will be correct after reopen scene.

Engine Version: 4.0.1.1401

Correct:
![image](https://user-images.githubusercontent.com/15343372/111630767-a77fab80-882d-11eb-9b4d-3fae424fb5c1.png)

Wrong:
![image](https://user-images.githubusercontent.com/15343372/111631036-e9a8ed00-882d-11eb-8b0a-83cb06bb14db.png)

Artromskiy:
Started doing this issue

PR #1260, opened 2022-01-03
Color space change bug fix

Artromskiy:
# PR Details

Changes of color space at GameSettings now applied correctly

## Description

GraphicsDeviceManager now applies changes to GraphicsDevice.ColorSpace before calling GraphicsDevice.Presenter.Resize (resize method uses color space info stored at graphics device to resize buffer with PixelFormat)
Texture.Direct3D now always converts textureDescrioptionFormat to srgb (most likely it was a bug, so only srgb format converted to srgb again)

## Related Issue

https://github.com/stride3d/stride/issues/1035

## Motivation and Context

https://github.com/stride3d/stride/issues/1035

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Graphics/Direct3D/Texture.Direct3D.cs
+++ b/sources/engine/Stride.Graphics/Direct3D/Texture.Direct3D.cs
@@ -95,7 +95,8 @@ namespace Stride.Graphics
             var newTextureDescription = ConvertFromNativeDescription(texture.Description);
 
             // We might have created the swapchain as a non-srgb format (esp on Win10&RT) but we want it to behave like it is (esp. for the view and render target)
-            newTextureDescription.Format = newTextureDescription.Format.ToSRgb();
+            if(isSrgb)
+                newTextureDescription.Format = newTextureDescription.Format.ToSRgb();
 
             return InitializeFrom(newTextureDescription);
         }
--- a/sources/engine/Stride.Graphics/Direct3D/Texture.Direct3D.cs
+++ b/sources/engine/Stride.Graphics/Direct3D/Texture.Direct3D.cs
@@ -95,8 +95,7 @@ namespace Stride.Graphics
             var newTextureDescription = ConvertFromNativeDescription(texture.Description);
 
             // We might have created the swapchain as a non-srgb format (esp on Win10&RT) but we want it to behave like it is (esp. for the view and render target)
-            if (isSrgb)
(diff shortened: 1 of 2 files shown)


sources/engine/Stride.Graphics/Direct3D/Texture.Direct3D.cs
```diff
@@ -95,8 +95,7 @@ internal Texture InitializeFromImpl(Texture2D texture, bool isSrgb)
             var newTextureDescription = ConvertFromNativeDescription(texture.Description);
 
             // We might have created the swapchain as a non-srgb format (esp on Win10&RT) but we want it to behave like it is (esp. for the view and render target)
-            if (isSrgb)
-                newTextureDescription.Format = newTextureDescription.Format.ToSRgb();
+            newTextureDescription.Format = newTextureDescription.Format.ToSRgb();
```
tebjan, line 98: This line seems to do the same as the one in the if branch

Eideren:
Removing the srgb test seems to significantly alter how textures are displayed:
Linear:
image
Gamma:
image
Gamma with srgb test:
image
Note the sphere getting darker while the background gets lighter.
Linear with srgb test looks the same as linear without.

I'm not that familiar with the subject so I might be wrong here, bare with me.
That background texture has srgb setup.
image
It is set as a 2d background, so it is not affected by lighting or shading.
In unity (I'm not saying that that's what we should emulate, just as a comparison) in gamma space, whether the texture is in sRGB or not doesn't affect the rendering of it, while in linear, with sRGB off it looks lighter and with it on it looks normal.
Is this expected behavior ?

Artromskiy:
I have not done tests with normal textures without lighting, just trying to achieve the same lighting of materials. I made two screenshots with different color space when cutting the scene (I used them as a reference). After changing the code to achieve the same lighting when changing the space in the game settings.

Eideren:
Alright, I'll create a new issue to make sure the different color-spaces behave as expected and merge this in then. Thanks again !

Artromskiy:
but i've reverted Texture.Direct3D, it's not working as expected now

Artromskiy:
@Eideren ?

> but i've reverted Texture.Direct3D, it's not working as expected now

Artromskiy:
It seems I was wrong, everything works fine. Last time, the color was incorrect unless the isSrgb condition was removed from the InitializeFromImpl Texture.Direct3D. Maybe something was wrong with the Visual Studio hot reload.

PR #1263, opened 2022-01-06
Add missing sdl keys

TheKeyblader:
# PR Details

Add missing key mapping for KeyboardSDL

## Related Issue

#1182 

## Types of changes

- [ ] Docs change / refactoring / dependency upgrade
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [ ] My change requires a change to the documentation.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #1265, opened 2022-01-07
[Input] Improve Input.Keys docs and fix a couple of keymapping in SDL

Eideren:
# PR Details
Related to PR #1263 
- Made ``NumPadDecimal`` obsolete, the US key mapped to character ``.`` and ``>`` is ``OemPeriod``, not ``Decimal``. ``Decimal`` already maps to the numpad in winforms and in rawinput most likely as well given that there isn't any specific logic for it. Made changes to the SDL keyboard to reflect that.

- Fixed a conflict between ``OemBackslash`` and ``OemPipe``.
- ``NumPadEnter`` and ``OemPlus`` mapped to the right key.
- Support for ``LaunchApplication1`` and ``LaunchApplication2`` keys.

- Swapped SDL keyboard key mapping from a dictionary to a switch, I'm not too sure why we would use a dictionary for this when a switch is as easy to use and maintain ... ? Separated that change into its own push in case that's something we don't want.

- Improved the summary for most of the ``Input.Keys``, it should be clearer now what those oems are and when two label map to the same value.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Input/Keys.cs
+++ b/sources/engine/Stride.Input/Keys.cs
@@ -906,6 +906,6 @@ namespace Stride.Input
         /// <summary>
         /// The numeric keypad 'decimal' key.
         /// </summary>
-        NumPadDecimal = 181,
+        [System.Obsolete($"Use {nameof(Decimal)} instead")] NumPadDecimal = 181,
     }
 }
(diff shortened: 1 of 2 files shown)


tebjan:
LGTM, thanks!

PR #1274, opened 2022-01-11
[Editor] Fix missing editor keyboard input

### LINKED ISSUE #94: Editor camera stutters/does not process inputs

Eideren:
# PR Details
See #94
Setting parent to none for the embedded windows (like the scene view) is enough to ensure that all inputs are received, one of the parent control is most likely eating some of the inputs. Debugging the WPF hierarchy would be a real undertaking so I'll leave it at that for now.
This PR sets the ``WS_CHILD`` flag off whenever the layout changes, not exactly sure how it works internally but the window still behaves as a child window while properly receiving inputs and testing the rest of the engine didn't result in any unexpected behavior.

Also removed ``MissingInputHack`` since it relied on rawinput which isn't part of the editor since the net5 changes I think ?

## Related Issue
Fix #94

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Input/Windows/InputSourceWinforms.cs
+++ b/sources/engine/Stride.Input/Windows/InputSourceWinforms.cs
@@ -48,7 +48,6 @@ namespace Stride.Input
             input = inputManager;
 
             uiControl.LostFocus += UIControlOnLostFocus;
-            MissingInputHack();
 
             // Hook window proc
             defaultWndProc = Win32Native.GetWindowLong(uiControl.Handle, Win32Native.WindowLongType.WndProc);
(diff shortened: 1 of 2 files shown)


Eideren:
I tested this one more thoroughly again today and it does work quite well. I'll merge it in for the upcoming 4.1, if anyone stumbles on issues in the mean time I'll roll it back.

johang88:
The style flag update breaks drag & drop of assets into the scene view

tebjan:
We have to revert this for now, and look for another fix or understand the issue better.

PR #1275, opened 2022-01-12
Expose RenderLight bounding boxes

ykafia:
# PR Details

## Description
Exposing the BoundingBox and BoundingBoxExt to users.

## Motivation and Context

Those fields are useful when implementing a new light externally. 

## Types of changes

## Checklist


Eideren:
Those fields are written to by that class when it calls ``Type.ComputeBounds`` as long as ``RenderLight.Type.HasBoundingBox`` returns true ? Just make sure that your custom light type inherits from ``IDirectLight`` and set ``Type`` to it ?
Is that not enough for your use case ?

ykafia:
You're right, `Type.ComputeBounds` computes the bounding box, it works but it looks like it would recompute the bounding box already updated and stored in DirectLight.
I might be wrong though

Eideren:
> it works but it looks like it would recompute the bounding box already updated and stored in DirectLight.

I'm not following, what do you mean ?

ykafia:
Oh sorry I wasn't very clear, on RenderLight you get an updated value of the bounding box thanks to this function :

https://github.com/stride3d/stride/blob/1f89849f57fd9e0e4ce8c2ddfe96771f320a1bc6/sources/engine/Stride.Rendering/Rendering/Lights/RenderLight.cs#L51-L65

So, instead of using the `RenderLight.Type.ComputeBounds(... )` method, I want to use the field  `RenderLight.BoundingBox`

Some Light renderers use this field directly (and it works since they're internal), so I guessed they might be helpful

Eideren:
Sure, that actually makes sense, thanks !

tebjan:
I would not convert the field to a property with method access, render objects are high performance and low level code that work with fields directly. I would just make the field public...

Eideren:
With that kind of property JIT will inline it into a field, see [Sharplab JIT Asm](https://sharplab.io/#v2:C4LghgzgtgPgAgJgIwFgBQcDMACR2DC2A3utmbjnACzYCyAFAJTGnlsBuYATtgMbYBebADsApgHdsAFVERg+ADaQITANys2ZXgDoAClwD2AB0HYkABnVpN5HfuOmkAVis2t2gGIBLUQoAmpgiWGtgAvujhaOh4MnKKyugk1uRY2F7CwNj2JkTYAOaiwKppGaJcwmAK2BCFxZFsqemZ3r5+VqFAA=)

tebjan:
Yes sure, it's just the only property on a render object, as far as I know. The code style is different now...

Eideren:
Sure, you can make a PR and I'll merge it if you want, you'll just want to specify that it would be preferable to use the ``Type``'s ``ComputeBounds`` instead of writing to that field directly

Earthmark:
I believe this broke CollectVisibleLights in \sources\engine\Stride.Rendering\Rendering\Lights\ForwardLightingRenderFeature.cs , it used the field as a ref field but it's now a property. Seems this may have broken the build!

The error in question:
```
Severity	Code	Description	Project	File	Line	Suppression State
Error	CS0206	A property or indexer may not be passed as an out or ref parameter	Stride.Rendering (net6.0)	<repo path>\sources\engine\Stride.Rendering\Rendering\Lights\ForwardLightingRenderFeature.cs	575	Active

```

PR #1281, opened 2022-01-15
[Test] Fixed compiler script test failures and oddities

Earthmark:
# PR Details
This fixed the script asset tests that were failing on my local machine. This also addresses a side effect of that failure taking out other tests.

## Description

Fixed the console output not getting reset on a test failure, resulting in other tests failing. Now the output is always reset. _Further improvements may be to avoid using console, as it's a singleton and they cause exactly this problem._

Fixed compilation failing in dotnet 6, due to missing a reference to an assembly containing Console or System.Runtime. The following dependencies are now used:
 * `System.Runtime.dll` - Needed for type Object and Decimal, these are specifically not System.Object, it's something else (I'm not sure what).
 * `System.Console.dll` - Needed for System.Console.
 * `System.Private.CoreLib.dll` - Needed for System.Object, System.Void, and other things.

My guess is previously these were all covered by mscorelib or a similar framework-ish assembly.

Testing to ensure these assemblies are available on the target platforms likely needs to happen, I'm curious what the CI system thinks of this.

## Related Issue

None found, this was an issue found while trying to get the tests running locally.

## Motivation and Context

This was found while trying to get the unit tests to run, they did not run successfully on my machine.

## Types of changes

## Checklist

_The tests pass locally where as they failed previously, CI verification is needed as this seems like a platform dependency thing._

### DIFF
--- a/sources/engine/Stride.Assets.Tests/TestVisualScriptCompiler.cs
+++ b/sources/engine/Stride.Assets.Tests/TestVisualScriptCompiler.cs
@@ -185,21 +185,26 @@ namespace Stride.Assets.Tests
 
             using (var textWriter = new StringWriter())
             {
-                Console.SetOut(textWriter);
+                try
+                {
+                    Console.SetOut(textWriter);
 
-                // Create class
-                var testInstance = CreateInstance(new[] { compilerResult.SyntaxTree });
-                // Execute method
-                testCode(testInstance);
+                    // Create class
+                    var testInstance = CreateInstance(new[] { compilerResult.SyntaxTree });
+                    // Execute method
+                    testCode(testInstance);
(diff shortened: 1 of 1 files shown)


Earthmark:
No CI system ran... is there a way to kick team city into running a pass over this PR?

xen2:
It's only active for collaborators/members (for security reason, otherwise anybody could trigger code to run on CI by creating a PR).
I have just added you as an outside collaborator. Not sure if it will start automatically or it needs a new commit to notice the change, let's see.

Earthmark:
No dice it seems.

Earthmark:
Let's try some things to see if TC will pick this up...

PR #1276, opened 2022-01-13
[Build] Fix nuget conflicts with latest VS update

Eideren:
# PR Details
This PR fixes issues that the latest visual studio update introduced when doing a fresh build. AFAICT there was a conflict between the nuget versions used when building and in the package references.

@xen2 the AssetComplier outputs the following warnings which I don't remember it doing before ?
Image `NullWidth` has an invalid image source file `[...]/sources/data/tests/Sphere.png`, resulting texture will be null.
Image `NegativeHeight` has an invalid image source file `[...]/sources/data/tests/Sphere.png`, resulting texture will be null.
But it seems to be expected behavior given that the line that creates those is as such:
```cs
// Invalid sprites
CreateSpriteEntity(debugSheet, "NullWidth");
CreateSpriteEntity(debugSheet, "NegativeHeight");
```
Is this fine or is the warning legit ?

Also might need some tweaks in teamcity ? Not too sure about that part @xen2 ?

## Types of changes
- [ ] Docs change / refactoring / dependency upgrade
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist
- [ ] My change requires a change to the documentation.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #1283, opened 2022-01-15
[Readme] Updated the readme to call out msbuild instead of dotnet.

Earthmark:
# PR Details

## Description

The dotnet command can't compile C++, which causes build failures. msbuild from the build tool (or vs) just fine on its own.

The previous step adds msbuild to the path, so this should be fine. To my knowedge dotnet is doing this same thing, just it doesn't have the libraries accessable to the msbuild install at the added-path-location.

Using the `dotnet` you get errors of the flavor:
```
$repoPath\sources\tools\Stride.Assimp.Translation\Stride.Assimp.Translation.vcxproj(34,3): error MSB4019: The imported project "C:\Microsoft.Cpp.Default.props" was not found. Confirm that the expression in the Import declaration "\Microsoft.Cpp.Default.props" is correct, and that the file exists on disk.
```

Using `msbuild` it compiles more successfully.

This may be due to me using a different build pipeline, for instance I need to use https://github.com/stride3d/stride/pull/1276 to build. I'm not sure how this would have worked in previous versions, but it's possible it did maybe? Perhaps something was directly including the cpp props files.

Using the build tools enviorments may avoid this error with dotnet, as that includes a bunch of other things. I didn't test that, and the readme doesn't suggest using that batch file though so I didn't!

## Related Issue

Unknown (didn't check)

## Motivation and Context

Trying to follow the readme didn't work, with these updated steps it works a bit better.

## Types of changes

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ Ask for help or report issues:
 1. Install [Visual Studio Build Tools](https://aka.ms/vs/17/release/vs_BuildTools.exe) with the same prerequisites listed above
 2. Add MSBuild's directory to your system's *PATH* (ex: `C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\MSBuild\Current\Bin`)
 3. Open a command prompt, point it to a directory and clone Stride to it: `git clone https://github.com/stride3d/stride.git`
-4. Navigate to `/Build` with the command prompt, input `msbuild /t:Restore Stride.sln` then `msbuild /t:Build Stride.sln` (Add `-m` for improved performance)
+4. Navigate to `/Build` with the command prompt, input `dotnet restore Stride.sln` then `compile.bat`
 
 If building failed:
 * If you skipped one of the `Prerequisites` thinking that you already have the latest version, update to the latest anyway just to be sure.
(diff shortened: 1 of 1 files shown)


Earthmark:
... team city ran on this one, I may need to remake https://github.com/stride3d/stride/pull/1281 to see if that kicks TC into running over that code.

Eideren:
Restoring from MSBuild instead of dotnet makes sense, but ``compile.bat`` does use MSBuild to compile projects, what does that change provide ?

Earthmark:
I see my confusion now, I thought the directions were incomplete and dropped the ball saying "ok, now compile" which I interrupted as `dotnet build`, which fails. It didn't occur to me to look that a file named compile was there.

I'll restructure this request to resolve that new confusion point for me, I'm curious if it makes it more understandable for others as well.

Eideren:
Please, do include your msbuild restore, that's more inline with the msbuild tools requirement than using dotnet. We'll move over to full dotnet when/once we don't have msbuild requirements.

Earthmark:
Swapped it back to msbuild for the restore command, I did notice that compile.bat doesn't report an error if a compilation error occurs, not sure if that's intended but ah well.

Eideren:
That's an MSBuild feature/bug, the bat compiles with minimal verbosity by default instead of normal, do `compile verbosity:n`.
Minimal does print errors and warnings, it just doesn't dump them again at the end of compilation.

Eideren:
Thanks !

Earthmark:
I'll make a pr to change the suggested verbosity later today, thanks for merging.

PR #1286, opened 2022-01-17
[Readme] Added a verbosity clamp for msbuild, so it's a bit less screamy.

Earthmark:
# PR Details

Adds a verbosity argument to msbuild restore based on the trailing comment in https://github.com/stride3d/stride/pull/1283 , personally I don't mind the extra output  but if it's the style of the project then cool.

## Description

Changes a command in a trivial way.

## Related Issue

See the conversation in https://github.com/stride3d/stride/pull/1283

## Motivation and Context

See the conversation in https://github.com/stride3d/stride/pull/1283

## Types of changes

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ Ask for help or report issues:
 1. Install [Visual Studio Build Tools](https://aka.ms/vs/17/release/vs_BuildTools.exe) with the same prerequisites listed above
 2. Add MSBuild's directory to your system's *PATH* (ex: `C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\MSBuild\Current\Bin`)
 3. Open a command prompt, point it to a directory and clone Stride to it: `git clone https://github.com/stride3d/stride.git`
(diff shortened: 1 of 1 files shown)


Eideren:
Just to clarify, what I meant to say here https://github.com/stride3d/stride/pull/1283#issuecomment-1014373944 was that the `compile.bat` file uses [m]inimal verbosity by default here https://github.com/stride3d/stride/blob/d0c8f2e5b88acc88f6295aa691a4b57094da9c0d/build/compile.bat#L7-L9 and that you could make a PR if you want to change it to use [n]ormal by default.
Specifying `minimal` for the restore command line seems like adding complexity without being that useful as minimal and normal dump almost as much to the console ? Sorry for making you go through this ...

Earthmark:
Your fine, I'm mostly making the readme changes on lunch breaks on my phone so it's nice and easy! I see what you mean, I'm curious why the default was used in the command though. The default msbuild build is way more information than is really consumable... 

I wonder where the happy medium is. I'll revert this change and think this over, I think the change needs to have a positive impact, and I'm not sure the build command being more verbose is a positive impact...

PR #1287, opened 2022-01-19
[WIP] Base spline support

Aggror:
# PR Details
This PR contains basic support for creating splines and visualizing those splines.

- 3 new components: SplineComponent, SplineNodeComponent and SplineTraverserComponent
- Visualisation is supported in the editor by viewing 
 - Curves
 - Boundingboxes
 - Nodes
 - 
- Other spline components are in the works as well, but purposely left out to focus on getting the core for Splines better in shape. That is also the reason there is no documentation written yet.
 - SplineMeshComponent
 - SplineDecoratorComponent  

## Motivation and Context
Splines are useful for camera paths, or any other object that requires bezier curve for movement, generating geometry and for distributing geometry (see [prototype video](https://www.youtube.com/watch?v=igm3s1NzNSo&t=8s)). This first PR should cover the creation and editing of the spline.

## Description
(outdate, should make a new one)
For a full overview see this video which is specifically recorded for this PR: 
https://youtu.be/auLS9RObIXY
Demo project: https://github.com/Aggror/SplineTools/tree/Spline-PR-1-demo

A spline entity exists out of SplineNodes (also entities) which are its children. These can be created entirely by code if so desired. To make them slightly more user friendly, these objects are utilised in components so that an editor user can view them inside the editor.

**Some details**
- A spline exists out out nodes. 1 node that connects to another is called a (cubic bezier) curve
- An entity contains the SplineComponent
- The children of this entity should contain the SplineNodeComponent
- The SplineComponent holds a reference to the children SplineNodeComponents
- There need to be at least 2 SplineNodeComponents (only 1 per entity allowed) before a spline is calculated
- The SplineComponent holds an internal SplineRenderer which generates spline visualisation geometry when the spline is updated. Previously generated meshes are discarded upon updating spline.

**Spline traverser**
An entity with a reference to the spline, can traverse the spline. A basic Spline traverser component has been added that support moving forward and backwards along the spline. 

## Types of changes

## Checklist


markusstephanides:
Is this still being worked on?

Aggror:
> Is this still being worked on?

I had to put in in the fridge in order to complete the tutorials project. Since then I havent had the time yet to work further on this.

Aggror:
Little update on this:

- Decided re-add the spline traverser component, since that is 1 of the common use cases to have a spline.
- A spline traverser can move forwards or backwards along the spline 
  - no rotation support
  - Physical based movement not yet supported
  
Some open bugs:
- opening a scene for the first time, doesn't render the spline trail. Updating the spline or 1 of its nodes or any of the settings, updates the spline.
- duplicating a spline or creating a new spline, breaks older splines in the scene. Even upon reloading the scene. Only the latest spline component added works for some reason.

https://user-images.githubusercontent.com/3499539/180619473-5b3f6bb0-8ea6-4662-a957-6ff04fecfdbe.mp4


Stride Fork: https://github.com/Aggror/Stride/tree/SplinePR
Test project: https://github.com/Aggror/SplineTools/tree/Spline-PR-1-demo

sources/engine/Stride.Engine/Engine/Splines/Processors/SplineTransformProcessor.cs
```diff
@@ -0,0 +1,117 @@
...
+        {
+            public SplineViewHierarchyTransformOperation TransformOperation;
+        }
+
+        public override void Draw(RenderContext context)
+        {
+            if (!SplineComponent.Spline.Dirty)
+            {
+                return;
+            }
```
Eideren, line 66: Hey, so the bug with only the last spline being rendered is because you only process the last one added to the `SplineTransformProcessor`, see lines attached.

At line 45 you keep a reference to the latest component added to the processor and at 63 you only consider that one for drawing.
A processor should process every component sent to it through the `OnEntityComponentX` pair of function. What I did to fix your issue is have add a `HashSet<SplineComponent> _components = new();` to the class, add and remove to and from that collection inside of the `OnEntityComponentX` pair of function and change the first couple of lines of draw to the following:
```cs
public override void Draw(RenderContext context)
{
    foreach (var SplineComponent in _components)
    {
        if (!SplineComponent.Spline.Dirty)
```
As for the issue with initial setup not drawing the spline, you should always force draw those that were just added to the processor.

Also you might want to be careful with how `Spline.Dirty` is used, if users read and write to it for whatever reason the transform processor won't draw/update as expected. You could fix that by adding an event to splines that the processor listens to, the processor could then push splines that fired those events to a collection for them to be redrawn next `Draw()`. It even has an added performance bonus since you won't need to iterate through all splines to check for dirty, just go through that collection of dirty splines and empty it after  you're finished updating/drawing them.

Aggror: > 
@Eideren  Thanks for your help and feedback.

It is so incredibly obvious now that you pointed it out. Using a hashset to keep track of all components that need updating is the way to go. I am also adding components to the hashset when they are dirty. Using an event for this is so much cleaner. I will do the same thing for the traverser component.

Aggror:
Alright another step closer to completion. SplineTraverserComponent references a SplineTraverser object. Traverser transform processor now deals with multiple traversers properly IN the editor. Running the game is not working yet.

Aggror:
Did some more testing with test project.

- I have tested the spline. spline nodes and spline traverser in the editor in a lot of ways:
  - Removing entities or removing spline node components of existing splines. 
  - Removing spline nodes from the Spline component reference
  - Doing al of this while there are 2 spline traverser going forward and backwards on the spline

Some open bugs:

- Something I still can't lay my finger on: opening a scene for the first time, doesn't render the spline trail. I can see that the spline is calculated and that the spline renderer is there. It just doesn't update the scene view.
   -  Updating the spline position or any of the splinecomponent/splinenodecomponent settings does update the spline.

Eideren:
The traverser has issues with suttery rotation and getting stuck at low frame rates:

[stutter.webm](https://user-images.githubusercontent.com/5742236/188520196-0a25b5e2-6ba8-4fec-bd6d-1da4de83a918.webm)

I understand that might be related to your comment over on the speed property:
> The speed at which the traverser moves over the spline. Use a negative value, to go in to the opposite direction
> Note: Using a high value, can cause jitters. 
> With a higher speed value, it is recommended to reduced the amount of spline points or segments
Could you elaborate on what the issue is exactly ? Perhaps I could find a solution for you

Aggror:
> Could you elaborate on what the issue is exactly ? Perhaps I could find a solution for you
With a high speed/velocity, a traverser can overshoot its target. If the traverser is not within a threshold range of the target position, the traverser continues with its current velocity.  I am not sure if a low FPS causes the same issue.
The check for the next node is just a simple distance check. 

` var distance = Vector3.Distance(entity.Transform.WorldMatrix.TranslationVector, targetBezierPoint.Position);

  if (distance < thresholdDistance)
  {
      SetNextTarget();`

In most scenarios, this code works fine, but with high amount of nodes or a high velocity, another check needs to be performed: If the added velocity overshoots the target position. Thus far I didn't find it worth it since it is a matter of tweaking the velocity.

Aggror:
Did not manage to get the jitter to go away when the game window is not selected or when fps is low. I also have some improvements for the spline renderer although that can also be a different PR when spline meshes are support. 
I am also interested in whether my code is architecturally sound (does that make sense?).

sources/engine/Stride.Engine/Engine/Splines/Models/BoundingBoxMesh.cs
```diff
@@ -0,0 +1,70 @@
...
+using Stride.Core.Mathematics;
+using Stride.Graphics;
+using Stride.Rendering;
+using Buffer = Stride.Graphics.Buffer;
+
+namespace Stride.Engine.Splines.Models
+{
+    public class BoundingBoxMesh
+    {
+        public MeshDraw MeshDraw;
```
manio143, line 13: Consider making this mesh private so that it can't be overwritten from outside this class and exposing it through a get-only property.

sources/engine/Stride.Engine/Engine/Splines/Models/BoundingBoxMesh.cs
```diff
@@ -0,0 +1,70 @@
...
+            }
+
+            // Sides
+            for (var i = 0; i < 4; i++)
+            {
+                indices[indexOffset++] = i;
+                indices[indexOffset++] = i + 4;
+            }
+
+            vertexBuffer = Buffer.Vertex.New(graphicsDevice, vertices);
```
manio143, line 60: Should you implement `IDisposable` on this class for those GPU buffers?

sources/engine/Stride.Engine/Engine/Splines/Models/Spline.cs
```diff
@@ -0,0 +1,291 @@
+//// Copyright (c) Stride contributors (https://Stride.com)
+//// Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using System.Collections.Generic;
+using Stride.Core;
+using Stride.Core.Mathematics;
+
+namespace Stride.Engine.Splines.Models
+{
+    [DataContract]
```
manio143, line 10: Why is this class serializable? In the component the property with this class is ignored.

sources/engine/Stride.Engine/Engine/Splines/Models/SplineNode.cs
```diff
@@ -0,0 +1,293 @@
...
+                bezierPointCount = segments + 1;
+                baseBezierPointCount = bezierPointCount > baseBezierPointCount ? baseBezierPointCount + 10 : baseBezierPointCount;
+
+                InvokeOnDirty();
+            }
+        }
+        #endregion
+
+        #region Out
+        private Vector3 tangentOut { get; set; }
```
manio143, line 66: These private properties should be fields.

sources/engine/Stride.Engine/Engine/Splines/Processors/SplineTransformProcessor.cs
```diff
@@ -0,0 +1,122 @@
...
+            component.Spline.OnSplineDirty += () => splineComponentsToUpdate.Add(component);
+
+            splineComponentsToUpdate.Add(component);
+
+            entity.Transform.PostOperations.Add(data.TransformOperation);
+        }
+
+        protected override void OnEntityComponentRemoved(Entity entity, SplineComponent component, SplineTransformationInfo data)
+        {
+            component.Spline.OnSplineDirty -= () => splineComponentsToUpdate.Add(component);
```
manio143, line 52: Note that this won't work, because you're trying to remove a lambda function which is a different object from the one in the OnAdding method. I would suggest adding an `Update()` method to the `SplineTransformationInfo` which would perform the action within lambda. But passing `data.Update` as a delegate should be reference preserving and properly deregister the event.

sources/engine/Stride.Engine/Stride.Engine.csproj
```diff
@@ -32,6 +32,9 @@
     <ProjectReference Include="..\Stride.Rendering\Stride.Rendering.csproj" />
     <ProjectReference Include="..\Stride.VirtualReality\Stride.VirtualReality.csproj" />
     <ProjectReference Include="..\Stride.Shaders.Compiler\Stride.Shaders.Compiler.csproj" />
+  </ItemGroup>
+   <ItemGroup>
+    <Folder Include="Rendering\Spline\" />
```
manio143, line 37: Empty folder included

manio143 (commented):
Hey, I took some time tonight to review parts of your PR. Good work overall!
I mostly looked at general stuff and did not dive into actual computation logic you have there.

Two general things to look at:
* Make sure all new files have the copyright notice and that it includes .NET Foundation
* Sort usings to ensure easier checking of referenced namespaces.

I hope I could provide some valuable comments. There's quite a few, but let's make sure they're not discouraging :D I do code review every week and may have a somewhat dry style of delivery.

manio143:
On your question on being architecturally sound - you've placed most of your update logic in the models rather than having them just be containers and placing logic in the component processors. I think for a feature of this scale it's not a bad idea, but would ensure there's decent logical separation between what is user facing and what is internal state, which would be normally placed in a class in the processor (such that ComponentData[component] can point towards that internal data).

Aggror:
> Hey, I took some time tonight to review parts of your PR. Good work overall! I mostly looked at general stuff and did not dive into actual computation logic you have there.
> 
> Two general things to look at:
> 
>     * Make sure all new files have the copyright notice and that it includes .NET Foundation
> 
>     * Sort usings to ensure easier checking of referenced namespaces.
> 
> 
> I hope I could provide some valuable comments. There's quite a few, but let's make sure they're not discouraging :D I do code review every week and may have a somewhat dry style of delivery.

Thanks @manio143 for the thorough review. It is good that you don't back, otherwise I don't learn. Will get back to improving things soon.

Aggror:
> On your question on being architecturally sound - you've placed most of your update logic in the models rather than having them just be containers and placing logic in the component processors. I think for a feature of this scale it's not a bad idea, but would ensure there's decent logical separation between what is user facing and what is internal state, which would be normally placed in a class in the processor (such that ComponentData[component] can point towards that internal data).

I had feedback from several users, that they don't want to user components. They just want to be able to do something simple like:

`var spline = new Spline();
spline.PointA = pointA;
spline.PointB = pointB;
spline.UpdateSpline();
`
The UpdateSpline could be move to the SplineProcessor, but that would require the user to make a SplineProcessor. Is that a better design principle?

`var spline = new Spline();
spline.PointA = pointA;
spline.PointB = pointB;

var splineProccesor = new SplineProccesor ();
splineProccesor.BuildSpline(spline);
`

manio143:
> I had feedback from several users, that not all of them want to use use components (which are now tied to processors).

I think we need to understand the different parts of how splines would be used.
So for example you can isolate maths parts completely from anything engine related and allow the user to work with those object even if they're not using Stride systems.
Then for things which interact with other parts of Stride you would put those in a component-processor style.
The component can just hold a spline object and expose it's public properties inline in the editor.

Aggror:
> > I had feedback from several users, that not all of them want to use use components (which are now tied to processors).
> 
> I think we need to understand the different parts of how splines would be used. So for example you can isolate maths parts completely from anything engine related and allow the user to work with those object even if they're not using Stride systems. Then for things which interact with other parts of Stride you would put those in a component-processor style. The component can just hold a spline object and expose it's public properties inline in the editor.

Thanks for replying. So that would bring me to a spline builder/service/generator.  Spline and SplineNode will contain nothing but data, but the service contains the logic to tie these 2 together.

sources/engine/Stride.Engine/Engine/Splines/Models/BoundingBoxMesh.cs
Aggror: I dont see this happening for the other Meshclasses in the Stride engine. Or is it missing there too?

sources/editor/Stride.Assets.Presentation/AssetEditors/Gizmos/Spline/SplineNodeGizmo.cs
```diff
@@ -0,0 +1,102 @@
...
+        private Vector3 previousOutTangent;
+        private Vector3 previousNodePosition;
+
+        public SplineNodeGizmo(EntityComponent component) : base(component)
+        {
+        }
+
+        protected override Entity Create()
+        {
+            inMaterial = GizmoUniformColorMaterial.Create(GraphicsDevice, Color.LightYellow);
```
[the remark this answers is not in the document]
Aggror: Good comment. I reused some of you code here. Thanks

sources/engine/Stride.Engine/Engine/Splines/Models/BoundingBoxMesh.cs
manio143: It may be missing. Not having an explicit using shouldn't cause memory leaks, but can delay freeing up native resources, rather than as soon as they're not needed.

sources/engine/Stride.Engine/Engine/Splines/Processors/SplineTransformProcessor.cs
Aggror: I you have the time, I could use your help here. I tried moving the action code to a update method and starting playing around with passing all types of delegates, but none of these seem to trigger the removal of the actual component.

manio143: Hmm, the main thing is that when you're removing the delegate it will look at two things - object instance and pointer to the method. For static methods object instance is null.
When you create a lambda which captures an object you're allocating a new closure each time you see the lambda in code. For static lambdas these are hoisted to a static method and the removal from an event would actually work.

So how do you go about fixing it? The simplest thing you can do is cache the created closure in the context for your component processor.

```csharp
// OnComponentAdding
Action onSplineDirty = () => update...;
data.OnSplineDirtyCallback = onSplineDirty;
component.Spline.OnSplineDirty += onSplineDirty;

// OnComponentRemoved
component.Spline.OnSplineDirty -= data.OnSplineDirtyCallback;
```

manio143: What I was suggesting earlier was to encapsulate the logic in `data.Update()` and avoid using a lambda expression entirely.

sources/engine/Stride.Engine/Engine/Splines/Processors/SplineProcessor.cs
```diff
@@ -0,0 +1,152 @@
...
+            entity.Transform.PostOperations.Add(data.TransformOperation);
+        }
+
+        protected override void OnEntityComponentRemoved(Entity entity, SplineComponent component, SplineTransformationInfo data)
+        {
+            //When a spline component is removed, we still need to destroy the Spline renderer
+            DestroySplineRenderer(null, entity);
+            component.Spline.OnSplineDirty -= data.OnSplineDirtyAction;
+            entity.Transform.PostOperations.Remove(data.TransformOperation);
+        }
```
xen2, line 59: Probably better to add `splineComponentsToUpdate.Remove(component)` in case it gets added and removed without a frame running in between (it can happen if components are manually removed after a load, etc.)

sources/engine/Stride.Engine/Engine/Splines/Processors/SplineTraverserProcessor.cs
```diff
@@ -0,0 +1,258 @@
...
+            component.SplineTraverser.Entity = entity;
+
+            entity.Transform.PostOperations.Add(data.TransformOperation);
+        }
+
+        protected override void OnEntityComponentRemoved(Entity entity, SplineTraverserComponent component, SplineTraverserTransformationInfo data)
+        {
+            component.SplineComponent.Spline.OnSplineDirty -= data.OnSplineDirtyAction;
+            entity.Transform.PostOperations.Remove(data.TransformOperation);
+        }
```
xen2, line 59: Missing `splineTraverserComponents.Remove()`

Eideren:
One thing to note is that Quadratic Bezier curves are significantly better for games as they don't require iterative solutions for common utility functions like finding the length of the segment, bounding box, closest point on the spline, direction, relative speed, etc.
Meaning that, as opposed to iterative methods, those functions are precise, and the cost to call them is always constant.
I can provide an implementation if you're interested.
They are slightly harder to fiddle around with though given that you have a single handle, but this is more of a tooling issue than anything since you can split your cubic spline into two quadratics.

xen2:

PR #1288, opened 2022-01-21
Switched the default build verbosity to normal

Earthmark:
# PR Details

Switched the default build verbosity to normal to better align with restore.

## Description

Either way it's somewhat opaque, but build processes should have more information this way.

This is a follow up from https://github.com/stride3d/stride/pull/1286 and https://github.com/stride3d/stride/pull/1283 .

This change doesn't really effect anything, and it might be making things more verbose, but it does align both steps to use the same command and verbosity. 

It's a bit confusing what the command is doing outside of just calling msbuild with some arguments that could be build file provided, but if it works it works.

## Related Issue

https://github.com/stride3d/stride/pull/1286 and https://github.com/stride3d/stride/pull/1283 both relate.

## Motivation and Context

This started with a change to make the build steps a bit clearer, because I misinterpreted then and figure someone else might have the same issue.

## Types of changes

## Checklist

### DIFF
--- a/build/compile.bat
+++ b/build/compile.bat
@@ -5,7 +5,7 @@ setlocal
 set STARTTIME=%TIME%
 set __SkipTestBuild=true
 set __BuildType=Debug
-set __BuildVerbosity=m
+set __BuildVerbosity=n
 set __BuildDoc=0
 set __ContinueOnError=false
 set __SelectedProject=Stride.sln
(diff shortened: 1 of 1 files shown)


Eideren:
> It's a bit confusing what the command is doing outside of just calling msbuild with some arguments that could be build file provided, but if it works it works.
If you do have something better in mind that's as easy to remember/type and provides the same functionality this one does feel free to share.

Thanks again !

Earthmark:
If the goal is easy to invoke, then I agree with having compile.bat, I was not sure what the file was trying to solve.

It felt to me like it's trying to act as a file that can be used instead of invoking msbuild, so in the future the implementation can pivot (like changing the sln name). But that kinda doesn't work as the command just before calls out the sln... perhaps the compile.bat could benefit from a restore argument, where it restores then builds?

As it stands the directions don't call restore on the android or ios builds, even though compile.bat can invoke the two. Perhaps adding restore functionality to compile.bat may make things more flexible build system wise then? If that's something that would be useful I can poke at that, shifting the directions to be:
1. Call `compile.bat /restore`
2. Call `compile.bat`

Personally I'm still trying to understand the build system, how it compiles some c++ in here and all that, I think I can ponder this

PR #1293, opened 2022-01-23
Default SceneInstance and Scene added for code-only approach

### LINKED ISSUE #1290: Default SceneInstance and Scene
We have been discussing here https://github.com/stride3d/stride/discussions/1253 code-only approach and we are looking for options to update engine if possible to simplify certain defaults being created without Stride Game Studio.

We are going to add else branch so we can ensure that Scene is never null here 

https://github.com/stride3d/stride/blob/b9b2dbb380f39ea02dab3c65cc0ae1b68e7c2f51/sources/engine/Stride.Engine/Engine/SceneSystem.cs#L115-L121

```c# 
else
{
   SceneInstance ??= new SceneInstance(Services) { RootScene = new Scene() };
}
```

We discussed an alternative to create a default Scene in the extension also described in the discussion above but decided rather to update engine instead in this case.

**Additional context**
Empty scene is created when using Stride Game Studio so if the code-only approach is used there should also be a default empty Scene created.

VaclavElias:
# PR Details

Default SceneInstance and empty Scene is created programmatically if the game is run without the default assets (MainScene.sdscene) created by Game Studio. 

## Description

Discussed here https://github.com/stride3d/stride/discussions/1253#discussioncomment-1979625

else branch added

```c#
else
{
    SceneInstance ??= new SceneInstance(Services) { RootScene = new Scene() };
}
```

## Related Issue

More details in the issue below.

Resolves #1290 

## Motivation and Context

We are working on code-only approach and to simplify this approach so a game can be run without a too much boilerplate. Default scene can be always updated/replaced as needed through the code. 

This will help to run a game with a very few lines of the code.

**This change shouldn't affect any existing approach through the Game Studio Editor.**

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Engine/Engine/SceneSystem.cs
+++ b/sources/engine/Stride.Engine/Engine/SceneSystem.cs
@@ -119,6 +119,10 @@ namespace Stride.Engine
                 else
                     SceneInstance = new SceneInstance(Services, content.Load<Scene>(InitialSceneUrl));
             }
+            else
+            {
+                SceneInstance ??= new SceneInstance(Services) { RootScene = new Scene() };
+            }
 
             if (InitialGraphicsCompositorUrl != null && content.Exists(InitialGraphicsCompositorUrl))
             {


Eideren:
Thanks !

VaclavElias:
My pleasure! 👍😀

PR #1298, opened 2022-01-24
[Graphics] Fix resource memory leak

Eideren:
# PR Details
It's unlikely that resources were ever manually released given that objects that hold and uses those buffers do not implement IDisposable, they are expected to be GC'ed not explicitly 'destroyed'.
The device's resources are now weakly referenced and native stuff is cleaned up when resource are GC'ed, through their destructor.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Graphics/Direct3D/GraphicsResourceBase.Direct3D.cs
+++ b/sources/engine/Stride.Graphics/Direct3D/GraphicsResourceBase.Direct3D.cs
@@ -59,7 +59,7 @@ namespace Stride.Graphics
         {
             Destroyed?.Invoke(this, EventArgs.Empty);
 
-            ReleaseComObject(ref nativeDeviceChild);
+            TryReleaseComObject(ref nativeDeviceChild);
             NativeResource = null;
         }
 
@@ -107,6 +107,21 @@ namespace Stride.Graphics
                 Debug.Assert(refCountResult >= 0);
                 comObject = null;
             }
+        }
+
+        /// <summary> Returns true if the object was previously allocated and has been released </summary>
+        internal static bool TryReleaseComObject<T>(ref T comObject) where T : CppObject, IUnknown
+        {
+            if(comObject != null && comObject.NativePointer != IntPtr.Zero)
+            {
+                var refCountResult = comObject.Release();
+                Debug.Assert(refCountResult >= 0);
+                comObject = null;
+                return true;
+            }
+            
+            comObject = null;
+            return false;
         }
     }
 }
--- a/sources/engine/Stride.Graphics/GraphicsResourceBase.cs
+++ b/sources/engine/Stride.Graphics/GraphicsResourceBase.cs
@@ -10,6 +10,8 @@ namespace Stride.Graphics
         internal GraphicsResourceLifetimeState LifetimeState;
         public Action<GraphicsResourceBase> Reload;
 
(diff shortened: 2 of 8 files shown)


sources/engine/Stride.Graphics/Direct3D/GraphicsResourceBase.Direct3D.cs
```diff
@@ -108,6 +108,21 @@ internal static SharpDX.Direct3D11.CpuAccessFlags GetCpuAccessFlagsFromUsage(Gra
...
+            {
+                var refCountResult = comObject.Release();
+                Debug.Assert(refCountResult >= 0);
+                comObject = null;
+                return true;
+            }
+            
+            comObject = null;
+            return false;
+        }
```
Eideren, line 125: Added this new signature instead of changing the one above, the comment on that one is cryptic, I don't have the time or patience to see if I break anything by removing it so I'll leave it in.

sources/engine/Stride.Graphics/GraphicsResourceBase.cs
```diff
@@ -90,6 +98,7 @@ protected internal virtual void OnResume()
         /// <inheritdoc/>
         protected override void Destroy()
         {
+            GC.SuppressFinalize(this);
```
manio143, line 101: Is this ok to call `SuppressFinalize` in here, when called from the finalizer?

Eideren: Yes, I tested it a bunch and [microsoft's docs](https://docs.microsoft.com/en-us/dotnet/api/system.gc.suppressfinalize?view=net-6.0) report that:
> If obj does not have a finalizer or the GC has already signaled the finalizer thread to run the finalizer, the call to the SuppressFinalize method has no effect.

Ethereal77:
Is this still relevant? What was the problem here?

Also, this may need updating, as `CppObject` is iirc a `SharpDX` type.

Do note that some of the graphics types that seem to require manual disposal do not have `Dispose()` themselves because they are reference-counted through `ComponentBase` and its `Destroy()` mechanism (which also takes into account finalization by the GC). It was a bit of a headache to understand this because I tent to like more explicit lifetime logic, but Stride used a lot the automatic ref-counted components.

In many of the places I ported to Silk.NET where a COM pointer should be released, I used the `SafeRelease` extension method, that checks for `null` pointer already. It does not check if the reference count is 0 because iirc that part is managed by the type itself being a `ComponentBase` and may be in use (referenced) still by some part of the renderer. Like I said above, a bit of a headache.

I can review this if you need it, but better in a few weeks, as I'm currently pretty busy

Eideren:
Like I mentioned over on discord, this was a very crude safety net put in place to catch undisposed resources. Given how old this PR is you'll have to forgive the lack of detail, but from what I remember at the time, ref to those resources were being held in memory solely through the affected list and never actually disposed of.

Looking back on it, this PR definitely isn't the right way to go about it, but now that we're here, let's clarify our intent as to how we expect users to dispose of buffers and other graphics resources. Whether the current way we do things make sense, and what kind of fallback or tool should we put in place for them to catch cases were they do not dispose of those resources appropriately.

Best to move over back to discord for this I feel.

xen2:
I will close this one.
Let's reopen new/existing github issue or discuss on discord as necessary for proper disposing strategy.

PR #1300, opened 2022-01-25
[Graphics] Fix and improve Resource.Reload

Eideren:
# PR Details
Reloading buffers wasn't ever called in editor or in engine I don't think but fixed it just in case someone ends up using it.
Also reduced the amount of unnecessary bloat allocated through some of the more common reload delegate capture, it's not that large but given that those resource are common objects it can quickly add up for a feature that's almost never used.

## Types of changes

## Checklist


tebjan:
Good find with the buffers! LGTM.

PR #1301, opened 2022-01-25
[Math] Optimize Matrix mult and invert

Eideren:
# PR Details
Passing operations to numerics where appropriate to benefit from SIMD, only did it for the two most commonly used function right now, we could look into writing it explicitly for column major later.
Also small cleanup for `TransformProcessor`.

## Types of changes

## Checklist

### DIFF
--- a/sources/core/Stride.Core.Mathematics/Matrix.cs
+++ b/sources/core/Stride.Core.Mathematics/Matrix.cs
@@ -997,16 +997,10 @@ namespace Stride.Core.Mathematics
         /// <param name="result">The product of the two matrices.</param>
         public static void Multiply(ref Matrix left, ref Matrix right, out Matrix result)
         {
-            unsafe
-            {
-                fixed (Matrix* outPtr = &result)
-                {
-                    ref MatrixDotnet l = ref UnsafeRefAsDotNet(in left);
-                    ref MatrixDotnet r = ref UnsafeRefAsDotNet(in right);
-                    *(MatrixDotnet*)outPtr = LayoutIsRowMajor ? l * r : r * l;
-                    return; // Explicit return until JIT optimization is fixed
-                }
-            }
+            ref MatrixDotnet l = ref UnsafeRefAsDotNet(in left);
+            ref MatrixDotnet r = ref UnsafeRefAsDotNet(in right);
+            Unsafe.SkipInit(out result);
+            UnsafeRefAsDotNet(in result) = LayoutIsRowMajor ? l * r : r * l;
         }
 
         /// <summary>
@@ -1019,16 +1013,10 @@ namespace Stride.Core.Mathematics
         /// <param name="result">The product of the two matrices.</param>
         public static void MultiplyIn(in Matrix left, in Matrix right, out Matrix result)
         {
-            unsafe
-            {
-                fixed (Matrix* outPtr = &result)
-                {
-                    ref MatrixDotnet l = ref UnsafeRefAsDotNet(in left);
-                    ref MatrixDotnet r = ref UnsafeRefAsDotNet(in right);
-                    *(MatrixDotnet*)outPtr = LayoutIsRowMajor ? l * r : r * l;
-                    return; // Explicit return until JIT optimization is fixed
-                }
-            }
+            ref MatrixDotnet l = ref UnsafeRefAsDotNet(in left);
+            ref MatrixDotnet r = ref UnsafeRefAsDotNet(in right);
+            Unsafe.SkipInit(out result);
+            UnsafeRefAsDotNet(in result) = LayoutIsRowMajor ? l * r : r * l;
         }
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


sources/core/Stride.Core.Mathematics/Matrix.cs
```diff
@@ -1014,39 +1017,42 @@ public static void Multiply(ref Matrix left, ref Matrix right, out Matrix result
...
-        /// <returns>The product of the two matrices.</returns>
-        public static Matrix Multiply(Matrix left, Matrix right)
+        /// <param name="result">The product of the two matrices.</param>
+        [Obsolete($"Use {nameof(Multiply)} instead")]
+        public static void MultiplyRef(ref Matrix left, ref Matrix right, ref Matrix result)
         {
-            Matrix result;
             Multiply(ref left, ref right, out result);
-            return result;
         }
```
Eideren, line 1046: Small diff fail here, `Matrix Multiply(Matrix left, Matrix right)` hasn't been removed and `MultiplyRef` isn't new, the former is now right below and the later was above but has been replaced.

tebjan:
this looks very interesting, my only concern is that the conversion overhead is making the improvement negligible. can you see in a benchmark that this is actually faster?

Eideren:
With a cpu supporting the SSE set multiply is ~40% faster and invert is ~33% faster.
Overhead has been mostly optimized out, stripping that bloat out leads to ~7% additional improvement but I'm not sure that's worth it considering dot net might improve support while we aren't looking.
Here's a [sharplab](https://sharplab.io/#v2:C4LghgzgtgPgAgBgAQFkzAE4EsAeARAe2ADsBTYJAXiTgEYA6AOQFcpTsBjCetTXAFhz8A3AFgAUIhoMASs2LAsbegGECUAA5YANuwDK7AG5YOpCGPESImZhwq9sOCQG8JSdzQDMSAGbaC6Ki0tBYeXr7+gSgATCFuHnDefgH2nnHiYYkRKaj86ZlJkfa00aEJhTkxpfHuWclRntUZ5dlR/E0FrcWeZbUVUdE9NeH1qUPNfV25452jQSLDdUWo0QsTI8sonmuzm/w7HsOL3hwExNZIAEYEBNpIADJgAJ4EzMAAkhAyBADuaABWBAwVF8YG0EFIvSQi1oADYkBhSD5UOhHIQSOQkABVc5gHykGRIgCCEHRjHIAAosMQUXwcEgoABKKgAPgRSOxuPx9BJAB4HLgADS0tFEMjAFkUxHInEQPGkHlfJFUmlMxlQujw6Ui3CcuX4wk+ABiGHUZMp1J1+DFmKZrPZMq5Cr5AutGOAwtdku1svlisNKoZjPVR3Wx2k8Lg/FQzG0ig02ieAHltAATKUc11IXQ+D0Oq0IrAAcwAFnnXvZUbrERBY8BGcNXOswjW6zxgiCKTngO3aEgAFSF0s9lDBZkAaiQXaRI5KA6HZZ4sQnU+77e8g+ww54aRX09z7ejm+Li5QeRDzY8rbjS771H3I9i863p7HSEnD6X0WfJ8ftD3a4xBuC4jru76rjOS5HiBPDnlCLZmG2Wx3hBB7IT+26jv+4Gfls37Hphy44YBnjAS+oHYR+JHQeRsH/vBV6ITeZ4obheQYa+lGoSO7QcX+AGQWeZG/juXFsTRIksReYQIbWzFzvegEdgRr7RAJaFzipj5qcRgm0MJmGNOps4SZh7TSTJ7jXtpnaAU+WntjpVGCdE+EwVUxlLgZp5GbpaGrHxsFqQxVlMaB36KYJaSBaOTncTubm0R5fmgd54WeVspmnuZIXsnJPERfFLExSUGW8Q5yXOWh/BpQlZVZQVFmWdZ662XpfYVfpGWae5gzdbVWyeN1DWwUNuUtUBbX+R17ldSlX4xX182TRVpEZQFFU1U1MkTaRU0USVQ3LY0i1HVVqUxWtx0jUJ22yUhNX7XRh1lYlklLedo2XWdRX7DFW3jWFh5PVhJX8N1b2EeDy36Zd0OfaON37HdjH5VBIP2bN8NFa5i3Y7hgxw+tSPg4DaOZSD0WdfjJGQ6eqwZXtq001FJMo6F5N/ZF1UzUltAs9VdOPgLPEDds9X/aTwwAL6hp0mo0NGKB1lgCZPO8xCBlm3bCpaWbkcKFYFtZDbrE2lmhcirrouK2Ygj6TqGiS5rAIG3bs/m1s2hQwLUA7+oEsSpJEOSruWuRHu+tyegANaqxrWCu0bJu5VHgc+M7IcWjSJsgo8LxvJ83x/GAgLAgA/Hbm5ICACLztoUKy2G6xZArWYEBo7DoECA5a1W9I60gev9yBpthObFval77p237HJp07wfAKHbszh7U/9zbmK+/mC9By7gYR7lcAAOx6vKhommaWeu/nFZF78AI95XdzV7XwKDto21N2E4YK1GSANaGHYK7DedIkCGDBMwUghs3jGzCmPDwE9LJp3oLHeOxBE4UmTgg3K09xT0CASAike8M5LxXpA7Q0DGSwIoKQzOy9KQm2/hIaWQA=) setup if you want to take a look at it.

ykafia:
Would it be safer to use [this library](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.intrinsics?view=net-6.0) ? 
It seems to work only on ARM et x86 architectures though

Eideren:
That's what numerics uses internally, so technically we already are using that. What do you mean by safer ? If you mean using them directly instead of passing through dotnet's types, sure, but as I said above I'm not sure that's worth it considering dot net might improve support while we aren't looking.

tebjan:
> multiply is ~40% faster and invert is ~33% faster.
> Overhead has been mostly optimized out, stripping that bloat out leads to ~7% additional improvement

that sounds great! i am ok with merging this, if it will cause any problems, we can still revert it.

Issue #1303, opened 2022-01-26
InputManager and  DownButtons return incorrect collection


dawnmichal:
**Release Type**: Official Release

**Version**: 4.1.0.1459-beta

**Platform(s)**: Windows

**Describe the bug**
When pressing two buttons and then only one is released Input.DownButtons is empty.

**To Reproduce**
1. Nothing is pressed = Input.DownButtons is empty
2. pressed RMB = Input.DownButtons contains Right
3. pressed LMB and still holding RMB = Input.DownButtons contains Right and Left. So far so good, but now...
4. release only LMB = Input.DownButtons is empty?

**Expected behavior**
Input.DownButtons should contain Right mouse button at step 4.

Eideren:
Could repro, see #1304

dawnmichal:
Just FYI study case. RMB is used for getting ready (rise) weapon from lower pose. LMB is for fire. So because of this bug character goes to lower state after fire though player still holding RMB.

PR #1304, opened 2022-01-27
[Input] Fix mouse release for Winforms

Eideren:
# PR Details
For some reason Winforms fires the `MouseCaptureChanged` when holding both and releasing just one of the two mouse buttons, see #1303
[Here's what the docs says about that event](https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.control.mousecapturechanged?view=windowsdesktop-6.0):
> Now, click and hold the left mouse button on the Button control. While still clicking the mouse, press ALT+TAB to switch to another program. Notice that the MouseCaptureChanged event is raised enabling you to potentially handle this scenario.

It makes sense then that the engine removes any mouse button held when that event occurs but afaict it shouldn't occur in this case. I swapped it to `LostFocus` instead anyway since it makes more sense.

## Related Issue
#1303

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Input/Windows/MouseWinforms.cs
+++ b/sources/engine/Stride.Input/Windows/MouseWinforms.cs
@@ -33,9 +33,9 @@ namespace Stride.Input
             uiControl.MouseDown += OnMouseDown;
             uiControl.MouseUp += OnMouseUp;
             uiControl.MouseWheel += OnMouseWheelEvent;
-            uiControl.MouseCaptureChanged += OnLostMouseCapture;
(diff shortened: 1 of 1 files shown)


Aggror:
Nice work Eideren!

VaclavElias:
Very quick response 🙏.

tebjan:
This LGTM. Is this issue also present for SDL?

Eideren:
I'm pretty sure I tested it at the time and sdl didn't have the same issue, most likely because mouse input is ignored by default when out of focus.

Issue #1312, opened 2022-02-01
Add .vsconfig


TheKeyblader:
**Is your feature request related to a problem? Please describe.**
Prerequisites for stride can be hard to found in vs installer
and even worst if installer is not in english

**Describe the solution you'd like**
add .vsconfig to build folder.
With .vsconfig, Visual Studio will automatically detect which components are missing and prompt you to install them.

**Additional context**
[Configure Visual Studio across your organization with .vsconfig](https://devblogs.microsoft.com/setup/configure-visual-studio-across-your-organization-with-vsconfig/)

PR #1314, opened 2022-02-01
[Build] Add .vsconfig to check missing requirements in VS

Eideren:
# PR Details
See #1312 , didn't/can't test but should be working

## Types of changes

## Checklist


tebjan:
The list of workloads looks a bit long to me... are you sure that we need all of these?

Eideren:
The engine might only need a subset of these but this list is what VS installer generates for the workloads listed in our readme, better safe than sorry so I kept it as is.
Just as a note, each requirement listed in the readme generates multiple workloads, `Desktop development with C++` alone for example takes around 1/3 of that list.

tebjan:
Ah I see, didn't know how that works in detail. maybe @xen2 would be better to review this.

Eideren:
I initially created it with VS tools but tools and studio configurations aren't shared, this version should work for Studio.

jrinker03:
I've tested this change with a fresh installation of VS with few of the prereqs installed and it works as expected:
image

Clicking on 'Install' opens the VS Installer with the workloads and components selected as expected:

image

After VS reopens, the solution builds completely.

Eideren:
Thanks a bunch for testing this out @jrinker03 ! I'll merge this in then since it's quite safe. If Xen ever has time to take a look later we'll update accordingly.

PR #1350, opened 2022-03-11
[Rendering] Fixes crash when changing DoF quality at runtime

tebjan:
# PR Details

Fixes crash when changing DoF quality at runtime.

## Description

Adds an `UpdateEffect()` call after setting the permutation parameter `CombineLevelsFromCoCKeys.LevelCount`. If the quality preset changes, this shader permutation key gets set, which changes the length of an array parameter. If the effect isn't updated, a subsequent `Set` on the array shader parameter throws an exception. The same is already done in the CoCMapBlur effect: https://github.com/stride3d/stride/blob/1c2969e3558228565431e267140557637033a9aa/sources/engine/Stride.Rendering/Rendering/Images/DepthOfField/CoCMapBlur.cs#L81

## Related Issue

https://github.com/vvvv/VL.Stride/issues/333

## Motivation and Context

Avoids a crash when changing the quality preset.

## Types of changes

## Checklist


Eideren:
Can you create a PR for `master` as well ? Thanks !

Issue #1361, opened 2022-03-17
Stride.Launcher - Replace deprecated System.Net.WebRequest & WebClient


jrinker03:
**Release Type**: GitHub

**Version**: main

**Platform(s)**: N/A 

**Describe the bug**
The Stride Launcher uses the recently deprecated System.Net.WebRequest to fetch various files from the web for display in the UX.   
Details:  https://docs.microsoft.com/en-us/dotnet/core/compatibility/networking/6.0/webrequest-deprecated

This results in deprecation warnings in the build output and in the VS editor when opening an affected .cs file.

Per the above article and compiler warnings, use of WebRequest should be replaced with System.Net.Http.HttpClient instead.

**To Reproduce**
1. sync to a recent commit in master and build the Stride.Launcher or open one of the affected .cs files in Visual Studio.

**Expected behavior**
No deprecation warnings.

**Screenshots**
See build log snippet below

**Log and callstacks**
From the build log:

1>H:\src\jrinker03\stride\sources\launcher\Stride.Launcher\ViewModels\DocumentationPageViewModel.cs(78,38,78,100): warning SYSLIB0014: 'WebRequest.Create(string)' is obsolete: 'WebRequest, HttpWebRequest, ServicePoint, and WebClient are obsolete. Use HttpClient instead.'

**Additional context**
None

jrinker03:
I'm going to put together a change for this soon.

jrinker03:
Launcher also uses WebClient which is also covered in the above article for deprecated classes.   Will also update the use of WebClient.

PR #1362, opened 2022-03-17
Fix for Issue #1361 - Replace deprecated WebRequest & WebClient

jrinker03:
# PR Details
1) Replaced all instances of WebRequest with HttpClient plus the appropriate Get + Read operations.

2) Replaced the use of WebClient with HttpClient plus Get + stream read.

3) Removed semi-problematic references and the SDK specifier as they caused the compiler to generate a number of warnings.   All these are not needed in .NET 6.

## Description

See details above and context below.

## Related Issue
#1361 

## Motivation and Context
The top-level motivation is to improve the developer experience when working with the code and ensure that the code is using the most current classes and practices.

This change replaces several instances of deprecated classes with current classes that handle the same functionality.  

Additionally, there were quite a number of warnings output by the compiler due to the .csproj file not being updated for .NET 6.   Updating the .csproj eliminated compiler confusion and the accompanying warning messages when selecting assemblies.  As a nit, the compiler also output a warning that <Project Sdk="Microsoft.NET.Sdk.Desktop"> could be simplified so I made that change to eliminate one more warning.

Also made some minor changes to eliminate other VS-generated suggestions in the affected files.

## Types of changes

## Checklist
There are no tests for the launcher so I have manually verified that the code results in the same experience.

### DIFF
--- a/sources/launcher/Stride.Launcher/Services/SelfUpdater.cs
+++ b/sources/launcher/Stride.Launcher/Services/SelfUpdater.cs
@@ -5,28 +5,28 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
+using System.Net.Http;
 using System.Reflection;
+using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
 using System.Windows;
-
 using Stride.Core;
 using Stride.Core.Extensions;
-using Stride.LauncherApp.Views;
 using Stride.Core.Packages;
 using Stride.Core.Presentation.Services;
 using Stride.Core.Presentation.ViewModel;
+using Stride.LauncherApp.Resources;
+using Stride.LauncherApp.Views;
 using MessageBoxButton = Stride.Core.Presentation.Services.MessageBoxButton;
 using MessageBoxImage = Stride.Core.Presentation.Services.MessageBoxImage;
-using System.Net;
-using Stride.LauncherApp.Resources;
-using System.Text.RegularExpressions;
 
 namespace Stride.LauncherApp.Services
 {
     public static class SelfUpdater
     {
         public static readonly string Version;
+        private static readonly HttpClient httpClient = new();
 
         private static SelfUpdateWindow selfUpdateWindow;
 
@@ -46,7 +46,7 @@ namespace Stride.LauncherApp.Services
                 {
                     await UpdateLauncherFiles(dispatcher, services.Get<IDialogService>(), store, CancellationToken.None);
                 }
-                catch (Exception e)
+                catch (Exception)
                 {
                     dispatcher.Invoke(() => selfUpdateWindow?.ForceClose());
                     throw;
@@ -214,10 +214,13 @@ namespace Stride.LauncherApp.Services
 
 
                 var strideInstaller = Path.Combine(Path.GetTempPath(), $"StrideSetup-{Guid.NewGuid()}.exe");
-                using (WebClient webClient = new WebClient())
-                {
-                    webClient.DownloadFile(strideInstallerUrl, strideInstaller);
-                }
+                var response = await httpClient.GetAsync(strideInstallerUrl);
+                response.EnsureSuccessStatusCode();
+
+                await using var responseStream = await response.Content.ReadAsStreamAsync();
+                await using var fileStream = File.Create(strideInstaller);
+                responseStream.Seek(0, SeekOrigin.Begin);
+                responseStream.CopyTo(fileStream);
 
                 var startInfo = new ProcessStartInfo(strideInstaller)
                 {
--- a/sources/launcher/Stride.Launcher/Services/SelfUpdater.cs
+++ b/sources/launcher/Stride.Launcher/Services/SelfUpdater.cs
@@ -214,13 +214,14 @@ namespace Stride.LauncherApp.Services
 
 
                 var strideInstaller = Path.Combine(Path.GetTempPath(), $"StrideSetup-{Guid.NewGuid()}.exe");
-                var response = await httpClient.GetAsync(strideInstallerUrl);
-                response.EnsureSuccessStatusCode();
+                using (var response = await httpClient.GetAsync(strideInstallerUrl))
+                {
+                    response.EnsureSuccessStatusCode();
 
-                await using var responseStream = await response.Content.ReadAsStreamAsync();
-                await using var fileStream = File.Create(strideInstaller);
-                responseStream.Seek(0, SeekOrigin.Begin);
-                responseStream.CopyTo(fileStream);
+                    await using var responseStream = await response.Content.ReadAsStreamAsync();
+                    await using var fileStream = File.Create(strideInstaller);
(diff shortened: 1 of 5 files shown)


sources/launcher/Stride.Launcher/ViewModels/ReleaseNotesViewModel.cs
```diff
@@ -68,26 +70,17 @@ internal ReleaseNotesViewModel([NotNull] LauncherViewModel launcher, [NotNull] s
...
-
-                        if (str != null)
-                        {
-                            using (var reader = new StreamReader(str))
-                            {
-                                releaseNotesMarkdown = reader.ReadToEnd();
-                            }
-                        }
-                    }
-                    // fetch the response Uri and update the base URL
```
jrinker03, line 89: This part seemed a bit odd since it appears that the BaseUrl isn't used again after this.   I left the functionality of the code intact, but I'd suggest that updating the BaseUrl isn't really necessary, even if there were a redirect.   I'm happy to remove this code as a part of this change or as a follow-on if reviewers agree it's not needed.

sources/launcher/Stride.Launcher/Services/SelfUpdater.cs
```diff
@@ -214,10 +214,13 @@ internal static async Task DownloadAndInstallNewVersion(IDispatcherService dispa
...
-                using (WebClient webClient = new WebClient())
-                {
-                    webClient.DownloadFile(strideInstallerUrl, strideInstaller);
-                }
+                var response = await httpClient.GetAsync(strideInstallerUrl);
+                response.EnsureSuccessStatusCode();
+
+                await using var responseStream = await response.Content.ReadAsStreamAsync();
+                await using var fileStream = File.Create(strideInstaller);
+                responseStream.Seek(0, SeekOrigin.Begin);
```
manio143, line 222: Do you need to seek here? Since no other reads happened on the `responseStream` this can be avoided (and prevent an issue if in the future the stream isn't seekable).

sources/launcher/Stride.Launcher/ViewModels/NewsPageViewModel.cs
```diff
@@ -63,17 +65,12 @@ private async Task OpenUrl()
...
-                using (var reponse = await request.GetResponseAsync())
-                {
-                    using (var str = reponse.GetResponseStream())
-                    {
-                        str?.CopyTo(rss);
-                    }
-                }
+                HttpResponseMessage response = await httpClient.GetAsync(Urls.RssFeed);
+                response.EnsureSuccessStatusCode();
+                rss = (MemoryStream)await response.Content.ReadAsStreamAsync();
```
manio143, line 73: You should use `CopyTo` here as well instead of casting the returned stream as you're not guaranteed it will always be a `MemoryStream`

sources/launcher/Stride.Launcher/ViewModels/ReleaseNotesViewModel.cs
```diff
@@ -68,26 +70,17 @@ internal ReleaseNotesViewModel([NotNull] LauncherViewModel launcher, [NotNull] s
...
         public async void FetchReleaseNotes()
         {
-            string releaseNotesMarkdown = null;
+            string releaseNotesMarkdown;
 
             try
             {
-                var request = WebRequest.Create($"{BaseUrl}{ReleaseNotesFileName}");
-                using (var response = await request.GetResponseAsync())
+                using (var response = await httpClient.GetAsync($"{BaseUrl}{ReleaseNotesFileName}"))
```
manio143, line 77: Can you check if the response is supposed to be explicitly disposed of for HttpClient - if so, we should do it in other places as well, if not, remove the using here.

sources/launcher/Stride.Launcher/ViewModels/ReleaseNotesViewModel.cs
manio143: I don't have a strong opinion on that, but if the code is mostly pointless then I have nothing against removing it.

sources/launcher/Stride.Launcher/Services/SelfUpdater.cs
jrinker03: No, we don't.  I had a moment of weakness and left it there. :)   Thanks for the clarity.

sources/launcher/Stride.Launcher/ViewModels/ReleaseNotesViewModel.cs
jrinker03: Yes, good point. Should be consistent one way or the other.

sources/launcher/Stride.Launcher/ViewModels/ReleaseNotesViewModel.cs
jrinker03: Code and set removed in second commit.

sources/launcher/Stride.Launcher/Services/SelfUpdater.cs
jrinker03: Removed in second commit.

sources/launcher/Stride.Launcher/ViewModels/NewsPageViewModel.cs
jrinker03: Updated in second commit.

sources/launcher/Stride.Launcher/ViewModels/ReleaseNotesViewModel.cs
jrinker03: It should be disposed so made it consistent.

sources/launcher/Stride.Launcher/Stride.Launcher.csproj
```diff
@@ -47,20 +47,6 @@
...
-      <HintPath>..\..\..\deps\Metrics\System.Net.Http.Formatting.dll</HintPath>
-    </Reference>
-    <Reference Include="System.Security" />
-    <Reference Include="System.Windows.Forms" />
-    <Reference Include="System.Xaml">
-      <RequiredTargetFramework>4.0</RequiredTargetFramework>
-    </Reference>
-    <Reference Include="WindowsBase" />
-    <Reference Include="PresentationCore" />
-    <Reference Include="PresentationFramework" />
```
jrinker03, line 63: This removal of explicit references to SDK assemblies should also happen in other projects targeting .NET 6 as the compiler can get confused and emit warnings of the form:
`14>C:\Program Files\Microsoft Visual Studio\2022\Enterprise\MSBuild\Current\Bin\amd64\Microsoft.Common.CurrentVersion.targets(2301,5): warning MSB3243: No way to resolve conflict between "System.Security, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" and "System.Security". Choosing "System.Security, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" arbitrarily.`

I'll open an issue for this.

manio143 (approved):

Issue #1306, opened 2022-01-29
Can't isntal Visual Studio extension from launcher


KuraiAndras:
**Release Type**: Official Release

**Version**:
- Launcher: 5.0.5
- Stride: 4.1.0.1459-beta

**Platform(s)**: Windows 11, Visual Studio 2022 17.0.5

**Describe the bug**
VSIX installer fails.

**To Reproduce**
1. Install Stride launcher
2. Click `Install Visual Studio plugin`

**Expected behavior**
VSIX installs without an error

**Screenshots**
![image](https://user-images.githubusercontent.com/19240945/151660522-d369dbbd-6f50-450c-876e-2b56f659b9b1.png)
![image](https://user-images.githubusercontent.com/19240945/151660552-85a60070-fa2f-43cb-927e-429028d5dd7b.png)

**Log and callstacks**
[dd_VSIXInstaller_20220129131228_8368.log](https://github.com/stride3d/stride/files/7963624/dd_VSIXInstaller_20220129131228_8368.log)
[dd_VSIXInstaller_20220129130651_0c78.log](https://github.com/stride3d/stride/files/7963625/dd_VSIXInstaller_20220129130651_0c78.log)

MistahJ26:
I ran into this same issue, I reinstalled VS 2019 instead and didn't get the error message. So it's an issue related to using the new version of VS.

ericwj:
Is it difficult to release the .vsix to Visual Studio Marketplace? I had issues before 4.1 released over first not being able to install the 2022 extension and second the button in the launcher only picking an old VS2017 which I didn't want to change. After 4.1 released I got stuck again with the new button disabled for non-apparent reasons. Third having it published to the marketplace will sync installation with my Visual Studio account, allow me to pick which exact version to install, allow changing the version between published and built from source. Etc

ericwj:
I think the project should have an account on the marketplace and just publish the vsix whenever Stride is released?

PR #1382, opened 2022-03-26
Add new button to launcher for VS2022 VSIX

manio143:
# PR Details

## Description

This PR adds a 3rd button to the launcher for VSIX installation - there's still TODOs that need to be figured out:

- [ ] PackageId - new id for the new package, while the previous one stays the same for VS2019
- [ ] Label formatting - Currently we take prefix of package id (to the first dot) as label for the button (Stride/Xenko)

## Related Issue

#1306

## Motivation and Context

VS2022 plugins are not backwards compatible with VS2019, therefore we need to deploy a new plugin. Support for VS2019 needs to be maintained for Stride 4.0

## Types of changes

## Checklist


tebjan:
Merged with #1411

Issue #1388, opened 2022-03-28
Tracking list for supporting VS 2019 and 2022 plugins


jrinker03:
**Summary**
Per discussion and related bugs, the Stride plugin for Visual Studio needs to have two published packages:  one for VS2019 and one for VS2022.    This issue is to track the work needed to cleanly support both versions.

As discussed in Discord, we will support the VS2019 plugin out of the 4.0 branch and the VS2022 plugin out of master.

**Work to support two plugins**
1. Update plugin code in master to support 2022 SDK - _**Done**_ #1221 
2. In the master-4.0 branch, change the VSIX package and installer exe to only target VS2019 - _**Done**_ #1393 
3. After 2 is complete, publish the updated **4.0.x** edition of Stride.VisualStudio.Package to NuGet.org
4. In the master branch, update the launcher to support both VS2022 and 2019 extensions from the appropriate NuGet packages.  It will need a new identity. #1405 
5. Publish the updated launcher.
6. Publish the **4.1.x** edition of Stride.VisualStudio.Package to NuGet.org.

jrinker03:
See also:
#1306 
#1375

jrinker03:
Closing this item as it is superseded by #1409

PR #1405, opened 2022-04-18
Support VS 2022 + 2019 Extensions in the launcher

jrinker03:
# PR Details

## Description

Since the architecture changed for extensions in 2022, the Stride extension needs to have a new identity and can only be installed on VS 2022.  VS 2019 extension support will come from the 4.0 branch and NuGet package.
Update the launcher to support both editions and remove button for old Xenko VS extension.

- This pulls in and updates the PR from @manio143 : #1382 

- Update the Launcher to only offer the VS Extension for VS2019 and VS2022.
- Limit the package editions eligible to install VS Extension.  For 2019, this is 4.0.x and for 2022 it's 4.1.x and later.
- Update VisualStudioVersions to support the two editions and clean up code a bit. 
- Remove dead code related to VS versions
- Other minor code clean up and dead code removal.
- Updated copyright headers in files touched to reflect the most recent discussion on Discord

Related issue: #1388 part 4.

Updated Extension in VS 2022:
image

Updated first-run dialog example:
image

Updated buttons on Launcher and **disabled** 2022 button since there is no published package for it yet:
image

## Types of changes

## Checklist


sources/core/Stride.Core.Design/VisualStudio/VisualStudioVersions.cs
```diff
@@ -68,102 +82,117 @@ public IDEInfo(Version version, string displayName, string installationPath, boo
...
+                vsixInstallerPath = null;
+            return vsixInstallerPath;
+        }
+
+        /// <summary>
+        /// Computes the installation path for Visual Studio Extensions for the current instance. 
+        /// This will always be the per-user location.
+        /// </summary>
+        /// <returns>The full path to the VSIX installation folder for this instance.</returns>
+        private string ComputeVsixInstallationPath()
```
jrinker03, line 123: Adding this to support detecting whether the extension is installed or not.  Should help with the 'Reinstall' case or if the user has manually removed the extension in VS.   In the latter case, the VSIX installer will throw a nasty error message so we should just detect and skip uninstall.

sources/engine/Stride.Assets/StrideConfig.cs
```diff
@@ -253,36 +253,6 @@ internal static bool IsVSComponentAvailableAnyVersion(IDictionary<Version, strin
             return false;
         }
 
-        /// <summary>
```
jrinker03, line 256: This code is no longer used in the code base and is incompatible with these changes so just removing it.

sources/launcher/Stride.Launcher/Resources/Strings.Designer.cs
```diff
@@ -19,7 +19,7 @@ namespace Stride.LauncherApp.Resources {
     // class via a tool like ResGen or Visual Studio.
     // To add or remove a member, edit your .ResX file then rerun ResGen
     // with the /str option, or rebuild your VS project.
-    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("System.Resources.Tools.StronglyTypedResourceBuilder", "16.0.0.0")]
+    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("System.Resources.Tools.StronglyTypedResourceBuilder", "17.0.0.0")]
```
jrinker03, line 22: Note that many of these changes were auto-generated in parallel with my change.  My changes were to rename the variables and strings to use 'Extension' rather than 'Plugin' to reduce confusion.  See related .resx file changes below.

Perhaps this file shouldn't be included in source control at all?

sources/launcher/Stride.Launcher/Stride.Launcher.csproj
```diff
@@ -1,10 +1,4 @@
 <Project Sdk="Microsoft.NET.Sdk">
-
```
jrinker03, line 2: Mostly removing commented or unneeded things from this .csproj.

sources/launcher/Stride.Launcher/ViewModels/PackageVersionViewModel.cs
```diff
@@ -317,7 +317,7 @@ private async Task DeleteInternal(bool displayErrorMessage)
 
         private void UpdateStatusInternal()
         {
-            CanBeDownloaded = LocalPackage == null || (LocalPackage != null && ServerPackage != null && LocalPackage.Version < ServerPackage.Version);
+            CanBeDownloaded = (LocalPackage == null && ServerPackage != null) || (LocalPackage != null && ServerPackage != null && LocalPackage.Version < ServerPackage.Version);
```
jrinker03, line 320: Somewhat ironic that the logic setting CanBeDownloaded never checked whether there was a server package to download or not.

I discovered this because there is no 4.1.x package for the VS extension.

sources/launcher/Stride.Launcher/ViewModels/VsixVersionViewModel.cs
```diff
@@ -104,8 +114,12 @@ protected override void UpdateInstallStatus()
         /// <inheritdoc/>
         protected override async Task UpdateVersionsFromStore()
         {
-            LocalPackage = await Launcher.RunLockTask(() => Store.GetLocalPackages(packageId).OrderByDescending(p => p.Version).FirstOrDefault());
-            ServerPackage = await Launcher.RunLockTask(() => Store.FindSourcePackagesById(packageId, CancellationToken.None).Result.OrderByDescending(p => p.Version).FirstOrDefault());
+            var versionRange = Store.VsixVersionToStrideRelease[this.vsixSupportedVsVersion];
```
jrinker03, line 117: See related changes to NugetStore.

sources/tools/Stride.VisualStudio.Package/source.extension.vsixmanifest
```diff
@@ -1,12 +1,12 @@
 <?xml version="1.0" encoding="utf-8"?>
 <PackageManifest Version="2.0.0" xmlns="http://schemas.microsoft.com/developer/vsx-schema/2011" xmlns:d="http://schemas.microsoft.com/developer/vsx-schema-design/2011">
   <Metadata>
-    <Identity Id="248ff1ce-dacd-4404-947a-85e999d3c3ea" Version="4.0.6" Language="en-US" Publisher="Stride" />
-    <DisplayName>Stride</DisplayName>
-    <Description>Stride VisualStudio Package</Description>
+    <Identity Id="Stride.VisualStudio.Package.2022" Version="4.1.0" Language="en-US" Publisher="Stride" />
```
jrinker03, line 4: Per https://docs.microsoft.com/en-us/visualstudio/extensibility/vsix-extension-schema-2-0-reference?view=vs-2022#metadata-element

Id - This attribute must be a unique ID for the package chosen by its author. The name should be qualified the same way CLR types are namespaced: Company.Product.Feature.Name. The Id attribute is limited to 100 characters.

sources/tools/Stride.VisualStudio.PackageInstall/Program.cs
```diff
@@ -45,7 +44,7 @@ static int Main(string[] args)
                     case "/uninstall":
                     {
                         // Note: we allow uninstall to fail (i.e. VSIX was not installed for that specific VIsual Studio version)
-                        RunVsixInstaller(visualStudioVersion.VsixInstallerPath, "/uninstall:248ff1ce-dacd-4404-947a-85e999d3c3ea");
+                        RunVsixInstaller(visualStudioVersion.VsixInstallerPath, "/uninstall:Stride.VisualStudio.Package.2022");
```
jrinker03, line 47: Corresponds to the vsixmanifest change above.

sources/core/Stride.Core.Design/VisualStudio/VisualStudioVersions.cs
jrinker03: I should mention that the work to detect/skip is not in iteration 1 of this PR.   Will include it later or as a separate PR.

sources/launcher/Stride.Launcher/ViewModels/LauncherViewModel.cs
```diff
@@ -514,32 +503,27 @@ public async Task CheckForFirstInstall()
                         var versionToInstall = StrideVersions.First(x => x.CanBeDownloaded);
                         await versionToInstall.Download(true);
                     }
-                    if (!VsixPackage.IsLatestVersionInstalled && VisualStudioVersions.AvailableVisualStudioInstances.Any())
+
```
jrinker03, line 506: I actually think that the VSIX install dialogs shouldn't be presented unless the user has elected to install some version of Stride.  It's a little bit of a corner case but makes a little more sense in my head that way.

Thoughts?

sources/core/Stride.Core.Design/VisualStudio/VisualStudioVersions.cs
jrinker03: Turns out the detection logic was pretty easy with my other change in place so the installer's uninstall operation will now detect if the VS2022 plugin is installed and skip if not.

sources/launcher/Stride.Launcher/ViewModels/LauncherViewModel.cs
Eideren: I might have misunderstood you but this is already the case? When downloading the first version of the engine, the launcher will ask after downloading it whether the user would like to install the VS plugin

sources/core/Stride.Core.Design/VisualStudio/VisualStudioVersions.cs
```diff
@@ -68,102 +82,117 @@ public IDEInfo(Version version, string displayName, string installationPath, boo
...
+        /// Thd path to the current user's Visual Studio extension installation directory.
+        /// </summary>
+        public string VsixInstallationPath { get; }
+
+        /// <summary>
+        /// Computes and verifies the location of devenv.exe relative to the given path or null if not found. 
+        /// </summary>
+        /// <param name="idePath">The path of this VS instance</param>
+        /// <returns>The full path to devenv.exe or null.</returns>
+        private string ComputeDevenvPath(string idePath)
```
Eideren, line 97: Wouldn't it make more sense as a `bool TryFindDevEnvPath(string, out string)`. Or even just inlined to the call site given that this function is very short and only called once.
Same with `ComputeVsixInstallerPath`.

sources/launcher/Stride.Launcher/ViewModels/LauncherViewModel.cs
jrinker03: It does, yes, but it will also ask if you _decline_ to install any version of Stride.   As I said, it's a bit of a corner case but feels weird to ask the user to install the extensions after they declined to install a first version of Stride.

sources/core/Stride.Core.Design/VisualStudio/VisualStudioVersions.cs
Eideren: I'm pretty sure you would just need to call the VSIXInstaller with `/quiet`, so:
```cs
RunVsixInstaller(ideInfo.VsixInstallerPath, "/quiet /uninstall:Stride.VisualStudio.Package.2022");
```
Inside of `sources/tools/Stride.VisualStudio.PackageInstall/Program.cs`

sources/launcher/Stride.Launcher/Resources/Strings.Designer.cs
Eideren: Yup, a bunch of other files shouldn't be tracked, we'll clean this one with the rest afterward. Let's leave this one in for now.

sources/tools/Stride.VisualStudio.Package/Stride.VisualStudio.Package.nuspec
```diff
@@ -9,8 +9,8 @@
     <projectUrl>http://stride3d.net/</projectUrl>
     <iconUrl>https://raw.githubusercontent.com/stride3d/media/master/images/mainlogo/nuget/logo.png</iconUrl>
     <repository type="git" url="https://github.com/stride3d/stride" />
-    <copyright>&#169; Stride contributors and Silicon Studio Corp.</copyright>
-    <description>Visual Studio Integration Plugin for Stride</description>
+    <copyright>&#169; Stride contributors and  the .NET Foundation</copyright>
+    <description>Visual Studio Extension for Stride</description>
```
Eideren, line 13: Same as copyright issue mentioned previously, probably should including contributors, dotnet and Silicon Studio

sources/assets/Stride.Core.Packages/NugetStore.cs
```diff
@@ -155,7 +155,29 @@ public static void UpdatePackageSource(ISettings settings, string name, string u
...
+        {
+            VS2019,
+            VS2022
+        }
+
+        /// <summary>
+        /// A mapping of the supported versions of VS to a Stride release version range.  
+        /// For each supported VS release, the first Version represents the included earliest Stride version eligible for the VSIX and the second Version is the excluded upper bound.
+        /// </summary>
+        public IReadOnlyDictionary<VsixSupportedVsVersion, (PackageVersion, PackageVersion)> VsixVersionToStrideRelease { get; } = new Dictionary<VsixSupportedVsVersion, (PackageVersion, PackageVersion)>
```
Eideren, line 173: You should be able to define names for the parameters of those tuples by defining the tuple like so 
```cs
public IReadOnlyDictionary<VsixSupportedVsVersion, (min:PackageVersion, max:PackageVersion)>
```
You can then use those inside of `sources/launcher/Stride.Launcher/ViewModels/VsixVersionViewModel.cs` instead of `Item1` and `Item2`

sources/launcher/Stride.Launcher/ViewModels/LauncherViewModel.cs
Eideren: Sure, that makes sense, we can go ahead with your idea.

Eideren (commented):
Thanks a bunch for working on this, there's a stray nupkg included in your PR, afaict it isn;'t necessary ?

sources/assets/Stride.Core.Packages/NugetStore.cs
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net) and Silicon Studio Corp. (https://www.siliconstudio.co.jp)
```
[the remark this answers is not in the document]
jrinker03: Oh, I misread the comment in the Discord discussion. You're right, it's new code.  Will update.

jrinker03:
> there's a stray nupkg included in your PR, afaict it isn't necessary ?

Shoot, I generated that for testing and missed removing it from my last commit.  Will remove.

sources/assets/Stride.Core.Packages/NugetStore.cs
jrinker03: Great, thanks for the tip.  Will update.

sources/core/Stride.Core.Design/VisualStudio/VisualStudioVersions.cs
jrinker03: Not sure about the TryFind semantic as the bool wouldn't be used.   I'll go with the suggestion to inline these methods as it makes things more compact without reducing readability.

sources/core/Stride.Core.Design/VisualStudio/VisualStudioVersions.cs
jrinker03: That would work but the difference is that it's still invoking VSIXInstaller which will take more than a few seconds vs. a few hundred milliseconds to check the file system and skip calling VSIXInstaller.

sources/assets/Stride.Core.Packages/NugetStore.cs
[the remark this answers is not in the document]
jrinker03: This should be resolved.

sources/core/Stride.Core.Design/VisualStudio/VisualStudioVersions.cs
Eideren: I don't think it matters that much given that it's only once per installation, easier to maintain and future proof ?

jrinker03: OK, it a good tradeoff: small perf degradation in some edge cases for better maintainability given Stride's limited contributor group.  I'll update and retest.

Since so much of the VS extension handling pieces are opaque to the public or explicitly not supported, Stride managing extensions on behalf of VS is probably more headache than it's worth.   As I mentioned in Discord, I believe the better direction is to publish the extension to the VS Marketplace and direct folks to install it from there.  VS has all the right support for this built in.

jrinker03: Code removed and it works as expected with the VSIX installed and uninstalled.

sources/assets/Stride.Core.Packages/NugetStore.cs
jrinker03: Just a note that the syntax was slightly different than what you mentioned but it works well and makes the code much more satisfying.

sources/core/Stride.Core.Design/VisualStudio/VisualStudioVersions.cs
Eideren: Regarding the VS marketplace, I do think you're right but I feel like we would need xen to chime in for this and it looks like he doesn't have a lot of time right now to do so. Something to investigate later.

xen2:
Great job!

> Shoot, I generated that for testing and missed removing it from my last commit. Will remove.

If you don't mind, can you rewrite the commits so that this file is not part of history please? (otherwise it will have to be downloaded forever by everybody since .nupkg were not part of LFS files).

PR #1401, opened 2022-04-09
Tutorial project added: C#intermediate tutorial

Aggror:
# PR Details
Adds a new tutorial project to the launcher: C# intermediate tutorials. 

## Description
Contains the first 11 tutorials. Preview with an explanation can seen here: https://www.youtube.com/watch?v=Fkd_FnzdThc

once this PR has been reviewed and feedback has been processed, videos will be recorded which added here under a new section: https://doc.stride3d.net/latest/en/tutorials/index.html

## Types of changes
- [x] Docs change will follow at a later point, once this PR has been approved
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist
- [x] My change requires a change to the documentation.

PR #1406, opened 2022-04-19
[Physics] Update WorldMatrix when forcing transformation on PhysicsComponent

Eideren:
# PR Details
`UpdatePhysicsTransformation()` is likely to set the transformations of the physics-engine side object based on previous frame transformation instead of latest, [see summary of `WorldMatrix` to explain why](https://github.com/stride3d/stride/blob/master/sources/engine/Stride.Engine/Engine/TransformComponent.cs#L45-L49).
It makes sense that an `Update` function works with up to date data, so this looks more like a bug to me.

More context here https://github.com/stride3d/stride/pull/1401#discussion_r850608711 and in the discussion further below.

Also took the time to clean things up a bit.

## Motivation and Context
Fix bug.

## Types of changes

## Checklist


tebjan:
This looks good to me and is an important fix, the physics transformation shouldn't lag a frame behind. I think you have merge rights, go ahead.

PR #1416, opened 2022-04-30
[WIP] Convert existing Samples to use a common base package.

jrinker03:
# PR Details
Current samples will use Stride package dependencies from a common version.
Also correct a couple of build errors.

## Motivation and Context
To try to reduce the number of files touched when publishing new package versions, all current samples will pull from a common package version number in a .props file.

Also while compiling the StrideSamples.sln, I found and corrected two build time errors.

## Types of changes

## Checklist
**Note that this change requires an update to the readme for Samples**


xen2:
I am not sure if it's a good idea:
- Samples have their own Stride version for a reason: This version should reflect the current version of the assets in the sample itself (aka minimum version that can open it). This version can be slightly out of sync with current one (until samples are upgraded). When we instantiate one (from New project dialog), they are automatically upgraded if necessary at the same time so most of the time upgrading the samples manually is not necessary (and better to keep it that way otherwise it will be an easy thing to forget before a release). Also, I am sure that if we have such a `StrideSamplesBasePackageVersion`, people will bump it without knowing it needs to be done by upgrading project with Stride).
- They might even later be distributed as separate package, listed in New Project from NuGet and downloaded on demand only if created by user (and for previously mentioned reason they can have their own version and require auto-upgrade, so we'll probably try to keep them as low-version as needed as long as auto-upgrade works, easier to maintain).
- I think current implementation wouldn't work as is, considering the way samples are instantiated (this `Stride.Samples.props` exists when you open samples through the project file, but it won't exist when we try to instantiate the samples from the New Project dialog). Also Stride probably won't understand that it needs to be saved back to this file when upgrading project.

So overall I don't think it will work nicely with the way we version, load and upgrade samples.

Hope that makes sense, let me know if you need further explanation/details.

jrinker03:
Thanks for the explanation, Xen2.  Makes sense.

Abandoning this proposal.

PR #1420, opened 2022-05-01
Update setup packages to detect and install .NET 6

### LINKED ISSUE #1410: StrideSetup.exe needs to be updated
**Release Type**: GitHub

**Version**: master

The current setup requires .NET 4.7.2 rather than .NET 6.  The setup.aip needs to be updated to require and download .NET 6.

We should probably also limit the supported OSes to Windows 10, 11 and Server 2019 and 2022.

jrinker03:
# PR Details
Update the setup prereqs and main setup installer projects to detect and, if needed, install .NET 6.   At the moment Advanced Installer only knows about 6.0.2 but that will be fine.

Also update to latest Advanced Installer, version 19.4.

## Related Issue
Fixes #1410 

## Checklist
Manually tested.

### DIFF
--- a/sources/launcher/Prerequisites/launcher-prerequisites.aip
+++ b/sources/launcher/Prerequisites/launcher-prerequisites.aip
@@ -212,11 +212,9 @@
   </COMPONENT>
   <COMPONENT cid="caphyon.advinst.msicomp.PreReqComponent">
     <ROW PrereqKey="A6D7F21EAEB4C4A901F0979CF110936" DisplayName=".NET Framework 4.7.2" SetupFileUrl="https://download.microsoft.com/download/6/E/4/6E48E8AB-DC00-419E-9704-06DD46E5F81D/NDP472-KB4054530-x86-x64-AllOS-ENU.exe" Location="1" ExactSize="83943272" WinNTVersions="Windows Vista RTM x86, Windows Server 2008 x86, Windows 7 RTM x86, Windows 10 version 1507 x86, Windows 10 version 1511 x86, Windows 10 version 1803 x86, Windows 10 version 1809 x86" WinNT64Versions="Windows Vista RTM x64, Windows Server 2008 x64, Windows 7 RTM x64, Windows Server 2008 R2 RTM x64, Windows 10 version 1507 x64, Windows 10 version 1511 x64, Windows 10 version 1803 x64, Windows 10 version 1809 x64" Operator="1" NoUiComLine="/q /norestart" Options="xy" MD5="87450cfa175585b23a76bbd7052ee66b" TargetName=".NET Framework 4.7.2"/>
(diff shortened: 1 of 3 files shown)


jrinker03:
@Xen2 is updating the build server to use AI 19.4 as well.

jrinker03:
Note that the Advanced Installer GUI made all these changes, and I can't claim to know the exact effect of each one.  Manual testing of a locally built setup does show it works as expected.   Once Xen2 gets the build server updated, I'll run a build and test from an official build as well.

manio143:
I looked through this XML and I think it's ok (lots of installer specific stuff) - so long as we can run it in a VM from a clean Windows installation and get to run Stride that's all we need.
I remember there being another requirement for VCRedist to run GameStudio, so maybe we could add that as well?

jrinker03:
>I remember there being another requirement for VCRedist to run GameStudio, so maybe we could add that as well?

While I'm sure that's so, it's missing from the [prereqs web page](https://doc.stride3d.net/latest/en/manual/requirements/index.html).   

Win 10+ comes with the Universal CRT so it shouldn't be an issue, it's only for older OSes that this might come into play.
And interestingly Advanced Installer doesn't even list a CRT as an option it supports out-of-the box.  

If needed, perhaps we should just add it later since it hasn't been in these projects previously?

jrinker03:
>so long as we can run it in a VM from a clean Windows installation and get to run Stride that's all we need.

I've verified a locally built installer EXE on a clean VM (win10-21h2-pro) and it works as expected:  it detects that .NET FW 4.x is installed and that .NET6  is not.  It properly downloads and installs the missing prereq and then properly installs the Launcher.    The Launcher is started at the completion of setup and works as expected.

I'll re-run on a new clean VM with a StrideSetup.exe from the build server once this PR is accepted and a build run.

jrinker03:
@manio143 - any objection with moving forward with the PR as-is since it seems to behave as expected?

manio143:
None from me. Not sure if we're waiting on @xen2 to comment before merging?

jrinker03:
OK, thanks.

Xen2 has the link to this PR so not sure.

xen2:
There's also sources/prerequisites/prerequisites.aip for Stride itself
Supposedly launcher should have installed it, but I think it's still good to add it there in case the launcher has different requirements in the future (and it doesn't install .NET 6, for example if we pack launcher with .NET 6 runtime).

xen2:
Thinking about it, we need to check if launcher already published with self-contained .NET 5/6 runtime.
If yes, both launcher setup don't need .NET 6.0, only Stride itself.

jrinker03:
Pushed the latest based on clean VM testing.  In the end:
- As @xen2 was thinking, since the launcher is a self-contained EXE, there is no need for a .NET 6 prereq.  I've removed that.
- I've updated the Stride prereq installer and on a Win10 21h2 machine it successfully evaluates everything and discovers that everything is good-to-go (CRT, DirectX and .NET dependencies) and exits without doing anything else.   This was verified with event log entries.

In the end, the update is really to use 19.4 version of Advanced installer and to change the Stride prereq installer to look for .NET 6 instead of 5.

jrinker03:
As an interesting aside, the Azure OS image comes with .NET 6 MSI in the OSes package cache, so it didn't wind up downloading it, just installing.

jrinker03:
Ah, and same with the CRT, it turns out.   Since it was a very fast silent install it tricked me into thinking it didn't do anything.

It might be worth considering moving to the Universal CRT but we can save that for another day since this works.

Eideren:
Sounds like this PR is good to go, ready to merge ?

manio143 (approved):

Issue #1383, opened 2022-03-27
Address hundreds of compiler warnings


jrinker03:
**Release Type**: GitHub

**Version**: master

**Platform(s)**: build

**Describe the bug**
When building the main stride solutions, there are hundreds of warnings emitted by the various tools.   As of this issue submission, there are ~290 such warnings.   This is fairly dangerous as it leads to developers ignoring potential bugs.

As examples of potential issues:
- warning CS8073: The result of the expression is always 'false' since a value of type 'Guid' is never equal to 'null' of type 'Guid?'
- warning C4244: 'initializing': conversion from 'double' to 'float', possible loss of data
- warning C4244: 'argument': conversion from 'float' to 'const fbxsdk::FbxLongLong', possible loss of data

These may be benign but perhaps not. The safest solution is to examine each and address the warning in an appropriate way.

Additionally, there are reasonably harmless messages but point at clutter that needs to be addressed to streamline the codebase.  Things like:
- warning CS0162: Unreachable code detected
- The variable 'e' is declared but never used
- The field '<foo>' is never used
- '<some class or field>' is obsolete

And the last class is use of deprecated items that the compiler points out.   Deprecated classes are usually replaced by something that is superior in some way (for example, security or perf are high value reasons). Granted, these are likely benign at the moment, but could come back to cause issues later.

See also #1381 

**To Reproduce**
1. Compile the stride.sln from the latest master code.

**Expected behavior**
No warnings on build.
Bonus would be that warnings are treated as errors.

**Screenshots**
N/A

**Log and callstacks**
See above for examples

**Additional context**
I realize that these might be time consuming to address completely and I am willing to take these on over a series of PRs.   From experience, having someone scrutinize these and bring up interesting cases for discussion is well worth the effort.

manio143:
I see there's even more warnings on the build server. There's SA* warnings - which leads to believe the target that turns off analyzers by default is actually disabled on the build server.

jrinker03:
From experience, some of those analyzer warnings are really good, too.  Shame they're off locally.  I'll have to take a peek at them one day.

PR #1421, opened 2022-05-02
Next round of addressing CS Warnings

jrinker03:
# PR Details
Continuing to address C# warnings.

## Description
This change addresses most or all of the following issues:
CS0414 - The field 'foo' is assigned but its value is never used
CS0612 - <member> is obsolete.   (see Note below)
CS0649 - Field 'foo' is never assigned to, and will always have its default value
CS0659 - 'foo' overrides Object.Equals(object o) but does not override Object.GetHashCode()
CS0675 - Bitwise-or operator used on a sign-extended operand; consider casting to a smaller unsigned type first
CS1998 - This async method lacks 'await' operators and will run synchronously.
CS4014 - Because this call is not awaited, execution of the current method continues before the call is completed.
CS8073 - The result of the expression is always 'false' since a value of type 'Guid' is never equal to 'null'

Note that because NodeViewModel.MemberInfo was always null, this caused 3 dependent providers to always have their MatchNode method return false.   I left the three classes in this iteration because they are still referenced in xaml and I'm unclear of the effect of removing the xaml at this time.

In the case of TestThumbnails.cs, it was half-finished and not containing any valid tests so it was removed.

Also, some other minor hygiene changes (sort/remove usings, remove unused package refs, etc.).

## Related Issue
#1383

## Motivation and Context
See related issue

## Types of changes

## Checklist


sources/engine/Stride.Engine/Shaders.Compiler/RemoteEffectCompilerClient.cs
```diff
@@ -82,7 +82,7 @@ public async Task<SocketMessageLayer> Connect(string packageName, CancellationTo
             // Register network VFS
             NetworkVirtualFileProvider.RegisterServer(socketMessageLayer);
 
-            Task.Run(() => socketMessageLayer.MessageLoop());
+            await Task.Run(() => socketMessageLayer.MessageLoop());
```
manio143, line 85: Same question as on `EffectCompilerServerSession.cs`

sources/engine/Stride.Rendering/Rendering/Shadows/ShadowMapRenderer.cs
```diff
@@ -44,9 +42,6 @@ public ShadowMapRenderer()
 
         public HashSet<RenderView> RenderViewsWithShadows { get; } = new HashSet<RenderView>();
 
-        // TODO
-        public IReadOnlyList<RenderStage> ShadowMapRenderStages => shadowMapRenderStages;
```
manio143, line 48: Clearly we don't know what was the idea behind this. Since this is part of public API, can you mark your PR as 'breaking changes' for awareness?

sources/tools/Stride.EffectCompilerServer/EffectCompilerServer.cs
```diff
@@ -87,11 +87,11 @@ protected override async void HandleClient(SimpleSocket clientSocket, string url
                     }
 
                     // Forward to game studio
-                    gameStudio.Send(packet);
+                    gameStudio.Send(packet).Wait();
                 });
             }
 
-            Task.Run(() => socketMessageLayer.MessageLoop());
+            await Task.Run(() => socketMessageLayer.MessageLoop());
```
manio143, line 94: Same comment here about message loop

sources/tools/Stride.SamplesTestServer/SamplesTestServer.cs
```diff
@@ -434,7 +434,7 @@ protected override async void HandleClient(SimpleSocket clientSocket, string url
                 await tester.TesterSocket.Send(new ScreenshotStored());
             });
 
-            Task.Run(async () =>
+            await Task.Run(async () =>
```
manio143, line 437: I don't think we should await it here. So to sum up, in all places where `Task.Run` is used to spawn a thread with a message loop, we should just use `_ = ` to ignore the value and suppress the warning.

manio143 (commented):
Generally alright. I'm a bit wary when it comes to awaiting `Task.Run`, see the last comment about message loops on my suggestion for change.

sources/core/Stride.Core.Design/PackageVersionRange.cs
```diff
@@ -36,7 +36,9 @@ namespace Stride.Core
     /// </code>
     /// </remarks>
     [DataContract("PackageVersionDependency")]
+#pragma warning disable CS0659 // Type overrides Object.Equals(object o) but does not override Object.GetHashCode()
     public sealed class PackageVersionRange : IEquatable<PackageVersionRange>
```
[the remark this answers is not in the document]
jrinker03: Sounds good.

sources/editor/Stride.Assets.Presentation/TemplateProviders/RenderStageReferenceTemplateProvider.cs
```diff
@@ -17,8 +13,7 @@ public class RenderStageReferenceTemplateProvider : NodeViewModelTemplateProvide
         public override bool MatchNode(NodeViewModel node)
         {
             // Everything that is not in GraphicsCompositorAsset.RenderStages should become a reference
-            return node.Type == typeof(RenderStage)
-                && !(node.Parent?.MemberInfo != null && node.Parent.MemberInfo.DeclaringType == typeof(GraphicsCompositorAsset) && node.Parent.MemberInfo.Name == nameof(GraphicsCompositorAsset.RenderStages));
+            return false;
```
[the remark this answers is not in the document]
jrinker03: Oh, you are correct!  Thanks for catching this.

sources/editor/Stride.Core.Assets.Editor/ViewModel/SessionViewModel.cs
```diff
@@ -636,7 +636,7 @@ private void EditSelectedAsset()
                 // Temporary code while we don't have an integrated text editor.
                 if (Editor.TextAssetTypes.Contains(asset.AssetType.Name))
                 {
-                    OpenWithTextEditor(asset, EditorSettings.ShaderEditor.GetValue());
+                    OpenWithTextEditor(asset, EditorSettings.ShaderEditor.GetValue()).Wait();
```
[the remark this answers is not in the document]
jrinker03: Right, TY.

sources/engine/Stride.Engine/Shaders.Compiler/Internals/NetworkVirtualFileProvider.cs
```diff
@@ -119,7 +119,7 @@ public NetworkWriteStream(SocketMessageLayer socketMessageLayer, string url)
 
             protected override void Dispose(bool disposing)
             {
-                socketMessageLayer.Send(new UploadFilePacket { Url = url, Data = memoryStream.ToArray() });
+                socketMessageLayer.Send(new UploadFilePacket { Url = url, Data = memoryStream.ToArray() }).Wait();
```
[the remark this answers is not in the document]
jrinker03: Yes, will update.

sources/engine/Stride.Rendering/Rendering/Shadows/ShadowMapRenderer.cs
jrinker03: Will do.

sources/tools/Stride.SamplesTestServer/SamplesTestServer.cs
jrinker03: Ah, that's a good, safe solution I should have thought of.  Will update.

sources/editor/Stride.Core.Assets.Editor/ViewModel/SessionViewModel.cs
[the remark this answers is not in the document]
jrinker03: I'm surprised the compiler didn't flag `OpenAssetEditorWindow` as well.

sources/tools/Stride.SamplesTestServer/SamplesTestServer.cs
xen2: Turns out we even made a method to ignore this warning, extension method `Forget()`:

https://github.com/stride3d/stride/blob/6a2e29b6aa4aeaf852631f9466b05b201d182142/sources/core/Stride.Core.Design/Extensions/TaskExtensions.cs#L14

You can see one example usage of it here:
https://github.com/stride3d/stride/blob/7e836297cb5930c01e6dfa0183e7f3cc64748fb6/sources/editor/Stride.Core.Assets.Editor/Components/DebugTools/UndoRedo/DebugUndoRedoViewModel.cs#L49

Please be very careful to not add any await in this PR, it is (usually) skipped in some specific places on purpose and might lead to lot of unexpected block, hang or other undesired side-effect.

sources/editor/Stride.Assets.Presentation/TemplateProviders/RenderStageReferenceTemplateProvider.cs
```diff
@@ -17,8 +14,7 @@ public class RenderStageReferenceTemplateProvider : NodeViewModelTemplateProvide
         public override bool MatchNode(NodeViewModel node)
         {
             // Everything that is not in GraphicsCompositorAsset.RenderStages should become a reference
-            return node.Type == typeof(RenderStage)
-                && !(node.Parent?.MemberInfo != null && node.Parent.MemberInfo.DeclaringType == typeof(GraphicsCompositorAsset) && node.Parent.MemberInfo.Name == nameof(GraphicsCompositorAsset.RenderStages));
```
xen2, line 21: I would prefer it if we don't suppress this code without any proper replacement (assuming it was there for a reason).
Either:
- fix it (not sure of exact equivalent of `MemberInfo.DeclaringType`, maybe `Root.Type`?).
- if not sure or difficult to test, I would leave the current broken code in, otherwise we will forget about it if this code is simply suppressed.
- in this specific case, it seems that `RenderStageReferenceTemplateProvider` might not be even used at all? so it might be easier to make sure of that and delete the class if true.

Hopefully, we're in the 3rd case (which would make sense, otherwise the behavior would have been broken).

Same applies for all uses of `MemberInfo`

ericwj:
Would you like to finish this @jrinker03? Would be best to see the result of this before merging a lot more warnings.

PR #1423, opened 2022-05-03
Fixes "key already exists" exception when preloading library containing uppercase letters 

azeno:
# PR Details

## Description

Fixes crash when calling `PreloadLibrary` multiple times with a library name containing uppercase letters (for example "libGLESv2"). Instead of using lowercase letters when storing the result in the dictionary, simply tell the dictionary to use a comparer ignoring character casing.

## Types of changes

## Checklist


Eideren:
Thanks !

Issue #1422, opened 2022-05-02
Problematic use of shared projects and code files


jrinker03:
**Release Type**: GitHub

**Version**: master

**Describe the bug**
The way in which multiple assemblies are reusing the same code files and shared projects within the same application leads to multiple duplicated symbols resulting in compiler warnings, slower compilation times and marginally larger assemblies.   

In the case of shared projects (ex. Stride.NuGetResolver), these are not being used in the way .shproj were intended.   Shared projects are for code and assets being reused across different applications, typically targeting different native platforms.   [More on that from Microsoft](https://docs.microsoft.com/en-us/xamarin/cross-platform/app-fundamentals/shared-projects?tabs=windows).  
The code within the existing shared projects should stand alone in their own assemblies or get rolled into one or more existing assemblies to eliminate the duplication issue.

Similarly, there are instances of individual .cs files being compiled into different assemblies which interact with each other in the same application.  The largest offender is the SharedAssemblyInfo, which anyone who has compiled Stride.sln will have seen the many duplicate symbol warnings.  

**To Reproduce**
Compile Stride.sln and observe many warnings of these type (the assemblies in the warnings will differ per build):
```
warning CS0436: The type 'StrideVersion' in 'H:\src\jrinker03\stride\sources\editor\Stride.GameStudio\..\..\shared\SharedAssemblyInfo.cs' conflicts with the imported type 'StrideVersion' in 'Stride.Core.Assets.Editor, Version=4.1.0.1, Culture=neutral, PublicKeyToken=null'. Using the type defined in 'H:\src\jrinker03\stride\sources\editor\Stride.GameStudio\..\..\shared\SharedAssemblyInfo.cs'.

warning CS0436: The type 'PublicKeys' in 'H:\src\jrinker03\stride\sources\editor\Stride.Editor\..\..\shared\SharedAssemblyInfo.cs' conflicts with the imported type 'PublicKeys' in 'Stride.Engine, Version=4.1.0.1, Culture=neutral, PublicKeyToken=null'. Using the type defined in 'H:\src\jrinker03\stride\sources\editor\Stride.Editor\..\..\shared\SharedAssemblyInfo.cs'.

[...]

PR #1424, opened 2022-05-04
Move or remove problematic shared code within the same application, part 1

jrinker03:
# PR Details

This change is related to duplicated code in the same application causing compiler issues. This change either moves or removes duplicate references to shared code within the same end application.   

Some of the shared code turns out to be obsolete related to live debugging in Studio.  In this case, this change also removes the related code as well.

## Description

There are three types of changes which are reasonably benign:
- Some shared projects were no longer needed in projects that referenced them, so the reference was removed.  For example, Stride.NuGetResolver
- In the case of Stride.Core.ShellHelper, moving the code to a common, existing assembly was the simple solution.
- The more wide-spread change was removing code related to an obsolete 'edit and continue'-like mode that leveraged shared code that itself becomes obsolete.   It's possible that I might have missed related code so welcome reviewer's input on this part of the work.

This PR does not try to tackle the entirety of #1422 as I expect the rest will require some continued thought and discussion.

## Related Issue

#1422 

## Motivation and Context

Continuing to chip away at improving the codebase's hygiene.

## Types of changes

## Checklist


sources/editor/Stride.GameStudio/StrideEditorPlugin.cs
```diff
@@ -82,11 +82,7 @@ public override void InitializeSession(SessionViewModel session)
             var thumbnailService = new GameStudioThumbnailService(session, settingsProvider, builderService);
             session.ServiceProvider.RegisterService(thumbnailService);
 
-            var strideDebugService = new StrideDebugService(session.ServiceProvider);
-            session.ServiceProvider.RegisterService(strideDebugService);
-
             GameStudioViewModel.GameStudio.Preview = new PreviewViewModel(session);
-            GameStudioViewModel.GameStudio.Debugging = new DebuggingViewModel(GameStudioViewModel.GameStudio, strideDebugService);
```
manio143, line 89: Do we need to keep the `DebuggingViewModel` class if we remove it from being created?

Issue #1061, opened 2021-05-11
Cant build / run any stride application on macOS machine


coman3:
**Release Type**: Official Release

**Version**: 4.1.0.1459-beta

**Platform(s)**: macOS

**Describe the bug**
Execution of build through Visual Studio for Mac results in error:
```
/Users/lachlan/.nuget/packages/stride.core.assets.compilerapp/4.1.0.1459-beta/buildTransitive/Stride.Core.Assets.CompilerApp.targets(5,5): Error MSB3073: The command "mono "/Users/lachlan/.nuget/packages/stride.core.assets.compilerapp/4.1.0.1459-beta/buildTransitive/../lib/net5.0-windows7.0/Stride.Core.Assets.CompilerApp.exe"  --disable-auto-compile --project-configuration "Release" --platform=Windows --project-configuration=Release --compile-property:StrideGraphicsApi=Direct3D11 --output-path="/Users/lachlan/Desktop/CSharpBeginner/Bin/macOS/Release/osx-x64/data" --build-path="/Users/lachlan/Desktop/CSharpBeginner/CSharpBeginner.macOS/obj/stride/assetbuild/data" --package-file="/Users/lachlan/Desktop/CSharpBeginner/CSharpBeginner.macOS/CSharpBeginner.macOS.csproj" --msbuild-uptodatecheck-filebase="/Users/lachlan/Desktop/CSharpBeginner/CSharpBeginner.macOS/obj/x64/Release/osx-x64/stride/assetcompiler-uptodatecheck"" exited with code 2. (MSB3073) (CSharpBeginner.macOS)
```
(Notice how platform is "windows" despite building on a mac.
**To Reproduce**
1. Create new stripe project on Windows machine through the Stripe UI (ensure you select macOS and Windows as a platform)
2. Ensure it builds and runs on windows
3. Copy assets over to macOS machine ensuring to remove all bin, obj and .vs folders
4. Open project in Visual Studio for Mac
5. Build macOS project

**Expected behavior**
The project builds

**Log and callstacks**
**Additional context**
If building the macOS project on windows first and attempting to run through macos using `dotnet run projectname.macos.dll` i am presented with the following error as well:
```
Unhandled exception. System.TypeInitializationException: The type initializer for '<Module>' threw an exception.
 ---> System.TypeInitializationException: The type initializer for '<Module>' threw an exception.
 ---> System.TypeInitializationException: The type initializer for '<Module>' threw an exception.
[...]

PR #1428, opened 2022-05-12
[Build] Native libraries get copied to runtimes/RID/native folder

azeno:
## Description

- In user project nuget will take care of copying required native libs to final output folder depending on used RuntimeIdentifier
- Native libraries can be registered as dependencies for a later lookup (used by Stride.NuGetLoader to register native dependencies from global nuget package cache)
- Fixes library not found exceptions on platforms other than windows
- Fixes generated Android project

It seems to work fine for Windows and Linux. Android worked so far that I was able to build and run in debug on a Fairphone 3. However it didn't feel stable - for example on the second attempt it started up in the wrong resolution (was only rendering on half of the screen) and after a hide/show it crashed with depth buffer having different size than render target. In release build I had to set the RID to android-arm64 or otherwise the asset compiler would get invoked multiple times tripping over itself with file in use exceptions. But these crashes seem unrelated to the changes I'm proposing here.

MacOS and iOS I couldn't test yet because of lack of Apple hardware.

## Related Issue

#1061 (I tested with Linux and not OSX)

## Types of changes

## Checklist


sources/core/Stride.Core/Native/NativeLibraryHelper.cs
```diff
@@ -33,6 +34,24 @@ public static void PreloadLibrary(string libraryName, Type owner)
...
+                    if (NativeDependencies.TryGetValue(libraryName, out var path) && NativeLibrary.TryLoad(path, out var result))
+                    {
+                        LoadedLibraries.Add(libraryName, result);
+                        return;
+                    }
+                }
+
+                // Try default loading mechanism (https://docs.microsoft.com/en-us/dotnet/core/dependency-loading/loading-unmanaged)
+                {
+                    if (NativeLibrary.TryLoad(libraryName, owner.Assembly, searchPath: null, out var result))
```
manio143, line 48: This is the only callsite in this helper that includes owner assembly, do you think we should include it everywhere?

sources/shared/Stride.NuGetResolver/NuGetAssemblyResolver.cs
```diff
@@ -145,13 +147,26 @@ internal static void SetupNuGet(string packageName, string packageVersion)
...
-                            var (request, result) = RestoreHelper.Restore(logger, nugetFramework, "win", packageName, versionRange);
+                            var (request, result) = RestoreHelper.Restore(logger, nugetFramework, RuntimeInformation.RuntimeIdentifier, packageName, versionRange);
                             if (!result.Success)
                             {
                                 throw new InvalidOperationException($"Could not restore NuGet packages");
                             }
 
                             assemblies = RestoreHelper.ListAssemblies(result.LockFile);
+
+                            // Register the native libraries
```
manio143, line 158: Can this be extracted into a method? And maybe add a new line after each if for readability.

manio143 (commented):
It's really nice to see us getting more in line with what NuGet offers :+1:

sources/shared/Stride.NuGetResolver/NuGetAssemblyResolver.cs
azeno: Done

xen2:
A very appreciated change!
Since we're now .NET 6.0 only (vs 5 different TFM until recently), it's great we can simplify all those custom stuff we had and replace them with standard/default approach.

azeno:
Some stuff will need to stay because there's no system yet which deals with native dependency consumption in project to project references. There are a couple of issues on dotnet regarding this, for example https://github.com/dotnet/sdk/issues/10575

sources/core/Stride.Core/Native/NativeLibraryHelper.cs
manio143: It's fine - I thought the `TryLoad` had more overrides and it's not completely clear to me what purpose the passed assembly serves. Like how would it be different if you used `TryLoad(libraryName, out var result)`?

azeno: Best description I found is in the source code: https://github.com/dotnet/corert/blob/master/src/System.Private.CoreLib/shared/System/Runtime/InteropServices/NativeLibrary.cs#L121
The assembly defines what AssemblyLoadContext to use. It can also provide search paths through attributes it seems.

azeno:
Now that 4.1 is released, and to my understanding the build agent has been updated - can we give this one another spin?

manio143:
I think I was happy with those changes. @xen2 is it fine to try and merge it?

azeno:
Seeing people struggling with building for Android (https://github.com/stride3d/stride/discussions/1601, https://github.com/stride3d/stride/pull/1456), I wonder what is the hold up here? If I remember correctly I had the example game running on Android.

PR #1441, opened 2022-06-14
OpenXR support

arturoc:
# PR Details

Adds support for OpenXR using DX11

## Description

Port of https://github.com/anotherdevguy/stride/tree/openxr to work with DX11 instead of Vulkan.
Supports the same functionality as currently supported by OpenVR except for the recenter function which is not supported by OpenXR itself and should be done manually from the engine.

## Motivation and Context

OpenXR superseeds OpenVR and supports many more features and devices. 

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Engine/Rendering/Compositing/ForwardRenderer.cs
+++ b/sources/engine/Stride.Engine/Rendering/Compositing/ForwardRenderer.cs
@@ -712,7 +712,7 @@ namespace Stride.Rendering.Compositing
                     }
 
                     //draw mirror to backbuffer (if size is matching and full viewport)
-                    if (VRSettings.CopyMirror && VRSettings.VRDevice.MirrorTexture != null)
+                    if (VRSettings.CopyMirror)
                     {
                         CopyOrScaleTexture(drawContext, VRSettings.VRDevice.MirrorTexture, drawContext.CommandList.RenderTarget);
                     }
--- a/sources/engine/Stride.Engine/Rendering/Compositing/ForwardRenderer.cs
+++ b/sources/engine/Stride.Engine/Rendering/Compositing/ForwardRenderer.cs
@@ -712,7 +712,7 @@ namespace Stride.Rendering.Compositing
                     }
 
                     //draw mirror to backbuffer (if size is matching and full viewport)
-                    if (VRSettings.CopyMirror)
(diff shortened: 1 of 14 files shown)


sources/engine/Stride.Engine/Rendering/Compositing/ForwardRenderer.cs
```diff
@@ -712,7 +712,7 @@ protected override void DrawCore(RenderContext context, RenderDrawContext drawCo
                     }
 
                     //draw mirror to backbuffer (if size is matching and full viewport)
-                    if (VRSettings.CopyMirror)
+                    if (VRSettings.CopyMirror && VRSettings.VRDevice.MirrorTexture != null)
```
tebjan, line 715: Was this still necessary in the end to avoid exceptions?

tebjan (commented):
Looks pretty good overall, very good work, thank you!

sources/engine/Stride.Engine/Rendering/Compositing/ForwardRenderer.cs
arturoc: Apparently OpenXR doesn't support this feature. Instead of checking the texture from the forward renderer I've added a CanMirrorTexture property to VRDevice that implementations can override to signal they don't support this feature.

Then VRSettings.CopyMirror will return false no matter what it's set to. Not sure if that's the best way to implement it but at least it's clearer than checking from the renderer

xen2:
Probably we should target `master` rather than `master-4.0`?

joreg:
@xen2 ideally we could have this in a 4.0 branch release so we can ship it with current versions of vvvv. or would this be too big a hassle on your end?

xen2:
> @xen2 ideally we could have this in a 4.0 branch release so we can ship it with current versions of vvvv. or would this be too big a hassle on your end?

If this works and no apparent regression, I can give it a try (build system changed quite a lot so not sure how CI will behave -- worst case I should be able to compile/publish it locally since it's a one time thing).

joreg:
if you could give that a shot it would be amazing! 
if it is too big a hassle, we've also added the same here already #1483

xen2:
FYI I had to revert because various platforms were broken (it might be difficult because older Xamarin platforms don't have proper Span support).
I will focus on `master` for now.

PR #1444, opened 2022-06-20
C#intermediate tutorial

Aggror:
# PR Details

## Description

Small improvements to intermediate tutorial code found during recording.

## Types of changes

### DIFF
--- a/samples/Tutorials/CSharpIntermediate/CSharpIntermediate/CSharpIntermediate.Game/05_Async/AsyncWebApi.cs
+++ b/samples/Tutorials/CSharpIntermediate/CSharpIntermediate/CSharpIntermediate.Game/05_Async/AsyncWebApi.cs
@@ -1,7 +1,6 @@
 // Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net)
 // Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.Net.Http;
 using System.Threading.Tasks;
 using Newtonsoft.Json;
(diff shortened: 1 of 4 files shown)


samples/Tutorials/CSharpIntermediate/CSharpIntermediate/CSharpIntermediate.Game/05_Async/AsyncWebApi.cs
```diff
@@ -12,43 +11,46 @@ namespace CSharpIntermediate.Code
...
-                if (Input.IsKeyPressed(Stride.Input.Keys.G)){
+                int drawX = 500, drawY = 600;
+                DebugText.Print($"Press A to get Api data from https://opencollective.com/stride3d", new Int2(drawX, drawY));
+
+                if (Input.IsKeyPressed(Stride.Input.Keys.G))
+                {
                     await RetrieveStrideRepos();
                 }
 
+                foreach (var openCollectiveEvent in openCollectiveEvents)
```
manio143, line 30: You might have to await NextFrame at the end of the `if` above to make sure to execute DebugText.Print on the game thread.

manio143 (approved):

samples/Tutorials/CSharpIntermediate/CSharpIntermediate/CSharpIntermediate.Game/05_Async/AsyncWebApi.cs
Aggror: @manio143 Hi!, thanks for the feedback. Would you add this specifically in the Retrieve method or right after the await? If I try this locally it don't see any behavior change.

manio143: I would say after the await in this method so that only one method does sync with the frames. I don't expect there to be changes in behavior in general, but since the continuation of the async call for a Task may run on another thread and Stride's APIs aren't thread safe, you can sometimes get into a race condition.

PR #1462, opened 2022-07-21
Fixes key not found exception when working with multiple devices

azeno:
See initial bug report here: https://discourse.vvvv.org/t/multiple-multitouch-screens-crashes-on-fast-touches/20683/5

## Types of changes

## Checklist


tebjan:
Typo `finderId` should probably be `fingerId`.

azeno:
Indeed, fixed. Btw this crash also happens with just two windows. Doesn't need multiple screens like in the initial report.

PR #1463, opened 2022-07-21
Fixes key not found exception when working with multiple devices

azeno:
See initial bug report here: https://discourse.vvvv.org/t/multiple-multitouch-screens-crashes-on-fast-touches/20683/5

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Input/SDL/PointerSDL.cs
+++ b/sources/engine/Stride.Input/SDL/PointerSDL.cs
@@ -21,7 +21,7 @@ namespace Stride.Input
         private static Sdl SDL = Window.SDL;
 
         private readonly Window uiControl;
-        private readonly Dictionary<long, int> touchFingerIndexMap = new Dictionary<long, int>();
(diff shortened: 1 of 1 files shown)


sources/engine/Stride.Input/SDL/PointerSDL.cs
```diff
@@ -62,25 +62,24 @@ private void OnSizeChanged(WindowEvent eventArgs)
             SetSurfaceSize(new Vector2(uiControl.ClientSize.Width, uiControl.ClientSize.Height));
         }
 
-        // TODO Code from PointeriOS with slight modifications, consider creating and referencing a utility class
```
manio143, line 65: Did you mean to remove this TODO comment?

xen2:
> @azeno I believe this PR got duplicated. : #1462 I couldn't find any differences in the PR's other than the latter one already had comments. I assume this is abnormal.

They target different branches (4.1 and 4.0 backport)

sources/engine/Stride.Input/SDL/PointerSDL.cs
azeno: Yes, because the code is different now. I imagine this issue doesn't exist at all on iOS.

Issue #1461, opened 2022-07-20
Migrate away from using `Stride.Core.Utilities`/`Stride.Core.Interop`


ericwj:
New features are available in recent versions of .NET that match functionality in `Stride.Core.Utilties` about exactly, employment of which can improve Stride in a number of ways.
* Removing P/Invoke overhead
* Migrating some C++ to C# or even eliminating the need for features implemented in C++
* Reducing the need for IL weaving
* Probably allow both the C# compiler and the JIT to do a better job optimizing the code

Getting rid of IL weaving is more or less a prerequisite or at least an advantage towards eventually fully supporting ahead of time compilation (AOT) during publish with minimal effort from the developer.

I would like to propose first to mark the public API in `Stride.Core.Utilities` as obsolete for as far as there are alternatives in C#/.NET and replace their use throughout Stride with the corresponding alternatives.

I deliberately scope this issue just to the class mentioned, although there are a series of other opportunities to follow the exact same line of thinking and that have about the same benefits. Those will come later.

PR #1469, opened 2022-07-23
Improve memory management, embrace modern .NET

ericwj:
# PR Details

Migrate away from IL weaving and take the first steps to improve memory management.

The mode of operation has generally been not to rewrite to safe code immediately.

## Description

* Fixed a buffer overrun in `ParameterCollection.CopyTo<T>`.
* Identified a bug affecting the copy operation in `ParameterCollection.UpdateLayout` caused by parameters changing type and size after being initialized. Modified the copy operation to right-size them for every element in the collection, pending fixing of the bug causing the types to differ.
* Replaces uses of methods that are IL weaved with their C# equivalents. Rewrites `Utilities` marking most methods in it as obsolete, remove `Interop`.
  * In particular using `Unsafe.CopyBlockUnaligned` instead of `Utilities.CopyMemory`.
* Reviewed most uses of `fixed`
  * To verify that code around it throws instead of dereferencing `null` if the array being fixed is `null` or `empty`.
  * To elide checks on `array.Length` if the `null` result of `fixed` is intended when the array is `null` or empty.
* Add `Debug.Assert` in some places to verify correctness of `CopyBlock` and similar operations.
* Review all implementations of `System.IO.Stream`:
  * Add synchronous overloads for `Read` and `Write` that take `Span<byte>` as argument.
  * Fix the exception behavior to match the documentation of `System.IO.Stream`; no longer throw `InvalidOperationException` or `EndOfStreamException`, but use `NotSupportedException` and `IOException`.
  * Obsolete and remove `NativeStream` and `NativeMemoryStream` from the inheritance graph of Strides `Stream` types.
    * Obsolete related types such as `NativeStreamWrapper`.
    * `SerializationStream.NativeStream` is of type `Stream`, no longer `NativeStream` and is obsoleted in favor of the new property `UnderlyingStream`.
    * Use `UnmanagedMemoryStream` to replace all usage of `NativeMemoryStream` in Stride.
    * `BlobStream` now inherits from `UnmanagedMemoryStream` instead of from `NativeMemoryStream`.
  * Polyfill the `virtual` methods in `NativeStream` with extension methods. These are obsolete themselves since they are inefficient.
* Obsolete `BinarySerialization` on grounds that it is inefficent and a very minor benefit to Stride users.
* Added handle types to the P/Invoke declarations of `Navigation`.
* Many methods dealing with primitives and e.g. image data have had their type constraints changed from `struct` to `unmanaged`.
* Minor improvements in performance such as vector-enabled `SequenceEqual` over `CompareMemory`, correctness, code style and use of new language features in particular `nint` and `nuint` over `IntPtr`.

## Related Issue

#1461

## Motivation and Context

See issue for the motivation for starting this PR. Issues with buffer overruns which have surfaced are hard to fix without tracking of owned memory throughout the code. Hence the scope of this PR is broadened to include basic memory management.

## Types of changes

- [x] ~Docs change~ / refactoring / ~dependency upgrade~
- [x] Bug fix (non-breaking change which fixes an issue) *Access violations have been reported and there are hidden buffer overruns which may or may not be separate of those.*
- [x] New feature (non-breaking change which adds functionality) *The changes in types and inheritance hierarchy are binary breaking changs, but will be source-compatible.*
  - [x] Public API affected by the described changes have been marked obsolete and will now generate warnings.

## Did not make it into this PR
  - [x] Many formal declarations using `nint`, `IntPtr`, `void*` or other unmanaged pointers will be obsoleted in favor or overloads that consistently carry length information, such as `Memory<T>` and `Span<T>`.
  - [x] Use of arrays in signatures, fields and implementations may be (further) discouraged and obsoleted.

## Checklist

- [x] This work in progress has completed.
  - [x] Documentation comments have been modified and added.
- [x] All new and existing tests passed. *All tests succeed, except the following. 3 test projects fail to run due to test infrastructure failures. 1 test project fails all tests due to expecting packages for samples in `$env:USERPROFILE\.nuget\packages` which don't exist.*


xen2:
I was thinking, shouldn't we directly use the underlying method rather than writing yet another wrapper on it?

It would make the code more direct, standard & easy to understand, esp. for people familiar with recent high-perf C# (no need to check how our low-level `Utilities`/`CoreUtilities` map to yet another low-level `Unsafe` call).

That's what I initially had in mind when I saw all the new recently `Unsafe` equivalent to our `Utilities` class.

ericwj:
> I was thinking, shouldn't we directly use the underlying Unsafe methods rather than writing yet another wrapper on it?
Ideally yes but there are AV's, this is perhaps a halfway measure to have *some* checks (right now also but sometimes). Since Stride usually works, can also leave it for now and come back to the AV's later getting rid of `AllocHGlobal` and using a better, verifyable allocation strategy.

ericwj:
I had `Span<T>.CopyTo` in a fair few cases. But testing it the first time it was clear it wasn't going to work.

ericwj:
Checked it twice, don't see anything wrong - most changes are mundane. Yet I can't start GameStudio anymore - when it sais Loading scene opening Starbreach there is an AV.

ericwj:
I have updated the title and the description to reflect the current plan as discussed with @xen2.

@manio143 would you review please as I go?

sources/core/Stride.Core/Utilities.cs
```diff
@@ -717,88 +611,34 @@ public static bool Compare<T>(ICollection<T> left, ICollection<T> right)
...
         {
-            var ms = (long)sleepTime.TotalMilliseconds;
-            if (ms < 0 || ms > int.MaxValue)
-            {
-                throw new ArgumentOutOfRangeException(nameof(sleepTime), "Sleep time must be a duration less than '2^31 - 1' milliseconds.");
-            }
-            NativeInvoke.Sleep((int)ms);
+            if (duration.Ticks < 0)
+                throw new ArgumentOutOfRangeException(nameof(duration));
+            Thread.Sleep(duration);
```
manio143, line 632: Is Thread.Sleep behaving the same as the native one - I think it was calling a C function?

sources/engine/Stride.Engine/Updater/UpdateEngineHelper.cs
```diff
@@ -12,6 +13,83 @@ internal static unsafe class UpdateEngineHelper
...
+                func = isPinnable.AddOrUpdate(type, IsReferenceOrContainsReferences(type), KeepExisting);
+            if (func() && !isPinnableBroken.ContainsKey(type))
+            {
+                isPinnableBroken.AddOrUpdate(type, true, KeepExisting);
+            }
+            #endif
+            return GCHandle.Alloc(@object, GCHandleType.Pinned);
+        }
+        #if CHECKED
+        private static readonly MethodInfo pinMethod = typeof(RuntimeHelpers).GetMethod(nameof(RuntimeHelpers.IsReferenceOrContainsReferences), BindingFlags.Static | BindingFlags.Public);
```
manio143, line 41: `pinMethod` doesn't match `IsReferenceOrContainsReference` and can be confusing

manio143 (commented):
Looked through the first half of this PR (incl Stride.Graphics) - looks good overall

sources/core/Stride.Core/Serialization/MemberSerializerGenerated.cs
```diff
@@ -4,6 +4,7 @@
 // <auto-generated/>
 using System;
 using System.Reflection;
+using System.Runtime.CompilerServices;
```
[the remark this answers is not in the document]
ericwj: Oh that explains the compiler error after merging. Haven't had 1 second of attention to `.tt`'s so far.

sources/core/Stride.Core/Storage/ObjectIdBuilder.cs
```diff
@@ -324,8 +330,73 @@ public void Write(byte* buffer, int length)
...
+            // Partial block to continue?
+            if (position != 0)
+            {
+                var remainder = 16 - position;
+
+                var partialLength = length;
+                if (partialLength > remainder)
+                    partialLength = remainder;
+
+                ref var dest = ref Unsafe.Add(ref currentBlock, position);
```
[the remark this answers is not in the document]
ericwj: Its just the monkey porting the reference assignment although this will copy byte for byte and thats not smart.
```
var dest = currentBlock + position;
```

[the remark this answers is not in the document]
ericwj: If you see other archaic C++'isms, let's put a warning and punt fixing it.

sources/core/Stride.Core/Utilities.cs
ericwj: I see [it](https://github.com/sinkingsugar/NativePath/blob/master/source/NativeThreading.cpp) uses `usleep` or `std::this_thread::sleep_for(std::chrono::milliseconds(milliseconds));`.

`System.Threading.Thread.Sleep` uses `Kernel32!Sleep` [here](https://github.com/dotnet/runtime/blob/57bfe474518ab5b7cfe6bf7424a79ce3af9d6657/src/libraries/System.Private.CoreLib/src/System/Threading/Thread.Windows.cs#L19) or on Unix its much more [involved](https://github.com/dotnet/runtime/blob/9d6396deb02161f5ee47af72ccac52c2e1bae458/src/libraries/System.Private.CoreLib/src/System/Threading/WaitSubsystem.ThreadWaitInfo.Unix.cs#L391) but it'll be compatbile.

The biggest difference imho is that .NET accepts `Timeout.Infinite` which is a negative value, so that's why I didn't obsolete `Utilities.Sleep` although most likely that check is not practically necessary.

sources/core/Stride.Core/StrideCoreExtensions.cs
```diff
@@ -0,0 +1,51 @@
...
+    /// <list type="bullet">
+    /// <item><paramref name="first"/> and <paramref name="second"/> are the same object.</item>
+    /// <item>Neither enumerable is <c>null</c> and they have the same length and each of the elements in the enumerables compare equal pairwise.</item>
+    /// </list>
+    /// <para><c>false</c> otherwise.</para></returns>
+    public static bool SequenceEqualAllowNull<T>(this IEnumerable<T> first, IEnumerable<T> second, IEqualityComparer<T> comparer)
+    {
+        if (ReferenceEquals(first, second)) return true;
+        if (first is null || second is null) return false;
+        if (first is List<T> llist && second is List<T> rlist)
```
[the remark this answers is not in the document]
ericwj: Let the JIT do it.
https://source.dot.net/#System.Linq/System/Linq/SequenceEqual.cs,27

manio143:
Please note that Vulkan builds are failing - VS by default only builds for D3D11 and you may need to modify `sources\targets\Stride.props` and set `<StrideDefaultGraphicsApi>Vulkan</StrideDefaultGraphicsApi>` on line 21 to have VS build Vulkan config for you.

ericwj:
"Works on my machine."

ericwj:
I changed this:
```XML
  <!-- Default GraphicsApi -->
  <PropertyGroup>
    <StrideGraphicsApis Condition="('$(StrideGraphicsApis)' == '' Or '$(StrideGraphicsApiDependentBuildAll)' == 'true') And ('$(TargetFramework)' == '$(StrideFramework)' Or '$(TargetFramework)' == '$(StrideFrameworkWindows)')">Direct3D11;Direct3D12;OpenGL;OpenGLES;Vulkan</StrideGraphicsApis>

    <StrideDefaultGraphicsApi Condition="'$(StrideGraphicsApis)' != ''">$(StrideGraphicsApis.Split(';', StringSplitOptions.RemoveEmptyEntries)[0])</StrideDefaultGraphicsApi>
    <StrideDefaultGraphicsApi Condition="'$(TargetFramework)' == '$(StrideFrameworkUWP)'">Direct3D11</StrideDefaultGraphicsApi>
    <StrideDefaultGraphicsApi Condition="'$(TargetFramework)' == '$(StrideFrameworkAndroid)'">OpenGLES</StrideDefaultGraphicsApi>
    <StrideDefaultGraphicsApi Condition="'$(TargetFramework)' == '$(StrideFrameworkiOS)'">OpenGLES</StrideDefaultGraphicsApi>
    <StrideDefaultGraphicsApi>Vulkan</StrideDefaultGraphicsApi>
  </PropertyGroup>
```
Stride builds but Starbreach gives me this:
```
Rebuild started...
Failed to restore C:\Operations\LargeFolders\Starbreach\Starbreach\Starbreach.csproj (in 605 ms).
Failed to restore C:\Operations\LargeFolders\Starbreach\VFXPackage\VFXPackage.csproj (in 605 ms).
Failed to restore C:\Operations\LargeFolders\Starbreach\Starbreach.Windows\Starbreach.Windows.csproj (in 1.04 sec).
1>------ Rebuild All started: Project: VFXPackage, Configuration: Debug Any CPU ------
NuGet package restore failed. Please see Error List window for detailed warnings and errors.
1>C:\Operations\LargeFolders\Starbreach\VFXPackage\VFXPackage.csproj : error NU1202: Package Stride.VirtualReality 4.1.0.1 is not compatible with net6.0 (.NETCoreApp,Version=v6.0). Package Stride.VirtualReality 4.1.0.1 does not support any target frameworks.
1>C:\Operations\LargeFolders\Starbreach\VFXPackage\VFXPackage.csproj : error NU1202: Package Stride.Games 4.1.0.1 is not compatible with net6.0 (.NETCoreApp,Version=v6.0). Package Stride.Games 4.1.0.1 does not support any target frameworks.
1>C:\Operations\LargeFolders\Starbreach\VFXPackage\VFXPackage.csproj : error NU1202: Package Stride.Input 4.1.0.1 is not compatible with net6.0 (.NETCoreApp,Version=v6.0). Package Stride.Input 4.1.0.1 does not support any target frameworks.
1>Done building project "VFXPackage.csproj" -- FAILED.
2>------ Rebuild All started: Project: Starbreach, Configuration: Debug Any CPU ------
2>C:\Operations\LargeFolders\Starbreach\Starbreach\Starbreach.csproj : error NU1202: Package Stride.Video 4.1.0.1 is not compatible with net6.0 (.NETCoreApp,Version=v6.0). Package Stride.Video 4.1.0.1 does not support any target frameworks.
2>C:\Operations\LargeFolders\Starbreach\Starbreach\Starbreach.csproj : error NU1202: Package Stride.VirtualReality 4.1.0.1 is not compatible with net6.0 (.NETCoreApp,Version=v6.0). Package Stride.VirtualReality 4.1.0.1 does not support any target frameworks.
2>C:\Operations\LargeFolders\Starbreach\Starbreach\Starbreach.csproj : error NU1202: Package Stride.Games 4.1.0.1 is not compatible with net6.0 (.NETCoreApp,Version=v6.0). Package Stride.Games 4.1.0.1 does not support any target frameworks.
2>C:\Operations\LargeFolders\Starbreach\Starbreach\Starbreach.csproj : error NU1202: Package Stride.Input 4.1.0.1 is not compatible with net6.0 (.NETCoreApp,Version=v6.0). Package Stride.Input 4.1.0.1 does not support any target frameworks.
2>Done building project "Starbreach.csproj" -- FAILED.
3>------ Rebuild All started: Project: Starbreach.Windows, Configuration: Debug Any CPU ------
3>C:\Operations\LargeFolders\Starbreach\Starbreach.Windows\Starbreach.Windows.csproj : error NU1202: Package Stride.Video 4.1.0.1 is not compatible with net6.0-windows7.0 (.NETCoreApp,Version=v6.0). Package Stride.Video 4.1.0.1 does not support any target frameworks.
3>C:\Operations\LargeFolders\Starbreach\Starbreach.Windows\Starbreach.Windows.csproj : error NU1202: Package Stride.VirtualReality 4.1.0.1 is not compatible with net6.0-windows7.0 (.NETCoreApp,Version=v6.0). Package Stride.VirtualReality 4.1.0.1 does not support any target frameworks.
3>Done building project "Starbreach.Windows.csproj" -- FAILED.
========== Rebuild All: 0 succeeded, 3 failed, 0 skipped ==========
```

xen2:
I think you just want to add `<StrideGraphicsApiDependentBuildAll>true</StrideGraphicsApiDependentBuildAll>` to `build/Stride.Build.props`.
If you specify them manually, you want to use `StrideGraphicsApis` (with a `s`) and have both Direct3D11 (mandatory) and the one you want, separated by semi-colons.
Everything should happen in `build/Stride.Build.props` (ideally we should have a .user file also on .gitignore so that it's more clear it should be the one to edit).

ericwj:
Yes now I get proper errors. Obsiously I have used Find All References and not regex :wink:

Would be great to use multi-targeting for that so that stuff compiles regardless of host machine.

sources/engine/Stride.Graphics/Vulkan/PipelineState.Vulkan.cs
```diff
@@ -158,22 +157,29 @@ private unsafe void Recreate()
                 viewportCount = 1,
             };
 
-            fixed (VkDynamicState* dynamicStatesPointer = &dynamicStates[0])
+            // TODO: Verify `(array.Length == 0) == (fixed (T* f = array) f == null)`;
```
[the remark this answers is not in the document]
ericwj: Yes, its obvious many people don't know this. The conditionals in the Vulkan code are a prime example. Its often hidden because its accompanied by `array.Length` or similar, but it could be causing some of the access violations.

ericwj:
Grr

ericwj:
Could just as well bring up now that the .NET SDK is broken also wrt the code analysis. Besides importantly the [default defines](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives#conditional-compilation) not working.

It started just working when I was fiddling with the build system and skipping swaths of it. But the number of warnings is absolutely massive, making it almost useless for that reason. Still perhaps enabling some of these rules to catch some pervasive patterns I encounter would be a much better solution than using `#warning` and `[Obsolete(DiagnosticID = ...)]`.

sources/core/Stride.Core.IO/VirtualFileStream.cs
```diff
@@ -203,6 +203,17 @@ public override void Write(byte[] buffer, int offset, int count)
             InternalStream.Write(buffer, offset, count);
         }
 
+        /// <inheritdoc/>
+        public override void Write(ReadOnlySpan<byte> buffer)
+        {
+            if (endPosition != -1 && buffer.Length > endPosition - InternalStream.Position)
+            {
+                throw new NotSupportedException("Can't write beyond end of stream.");
```
manio143, line 211: This should be an `InvalidOperationException`.
> The InvalidOperationException is thrown when a method call is invalid for the object's current state.

> The NotSupportedException is thrown when an invoked method is not supported, or when there is an attempt to read, seek, or write to a stream that does not support the invoked functionality.

sources/core/Stride.Core/IO/NativeMemoryStream.cs
```diff
@@ -84,6 +98,20 @@ public override unsafe int Read(byte[] buffer, int offset, int count)
             return count;
         }
 
+        /// <inheritdoc/>
+        public override void Write(ReadOnlySpan<byte> buffer)
+        {
+            Debug.Assert(dataEnd >= dataCurrent);
+            var written = (nint)(dataEnd - dataCurrent);
+            if (written < buffer.Length)
+                throw new IOException();
```
manio143, line 107: You need to provide proper explanation in the exception message.

sources/core/Stride.Core/IO/NativeStream.cs
```diff
@@ -9,14 +9,17 @@ namespace Stride.Core.IO
     /// <summary>
     /// A <see cref="Stream"/> with additional methods for native read and write operations using <see cref="IntPtr"/>.
     /// </summary>
+    [Obsolete]
```
manio143, line 12: If this abstract class is made obsolete should all classes deriving from it be obsolete as well? I'm not sure if this is right.

sources/core/Stride.Core/Serialization/Binary/BinarySerialization.cs
```diff
@@ -11,6 +12,7 @@ namespace Stride.Core.Serialization
     /// <remarks>
     /// This class is a simple front end to <see cref="BinarySerializationReader"/> and <see cref="BinarySerializationWriter"/>.
     /// </remarks>
+    [Obsolete] // not used, inefficient
```
manio143, line 15: I'm pretty sure this is the class users would use in their code to use the serialization system - it should not be deprecated unless a better backwards compatible alternative is provided.

sources/core/Stride.Core.IO/VirtualFileStream.cs
ericwj: It is what `void Write(byte[] buffer, int offset, int count)` does. You want me to change them in both?

sources/core/Stride.Core/IO/NativeStream.cs
ericwj: That's why I asked. I will be removing it from the inheritance hierarchy shortly.

sources/core/Stride.Core/Serialization/Binary/BinarySerialization.cs
ericwj: It is silly to new up two objects in some two line static method which returns neither of them. Shouldn't they be requesting some kind of service from DI instead? Or they should be directed to just create the reader or writer themselves imho.

sources/core/Stride.Core.IO/VirtualFileStream.cs
manio143: Yes, please do. If I encountered this exception I'd be confused.

sources/core/Stride.Core/Serialization/Binary/BinarySerialization.cs
manio143: Ok, fair enough. Since the change isn't very complicated for the user conceptually let's go with your recommendation in the Obsolete attribute. Usually you just need a simple static [de]serialization method and this what provided here.

sources/core/Stride.Core.IO/VirtualFileStream.cs
ericwj: Then I make it `IOException`. `System.IO.Stream` doesn't throw `InvalidOperationException` except for:

`Stream.WriteAsync`/`Stream.WriteAsync`:
* [InvalidOperationException](https://docs.microsoft.com/en-us/dotnet/api/system.invalidoperationexception?view=net-6.0)
The stream is currently in use by a previous read/write operation.

Making that consistent would affect `VirtualFileStream`, `LZ4Stream` and the obsolete `NativeMemoryStream`.

sources/core/Stride.Core.Serialization/LZ4/LZ4Stream.cs
```diff
@@ -552,16 +520,13 @@ public override unsafe void Write(byte[] buffer, int offset, int count)
 			}
 		}
 
-        /// <summary>When overridden in a derived class, writes a sequence of bytes to the current stream and advances the current position within this stream by the number of bytes written.</summary>
-        /// <param name="buffer">An array of bytes. This method copies <paramref name="count" /> bytes from <paramref name="buffer" /> to the current stream.</param>
-        /// <param name="offset">The zero-based byte offset in <paramref name="buffer" /> at which to begin copying bytes to the current stream.</param>
-        /// <param name="count">The number of bytes to be written to the current stream.</param>
+	    /// <inheritdoc/>
```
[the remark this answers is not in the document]
ericwj: Yeah I saw that, sorry - must've hit the wrong key combo and not notice it. To my defence I don't think I have used *Format Selection* (`Ctrl+E`, `F`) before in my life.

sources/core/Stride.Core.IO/VirtualFileStream.cs
```diff
@@ -199,18 +203,20 @@ public override void SetLength(long value)
         {
             if (endPosition != -1)
             {
-                throw new InvalidOperationException("Can't resize VirtualFileStream if endPosition is not -1.");
+                throw new NotSupportedException("Can't resize VirtualFileStream if endPosition is not -1.");
```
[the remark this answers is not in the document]
ericwj: I just follow the [docs](https://docs.microsoft.com/en-us/dotnet/api/system.io.stream.setlength?view=net-6.0).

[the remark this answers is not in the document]
Basewq: [Docs don't agree with you in this case](https://docs.microsoft.com/en-us/dotnet/api/system.notsupportedexception?view=net-6.0)

> [InvalidOperationException](https://docs.microsoft.com/en-us/dotnet/api/system.invalidoperationexception?view=net-6.0)
This exception is thrown in scenarios in which it is generally sometimes possible for the object to perform the requested operation, and the object state determines whether the operation can be performed.

InvalidOperationException is the correct exception.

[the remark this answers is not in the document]
ericwj: I'm surprised to find reference to stream in that location in the docs. But yeah, the argument is that people shouldn't have to read docs on custom streams to figure out why it throws yet another exception.

ericwj:
Documented the changes in this PR in the description, since the intent is to merge and continue in a future PR.

sources/engine/Stride/Rendering/ParameterCollection.cs
```diff
@@ -21,6 +21,8 @@ namespace Stride.Rendering
     [DebuggerTypeProxy(typeof(ParameterCollection.DebugView))]
     public class ParameterCollection
     {
+        private const int Alignment = 16;
```
manio143, line 24: Can you add a comment on this line `// Important! Alignment must be a power of 2` just in case?

sources/engine/Stride/Rendering/ParameterCollection.cs
```diff
@@ -611,33 +628,50 @@ public unsafe void UpdateLayout(ParameterCollectionLayout collectionLayout)
...
+                        elementCount: oldParameterKeyInfo.Count);
+                    var oldSpan = DataValues.AsSpan(oldParameterKeyInfo.Offset, oldTotalSize);
+                    var newSpan = newDataValues.AsSpan(newParameterKeyInfo.Offset, newTotalSize);
+
+                    if (oldParameterKeyInfo.Key.Size == newParameterKeyInfo.Key.Size)
+                    {
+                        var minTotalSize = Math.Min(oldTotalSize, newTotalSize);
+                        oldSpan[..minTotalSize].CopyTo(newSpan);
+                        if (newTotalSize > oldTotalSize) newSpan[minTotalSize..].Fill(0);
+                    } else
```
manio143, line 650: nitpick: new line before `else`

manio143 (approved):

xen2:
I didn't have the time to go through every single change yet, but from what I saw so far it's a great PR!
It simplifies a lot of code and put it much more in line with recent C#! (esp. unsafe and lot of custom workaround we had)

xen2 (approved):

PR #1470, opened 2022-07-23
Make editing projects easier.

ericwj:
Make editing projects easier.

Guess what Visual Studio does to the poor people whose indentation is set to Tabs by default...

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -13,13 +13,16 @@ charset = utf-8
 insert_final_newline = false
 
 ##################################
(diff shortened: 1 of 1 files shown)


NicusorN5:
Tabs > Spaces

tebjan:
do they all use only two spaces?

ericwj:
The one `.pbxproj` uses tabs.

tebjan:
Great, thanks a lot!

PR #1315, opened 2022-02-02
[WIP] Custom Render Context

TheKeyblader:
# PR Details

Add missing extension point to support custom render context
- Custom Window
- Custom Platform

## Description

Mostly putting internal things in public

## Related Issue

#870

## Motivation and Context

These changes allow stride to be hosted in a custom window platform like:
- Headless
- Remote
- GLFW
- vvvv
- Avalonia

## Types of changes

- [ ] Docs change / refactoring / dependency upgrade
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [ ] My change requires a change to the documentation.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

Related Open Colletive Project
https://opencollective.com/stride3d/projects/embed-stride-ui-frameworks

Issue #870, opened 2020-10-07
Support for Custom render context


W-Sebastian:
I want to use Stride as a 3D renderer for a cross platform project where the render window is integrated in the Avalonia UI.
Now both Stride and Avalonia, theoretically, would support this quite nicely but the issue appears with extending the Stride engine with a new rendering context (this means custom GameContext, GamePlatform, GameForm and GameWindow) which would be based on Avalonia native controls (rather then, for WinForms, based on System.Windows.Forms.Control). This is not a complicated extension as it would be, more or less, equal in implementation with the current WinForms. 

The problem is that when using the Stride nuget packages (or the old Xenko ones), consumers of these packages are not allowed to provide custom instances to these classes because:
- GamePlatforms are created in a static internal method and based on an enumeration so there is no way to provide a custom implementation here;
- GamePlatform itself is marked as internal which means we couldn't inherit from it;
- GameWindow contains internal abstract methods which cannot be overloaded;
- There are various other properties or methods marked as internal that might be required in such an implementation but I don't currently have an exhaustive list for those.

Is there a design limitation for why a consumer of these assemblies couldn't provide custom classes for rendering concepts? From what I can tell, the rest of the code is neatly abstracted and should work regardless of what implementations we might have there.

Exposing the internal members outside the assembly and introducing a new overridable factory pattern for creating game platforms would allow one to fully implement these concepts in separate assemblies without having the need to modify the internal Stride code.

Of course, this would not have any impact on the GameEngine itself but would allow Xenko to be embedded in any kind of contexts.

Lastly, if there are no design limitations with my proposals or other issues I may not notice now, would a merge request be accepted? I can pretty easily fork the repo and maintain my own implementation of it but I feel it would be much better, if there are no downsides, to have it contained in the main repository.

Eideren:
There's no reason I know of for it to be closed off like it is right now, and if there is I'm sure we can fix or work around that. You can go ahead and make a pr when it's ready :)

tebjan:
That would be great! It has been asked frequently to integrate a 3d surface rendered by stride into Avalonia. In our fork, we have done some work regarding multiple windows, which also includes some changes to GamePlatform and GameWindow. You can look at our changes here: https://github.com/vvvv/stride/pull/7/files

It would be a very welcome change to open the entry point and be able to support more scenarios.

tebjan:
[...]

PR #1474, opened 2022-07-24
[WIP] Embedding Game in WPF and WinForms

juyanith:
Updates Window so that it will accept a parent HWND and create a child in which to render.

## Description

This is a Windows (OS) specific update. By passing in the parent window, using SDL CreateWindowFrom(), and some calls to native windows code I was able to host a Game inside WPF and WinForms. I used the Particles Sample project to create example implementations: ParticlesSample.Wpf and ParticlesSample.WinForms.

The WPF example makes use of GameEngineHost while the WinForms example doesn't need it. Both make use of SDL.Window which has been modifed to create child windows when passed a parent HWND. There is also an update to GameBase that prevents the PreferredBackBufferWidth and PreferredBackBufferHeight from overriding the Context RequestedWidth and RequestedHeight.

I'm not sure how you would go about doing something similar in Avalonia Given but how similar it is to WPF it may not be too difficult, although I suspect an equivalent class to GameEngineHost would need to be created.

## Related Issue

#870 
#1315 

## Motivation and Context

I would like to use Stride as a visualization tool inside an existing WPF application. 

## Types of changes

## Checklist


tebjan:
Nice one, I wonder whether we can introduce the parent window idea to the platform-independent game window class. So that it works for SDL, Winforms, and whatever windowing backend will be used. Or do you think the abstraction would creat too much work?

ericwj:
Would be absolutely awesome if the sample would consist of three projects, for WPF, WinUI3 and WinForms. For WinUI3 I would just try and start with a MAUI app and see if appropriate use of `#if` can actually yield a solution listed for WinUI3, but its probably fine to start with WinUI3 and leave the crossplat for later.

See [Retrieve a window handle (HWND) - Windows apps | Microsoft Docs](https://docs.microsoft.com/en-us/windows/apps/develop/ui-input/retrieve-hwnd)

juyanith:
I added a WinForms sample. It works, but something is off on user input. Specifically, the right side arrows don't respond to mouse clicks. It's also sluggish, but I figured it was worthwhile to show something partially working.

tebjan:
For Windforms there is another context, that also uses the winforms message loop. That might fix the issues. https://github.com/stride3d/stride/blob/master/sources/engine/Stride.Games/GameContextWinforms.cs

ericwj:
Its not sluggish on my machine. Maybe thats much better if you build in Release.
> For Windforms there is another context
Interesting though that right now this WinForms project is the only one not affected by the freezing while dragging.

juyanith:
> Interesting though that right now this WinForms project is the only one not affected by the freezing while dragging.

I experimented with this and running the WPF game loop in a task accomplished the same thing. It even shows the same behavior on clicking the right arrows. However, what is interesting is that after you resize the window that problem goes away. I tried forcing a refresh after starting the game loop but it didn't help. It is a clue though and may help figure out what is happening of the WinForms side as well.

SoggyBottomBoy:
Would this also address support for winui3?

ericwj:
Maybe change the title to include WinForms, Win32 (if that is what Particles.Windows is) and WinUI3 once its in?

juyanith:
I at least figured a workaround for the game not responding to mouse clicked until resized. Basically, the game did not have the correct values set for width and height of the render area. 

I'm still trying to work my way through the generic classes (e.g. GameWindow) to see if there is a way to encapsulate this in a platform independent way. I'm not having a great deal of luck however. That may be a task better suited to someone more familiar with Stride.

@SoggyBottomBoy I may try to add something for winui3. I'm not sure how much will have to be added since the WPF example makes use of the already existing GameEngineHost.

@ericwj I probably should adjust the title. I don't believe Particles.Windows was modified in this PR.

ericwj:
No indeed. The point would be to have people find this searching for Win32. But on second thought its only a temporary advantage; people won't normally find the PR once its merged. Particles should eventually be listed in the docs with these terms instead.

ericwj:
> I'm not sure how much will have to be added since the WPF example makes use of the already existing GameEngineHost.

Forget about WPF. Both WinUI3 and WPF have XAML but technically WPF is special. I think you should look at the Windows project to figure out how it'd work with WinUI3.

samples/Particles/ParticlesSample/ParticlesSample.Wpf/Win32GameControl.cs
```diff
@@ -0,0 +1,119 @@
...
+            if (host is null) return;
+            var wih = new WindowInteropHelper(host);
+            var hostHandle = wih.Handle;
+
+            // Create a child window in which to host the game
+            var className = GetType().Name;
+            var wndClass = new NativeMethods.WndClassEx();
+            wndClass.cbSize = (uint)Marshal.SizeOf(wndClass);
+            wndClass.hInstance = NativeMethods.GetModuleHandle(null);
+            wndClass.lpfnWndProc = NativeMethods.DefaultWindowProc;
```
ericwj, line 59: I think this causes the problems with resizing and dragging. It shouldn't be needed to run the rendering in a task, either.

Take a look at [this](https://github.com/terrafx/terrafx/blob/3e3fa918543db50fd681debddddcb555e838d2d9/sources/UI/UI/UIWindow.cs#L392) window procedure and then add to that a handler for `WM_PAINT` which *doesn't* use `BeginPaint`, `EndPaint`, but calls into the rendering instead. You can see how it is done in C++ in the DirectX samples [here](https://github.com/microsoft/DirectX-Graphics-Samples/blob/1704d8883ea74a73d332516bb6d990a134abaebd/Samples/Desktop/D3D12Fullscreen/src/Win32Application.cpp#L226).

The way to get `WndProc` to be called for a specific window instance is by forwarding like it is done [here](https://github.com/terrafx/terrafx/blob/3e3fa918543db50fd681debddddcb555e838d2d9/sources/UI/UI/UIService.cs#L144). Notice the `[UnmanagedCallersOnly]` and the use in there of `CREATESTRUCTW` to carry a `GCHandle` which targets the Window itself.

xen2:

PR #1482, opened 2022-07-26
[Rendering] Adds support for DrawAuto and ArgumentBuffer to MeshDraw

### LINKED ISSUE #1481: [Rendering] Add DrawAuto and ArgumentBuffer for mesh drawing
Generate geometry on GPU.

Allow `DrawAuto` and use `ArgumentBuffer` as draw call arguments.

Couldn't find an alternative.

**Additional context**
When generating geometry on the GPU the draw count is often not known on the CPU so the vertex count or instance count needs to be set by an ArgumentBuffer.

**Here are some questions that can aid in the description of the feature request:**
The class `MeshDraw` carries the draw call arguments for geometry. It could allow having a flag for `DrawAuto` and am `ArgumentBuffer` property.

tebjan:
# PR Details

Adds a `DrawAuto` flag and `ArgumentBufferBinding` to `MeshDraw`.

## Description

The new fields will be read by the `MeshRenderFeature` to select the right draw call method.

Questions:
I've just copied the IndexBufferBinding, does this even need a serializer? Also, does it need an asset version increase?
Is the args buffer even placed correctly there? Other ideas on how to solve this are welcome.

## Related Issue

Fixes #1481

## Motivation and Context

Generate geometry on the GPU.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Graphics/ArgumentBufferBinding.cs
+++ b/sources/engine/Stride.Graphics/ArgumentBufferBinding.cs
@@ -1,0 +1,41 @@
+// Copyright (c) Stride contributors (https://stride3d.net) and Silicon Studio Corp. (https://www.siliconstudio.co.jp)
+// Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
+using System;
+using Stride.Core.Serialization;
+using Stride.Core.Serialization.Serializers;
+
+namespace Stride.Graphics
+{
+    [DataSerializer(typeof(ArgumentBufferBinding.Serializer))]
+    public class ArgumentBufferBinding
+    {
+        public ArgumentBufferBinding(Buffer indexBuffer, int alignedByteOffset = 0)
+        {
+            if (indexBuffer == null) throw new ArgumentNullException("argmentBuffer");
+            Buffer = indexBuffer;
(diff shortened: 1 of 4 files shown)


johang88:
Feels like it could be ignored by serialization as it's runtime only? Perhaps `DrawIndirect` instead of `DrawAuto`?

tebjan:
> Feels like it could be ignored by serialization as it's runtime only?
You are right, it is a runtime technique. So then it needs the `[DataMemberIgnore]` attribute, is that correct?

> Perhaps `DrawIndirect` instead of `DrawAuto`?
Yes, sorry didn't specify that correctly. The way I thought about it is that [`DrawAuto`](https://docs.microsoft.com/en-us/windows/win32/api/d3d11/nf-d3d11-id3d11devicecontext-drawauto) uses a vertex buffer from the stream out stage of another shader and with `ArgumentBuffer` I meant [`DrawIndirect`](https://docs.microsoft.com/en-us/windows/win32/api/d3d11/nf-d3d11-id3d11devicecontext-drawinstancedindirect), which takes the draw args like vertex count, instance count, and vertex buffer offsets as a buffer. I've corrected the title.

PR #1499, opened 2022-08-08
Added overload for Entity constructor

palkercode:
# PR Details
Added overload for Entity constructor.

## Description

Now you can create entities with rotation parameter

## Motivation and Context

In my game with procedural generation, many prefabs of rooms are created. I wanted them to randomly rotate, but I did not find a parameter with rotation.

## Types of changes

## Checklist


tebjan:
Nice one. LGTM, no objections from my side.

tebjan (dismissed):
Nice one. LGTM, no objections from my side.

manio143:
Hmm, I don't know if that's the best move here - the next person will want a position and a scale and another all three - so maybe there's a way to just have one constructor with optional parameters? We could break binary compatibility here and modify the constructor that currently takes only position.

My main issue with it is how easy it is for the user to write a wrapper method that would handle setting the rotation on an entity without modifying the source code.

VaclavElias:
Maybe we could have an experimental/preview quality of life extension(s), and if we knew that people are using some of them heavily and frequently, and it would be highly requested (we could vote after certain period), then it could move to the core.. 

But for this, would be great to have some docs so these are listed and visible as experimental?

tebjan:
> position and a scale and another all three

Good point, it should also have the scale parameter as an optional parameter set to 1. then you got one with position and one with all three. that should do...

VoidEssy (approved):
Overloads are straight forward and simple so looks like a fine solution to me.

Eideren:
In general I would be against such a change but for this particular case I think it makes sense, setting an initial rotation for an entity is a fairly common operation, I do think that this overload is worth the very small amount of bloat it introduces.

We could repurpose the second constructor for this though;
```cs
public Entity(string name)
    : this(Vector3.Zero, name)
{
}
```
to this
```cs
public Entity(string name = null, Vector3 position = default, Quaternion? rotation = null, Vector3? scale = null)
    : this(name, false)
{
    Id = Guid.NewGuid();
    TransformValue = new TransformComponent { Position = position, Rotation = rotation ?? Quaternion.Identity, Scale = scale ?? Vector3.One };
    Components.Add(TransformValue);
}
```

Users would then pick and choose what they want with something like the following
```cs
new Entity(rotation:Quaternion.RotationY(180f));
new Entity("Some Entity", scale:Vector3.One * 2f);
new Entity("Some Entity", new Vector3(10f, 10f, 10f), Quaternion.RotationY(180f));
```

VaclavElias:
I have got a few side questions (probably missing in our docs): 

- Is it a good practise to name the entities? 
- When you would name them?
- When you wouldn't name them? 

I would use your input for our docs. Thanks :)

Eideren:
Hey @VaclavElias, this is a bit off topic, let's go over that on discord. I'll hide your message to avoid derailing the conversation.

manio143:
@palkercode would be up for updating this PR with Eideren's suggestion in https://github.com/stride3d/stride/pull/1499#issuecomment-1223764151 ?

PR #1515, opened 2022-08-25
DebugTextSystem configurable time on-screen

insomnyawolf:
# PR Details

## Description

Changes a queue into a list so it can be iterated.

Added variable to hold the default time on screen with an arbitrary value of 1 second.

Added methods to expose the configurable times to the user.

Removed clearing the message queue on every update.

Deleted duplicated if check.

Changed foreach into a for loop that can remove entries after they "expire".

## Related Issue

None that i know

## Motivation and Context

Debug text dissapear too fast to be useful by itself in a lot of cases

## Types of changes

The changes made on the code allows the user set how much do they want the text to stay on screen in each message and the default value when it's not specified.

## Checklist

### DIFF
--- a/sources/engine/Stride.Engine/Profiling/DebugTextSystem.cs
+++ b/sources/engine/Stride.Engine/Profiling/DebugTextSystem.cs
@@ -17,10 +17,11 @@ namespace Stride.Profiling
             public string Message;
             public Int2 Position;
             public Color4 TextColor;
+            public TimeSpan RemeaningTime;
         }
         
         private FastTextRenderer fastTextRenderer;
-        private readonly Queue<DebugOverlayMessage> overlayMessages = new Queue<DebugOverlayMessage>();
+        private readonly List<DebugOverlayMessage> overlayMessages = new List<DebugOverlayMessage>();
 
         public DebugTextSystem(IServiceRegistry registry) : base(registry)
         {
@@ -49,13 +50,36 @@ namespace Stride.Profiling
         /// <param name="color"></param>
         public void Print(string message, Int2 position, Color4 color)
         {
-            var msg = new DebugOverlayMessage { Message = message, Position = position, TextColor = color };
-            overlayMessages.Enqueue(msg);
+            Print(message, position, color, TimeSpan.Zero);
+        }
+
+        /// <summary>
+        /// Print a custom overlay message
+        /// </summary>
+        /// <param name="message"></param>
+        /// <param name="position"></param>
+        /// <param name="timeOnScreen"></param>
+        public void Print(string message, Int2 position, TimeSpan timeOnScreen)
+        {
+            Print(message, position, TextColor, timeOnScreen);
+        }
+
+        /// <summary>
+        /// Print a custom overlay message
+        /// </summary>
+        /// <param name="message"></param>
+        /// <param name="position"></param>
+        /// <param name="color"></param>
+        /// <param name="timeOnScreen"></param>
+        public void Print(string message, Int2 position, Color4 color, TimeSpan timeOnScreen)
+        {
+            var msg = new DebugOverlayMessage { Message = message, Position = position, TextColor = color, RemeaningTime = timeOnScreen };
+            overlayMessages.Add(msg);
 
             //drop one old message if the tail size has been reached
             if (overlayMessages.Count > TailSize)
             {
-                overlayMessages.Dequeue();
+                overlayMessages.RemoveAt(0);
             }
         }
 
@@ -71,7 +95,8 @@ namespace Stride.Profiling
 
         public override void Update(GameTime gameTime)
         {
-            overlayMessages.Clear();
+            // if you keep this, the timmers would not work
+            //overlayMessages.Clear();
         }
 
         public override void Draw(GameTime gameTime)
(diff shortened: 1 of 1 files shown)


sources/engine/Stride.Engine/Profiling/DebugTextSystem.cs
```diff
@@ -64,14 +88,20 @@ public void Print(string message, Int2 position, Color4 color)
         /// </summary>
         public Color4 TextColor { get; set; } = Color.LightGreen;
 
+        /// <summary>
+        /// Sets or gets the time that messages will stay on screen by default.
+        /// </summary>
+        public TimeSpan DefaultOnScreenTime { get; set; } = TimeSpan.FromSeconds(1);
```
Eideren, line 94: I think this should be set to zero to behave in the same way that it was before for messages without any time specified

sources/engine/Stride.Engine/Profiling/DebugTextSystem.cs
```diff
@@ -64,14 +88,20 @@ public void Print(string message, Int2 position, Color4 color)
...
         /// <summary>
         /// Sets or gets the size of the messages queue, older messages will be discarded if the size is greater.
         /// </summary>
         public int TailSize { get; set; } = 100;
 
         public override void Update(GameTime gameTime)
         {
-            overlayMessages.Clear();
+            // if you keep this, the timmers would not work
+            //overlayMessages.Clear();
```
Eideren, line 104: Feel free to remove this given that it's not applicable anymore

sources/engine/Stride.Engine/Profiling/DebugTextSystem.cs
```diff
@@ -90,16 +120,15 @@ public override void Draw(GameTime gameTime)
                 };
             }
 
-            if (overlayMessages.Count == 0)
-                return;
-
```
Eideren, line 95: I think you can still keep this, if you don't have any message scheduled you don't need the rest of your logic to run ?

sources/engine/Stride.Engine/Profiling/DebugTextSystem.cs
```diff
@@ -17,10 +17,11 @@ internal struct DebugOverlayMessage
             public string Message;
             public Int2 Position;
             public Color4 TextColor;
+            public TimeSpan RemeaningTime;
```
Eideren, line 20: Typo, should be `RemainingTime` :)

sources/engine/Stride.Engine/Profiling/DebugTextSystem.cs
```diff
@@ -49,13 +50,36 @@ public void Print(string message, Int2 position)
...
+            var msg = new DebugOverlayMessage { Message = message, Position = position, TextColor = color, RemeaningTime = timeOnScreen };
+            overlayMessages.Add(msg);
 
             //drop one old message if the tail size has been reached
             if (overlayMessages.Count > TailSize)
             {
-                overlayMessages.Dequeue();
+                overlayMessages.RemoveAt(0);
             }
         }
```
Eideren, line 84: The amount of signatures is getting a bit out of hand, how about we remove all of them and add this one in instead:
```cs
public void Print(string message, Int2 position, Color4? color = null, TimeSpan? timeOnScreen = null)
{
    var msg = new DebugOverlayMessage { Message = message, Position = position, TextColor = color ?? TextColor, RemeaningTime = timeOnScreen ?? DefaultOnScreenTime };
    overlayMessages.Add(msg);
...
```
With that one you could call it in the following ways:
```cs
Print("msg", pos, timeOnScreen: TimeSpan.FromSeconds(1));
Print("msg", pos, Color.LightGreen, TimeSpan.FromSeconds(1));
Print("msg", pos, null, TimeSpan.FromSeconds(1));
Print("msg", pos);
```

Eideren (commented):
Hey, thanks a bunch for taking the time to go through this, there's a couple of things I want to discuss before merging this in if you don't mind

sources/engine/Stride.Engine/Profiling/DebugTextSystem.cs
Eideren: Never mind me, this is fine !

sources/engine/Stride.Engine/Profiling/DebugTextSystem.cs
insomnyawolf: Makes sense, i edited it to be TimeSpan.Zero

sources/engine/Stride.Engine/Profiling/DebugTextSystem.cs
insomnyawolf: Removed, shall we keep the Update method even if it's empty anyways?

Eideren: Nope, you can remove it too

sources/engine/Stride.Engine/Profiling/DebugTextSystem.cs
insomnyawolf: Sorry about it, and thank you for pointing it out to me so i can fix it.

sources/engine/Stride.Engine/Profiling/DebugTextSystem.cs
insomnyawolf: On the beggining i tried to follow what was already there but doing it that way you can do the same things with less code.
Also you can also use named parameters and just ommit the ones that you don't care about too
```cs
Print(message: "test", position: new Int2(), color: null, timeOnScreen: null);
Print(message: "test", position: new Int2(), timeOnScreen: null);
Print(message: "test", position: new Int2(), color: null);
Print(message: "test", position: new Int2());
```

insomnyawolf:
I've changed what you pointed me out, it all made sense, i also decided to add a brief comment about why the loop is done the way it is, just in case it can be helpful to someone in the future.

Eideren:
Thanks !

Issue #158, opened 2018-08-26
Drawing grids in different planes (GridXY | GridXZ | GridYZ)


nikserdev:
Now the grid is drawn only in one plane XZ. 
I suggest doing three checkboxes:

GridXY | GridXZ | GridYZ

PR #1516, opened 2022-08-26
Enable multiple selections of the axes of grids to be drawn

garychia:
# PR Details

Originally, the grid could be only drawn in one plane (X, Y or Z).
This PR enables the grids to be drawn in multiple planes.

image
image
image
image

## Related Issue
#158

## Types of changes

## Checklist

### DIFF
--- a/sources/editor/Stride.Assets.Presentation/AssetEditors/GameEditor/ViewModels/EditorGridViewModel.cs
+++ b/sources/editor/Stride.Assets.Presentation/AssetEditors/GameEditor/ViewModels/EditorGridViewModel.cs
@@ -10,6 +10,14 @@ using Stride.Core.Presentation.ViewModel;
 
 namespace Stride.Assets.Presentation.AssetEditors.GameEditor.ViewModels
 {
+    [Flags]
+    public enum ViewportGridAxis
+    {
+        X = 1,
+        Y = 2,
+        Z = 4,
+    }
+
     /// <summary>
     /// A view model controlling a grid in an editor game.
     /// </summary>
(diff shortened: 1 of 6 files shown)


Eideren:
How about changing this to toggle grids for each axis instead of a radio button for each combination of axis ?
i.e.:
- [ ] x
- [x] y
- [x] z
Instead of 
- [ ] x
- [ ] y
- [ ] z
- [ ] xy
- [ ] xz
- [x] yz

garychia:
That's a good idea. I'll give it a try.

sources/editor/Stride.Assets.Presentation/AssetEditors/GameEditor/ViewModels/EditorGridViewModel.cs
```diff
@@ -50,19 +50,67 @@ public EditorGridViewModel([NotNull] IViewModelServiceProvider serviceProvider,
...
+            return 7;
+        }
+
         public void LoadSettings([NotNull] SceneSettingsData sceneSettings)
         {
             IsVisible = sceneSettings.GridVisible;
             Color = sceneSettings.GridColor;
             Alpha = sceneSettings.GridOpacity;
             AxisIndex = sceneSettings.GridAxisIndex;
+            AxisSelection = ConvertAxisIndexToSelection(sceneSettings.GridAxisIndex);
```
manio143, line 113: Is the `AxisIndex` used for anything else? If not, then I would suggest replacing it completely with the `AxisSelection` and passing that to the ViewportGridGizmo. Moreover, the bit mask can be encoded in an enum:
```
[Flags]
public enum ViewportGridAxis
{
    X = 1,
    Y = 2,
    Z = 4,
}
```
Which allows you to do combined enum values like `ViewportGridAxis.X & ViewportGridAxis.Y`.

sources/editor/Stride.Assets.Presentation/AssetEditors/Gizmos/ViewportGridGizmo.cs
```diff
@@ -167,14 +176,14 @@ private Material CreateColoredTextureMaterial(Texture texture, float textureScal
             });
         }
 
-        protected override void UpdateBase(Color3 gridColor, float alpha, int gridAxisIndex, float sceneUnit)
+        private int UpdateGrid(Color3 gridColor, float alpha, int gridAxisIndex, float sceneUnit, int gridIndex)
```
manio143, line 179: Why are you not reusing `gridAxisIndex` for `gridIndex`?

manio143 (commented):
Sorry for such a late review. Got a couple comments, because currently the code feels a little over complicated, but otherwise good job!

sources/editor/Stride.Assets.Presentation/AssetEditors/GameEditor/ViewModels/EditorGridViewModel.cs
garychia: You are right. Thanks for your recommendation!

sources/editor/Stride.Assets.Presentation/AssetEditors/Gizmos/ViewportGridGizmo.cs
garychia: I'll try to simplify "AxisSelection" first and then refactor this part later.

manio143:
Ok, just to be clear - is the `originAxes` array about the colored line going along the unit axis on origin point? So in the current behaviour the flat grid has both X and Z axis. I would expect each singleton grid to have 2 axis marked, and any 2+ combination of grids to have all the axis marked. Looking at your screenshots that is not the case. Am I getting this right?

garychia:
Actually I was not sure how origin axes should be marked in this case. My understanding is that, for example, if you select X grid axis, all the origin axes excluding X axis will be marked. So in the UpdateOriginAxes function, I simply disable the origin axes selected as the grid axes, which results in the behavior the screenshots just show.

manio143:
I got to compile your code and first thing I noticed was that the default behavior has changed
image
Which will be unexpected for the users. I'll let you figure it out, while I looked at modifying the axis behavior in a PR onto your branch: https://github.com/garychia/stride/pull/1

garychia:
It seems that I forgot to change the default scene setting. Thanks for pointing it out. I'll try to solve it.

PR #1521, opened 2022-09-10
Add build support to referencing projects that use SemVer 2.0.0 format

Basewq:
# PR Details

http://semver.org now has Semantic Versioning 2.0.0 and if you reference a csproj that contains something like `<PackageVersion>1.0.0-rc.3</PackageVersion>` (a valid SemVer 2.0.0 format), Stride's Asset compiler will error and you can't build your project.

## Description

Expand regex for supporting `.` within "pre-release version", and ignore "build metadata".
Also note that the regex pattern for the "pre-release version" has expanded to allow alphanumeric instead of just an alphabetic char (as allowed by SemVer).
As per the documentation "Build metadata" has no ordering precedence, so can be safely discarded.

Tests have been added to ensure nothing is broken (hopefully!).

## Types of changes

## Checklist

### DIFF
--- a/sources/core/Stride.Core.Design.Tests/TestPackageVersion.cs
+++ b/sources/core/Stride.Core.Design.Tests/TestPackageVersion.cs
@@ -1,0 +1,41 @@
+// Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net) and Silicon Studio Corp. (https://www.siliconstudio.co.jp)
+// Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
+using System.Collections.Generic;
+
+using Xunit;
+using Stride.Core;
+
+namespace Stride.Core.Design.Tests
+{
+    public class TestPackageVersion
+    {
+        [Theory]
+        [InlineData("1.0.0")]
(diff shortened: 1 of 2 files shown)


manio143 (approved):

sources/core/Stride.Core.Design.Tests/TestPackageVersion.cs
```diff
@@ -0,0 +1,65 @@
...
+using Stride.Core;
+
+namespace Stride.Core.Design.Tests
+{
+    public class TestPackageVersion
+    {
+        [Fact]
+        public void TestParseVersionsByValidVersionStrings()
+        {
+            var versionTests = new VersionTest[]
```
ericwj, line 15: You can change this `[Fact]` to `[Theory]`, remove the `VersionTest` structure and instead have the test accept the four values as arguments, then search replace each `new(...)` with `[InlineData(...)]` and put them under `[Theory]`. This greatly reduces the amount of code and allows running individual tests from Test Explorer.

Basewq: Thanks for the advice, have changed to this style.

Issue #1394, opened 2022-03-29
Getting the stride dependencies under control


IceReaper:
Building stride is somewhat a pain in itself. A lot of dependencies are pre-compiled. Others are build-in and hugely outdated. And again others have proper replacements. Sometimes you need to find obscure old natives, and find out how to build those for the particular platform and cpu architecture. It would be nice to get this a bit under control, by moving to nuget based dependencies where possible and ensuring stride itself comes without any native (besides the ones the nuget packages provide).

Benefits i see here:
- Getting dependencies under control
- No need to struggle with any msbuild / make / gcc / whatever stuff to get the engine build
- Allow to build the engine (without gui ofcourse) on non-windows systems
- Allow to build the whole stack without having visual studio installed
- Easier upgrading of dependencies and implementation of features
- Making stride itself pure c#

### libCore (Stride.Core)

This natives basicaly contains two things:
- A thread sleep command. Can be replaced with .NET its own Thread.Sleep
. An LZ4 implementation. Can be replaced with https://www.nuget.org/packages/K4os.Compression.LZ4/ (if stide needs lz4 after all! c# has some buildin compression algorythms)

In my eyes this whole native can be replaced by c# stuff?

### libStrideAudio (Stride.Audio)

This native basicaly wraps 4 different audio systems:
- Celt. Can be replaced with https://www.nuget.org/packages/Concentus/
- OpenAL. Can be replaced with https://www.nuget.org/packages/Silk.NET.OpenAL/
- OpenSLES
- XAudio2. Can be replaced with https://www.nuget.org/packages/Silk.NET.XAudio/

Technically OpenAL should be one-to-rule-them-all?
Alternatively having XAudio for the Windows world is a nice addition, but this means maintaining two audio systems.
Are Celt and OpenSL ES realy required and do they provide any benefit on supporting even more audio libraries?

### libStride (Stride.Native)

This native wraps the following components:
- XInput. Can be replaced with https://www.nuget.org/packages/Silk.NET.XInput/
- FastTextRenderer. This one could be ported to c#
- Lightprobes. This one is complex. Leaving out for the moment, need to check what parts are used.
- SpriteBatchNative. This one could be ported to c#
- StrideNative. This one can be replaced with the c# buildin intrinsics

### libStrideNavigation (Stride.Navigation)

This one is basicaly the navmesh builder and pathfinder.
While being dead, someone already started a managed version: https://www.nuget.org/packages/SharpNav/1.0.0-alpha2
[...]

PR #1523, opened 2022-09-11
Replace native ThreadSleep with .NET Thread.Sleep

Jklawreszuk:
# PR Details
## Description

This PR removes native ThreadSleep function and replaces with .NET Thread.Sleep

## Related Issue
#1394 

## Types of changes

## Checklist

### DIFF
--- a/sources/core/Stride.Core.Tests/TestProfiler.cs
+++ b/sources/core/Stride.Core.Tests/TestProfiler.cs
@@ -6,6 +6,7 @@ using System.Text;
 using System.Text.RegularExpressions;
 using Xunit;
 using Stride.Core.Diagnostics;
+using System.Threading;
 
 namespace Stride.Core.Tests
 {
@@ -23,7 +24,7 @@ namespace Stride.Core.Tests
             {
                 using (var profile = Profiler.Begin(TestKey))
                 {
-                    Utilities.Sleep(100);
+                    Thread.Sleep(100);
                 }
             }
             watcher.Finish();
@@ -44,7 +45,7 @@ namespace Stride.Core.Tests
                 Profiler.Enable(TestKey);
                 using (var profile = Profiler.Begin(TestKey))
                 {
-                    Utilities.Sleep(timeToWait);
+                    Thread.Sleep(timeToWait);
                 }
             }
             watcher.Finish();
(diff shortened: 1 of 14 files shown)


tebjan:
Thanks for the PR. Did you check whether the .NET `Thread.Sleep` also works on Android and iOS in a standard way?

At the top of the `NativeInvoke.cs` file there is this `IFDEF`:

```csharp
#if STRIDE_PLATFORM_IOS
        internal const string Library = "__Internal";
        internal const string LibraryName = "libcore.so";
#else
        internal const string Library = "libcore";
#if STRIDE_PLATFORM_WINDOWS
        internal const string LibraryName = "libcore.dll";
#else
        internal const string LibraryName = "libcore.so";
#endif
#endif
```

Jklawreszuk:
Hi, I tested it only on Windows. Sadly, I'm currently have no opportunity to test mobile platforms at the moment.

ericwj:
There is already [this](https://github.com/stride3d/stride/pull/1469/files#diff-b280b1e408a1e68679d668713d8f036ce7a81c722987e747b035f5a986f30e9bR637) which changes `Utilities` to use `Thread.Sleep`.

I also checked the implementation of `Thread.Sleep` on `source.dot.net` and you can find comments about that in that PR in [this](https://github.com/stride3d/stride/pull/1469#discussion_r933811126) comment.

Jklawreszuk:
Looks cool, I didn't notice these changes.Close then?

ericwj:
I think this PR does a nice cleanup so I would say no keep it.

It is preferable to have `Thread.Sleep` in code, which my PR doesn't change. Also I don't think I have actually removed any of the native code which effectively became dead code, which you did here.

The only thing is there might be occurrences of `Utilities.Sleep` that *might* receive negative values depending on the code surrounding the call if the value is not constant. You should probably verify the call sites to rule that out.

Jklawreszuk:
Alright, So I decided to bring back the Sleep method and rename it Usleep, since it no longer accepts negative values. Logic is very inspired to the one in eric's pr (hope you don't mind :D ). But I found that it is enough to check that the specified value is equal to -1. Otherwise it will throw an ArgumentException anyway.

ericwj:
This is C# not C++. `Usleep` is a horrible name.

Jklawreszuk:
Ok,I've undone changes.

Jklawreszuk:
Conflicts were resolved. Can my changes be merged? @manio143

PR #1532, opened 2022-09-29
Fix rotation yaw/pitch/roll decompose calculation bugs on edge cases

### LINKED ISSUE #174: Entity Property Rotation Value Bug
**Version**: 3.0.0.4

**Describe the bug**
The rotation values have a rounding issue. 
Set 90 degree and see 89.98 degree after reselect the entity.
This behaves on different values.

**To Reproduce**
1. Create a entity (e.g. ribbon particle)
2. Change the transform rotation to X: -90 deg.
3. Deselect the entity.
4. Select the same entity again.
5. See the X rotation value: -89.98 deg.

**Expected behavior**
The entered value should not be modified. -90 should be -90.
Maybe just a display bug. But anyway ...

**Screenshots**
bug

**Additional context**
Windows 10 64-bit

### LINKED ISSUE #1528: Matrix.Decompose(yaw/pitch/roll) method not always correct

**Version**: 4.1.0.1734 (and any previous)

**Platform(s)**: Windows

**Describe the bug**
Rotation matrix won't always output correct yaw/pitch/roll values on certain angles from a Quaternion for the method `Decompose(out float yaw, out float pitch, out float roll)`
[This method is used by Game Studio in the entity transform Rotation property](https://github.com/stride3d/stride/blob/7e836297cb5930c01e6dfa0183e7f3cc64748fb6/sources/presentation/Stride.Core.Presentation/Controls/RotationEditor.cs) which is why I noticed it.

**To Reproduce**
This can be seen in the Editor:
1. Select an entity
2. Change Transform.Rotation to (90, 0, 90)
3. Select another entity
4. Select previous entity and review the Transform.Rotation (which is now shows (89.972, 0, 0))

Alternatively, code to reproduce this issue:
```
var yaw = MathUtil.DegreesToRadians(90);
var pitch = 0f;
var roll = MathUtil.DegreesToRadians(90);
var quat = Quaternion.RotationYawPitchRoll(yaw, pitch, roll);

var mat = Matrix.RotationQuaternion(quat);
mat.Decompose(out float decomposedYaw, out float decomposedPitch, out float decomposedRoll);

Assert.Equal(yaw, decomposedYaw);    // Technically fails due to precision loss
Assert.Equal(pitch, decomposedPitch);
Assert.Equal(roll, decomposedRoll);  // Not matching (returns 0 instead of Pi/2)
```

**Expected behavior**
Decompose should output the same as input (or close enough since precision loss may occur).

**Additional context**
Note that `Quaternion.RotationYawPitchRoll` and `Matrix.RotationQuaternion` are probably not the issue (except for the low precision due to being floats), as they are fairly standard code (I've looked around online). 
The relevant method is this:
https://github.com/stride3d/stride/blob/ca3dc903f72d586c95311049bdc8c16d571ac718/sources/core/Stride.Core.Mathematics/Matrix.cs#L630-L643

I have no idea how this method actually works and does not appear to be some standard solution, I suspect the code is not correctly handling certain edge case values, which may be exacerbated due to the precision loss from using floats (eg. `MathUtil.DegreesToRadians(90)` involves Pi).

Basewq:
# PR Details

Fixes the edge cases of the yaw/pitch/roll method variants of `Matrix.Decompose` & `Quaternion.RotationYawPitchRoll` and `Matrix.DecomposeXYZ` where pitch (or yaw) is 90 degrees, but not being handled by the correct edge case logic.

Also fixed a minor bug in the editor where rotation X displays `-0` instead of `0`.

## Description

Had to fish through google to try and track down what formulas are being used.
The closest I found was ['Euler Angle Formulas' by David Eberly](https://www.geometrictools.com/Documentation/EulerAngles.pdf)
The existing code checked for the +/- 1 edge cases using Asin and Cos, however the loss of precision in operating on Pi/2 when constructing the rotation matrix meant it'd never hit Cos(Angle) == 0, even with ZeroTolerance check. According to Eberly's pdf, just checking `M32` (or `M13` for DecomposeXYZ) directly is enough (and this won't degrade by going through Asin then Cos like the existing code).
Also +1 and -1 changes signs, which the current code doesn't handle (it incorrectly treated both cases as the +1 case).

For `Matrix.Decompose`, the matrix fields used for the common case & edge cases were consistent with Eberly's formula (except the -1 case)

For `Matrix.DecomposeXYZ`, the matrix fields used for the common case was consistent with Eberly's, but the edge case used different fields, ie.
Existing: `rotation.Z = MathF.Atan2(-M21, M31);`
New: `rotation.Z = MathF.Atan2(-M32, M22);`

Unfortunately I don't know where those fields came from, so just changed it to match Eberly's version.

The unit tests confirms the new code is working (though that assumes the tests themselves are correct...)

Not sure if `RotationCurveViewModel` is actually used anywhere. I've changed back it to use `Matrix.DecomposeXYZ`, as I suspect the 'better precision' version actually isn't better because the problem already starts with the incoming matrix.

## Related Issue

Fixes #1528
Partial fix to #174 (to some extent)

## Other Notes
Even though this also fixes the immediate problem of #174 (since it'll detect 90 degrees better now), there is still a potential for loss of precision because we're dealing with floats (probably not really solvable).
Also a hidden 'issue' is that we don't save the user's XYZ rotation value, so it is possible the editor may reload the rotation (orientation) with different XYZ values (because multiple XYZ combinations can achieve the same final rotation, eg. `X: 90, Y: 90, Z: 0` is the same orientation as `X: 90, Y: 0, Z: -90`) - this is beyond the scope of this pull.

---

Please advise if the attribution to Eberly's pdf & license needs to be changed, as I'm not sure of the right way to do it.

Also, the tests generates ~3800 tests each (due to trying to test many different combinations of yaw/pitch/roll). Not sure if that's an issue or if there's a better way to code this (though there's no performance problem since they're all simple calculations).

---

## Types of changes

## Checklist

### DIFF
--- a/sources/core/Stride.Core.Mathematics.Tests/TestMatrix.cs
+++ b/sources/core/Stride.Core.Mathematics.Tests/TestMatrix.cs
@@ -1,0 +1,79 @@
+// Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net)
+// Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
+using Xunit;
+
+namespace Stride.Core.Mathematics.Tests
+{
+    public class TestMatrix
+    {
+        /* Note: As seen in the TestCompose* tests, we check both expectedQuat == decompedQuat and expectedQuat == -decompedQuat
+         * This is because different combinations of yaw/pitch/roll can result in the same *orientation*, which is what we're actually testing.
+         * This means that decomposing a rotation matrix or quaternion can actually have multiple answers, but we arbitrarily pick
+         * one result, and this may not have actually been the original yaw/pitch/roll the user chose.
+         */
+
+        [Theory, ClassData(typeof(TestRotationsData.YRPTestData))]
+        public void TestDecomposeYawPitchRollFromQuaternionYPR(float yawDegrees, float pitchDegrees, float rollDegrees)
+        {
+            var yawRadians = MathUtil.DegreesToRadians(yawDegrees);
+            var pitchRadians = MathUtil.DegreesToRadians(pitchDegrees);
+            var rollRadians = MathUtil.DegreesToRadians(rollDegrees);
+
+            var rotQuat = Quaternion.RotationYawPitchRoll(yawRadians, pitchRadians, rollRadians);
+            var rotMatrix = Matrix.RotationQuaternion(rotQuat);
+            rotMatrix.Decompose(out float decomposedYaw, out float decomposedPitch, out float decomposedRoll);
+
+            var expectedQuat = rotQuat;
+            var decompedQuat = Quaternion.RotationYawPitchRoll(decomposedYaw, decomposedPitch, decomposedRoll);
+            Assert.True(expectedQuat == decompedQuat || expectedQuat == -decompedQuat, $"Quat not equals: Expected: {expectedQuat} - Actual: {decompedQuat}");
+        }
+
+        [Theory, ClassData(typeof(TestRotationsData.YRPTestData))]
+        public void TestDecomposeYawPitchRollFromMatrixYPR(float yawDegrees, float pitchDegrees, float rollDegrees)
+        {
+            var yawRadians = MathUtil.DegreesToRadians(yawDegrees);
+            var pitchRadians = MathUtil.DegreesToRadians(pitchDegrees);
+            var rollRadians = MathUtil.DegreesToRadians(rollDegrees);
+
+            var rotMatrix = Matrix.RotationYawPitchRoll(yawRadians, pitchRadians, rollRadians);
+            rotMatrix.Decompose(out float decomposedYaw, out float decomposedPitch, out float decomposedRoll);
+
+            var expectedQuat = Quaternion.RotationYawPitchRoll(yawRadians, pitchRadians, rollRadians);
+            var decompedQuat = Quaternion.RotationYawPitchRoll(decomposedYaw, decomposedPitch, decomposedRoll);
+            Assert.True(expectedQuat == decompedQuat || expectedQuat == -decompedQuat, $"Quat not equals: Expected: {expectedQuat} - Actual: {decompedQuat}");
+        }
+
+        [Theory, ClassData(typeof(TestRotationsData.YRPTestData))]
+        public void TestDecomposeYawPitchRollFromMatricesZXY(float yawDegrees, float pitchDegrees, float rollDegrees)
+        {
+            var yawRadians = MathUtil.DegreesToRadians(yawDegrees);
+            var pitchRadians = MathUtil.DegreesToRadians(pitchDegrees);
+            var rollRadians = MathUtil.DegreesToRadians(rollDegrees);
+
+            // Yaw-Pitch-Roll is the intrinsic rotation order, so extrinsic is the reverse (ie. Z-X-Y)
+            var rotMatrix = Matrix.RotationZ(rollRadians) * Matrix.RotationX(pitchRadians) * Matrix.RotationY(yawRadians);
+            rotMatrix.Decompose(out float decomposedYaw, out float decomposedPitch, out float decomposedRoll);
+
+            var expectedQuat = Quaternion.RotationYawPitchRoll(yawRadians, pitchRadians, rollRadians);
+            var decompedQuat = Quaternion.RotationYawPitchRoll(decomposedYaw, decomposedPitch, decomposedRoll);
+            Assert.True(expectedQuat == decompedQuat || expectedQuat == -decompedQuat, $"Quat not equals: Expected: {expectedQuat} - Actual: {decompedQuat}");
+        }
+
+        [Theory, ClassData(typeof(TestRotationsData.XYZTestData))]
+        public void TestDecomposeXYZFromMatricesXYZ(float yawDegrees, float pitchDegrees, float rollDegrees)
+        {
+            var yawRadians = MathUtil.DegreesToRadians(yawDegrees);
+            var pitchRadians = MathUtil.DegreesToRadians(pitchDegrees);
+            var rollRadians = MathUtil.DegreesToRadians(rollDegrees);
+
+            var rotMatrix = Matrix.RotationX(pitchRadians) * Matrix.RotationY(yawRadians) * Matrix.RotationZ(rollRadians);
+            rotMatrix.DecomposeXYZ(out Vector3 eulerAngles);
+
+            var decompedRotMatrix = Matrix.RotationX(eulerAngles.X) * Matrix.RotationY(eulerAngles.Y) * Matrix.RotationZ(eulerAngles.Z);
+            var decompedQuat = Quaternion.RotationMatrix(decompedRotMatrix);
+
+            var expectedQuat = Quaternion.RotationX(pitchRadians) * Quaternion.RotationY(yawRadians) * Quaternion.RotationZ(rollRadians);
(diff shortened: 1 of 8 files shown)


sources/core/Stride.Core.Mathematics.Tests/TestMatrix.cs
```diff
@@ -0,0 +1,109 @@
+// Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net) and Silicon Studio Corp. (https://www.siliconstudio.co.jp)
```
manio143, line 1: When adding new code to Stride (in a new file) don't include Silicon Studio Corp in the copyright notice.

sources/core/Stride.Core.Mathematics.Tests/TestMatrix.cs
```diff
@@ -0,0 +1,109 @@
...
+                    for (int pitchIdx = 0; pitchIdx <= angleTestCount; pitchIdx++)
+                    {
+                        for (int rollIdx = 0; rollIdx <= angleTestCount; rollIdx++)
+                        {
+                            float yawDegrees = yawIdx * angleIncrementDegrees;
+                            float pitchDegrees = pitchIdx * angleIncrementDegrees;
+                            float rollDegrees = rollIdx * angleIncrementDegrees;
+
+                            result.Add(new object[] { yawDegrees, pitchDegrees, rollDegrees });
+                            if (yawDegrees != 0 && pitchDegrees != 0 && rollDegrees != 0)
```
manio143, line 98: Why are we only adding negative case if all three variables are positive? It's not testing (0, 0, -15) for example.

sources/core/Stride.Core.Mathematics.Tests/TestMatrix.cs
```diff
@@ -0,0 +1,109 @@
...
+            rotMatrix.DecomposeXYZ(out Vector3 eulerAngles);
+
+            var decompedRotMatrix = Matrix.RotationX(eulerAngles.X) * Matrix.RotationY(eulerAngles.Y) * Matrix.RotationZ(eulerAngles.Z);
+            var decompedQuat = Quaternion.RotationMatrix(decompedRotMatrix);
+
+            var expectedQuat = Quaternion.RotationX(pitchRadians) * Quaternion.RotationY(yawRadians) * Quaternion.RotationZ(rollRadians);
+            Assert.True(expectedQuat == decompedQuat || expectedQuat == -decompedQuat, $"Quat not equals: Expected: {expectedQuat} - Actual: {decompedQuat}");
+        }
+
+        public static IEnumerable<object[]> RotationAngles
```
manio143, line 80: While I don't object testing a lot of the cases from a moral standpoint, this producing over 3k test cases may kill Visual Studio test viewer - I'd recommend checking which values hit which code pathways to ensure 100% coverage of logic in those method and keep a few example per pathway.

sources/core/Stride.Core.Mathematics/Matrix.cs
```diff
@@ -622,43 +622,91 @@ public void DecomposeLQ(out Matrix L, out Matrix Q)
...
+        /// This rotation matrix can be represented by <b>intrinsic</b> rotations in the order <paramref name="yaw"/>, <paramref name="pitch"/>, then <paramref name="roll"/>.
+        /// <br/>
+        /// Therefore the <b>extrinsic</b> rotations to achieve this matrix is the reversed order of operations,
+        /// ie. Matrix.RotationZ(roll) * Matrix.RotationX(pitch) * Matrix.RotationY(yaw)
+        /// </remarks>
         public void Decompose(out float yaw, out float pitch, out float roll)
         {
-            pitch = MathF.Asin(-M32);
-            if (MathF.Cos(pitch) > MathUtil.ZeroTolerance)
+            // Adapted from 'Euler Angle Formulas' by David Eberly - https://www.geometrictools.com/Documentation/EulerAngles.pdf
```
manio143, line 638: I believe this is a good way of attributing license for this kind of work. @xen2 does this qualify to be added to `THIRD PARTY.md`?

sources/editor/Stride.Assets.Presentation/CurveEditor/ViewModels/RotationCurveViewModel.cs
```diff
@@ -204,26 +204,26 @@ protected internal override void UpdateMaxMin()
...
-                rotation.Z = MathF.Atan2(-matrix.M21, matrix.M31);
-                rotation.X = 0.0f;
-            }
-        }
+        ///// <summary>
+        ///// This method should have better precision than <see cref="Matrix.DecomposeXYZ(out Vector3)"/>, especially on the y-axis.
+        ///// </summary>
+        ///// <param name="matrix"></param>
+        ///// <param name="rotation"></param>
+        //private static void DecomposeXYZ(Matrix matrix, out Vector3 rotation)
```
manio143, line 212: Can you test if this code provides any significant difference in output values from the new one? If not, it should be removed instead of leaving it commented. If yes, we should try to understand what is the difference here.

manio143 (commented):
Looks pretty good 👍
Regarding the 174 issue - if your PR is not fully fixing it, can you change PR description so instead of `Fixes` it says `Partial` - this way GitHub won't auto close the issue when this PR is merged.

sources/core/Stride.Core.Mathematics.Tests/TestMatrix.cs
Basewq: My mistake, have changed this now (see the other comment).

sources/core/Stride.Core.Mathematics.Tests/TestMatrix.cs
Basewq: Ok, I've sat down and rethought this. I only really need to concern myself with when the singularity (ie. edge cases occur), which is +/-90 degrees on the second rotation. Other angles are for coverage.

sources/editor/Stride.Assets.Presentation/CurveEditor/ViewModels/RotationCurveViewModel.cs
Basewq: Reinstated but fixed the edge case values. Thanks for making me double check.

So the `M13` test in `RotationCurveViewModel` was already testing the +1 case properly (`Math.Abs(matrix.M13) - 1 < 1e-6f` is exactly the same as doing `MathUtil.IsOne(Math.Abs(M13)`).

So doc was right in that the yaw value can be more precise. This is because it does `Math.Acos(Math.Max(Math.Abs(matrix.M11), Math.Abs(matrix.M33)))` to get the yaw value as it approaches the edge case (eg. when the yaw is close to Pi/2) whereas the Matrix's code just sets it straight to `+/-PiOverTwo` (this is both the old version, and in Eberly's solution). I tested angles approaching Pi/2 (and a bit past it), and it does also only hit the edge case exactly on Pi/2 whereas the Matrix's code would treat the angle as an edge case as it got closer to Pi/2.
For the sake of consistency with the old code (and performance), I think the Matrix version can still just set it to PiOverTwo instead of the additional Acos/Max/Abs calls.

manio143 (approved):

Eideren:
Awaiting @xen2's input before merge. Crediting the source of this implementation, see https://github.com/stride3d/stride/pull/1532#discussion_r984304444

sources/core/Stride.Core.Mathematics/Matrix.cs
Eideren: Could also be moved to the top of the file next to the copyright notice, let's see what Xen2 has to say in this regard.

manio143:
It's been 3 months now, if @xen2 doesn't have the time I'd be up for merging it as is - the in code attribution is good enough to meet the minimum bar.

PR #1545, opened 2022-10-06
[Misc] Addressing PVS report

### LINKED ISSUE #1543: Suspicious code fragments in Stride Game Engine.
Release Type: GitHub
Version: 4.1.0.1734
Platform(s): Windows
Description:
Hello! I’ve checked your project with a static analyzer and found several suspicious fragments. I was wondering if you could take a look at them.

Code fragment N1:

```
private void RemoveEditor(....)
{
  ....
  if (....)
  {
    multiEditor.OpenedAssets.CollectionChanged -= (_, e) => 
      MultiEditorOpenAssetsChanged(multiEditor, e);
    ....
  }
  ....
}
```
Location: [AssetEditorsManager.cs](https://github.com/stride3d/stride/blob/c7096103697894218a7617bd39478a13cef61041/sources/editor/Stride.GameStudio/AssetEditorsManager.cs#L444) 
In this code fragment, an anonymous function is used to unsubscribe from the 'CollectionChanged' event. However, this won’t work since the anonymous function is used as a reference to a handler object. When the function is declared, a new delegate instance is created. This instance is different from the one previously subscribed to 'CollectionChanged'. Even if the bodies of these functions are identical, the functions still represent different objects. Thus, a handler won’t unsubscribe via an anonymous function, and the value of 'CollectionChanged' remains unchanged.
Code fragment N2:

```
public TexAtlas(...., TexImage atlas) : base(....)
{
  ....
  Name = atlas.Name;
  ....
  Name = "";
}
```
Location: [TexAtlas.cs](https://github.com/stride3d/stride/blob/c7096103697894218a7617bd39478a13cef61041/sources/tools/Stride.TextureConverter/Frontend/TexAtlas.cs#L48)
Inside the constructor, an empty string is assigned to the 'Name' property. This happens right after an actual value is assigned to this property. The developer may have decided to do this for the sake of debugging, but they didn’t delete it after that.
Code fragment N3:

```
public override void AddPoint(....)
{
  ....
  switch (Component)
  {
    case VectorComponent.X:
      value.X = realPoint.Y;
      break;

    case VectorComponent.Y:
      value.Y = realPoint.Y;
      break;

    case VectorComponent.Z:
      value.Z = realPoint.Y; // <=
      break;

    case VectorComponent.W:
      value.Z = realPoint.Y; // <=
      break;

    default:
      throw new NotSupportedException();
  }
  ....
}
```
Location: [RotationComponentCurveViewModel.cs](https://github.com/stride3d/stride/blob/c7096103697894218a7617bd39478a13cef61041/sources/editor/Stride.Assets.Presentation/CurveEditor/ViewModels/RotationComponentCurveViewModel.cs#L57)
In the last case operator, the 'realPoint.Y' value is assigned to the 'value.Z' property. Perhaps, there should be the 'value.W' property instead of 'value.Z'. We can draw this conclusion by inspecting the implementation of other case operators.

I found these issues with the PVS-Studio static analyzer. The analyzer found other suspicious fragments. You can find their description [here](https://pvs-studio.com/en/blog/posts/csharp/0994/).

Eideren:
# PR Details
Addressing the issues reported in #1543, will describe each changes below.

## Related Issue
Fixes #1543

## Types of changes

## Checklist

### DIFF
--- a/sources/presentation/Stride.Core.Presentation.Quantum/ViewModels/GraphViewModel.cs
+++ b/sources/presentation/Stride.Core.Presentation.Quantum/ViewModels/GraphViewModel.cs
@@ -41,7 +41,6 @@ namespace Stride.Core.Presentation.Quantum.ViewModels
             GraphViewModelService = serviceProvider.TryGet<GraphViewModelService>();
             if (GraphViewModelService == null) throw new InvalidOperationException($"{nameof(GraphViewModel)} requires a {nameof(GraphViewModelService)} in the service provider.");
             Logger = GlobalLogger.GetLogger(DefaultLoggerName);
-            if (rootPresenters == null) throw new ArgumentNullException(nameof(rootNode));
             var viewModelFactory = serviceProvider.Get<GraphViewModelService>().NodeViewModelFactory;
             viewModelFactory.CreateGraph(this, type, rootPresenters);
         }
--- a/sources/presentation/Stride.Core.Quantum/References/ObjectReference.cs
+++ b/sources/presentation/Stride.Core.Quantum/References/ObjectReference.cs
@@ -134,9 +134,6 @@ namespace Stride.Core.Quantum.References
                 if (targetValue != null && !type.IsInstanceOfType(targetValue))
                     throw new InvalidOperationException(@"The type of the retrieved node content does not match the type of this reference");
 
-                if (targetValue != null && !type.IsInstanceOfType(targetValue))
-                    throw new InvalidOperationException("TargetNode type does not match the reference type.");
-
                 TargetNode = targetNode;
                 TargetGuid = targetNode.Guid;
             }
--- a/sources/tools/Stride.TextureConverter/Backend/Wrappers/FINetWrapper/Classes/MetadataModels.cs
+++ b/sources/tools/Stride.TextureConverter/Backend/Wrappers/FINetWrapper/Classes/MetadataModels.cs
@@ -1220,7 +1220,7 @@ namespace FreeImageAPI.Metadata
 
         /// <summary>
         /// Gets or sets the exposure bias. The unit is the APEX value.
-        /// Ordinarily it is given in the range of 99.99 to 99.99.
+        /// Ordinarily it is given in the range of -99.99 to 99.99.
         /// </summary>
         /// <remarks>
         /// <b>Handling of null values</b><para/>
--- a/sources/tools/Stride.TextureConverter/Backend/Wrappers/FINetWrapper/Classes/MetadataModels.cs
+++ b/sources/tools/Stride.TextureConverter/Backend/Wrappers/FINetWrapper/Classes/MetadataModels.cs
@@ -129,7 +129,7 @@ namespace FreeImageAPI.Metadata
             }
             set
             {
-                SetTagValue("GlobalPalette", (value != null) ? null : value.Data);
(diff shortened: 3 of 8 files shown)


sources/presentation/Stride.Core.Presentation.Quantum/ViewModels/GraphViewModel.cs
```diff
@@ -41,7 +41,6 @@ private GraphViewModel([NotNull] IViewModelServiceProvider serviceProvider, [Not
             GraphViewModelService = serviceProvider.TryGet<GraphViewModelService>();
             if (GraphViewModelService == null) throw new InvalidOperationException($"{nameof(GraphViewModel)} requires a {nameof(GraphViewModelService)} in the service provider.");
             Logger = GlobalLogger.GetLogger(DefaultLoggerName);
-            if (rootPresenters == null) throw new ArgumentNullException(nameof(rootNode));
```
Eideren, line 44: I removed this line entirely instead of implementing the suggested change as `rootNode` is supposed to be null at this point.

sources/presentation/Stride.Core.Quantum/References/ObjectReference.cs
```diff
@@ -134,9 +134,6 @@ internal IObjectNode SetTarget(IObjectNode targetNode)
                 if (targetValue != null && !type.IsInstanceOfType(targetValue))
                     throw new InvalidOperationException(@"The type of the retrieved node content does not match the type of this reference");
 
-                if (targetValue != null && !type.IsInstanceOfType(targetValue))
-                    throw new InvalidOperationException("TargetNode type does not match the reference type.");
-
```
Eideren, line 139: I removed this line entirely instead of implementing the suggested change as testing `targetNode` against the type doesn't make much sense in that context.

sources/tools/Stride.TextureConverter/Frontend/TexAtlas.cs
```diff
@@ -45,7 +45,6 @@ public TexAtlas(TexLayout layout, TexImage atlas)
             CurrentLibrary = atlas.CurrentLibrary;
             LibraryData = atlas.LibraryData;
             Layout = layout;
-            Name = "";
```
Eideren, line 48: Most properties are copied, even reference types, it seemed to make more sense to copy the name as well.

manio143 (approved):

manio143:
Note that one of the unit tests failed

sources/tools/Stride.TextureConverter/Backend/Wrappers/FINetWrapper/Classes/MetadataModels.cs
```diff
@@ -1220,7 +1220,7 @@ public byte[] OECF
 
         /// <summary>
         /// Gets or sets the exposure bias. The unit is the APEX value.
-        /// Ordinarily it is given in the range of �99.99 to 99.99.
+        /// Ordinarily it is given in the range of 99.99 to 99.99.
```
Basewq, line 1223: The encoding is showing a variation of '-' in Visual Studio, as it looks like it's meant to say "-99.99 to 99.99"

Eideren:
That unit test passed on my machine before and does pass again after repeating the tests more than three times under both dotnet and VS's environment, not sure why it broke for teamcity before but it doesn't anymore ...

PR #1546, opened 2022-10-06
[Rendering] Addressing PVS report - SDL

Eideren:
# PR Details
Addressing the issue reported in #1543 - this time only regarding SDL.
Cleaned up the code and tested it, pinging @tebjan since that's kind of his thing. I would like your input on the change I introduced that removed the tests for `GameFormSDL` to always operate on window, is there a specific reason for it ?

## Related Issue
#1543

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Games/SDL/GameWindowSDL.cs
+++ b/sources/engine/Stride.Games/SDL/GameWindowSDL.cs
@@ -61,45 +61,23 @@ namespace Stride.Games
             if (!deviceChangeWillBeFullScreen.HasValue)
                 return;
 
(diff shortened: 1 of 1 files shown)


manio143 (approved):
I looked it over and it looks good - the order of property invocations is the same. I don't see a reason why there was a cast onto `GameFormSDL` - the `IsFullScreen` is in no way overridden on it - and moreover it wasn't being applied consistently for calling `BrintToFront()`.
I'm happy to merge this, but pinging @tebjan again for visibility.

tebjan:
Looks good, that got rid of some duplicated code. Can be merged from my side.

PR #1587, opened 2023-01-18
Update stride-logo-readme.png

chkworks:
Removes transparent background so it is display correctly in Github page

# PR Details

Really something minor, but still an imporovement over the look of the Github page.

## Description

Just added solid color to the background of the PNG.

## Types of changes

## Checklist


Jklawreszuk:
Can I ask you what is the reason behind adding new background ? I have a concern that on a light background it might look a bit off.
Please, take a look:
screen

chkworks:
Okay... well, in my case it looks like this:

Bildschirmfoto 2023-01-18 um 15 58 09

chkworks:
I added a bit of padding to the image, so now its not that misplaced. Also I obviously had to choose between dark on white background or white on dark background, and thought the version white on dark works for both backgrounds a bit better. Hope it also works for you guys.

Bildschirmfoto 2023-01-18 um 16 22 46
Bildschirmfoto 2023-01-18 um 16 22 52

Jklawreszuk:
I still think there is a better solution for supporting dark themes. Maybe instead of adding an unnecessary background, we could just add an outline to the logo in a similar way Godot does?

Reference (dark/light theme) : 
comparsion
What do you think?

chkworks:
Good call! I added an outer glow to the logo, that worked better than an outline IMO.

manio143:
Currently this is what I see on your branch:
image
While I guess it could make sense to add some border to make the logo more visible in dark themed GitHub, IMO it should be minimal - (2-3px) - the glow I see looks very thick on the dark version and because it's feathered it makes the text less readable. I think a hard edge or much less intense feathering would be preferred.

But actually, I see there's a way to use different image for different theme color - https://github.com/GiorgosXou/Random-stuff/tree/main/Programming/StackOverflow/Answers/70200610_11465149
As such it would be advisable to use the correct logos. I pulled this zip of SVGs from Discord (author sebl):
[stridesvgs.zip](https://github.com/stride3d/stride/files/10776844/stridesvgs.zip)
We should use the SVGs to generate the PNGs without loss of quality. The current one is `solid.svg`, the white one is `solid-1.svg`

chkworks:
Okay, that seems to be a good way to go.

I have pushed the black and white versions made from sebls SVGs. What I don't know is how to link them properly, right now the path seems to be set dynamically but in the example a direct path is used. Could you link them in the Readme?

Jklawreszuk:
@chkworks Well, I think its a matter of choice. I would rather go with direct path because, less megs would be downloaded :v:

chkworks:
Alright, I have reverted the original logo to the design without the outer glow, but left the black and white versions in this PR. When they are pushed into the main branch, I can try to include them with the provided method using direct paths to the versions, depending on the theme.

Jklawreszuk:
> When they are pushed into the main branch, I can try to include them

@chkworks Why not directly in your current PR?

chkworks:
Alright, I initially thought that the URL will be linked to my user name, but I think the path is now taken from the original repository... hope that works now?

Jklawreszuk:
Yes it does! Looks good to me.

Jklawreszuk (approved):

PR #1591, opened 2023-01-21
Cleanup legacy platforms

Jklawreszuk:
# PR Details

This PR removes files associated with unused or long-unsupported platforms.

## Motivation and Context

My goal is to simplify and debloat build system.

## Types of changes

## Checklist

### Todo

- [x] Remove Win8, Win10, Windows Store, WinPhone, Silverlight related files/lines
- [x] Remove other legacy platforms (NET MicroFramework, NET Portable etc)
- [x] Test changes


Jklawreszuk:
I think I've finished my PR and It's ready for review.

samples/Tests/Stride.Samples.Tests.csproj
```diff
@@ -1,6 +1,5 @@
 <Project>
   <!-- NuGet resolved needed for StrideDefaultAssetsPlugin.LoadDefaultTemplates() -->
-  <Import Project="..\..\sources\shared\Stride.NuGetResolver\Stride.NuGetResolver.projitems" Label="Shared" />
```
manio143, line 3: Can you check if this was added by me as a workaround for the tests not running on a clean system?

sources/core/Stride.Core/PlatformType.cs
```diff
@@ -59,13 +59,6 @@ public enum PlatformType
...
         /// </summary>
-        macOS,
-#pragma warning restore SA1300 // Element must begin with upper-case letter
-
-        /// <summary>
-        /// The Universal Windows Platform (UWP). Please use <see cref="UWP"/> intead.
-        /// </summary>
-        [Obsolete("Please use UWP instead")]
-        Windows10 = UWP,
+        macOS
```
manio143, line 62: Nitpick: you should leave a trailing comma in case someone adds another value to the enum and git doesn't need to touch the last value for the diff

sources/engine/Stride.Assets.Tests2/Stride.Assets.Tests2.csproj
```diff
@@ -12,7 +12,6 @@
   <ItemGroup>
     <PackageReference Include="xunit" Version="2.4.2-pre.12" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.4.3" />
-    <PackageReference Include="TeamCity.VSTest.TestAdapter" Version="1.0.32" />
```
manio143, line 15: We should verify this test still runs fine on TeamCity - I'm not really sure what this adapter is meant to do here

VaclavElias:
(Un)related: 

@Jklawreszuk, as you are working on some clean up, wouldn't you mind checking for me just briefly if we need `ServiceWireSerializer.cs`. This file is causing me issues when I build Stride Docs. It seems there are no references and usage of this file. When I remove it I can build Stride but with my experience I can't confirm any further consequences. 

The new Stride Docs seems to be very sensitive on certain things. Because there is no reference to this file, probably when Stride  is built the `ServiceWire` package reference is removed and then docfx flags it as error.

I will create a separate issue with my PR, where I had to fix some references or remove things, but always good if someone can double check. https://github.com/VaclavElias/stride/tree/stride-docs-fix

Thanks.

image

manio143:
@VaclavElias it's offtopic from this PR - a separate issue for this question would be more appropriate - I'd suggest you open it if you want to respond to this comment.
The ServiceWireSerializer is marked with an attribute used by the AssemblyProcessor to create code that registers it at runtime in the serialization system. It handles serializing ServiceSyncInfo object. I haven't found any direct references to that class in the repo (other than serializer) but it may be included transitively or something (not clear).

samples/Tests/Stride.Samples.Tests.csproj
Jklawreszuk: @manio143 Indeed, this was added to "auto resolve packages correctly" so I've undone that change.

sources/core/Stride.Core/PlatformType.cs
Jklawreszuk: Fixed

sources/engine/Stride.Assets.Tests2/Stride.Assets.Tests2.csproj
Jklawreszuk: I'm afraid this duplicate package reference was also intentional [(see here)](https://github.com/stride3d/stride/commit/c21cf0224153c8d1a04a9bab6e9d37f0ea2f19f0#diff-c7b44c06ee0a39a3aa90ea0467d9fde7415f6f801ba2d45c1fe54174b1f46b78). I will undo this also for now.

PR #1592, opened 2023-01-27
Added implicit conversion of equivalent ValueTuples for vector structs

skohanowski:
More syntactic sugar.

## Description

Adds implicit conversion from ValueTuples to the equivalent vector struct.
(int, int) -> Int2 
(float, float, float) -> Vector3
etc..

## Motivation and Context

This was specifically to support cleaner data inlining such as:
```
public readonly ReadOnlyMemory<Int2> Corners = new Int2[]
{   (0, 0)
,   (1, 0)
,   (0, 1)
,   (1, 1)
};
```
And operations like:
```
var x = corners[0] + (2, 1);
```

## Types of changes

## Checklist

- [x] All new and existing tests passed. (I have failing tests, but they are tests that have continuously failed on my workstation going back to last July. No new issues.)

### DIFF
--- a/sources/core/Stride.Core.Mathematics/Color.cs
+++ b/sources/core/Stride.Core.Mathematics/Color.cs
@@ -1125,5 +1125,12 @@ namespace Stride.Core.Mathematics
             var value = (int)(component * 255.0f);
             return (byte)(value < 0 ? 0 : value > 255 ? 255 : value);
         }
+
+        /// <summary>
+        /// Implicitly convert equivalent ValueTuple.
+        /// </summary>
+        /// <param name="values"></param>
+        public static implicit operator Color((byte R, byte G, byte B, byte A) values) => new Color(values.R, values.G, values.B, values.A);
+
     }
 }
--- a/sources/core/Stride.Core.Mathematics/Color3.cs
+++ b/sources/core/Stride.Core.Mathematics/Color3.cs
@@ -831,6 +831,13 @@ namespace Stride.Core.Mathematics
             b = B;
         }
 
+        /// <summary>
+        /// Implicitly convert equivalent ValueTuple.
+        /// </summary>
+        /// <param name="values"></param>
+        public static implicit operator Color3((float R, float G, float B) values) => new Color3(values.R, values.G, values.B);
+
+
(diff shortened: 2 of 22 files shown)


manio143:
I'm not sure this is a good idea - it could lead to ambiguities. Do you know if there's target typing of numeral tuples? E.g. if a function takes `(float, float, float)` can it take a tuple with integer literals `(1, 1, 1)`?

A small amount of explicitness wouldn't hurt if otherwise we're faced with ambiguity and rely on compiler rules to choose the right cast for us.

For this list example you showed I believe you can use the short `new` syntax 

```
public readonly ReadOnlyMemory<Int2> Corners = new Int2[]
{   new(0, 0)
,   new(1, 0)
,   new(0, 1)
,   new(1, 1)
};
```

skohanowski:
> I'm not sure this is a good idea - it could lead to ambiguities. Do you know if there's target typing of numeral tuples? E.g. if a function takes `(float, float, float)` can it take a tuple with integer literals `(1, 1, 1)`?
> 
> A small amount of explicitness wouldn't hurt if otherwise we're faced with ambiguity and rely on compiler rules to choose the right cast for us.
> 
> For this list example you showed I believe you can use the short `new` syntax
> 
> ```
> public readonly ReadOnlyMemory<Int2> Corners = new Int2[]
> {   new(0, 0)
> ,   new(1, 0)
> ,   new(0, 1)
> ,   new(1, 1)
> };
> ```

Well now, that's interesting. I was going to argue that C# didn't perform implicit conversion chaining, and the two types where semantically and structurally the same but I did a little test and what'a'know. I just learned something about ValueTuples. It does in fact do just that. 

```
void Main()
{
	(int, int) x = (1, 1);
	
	//I expected both of these to be parse exceptions, but aren't.
	Test1 y = x;
	F1(x);

	//These fail as expected.
	Test2 w = x;
	F2(x);

	void F1(Test1 t) { }
	void F2(Test2 t) { }
}

public struct Test1
{
	public static implicit operator Test1((float, float) value) => new();
}

public struct Test2
{
	public static implicit operator Test2(Test1 value) => new();
}
```

PR #1593, opened 2023-02-02
Dotnet 7

Doprez:
# PR Details

Updated Stride projects to .NET 7.

## Description

- changed any references from .NET 6 to .NET 7 in all csproj files.
- Added unsafe keyword to OpenVR due to changes in C# 11 regarding ref.
- Added #pragma warning disable SYSLIB0011 to Binary formatter due to security error found from dotnet 7 (may need to be looked into further)

## Types of changes

## Checklist


sources/core/Stride.Core.Tests/TestSerialization.cs
```diff
@@ -89,6 +89,8 @@ public struct S
             public int A { get; set; }
         }
 
+        // TODO change binary serialization due to microsoft security concerns.
```
manio143, line 92: Can you open a bug for this?

sources/core/Stride.Core/Stride.Core.csproj
```diff
@@ -19,6 +19,14 @@
     <ExtrasUwpMetaPackageVersion>6.2.12</ExtrasUwpMetaPackageVersion>
   </PropertyGroup>
   
+  <PropertyGroup Condition="'$(Configuration)|$(TargetFramework)|$(Platform)'=='Debug|net7.0|AnyCPU'">
+    <WarningLevel>7</WarningLevel>
```
manio143, line 23: Can you explain this change and why it's conditional?

sources/engine/Stride.VirtualReality/OpenVR/OpenVR.cs
```diff
@@ -24,8 +24,11 @@ internal static class OpenVR
         /// <remarks>Take care to ensure that the struct has been initialized appropriately, otherwise the struct's fields could contain uninitialized data from the stack.</remarks>
         private static ref T SkipInit<T>(out T value)
         {
-            Unsafe.SkipInit(out value);
-            return ref value;
+            unsafe
```
manio143, line 27: I think it'd be nicer to put unsafe on the method declaration

Doprez: Fixed. I agree

sources/core/Stride.Core/Stride.Core.csproj
Doprez: I Must have been messing with too much when testing I have removed this as its not needed.

sources/core/Stride.Core.Tests/TestSerialization.cs
Doprez: https://github.com/stride3d/stride/issues/1596
added to Issues.

azeno:
Would it be possible to do one more stable release before switching to net7?

Doprez:
I would be curious if there was a pipeline to create a beta build or something instead of pushing to master? Maybe I was too hasty on this PR especially since .NET 7 support is a few months shorter maybe even skip this PR and just keep this as a reference for when .NET 8 comes out?

Perksey:
Yeah Silk.NET’s policy going forward is to exclusively support the latest LTS only, so every even numbered release.

tebjan:
I'd also vote for only going with LTS, but this PR is super helpful for the transition and for anyone who needs .NET7 now.

Doprez:
I think that makes sense as well to me. Ill decline the PR but the branch will be here for november.

tebjan:
We can keep this open so others will find it easier. I'll convert it to draft.

Doprez:
closing as this was superseded with https://github.com/stride3d/stride/pull/1616

Issue #1344, opened 2022-03-05
Use new flip model for dx11 swapchain


tebjan:
**Is your feature request related to a problem? Please describe.**

According to this article, 3d applications shouldn't use the old "blt" mode anymore:
https://devblogs.microsoft.com/directx/dxgi-flip-model/

**Describe the solution you'd like**

Change the swapchain mode when flip model is available.

**Additional context**

This would be a free performance improvement.

azeno:
Just did a quick test rendering a quad in a 1920x1200 window and the GPU load on a RTX 2070 went down from 20% to 0% using `FlipDiscard`.

PR #1594, opened 2023-02-02
Use flip model in Direct3D11

azeno:
# PR Details

## Description

As recommended by Microsoft this PR uses the flip model in Direct3D11. A rather good overview with useful information & links can be found [here](https://walbourn.github.io/care-and-feeding-of-modern-swapchains/).

## Related Issue

#1344 

## Motivation and Context

Comparison of the example game
without flip model
image

with flip model
image

In another test where we only rendered a white quad in 1920x1200 (no post-effects) the GPU usage dropped from 20% to 0%.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Graphics/Direct3D/SwapChainGraphicsPresenter.Direct3D.cs
+++ b/sources/engine/Stride.Graphics/Direct3D/SwapChainGraphicsPresenter.Direct3D.cs
@@ -380,26 +380,30 @@ namespace Stride.Graphics
 
         private SwapChain CreateSwapChainForDesktop(IntPtr handle)
         {
-            bufferCount = 1;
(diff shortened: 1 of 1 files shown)


manio143 (approved):

tebjan (approved):

tebjan:
Awesome, thanks!

Eideren:
Just a small question, it looks like this is disabled if we're even using msaa - unless ours already resolves and the description is the one of the resolved buffer ?

azeno:
True - I was under the impression in multi sampling the resolve was already handled in a separate target anyways? But you're right should be checked.

azeno:
Ah wait, this PR is ok, with msaa it simply falls back to the old behaviour. It would be a separate quest to enable the flip model with msaa.

Issue #1508, opened 2022-08-21
Clean UI with no post FX broken when no model present in the scene.


mmujic2:
### Release Type:
Official Release

### Version: 
4.1.0.1457-beta

### Platform(s): 
Windows

### Describe the bug:
After performing steps for clean UI with no post processing (steps can be found here: https://github.com/herocrab/StrideCleanUI ), if a scene without a model is created (scene contains only camera and entity with UI component) the post FX still happens, and after loading a new scene a warning pops up that says: 

`[SceneCameraRenderer]: Warning: SceneCameraRenderer [82781b1c-a13f-4f6c-90fa-fce37b46a656] has no camera assigned to its Slot[Main]. Make sure a camera is enabled and assigned to the corresponding Slot.`

### To Reproduce:
1. Create new project and perform the steps as shown in the link above.
2. Create new Scene and set the created scene as the default scene in game settings.
3. In the newly created scene add a camera and an Entity with a UI component attached to it.
4. Create basic page with one button and text and attach the page to the UI component of the entity.
5. Create a new sync script which will set the button.Click method to change scenes, attach script to camera (The scene has to be changed through the Click method, otherwise the warning doesn't pop up). 
6. After launching game the post FX should still happen and after clicking the button the warning should pop up.
7. Add any model or entity containing model component to the created scene and the problem should be resolved.

### Expected behavior:
The post FX shouldn't happen even if a model isn't added to the scene and the warning shouldn't pop up.

### Screenshots:
No model added
![image](https://user-images.githubusercontent.com/73075272/185807160-44f9c220-39a6-418a-bcc0-590c561ff8e5.png)
![image](https://user-images.githubusercontent.com/73075272/185807101-75522fe6-86d1-4232-8045-5dfeed9ea0d2.png)

Warning that pops up after changing scene
![image](https://user-images.githubusercontent.com/73075272/185807631-6b626828-546e-4481-bfaa-7379d0dff06f.png)

After adding the default ground model that comes with the project template:
![image](https://user-images.githubusercontent.com/73075272/185807174-06c41d58-9cd6-4fff-b66e-eb8388d6c772.png)
![image](https://user-images.githubusercontent.com/73075272/185807185-2b6c0de4-7bc8-4f3c-8ba1-612106326c3a.png)

NicusorN5:
Did you try using a background skybox? Or even a background sprite?

Issue #1154, opened 2021-09-29
Exclude UI from PostFx by default


manio143:
**Is your feature request related to a problem? Please describe.**
It's not expected by users that UI will be rendered with PostFx applied in the default project setup. This comes up often and we direct people to a tutorial on changing it (https://github.com/herocrab/StrideCleanUI https://www.youtube.com/watch?v=cazIR97VPcg)

**Describe the solution you'd like**
Modify default GraphicsCompositor according to the tutorial and set the default render group in the UI component accordingly.

Ethereal77:
As UI can be used also in 3D positioned in the scene, it should be excluded from post-FX only if they are marked as fullscreen or something. If not, elements in the scene would be overly bright and incorrect as they avoid tonemapping and other effects the surrounding geometry are affected by.

VaclavElias:
I am new to Stride, did a bit experimenting with Unity, I am beginner and I agree with @manio143. When we set white colour font for the UI you can't send us to a tutorial and do some ceremony to get the white colour working 🤣.

Aggror:
I think this can be implemented.

VaclavElias:
I also want it for code-only approach. There is one already in Stride ```GraphicsCompositorHelper``` which I can use and edit before it is used so the white is correct.

```c#
var graphicsCompositor = GraphicsCompositorHelper.CreateDefault(true);

// edit graphicsCompositor so the white is white

```

or I can just create another one and correct the white there 
https://github.com/VaclavElias/Stride3DTutorials/blob/main/Other/Minimal/Stride.GameDefaults/GraphicsCompositorBuilder.cs

In my case, it would be good to know how this GraphicsCompositor affects the performance but I can play with that once there is a time for that.

Not sure what needs to be changed in the Stride Studio so the template also have it, or even having a UI to switch it on if there is a reason for some to have it off?

PR #1603, opened 2023-02-04
Change default graphics compositor settings so UI isn't affected by post processing or shown in editor

JackPilley:
# PR Details

This PR updates the default graphics compositor setup so that UI is in its own render stage that isn't affected by post processing and also isn't shown in the editor view.

## Description

By default there is now a render stage called UserInterface which the UIRenderFeature is assigned to. The UIRenderFeature is also no longer assigned to the Transparent render stage.

The Editor Renderer entry point is now a SceneRenderCollection with one child, a shared forward renderer which doesn't use the UserInterface stage. This stops the UI from being shown in the editor. SceneRenderCollection was chosen in case a user wants the UI to be visible in the editor. They just need to add a SingleStageRenderer set to the UserInterface stage to the collection.

The Game Renderer remains mostly the same, using a Camera Renderer with a SceneRenderCollection as the child. The collection has an added SingleStageRenderer set to the UserInterface stage.

I consider this setup much simpler than the one described [here](https://github.com/herocrab/StrideCleanUI) while achieving the same results.

A small additional change this PR makes is giving proper names to the render features, they were all just test or null before.

There's a risk that this is a breaking change, so I'd like some feedback on that before marking this as ready to merge.

## Related Issue

Addresses https://github.com/stride3d/stride/issues/1154
This setup also avoids the problem from https://github.com/stride3d/stride/issues/1508

## Motivation and Context

See related issues.

## Types of changes

## Checklist

## To do
- [x] Update the level 10 archetype file
- [x] Update the level 9 archetype file
- [ ] Update the create default function in the [GraphicsCompositorHelper class](https://github.com/stride3d/stride/blob/master/sources/engine/Stride.Engine/Rendering/Compositing/GraphicsCompositorHelper.cs)


MechWarrior99:
Hey @JackPilley do you happen to remember what the status of this was? Does some other work need to be done on it before it could be merged? I'd be happy to take a go at it if something is still needed.

JackPilley:
> Hey @JackPilley do you happen to remember what the status of this was? Does some other work need to be done on it before it could be merged? I'd be happy to take a go at it if something is still needed.

There's a couple things. I only updated the empty new game project, so I think the samples and templates would also need to be updated. I'm also not sure if the changes would only apply to new projects, or if existing projects will be affected by the changes. And there's always the possibility I did something wrong and it'll break in some situation I didn't think to test.

VaclavElias:
Not sure if related. I am adding Clean UI through the code.

In the community toolkit, I am adding default `GraphicsCompositor` like so

`var graphicsCompositor = GraphicsCompositorHelper.CreateDefault(true);`

https://github.com/stride3d/stride-community-toolkit/blob/f774f0bb5a36810e622ceefc5d7ff1dacfd34efb/src/Stride.CommunityToolkit/Engine/GameExtensions.cs#L171-L180

And then, I am applying CleanUI with the extension `AddCleanUIStage()`.

`game.AddGraphicsCompositor().AddCleanUIStage();` 

https://github.com/stride3d/stride-community-toolkit/blob/f774f0bb5a36810e622ceefc5d7ff1dacfd34efb/src/Stride.CommunityToolkit/Rendering/Compositing/GraphicsCompositorExtensions.cs#L37-L43

And making adjustments

https://github.com/stride3d/stride-community-toolkit/blob/f774f0bb5a36810e622ceefc5d7ff1dacfd34efb/src/Stride.CommunityToolkit/Rendering/Compositing/GraphicsCompositorExtensions.cs#L37-L43

```csharp
AddPostEffects(graphicsCompositor);
AddRenderStagesAndFeatures(graphicsCompositor);
```

That means, on the top what is suggested in this issue, not sure if the Stride's ` GraphicsCompositorHelper.CreateDefault(true);` should be also updated so we get the same result, without me having extensions.

Note, that my code implementation adds CleanUI only to `Group31` if I am correct. Not sure if it should be across all groups.

JackPilley:
> Note, that my code implementation adds CleanUI only to Group31 if I am correct. Not sure if it should be across all groups.

Since the UI is in its own stage, it shouldn't need to be assigned to any particular group.

MechWarrior99:
So I messed around with it and the setup seems to work fine, couldn't find any issues, so all good there. I was looking at applying these changes to the samples. And it looks like they use the Level10 graphics compositor setup. But not quite sure the best way to go and actually edit them. Can't copy-paste due to the ref ids all being different, and not sure how to open the samples directly if possible even). Any suggestions?

JackPilley:
I spent some time today doing some testing. It seems like the changes apply to all of the templates and samples already. So they don't actually need to be updated.

I also tested creating a project on a previous version of Stride, editing the graphics compositor, and then opening the project again after recompiling Stride with these changes. It definitely does pose a risk of breaking existing projects if they've edited their compositor.

MechWarrior99:
>  It seems like the changes apply to all of the templates and samples already

Oh, just tested and you're right! YAY!

> I also tested creating a project on a previous version of Stride, editing the graphics compositor, and then opening the project again after recompiling Stride with these changes. It definitely does pose a risk of breaking existing projects if they've edited their compositor.

Hmm deffinitly shouldn't, it creates the assets in the project and then keeps no reference to the template used.
I tested it my self to be sure and it didn't apply to an existing compositor.
Steps:
1. Create new template project in current version of stride.
2. Open project, open GraphicsCompositor editing window.
3. Close stride, and click the "Save" button when prompted.
4. Open the same project with stride compiled with these changes.
5. See GraphicsCompositor unchanged.

Is it possible something else was going on with your setup?

JackPilley:
Here's what happened for me, the modifications I made the compositor are basically the setup for removing post processing from the UI that Hero Crab documented.

I added a new render stage called EXAMPLE

image

I modified the UIRenderFeature so there's groups going to two different stages

image

I added the new render stage to the entry points

image

Now after applying the changes and recompiling the engine I get this:

The UserInterface render stage got added automatically

image

The render feature still sends groups to different stages, but where the first one is going is now different, also the selected groups got reset for the first one

image

And there's now an entry point for the UserInterface stage

image

Make sure you clean the solution and fully rebuild the project when you're testing, Visual Studio doesn't always detect the changes in the files I'm modifying.

MechWarrior99:
I first opened and created the project via the stride launcher. And then reopened it from source. Are you able to test in another project or something? And maybe somone esle can as well?

JackPilley:
Ok, creating the project from an engine version downloaded through the launcher and then reopening from an engine compiled from source seems to work fine.

It seems weird to me that there would be a difference in behaviour though. Maybe it's because the version from the launcher is a different release number?

JackPilley:
It just occurred to me that this might also mean the changes won't apply to the samples and templates.

MechWarrior99:
Well @JackPilley I figured out what is going on... the graphics compositors that are created in new projects have one of the two default graphics compositors as their archetype. That is why they update!

JackPilley:
So if I'm understanding correctly, the samples and templates don't need to be updated, but this is a breaking change since existing projects will use the new archetype?

VaclavElias:
> Well @JackPilley I figured out what is going on... the graphics compositors that are created in new projects have one of the two default graphics compositors as their archetype. That is why they update!

Is it the one I mentioned in my comment `var graphicsCompositor = GraphicsCompositorHelper.CreateDefault(true);`

If that one is sorted as well, I can remove my extension 🙂.

JackPilley:
@VaclavElias No, it looks like the create default function doesn't use the archetype file, so it'll need to be updated as well. Good thing you mentioned it otherwise I wouldn't have known to look for it.

xen2:

PR #1616, opened 2023-02-23
Update dotnet 8

Doprez:
# PR Details

Updated Stride projects to .NET 8.

## Description

- changed any references from .NET 6 to .NET 8 in all csproj files.
- Added unsafe keyword to OpenVR due to changes in C# 11 regarding ref.
- Added #pragma warning disable SYSLIB0011 to Binary formatter due to security error found from dotnet 7 (may need to be looked into further)
- worked off of https://github.com/stride3d/stride/pull/1593

Marked draft as the CPP projects cant seem to be found by the dotnet build command.

## Types of changes

## Checklist


VaclavElias:
Just wondering for the future updates, would it make sense to have this `net6.0-windows7.0` as a parameter so it can be updated just in one place?

Doprez:
The reason these were done manually was because I was modifying threw a find and replace all in VSCode. all of these are outside the scope of the Stride.sln file where the references are managed by a couple of variables inside `StrideEditorTargetFramework` and `StrideFrameworkWindows` for `net8.0-windows7.0`. 

Just noting this for myself in case I dont update this today. I will try to take a look later to see if those can be referenced outside of the base Stride solution.

sources/engine/Stride.Assets/StridePackageUpgrader.cs
```diff
@@ -228,13 +228,13 @@ public override bool UpgradeBeforeAssembliesLoaded(PackageLoadParameters loadPar
                             {
                                 // In case it's a single TargetFramework, add the "s" at the end
                                 tfm.Xml.Name = "TargetFrameworks";
-                                tfm.Xml.Value = "net6.0";
+                                tfm.Xml.Value = "net8.0";
                                 isProjectDirty = true;
                             }
                             // Executable
                             else if ((tfm.EvaluatedValue.StartsWith("net4") || tfm.EvaluatedValue.StartsWith("net5")) && solutionProject.Type == ProjectType.Executable)
```
cNoNim, line 235: and here too

sources/engine/Stride.Assets/StridePackageUpgrader.cs
```diff
@@ -228,13 +228,13 @@ public override bool UpgradeBeforeAssembliesLoaded(PackageLoadParameters loadPar
                             {
```
cNoNim, line 228: Imho PackageUpdater should update from net5, net6, and net7 may be or no?

cNoNim, line 228: Imho PackageUpdater should update from net5, net6, and net7 may be or no?

sources/engine/Stride.Assets/StridePackageUpgrader.cs
cNoNim, line 235: and here too

cNoNim:
Why stride use windows7.0 TFM?

sources/engine/Stride.Assets/StridePackageUpgrader.cs
Doprez: for the net4 reference here is due to backwards compatability for .NET framework. This may not be needed anymore since this has changed since Stride 4.1. @Eideren would you be against removing this?

Eideren: Looks like we should be able to replace the whole
`if (dependency.Version.MinVersion < new PackageVersion("4.1.0.0") && solutionProject != null){...}`
with something like
`dependency.Version.MinVersion < this package version, swap to this tfm`
The additional checks for handwritten version seem unnecessary.

xen2: What you'll want to do is bump to 4.2.0.1 in `sources\shared\SharedAssemblyInfo.cs`.
4.1 => net6.0
4.2 => net8.0

Then ideally, to ease maintenance, we want to keep the version upgrade atomic. This makes sure all upgrade paths are tested (since they're all run for older projects) and sequentially applied. Also, it's later easier to remove support for older version upgrades if later we say we want to support only upgrading from 4.1+ and 4.0 can't be opened anymore.

So what you can do is add a separate check after existing one, only for 4.1 to 4.2 (old code stay the same, then you add another `if (dependency.Version.MinVersion < new PackageVersion("4.2.0.0")` after the `if (4.1)` branch to do the upgrade to net8.0).

cNoNim:
I guess Stride.Core.AssemblyProcessor also should be updated and recompiled to net8.0 in deps\AssemblyProcessor.

Doprez:
> I guess Stride.Core.AssemblyProcessor also should be updated and recompiled to net8.0 in deps\AssemblyProcessor.

Good catch, should be fixed.

cNoNim:
> > I guess Stride.Core.AssemblyProcessor also should be updated and recompiled to net8.0 in deps\AssemblyProcessor.
> 
> Good catch, should be fixed.

But not so simple
https://github.com/stride3d/stride/pull/1840
AssemblyProcessor used in build pipeline with netstandard2.0 TF and as binary dependency.
I'm not sure why exists net6.0 version, and for what it should be updated to net8.0.
Maybe we should leave only the netstandard2.0 version.

But for now you should recompile AssemblyProcessor and commit produced files. I guess.

manio143:
> AssemblyProcessor used in build pipeline with netstandard2.0 TF and as binary dependency. I'm not sure why exists net6.0 version, and for what it should be updated to net8.0. Maybe we should leave only the netstandard2.0 version.

The reason there's 2 variants is because `dotnet build` loads the newer version (which can use modern CLR features), while Visual Studio in some cases uses a legacy .NET Framework based MSBuild that can only load up to `netstandard2.0` libraries. And we use the AssemblyProcessor for some MSBuild Task definition so it has to be loadable in its process - maybe a look needs to be have if that task could be removed.

warappa:
**Suggestion**
As this PR already touches all `.csproj` files, wouldn't it make sense to also introduce [Central Package Management](https://learn.microsoft.com/en-us/nuget/consume-packages/Central-Package-Management) with its `Directory.Packages.props` to manage all package _versions_ in 1 file?
This way misaligned `PackageReference` `Version` attributes will be impossible within Stride's solution.

It's very well supported since .NET 7 (including tooling) and eases upgrading NuGet dependencies tremendously.

tebjan:
> As this PR already touches all `.csproj` files, wouldn't it make sense to also introduce [Central Package Management](https://learn.microsoft.com/en-us/nuget/consume-packages/Central-Package-Management)

I am a big fan of this, It simplifies version management a lot! We can define one property for the Silk.NET version and have it consistent everywhere.

sources/targets/Stride.Core.TargetFrameworks.Editor.props
```diff
@@ -2,7 +2,7 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
   <PropertyGroup>
-    <StrideEditorTargetFramework>net6.0-windows7.0</StrideEditorTargetFramework>
+    <StrideEditorTargetFramework>net8.0-windows7.0</StrideEditorTargetFramework>
```
Kryptos-FR, line 5: Cherry-pick/adapt/fix the changes from #1908, so that windows dependencies are removed for some libraries.
Or it can be done just after this PR, whichever feels safer (maybe better since it seems to have caused issues in main).

sources/targets/Stride.Core.props
```diff
@@ -44,8 +44,8 @@
     Settings StridePlatform specific
   -->
   <PropertyGroup>
-    <!-- Using C# version 10.0 which targets Net6.0 -->
-    <LangVersion>10.0</LangVersion>
+    <!-- Using C# version 11.0 which targets net8.0 -->
+    <LangVersion>11.0</LangVersion>
```
Kryptos-FR, line 48: We could use `latest` instead so that we are always on the latest with regards to the compiler capabilities.
```suggestion
    <LangVersion>latest</LangVersion>
```

sources/engine/Stride.Assets/StridePackageUpgrader.cs
Doprez: I finally got around to updating this. I left out net7 since it was never officially supported anyway.

sources/targets/Stride.Core.props
Doprez: I just removed since it should use latest available from the compiler by default.

sources/targets/Stride.Core.TargetFrameworks.Editor.props
Doprez: I'm going to leave this out for this PR at least. I did try already and there was a mess of configuration I wasnt able to figure out and it broke everything when building projects.

Doprez:
I think the last suggestion left is the property centralization for version from Vaso, Warappa and Tebjan. Would it be possible to just use the existing properties or should there be a new one?

I also left the Compiler service alone as it was since there needs to be some extra work to move it away from standard2.0 as well.

sources/targets/Stride.Core.TargetFrameworks.Editor.props
Kryptos-FR: Fair enough. Better to not complicate this PR.

IXLLEGACYIXL:
> I think the last suggestion left is the property centralization for version from Vaso, Warappa and Tebjan. Would it be possible to just use the existing properties or should there be a new one?
> 
> I also left the Compiler service alone as it was since there needs to be some extra work to move it away from standard2.0 as well.

you cant write roslyn in non netstandard2.0

deps/AssemblyProcessor/net8.0/Stride.Core.AssemblyProcessor.Packed.dll
xen2, line 1: Do we still need to keep `deps\AssemblyProcessor\net6.0` since we now have the net8.0 version?

samples/Audio/SimpleAudio/SimpleAudio.Game/SimpleAudio.Game.csproj
xen2, line 1: Since samples are updated, we probably want to bump the sample package version to 4.2.0.1-dev and 4.2.0.1
`sources\editor\Stride.Samples.Templates\ThisPackageVersion.DevBuild.cs`
`sources\editor\Stride.Samples.Templates\ThisPackageVersion.PackageBuild.cs`

sources/assets/Stride.Core.Assets/Stride.Core.Assets.csproj
```diff
@@ -29,6 +29,7 @@
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\..\buildengine\Stride.Core.BuildEngine.Common\Stride.Core.BuildEngine.Common.csproj" />
+    <ProjectReference Include="..\..\core\Stride.Core\Stride.Core.csproj" />
```
xen2, line 32: Curious, you remember why it was necessary?

sources/core/Stride.Core.Mathematics/Stride.Core.Mathematics.csproj
```diff
@@ -12,6 +12,7 @@
     <StrideAssemblyProcessorOptions>--auto-module-initializer --serialization</StrideAssemblyProcessorOptions>
     <StridePublicApi>true</StridePublicApi>
     <StrideCodeAnalysis>true</StrideCodeAnalysis>
+    <LangVersion>12.0</LangVersion>
```
xen2, line 15: Is this still necessary?
cf https://github.com/stride3d/stride/pull/1616#discussion_r1372368581

sources/core/Stride.Core/Stride.Core.csproj
```diff
@@ -16,6 +16,11 @@
     <StrideBuildTags>*</StrideBuildTags>
     <RestorePackages>true</RestorePackages>
     <ExtrasUwpMetaPackageVersion>6.2.12</ExtrasUwpMetaPackageVersion>
+    <LangVersion>12.0</LangVersion>
```
xen2, line 19: Is this still necessary?
cf https://github.com/stride3d/stride/pull/1616#discussion_r1372368581

sources/core/Stride.Core/Stride.Core.csproj
```diff
@@ -16,6 +16,11 @@
...
+  </PropertyGroup>
+  
+  <PropertyGroup Condition="'$(Configuration)|$(TargetFramework)|$(Platform)'=='Release|net8.0|AnyCPU'">
+    <WarningLevel>7</WarningLevel>
```
xen2, line 23: What is this for?

xen2 (commented):
Nice job, thanks again!
I tried to run it and it worked fine. I have just commented on a few small details.
One last thing to do is to update the prerequisite installer. I can do that part if you want.

deps/AssemblyProcessor/net8.0/Stride.Core.AssemblyProcessor.Packed.dll
Doprez: removed

sources/assets/Stride.Core.Assets/Stride.Core.Assets.csproj
Doprez: it used to be broken references? seems to work without now.

sources/core/Stride.Core.Mathematics/Stride.Core.Mathematics.csproj
Doprez: nope these were old net7 workarounds

sources/core/Stride.Core/Stride.Core.csproj
Doprez: it was due to an old net7 issue, seems to work without now.

Doprez:
> Nice job, thanks again! I tried to run it and it worked fine. I have just commented on a few small details. One last thing to do is to update the prerequisite installer. I can do that part if you want.

No problem! I think I got everything you suggested done. I don't know what the prereq installer is or where it is? The only thing I see is a VS2019 and a DriectX installer.

If you want to do it then we could the the peas in a pod achievemnt in GitHub... wink wink. But I dont mind looking into it lol

Jklawreszuk:
@Doprez Currently we are using propertary installer : _Advanced Installer 19.4_ and prereq file is _prerequisites.aip_

Doprez:
Ah I see it now, thank you. should this not be downloading the SDK instead of the Runtime? or is this not for the GameStudio prereqs
```
    <ROW PrereqKey="FAC9C433D3044829E6669A2C6FF54E6" DisplayName=".NET Desktop Runtime 6.0.2 x64" VersionMin="6.0" SetupFileUrl="https://download.visualstudio.microsoft.com/download/pr/efa32b7a-6eec-4d97-9cdc-c7336a29a749/3df4296170397cf60884dae1be3d103b/windowsdesktop-runtime-6.0.2-win-x64.exe" Location="1" ExactSize="57296456" WinNTVersions="Windows 9x/ME/NT/2000/XP/Vista/Windows 7/Windows 8 x86/Windows 8.1 x86/Windows 10 x86" WinNT64Versions="Windows Vista x64, Windows Server 2008 x64, Windows 7 RTM x64, Windows Server 2008 R2 x64, Windows 8 x64, Windows Server 2012 x64, Windows 10 version 1507 x64, Windows 10 version 1511 x64" Operator="1" NoUiComLine="/q /norestart" Options="xy" MD5="ce032ed3b098ceb4054eaf9c9158e2f5" TargetName=".NET 6.0"/>
```

xen2:
I have triggered some teamcity builds.
I think the OpenGL errors will be fixed once latest `master` is merged.
Not sure about the Android build issues though. Something might have changed in dotnet build system.

Doprez:
> I have triggered some teamcity builds. I think the OpenGL errors will be fixed once latest `master` is merged. Not sure about the Android build issues though. Something might have changed in dotnet build system.

Updated with master so the opengl should hopefully work now. I can try later to build an android project when I have time to see if something changed.

Kryptos-FR:
Consider updating the `README.md` with the correct dependencies for building with .NET 8.

sources/targets/Stride.Core.props
```diff
@@ -60,12 +58,12 @@
   </PropertyGroup>
 
   <PropertyGroup Condition=" '$(StrideRuntime)' == 'true' ">
-    <!-- Add net6.0 no matter what (needed for references) -->
-    <StrideRuntimeTargetFrameworks>net6.0</StrideRuntimeTargetFrameworks>
-    <StrideRuntimeTargetFrameworks Condition="$(_StridePlatforms.Contains(';Windows;')) And '$(StrideRuntimeWindowsNet6)' == 'true'">$(StrideRuntimeTargetFrameworks);net6.0-windows7.0</StrideRuntimeTargetFrameworks>
+    <!-- Add net8.0 no matter what (needed for references) -->
+    <StrideRuntimeTargetFrameworks>net8.0</StrideRuntimeTargetFrameworks>
+    <StrideRuntimeTargetFrameworks Condition="$(_StridePlatforms.Contains(';Windows;')) And '$(StrideRuntimeWindowsNet6)' == 'true'">$(StrideRuntimeTargetFrameworks);net8.0-windows7.0</StrideRuntimeTargetFrameworks>
```
Kryptos-FR, line 63: Shouldn't `$(StrideRuntimeWindowsNet6)` be renamed to `$(StrideRuntimeWindowsNet8)`?

Doprez: good point, maybe even `StrideRuntimeWindowsNetCore`?

Kryptos-FR:
@Doprez can you reset your branch to the commit before merging `master` (should be76186a2737653a449987425c3a19dbb0d023f51c)? And then force-push your branch.
`master` currently contains binary files where it shouldn't (they should have been added through git LFS).

xen2:
I took the liberty to do the merge again properly.
Howvever I didn't have the permissions to push it to your repo.
You can reset on this branch: https://github.com/stride3d/stride/tree/update-dotnet-8
or this commit: https://github.com/stride3d/stride/commit/e1d97deaf5fb16658d5b4418f21bb80ddc763dc6

xen2:
I have also added another commit which should hopefully fix android build

PR #1631, opened 2023-03-16
[Graphics] Fixes Vsync off when using new flip model

azeno:
# PR Details

## Description

Vsync off was no longer working since we use the new flip model (#1594).
This PR re-enables that feature, taking most of the code from https://learn.microsoft.com/en-us/windows/win32/direct3ddxgi/variable-refresh-rate-displays

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Graphics/Direct3D/SwapChainGraphicsPresenter.Direct3D.cs
+++ b/sources/engine/Stride.Graphics/Direct3D/SwapChainGraphicsPresenter.Direct3D.cs
@@ -47,6 +47,10 @@ namespace Stride.Graphics
 
         private int bufferCount;
 
+        private readonly bool flipModelSupport;
+
+        private readonly bool tearingSupport;
+
 #if STRIDE_GRAPHICS_API_DIRECT3D12
         private int bufferSwapIndex;
 #endif
@@ -56,6 +60,9 @@ namespace Stride.Graphics
         {
             PresentInterval = presentationParameters.PresentationInterval;
 
+            flipModelSupport = CheckFlipModelSupport(device);
+            tearingSupport = CheckTearingSupport(device);
+
             // Initialize the swap chain
             swapChain = CreateSwapChain();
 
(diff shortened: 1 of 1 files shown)


sources/engine/Stride.Graphics/Direct3D/SwapChainGraphicsPresenter.Direct3D.cs
```diff
@@ -426,21 +465,26 @@ private SwapChain CreateSwapChainForDesktop(IntPtr handle)
...
-            }
+        private SwapChainFlags GetSwapChainFlags()
+        {
+            var flags = SwapChainFlags.None;
+            if (Description.IsFullScreen)
+                flags |= SwapChainFlags.AllowModeSwitch;
+
+            // From https://learn.microsoft.com/en-us/windows/win32/direct3ddxgi/variable-refresh-rate-displays
+            // It is recommended to always use the tearing flag when it is supported.
+            if (tearingSupport)
```
manio143, line 482: Should you check Is Fullscreen here as well (based on the comment on tearing above)?

xen2:
LGTM, thanks!

sources/engine/Stride.Graphics/Direct3D/SwapChainGraphicsPresenter.Direct3D.cs
```diff
@@ -56,13 +60,39 @@ public SwapChainGraphicsPresenter(GraphicsDevice device, PresentationParameters
...
 
             // Reload should get backbuffer from swapchain as well
             //backBufferTexture.Reload = graphicsResource => ((Texture)graphicsResource).Recreate(swapChain.GetBackBuffer<SharpDX.Direct3D11.Texture>(0));
+
+            static bool CheckFlipModelSupport(GraphicsDevice device)
+            {
+                // From https://github.com/walbourn/directx-vs-templates/blob/main/d3d11game_win32_dr/DeviceResources.cpp#L138
+                using var dxgiDevice = device.NativeDevice.QueryInterface<SharpDX.DXGI.Device>();
+                using var dxgiAdapter = dxgiDevice.Adapter;
+                using var dxgiFactory = dxgiAdapter.GetParent<SharpDX.DXGI.Factory4>();
```
Basewq, line 79: Technically this bug was introduced in the #1594 pull, but can you wrap this around a try-catch block?
On Win7, this call throws a SharpDXException with `E_NOINTERFACE/No such interface supported`, so should return false on exception (this is because FlipDiscard is only supported on Win10+.

sources/engine/Stride.Graphics/Direct3D/SwapChainGraphicsPresenter.Direct3D.cs
```diff
@@ -56,13 +60,39 @@ public SwapChainGraphicsPresenter(GraphicsDevice device, PresentationParameters
...
+                return dxgiFactory != null;
+            }
+
+            static unsafe bool CheckTearingSupport(GraphicsDevice device)
+            {
+                // From https://learn.microsoft.com/en-us/windows/win32/direct3ddxgi/variable-refresh-rate-displays
+                using var dxgiDevice = device.NativeDevice.QueryInterface<SharpDX.DXGI.Device>();
+                using var dxgiAdapter = dxgiDevice.Adapter;
+                using var dxgiFactory = dxgiAdapter.GetParent<SharpDX.DXGI.Factory5>();
```
Basewq, line 88: I suspect this would also need to be wrapped in try-catch.

sources/engine/Stride.Graphics/Direct3D/SwapChainGraphicsPresenter.Direct3D.cs
azeno: Thanks, try/catch has been added.

sources/engine/Stride.Graphics/Direct3D/SwapChainGraphicsPresenter.Direct3D.cs
azeno: Thanks, try/catch has been added.

manio143 (approved):

PR #1639, opened 2023-03-31
[Graphics] Fixes crash when flip model is not supported

azeno:
# PR Details

This is a follow up on #1631 - while debugging a different issue it turned out that the `AllowTearing` flag must not be set when not using the flip model.

## Types of changes

## Checklist


manio143 (approved):
Looks good. The failing unit test seems unrelated.

PR #1656, opened 2023-05-03
fix Crash when NativeDeviceContext is null

Doprez:
# PR Details

I added a simple null check as when I was working with the source built version of stride this caused crashes about 3 times.

## Description

I am not 100% sure what causes this to be null since this seems to be called per frame but there were a few times when I added some new models that when I selected them the Gamestudio would pop up the crash message. 

I could not recreate this consistantly but after adding this I wasnt getting anymore issues.

## Motivation and Context

Should fix some crash issues when NativeDeviceContext is null

## Types of changes

## Checklist


manio143:
While it would be good to understand better the root cause, this change seems ok - you can only clear the state if there's an object otherwise it's a no op.

PR #1658, opened 2023-05-07
Change physics null exceptions

Doprez:
# PR Details

Added some null checks and changed from throwing errors to just returning the function and logging the error.

## Description

Obsdark and I were getting errors from trying to remove physics entities from a scene however there seems to be an issue with the components still trying to update with null data. 

I would love some input on this as I feel like this is 100% a bandaid fix for a larger issue with how Synced entities are removed from a Scene.

## Motivation and Context

This should prevent the engine from hard crashing due to null exceptions.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Physics/Elements/CharacterComponent.cs
+++ b/sources/engine/Stride.Physics/Elements/CharacterComponent.cs
@@ -26,7 +26,8 @@ namespace Stride.Physics
         {
             if (KinematicCharacter == null)
             {
-                throw new InvalidOperationException("Attempted to call a Physics function that is avaliable only when the Entity has been already added to the Scene.");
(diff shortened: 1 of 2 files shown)


Doprez:
image
one thing to note with this current PR is the logs do get spammed with the above error but it seems to stop after a few seconds, I assume, due to the garbage collector?

NicusorN5:
> I assume, due to the garbage collector?

Clearly not, this is probably just buggy multi threaded code. The C# garbage collector just frees variables and handles that are out of scope.

Doprez:
After sleeping and coming back to this I realized my issue was my own fault. I had a custom updater that was calling the empty references.

PR #1673, opened 2023-06-02
[WIP] Removing the NativeStream

IXLLEGACYIXL:
# PR Details

The Target is to remove entirely NativeStream, IO namespace and StreamExtensions

## Description
1. removed NullSerializer, as it didnt get implemented in 5 years and was unused

1. Only StreamExtension is used of the NativeStream, which doesnt use BinaryPrimitives.
2. Replacing the StreamExtension with locally overwriting the methods with BinaryPrimitives
3. Using more accurate types to have "less" casting
4. Starting with the Serialize methods. ( idk how to write the code more efficient and with less duplication )
5. Replacing write methods
6. Replacing read methods
7. Deleting IO namespace, its not used anymore

## Motivation and Context

NativeStream is Obsolete and can be removed afterr this, same with StreamExtension and the entire IO namespace in Stride.Core
Having less Obsolete stuff laying around in the engine,

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue ( issue is being obsolete) )

## Checklist

I dont know if NativeStream or StreamExtension appear in any Documentation, i couldnt find any, it could be.

- [x] All new and existing tests passed. ( i couldnt run assets tests )

## Help

I need help in testing this, i cant build the assets tests.
Ive ran all tests but idk how much the Reader is covered by the tests.
So some extra tests would be appreciated

Additional to that there is alot of redundant code, but idk how to reduce it, if there is an option to do it then i would appreciate it, then i can do it myself too but i dont have enough knowledge for that

In some places a BinaryReader can improve the redundancy, for example the float reader could be implemented as 
```cs
using var reader = new BinaryReader(UnderlyingStream);
float value = reader.ReadSingle();
```

but idk if thats more efficient/better in the long run. i followed the standard given in the StreamExtension as this standard worked before.


IXLLEGACYIXL:
@dotnet-policy-service agree

IXLLEGACYIXL:
conversion of double has a faulty converter.. allocates size of float instead of double

IXLLEGACYIXL:
double serializer is not covered by any test and not used at all anywhere..

manio143 (changes requested):
All of the array allocations must go 🙂
As for testing - I suggest making a small program that writes a file using old version and reads with the new version and vice versa.

sources/core/Stride.Core/Serialization/Binary/BinarySerializationReader.cs
```diff
@@ -33,55 +33,86 @@ public override void Serialize(ref bool value)
             value = result != 0;
         }
 
-#pragma warning disable CS0618 // Type or member is obsolete
         /// <inheritdoc />
         public override unsafe void Serialize(ref float value)
         {
-            fixed (float* valuePtr = &value)
-                *((uint*)valuePtr) = UnderlyingStream.ReadUInt32();
+            var buffer = new byte[sizeof(float)];
```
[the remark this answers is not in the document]
IXLLEGACYIXL: oke, i will look into it

IXLLEGACYIXL:
is the new BinaryWriter now surely allocation free?
I made a benchmark test, that i posted in discord.
Memory Allocation stays always the same, no matter how much i write to the serialization.

Also added some tests for the writer

IXLLEGACYIXL:
the string tests fail.. but i didnt even change that
```
 Stride.Core.Tests.TestSerialization.TestSerializationBaseTypes(serializationBackend: Binary)
   Quelle: TestSerialization.cs Zeile 186
   Dauer: 13 ms

  Nachricht: 
System.IO.EndOfStreamException : Unable to read beyond the end of the stream.

  Stapelüberwachung: 
BinaryReader.ReadString()
BinarySerializationReader.Serialize(String& value) Zeile 120
StringSerializer.Serialize(String& obj, ArchiveMode mode, SerializationStream stream) Zeile 44
MemberNullableSerializer`1.Serialize(T& obj, ArchiveMode mode, SerializationStream stream) Zeile 153
TestSerialization_SerializeBaseTypeTestSerializer.Serialize(SerializeBaseTypeTest& obj, ArchiveMode mode, SerializationStream stream)
TestSerialization.CopyBySerialization[T](T obj, SerializationBackend serializationBackend) Zeile 129
TestSerialization.TestSerializationBaseTypes(SerializationBackend serializationBackend) Zeile 210
```

IXLLEGACYIXL:
All tests work now again

IXLLEGACYIXL:
alot of the writes and reads are unnecessary as teh binarywriter allready implements it the same way

sources/core/Stride.Core/Serialization/Binary/BinarySerializationReader.cs
```diff
@@ -33,57 +34,85 @@ public override void Serialize(ref bool value)
...
         /// <inheritdoc />
         public override unsafe void Serialize(ref float value)
         {
-            fixed (float* valuePtr = &value)
-                *((uint*)valuePtr) = UnderlyingStream.ReadUInt32();
+            Span<byte> buffer = MemoryMarshal.Cast<float, byte>(MemoryMarshal.CreateSpan(ref value, 1));
+            var read = UnderlyingStream.Read(buffer);
+            if (read != sizeof(float))
+                throw new EndOfStreamException();
+            value = BinaryPrimitives.ReadSingleLittleEndian(buffer);
```
manio143, line 44: Is this still necessary? The read should directly populate the memory of the `value` - applies to every method.

sources/core/Stride.Core/Serialization/Binary/BinarySerializationReader.cs
```diff
@@ -100,19 +129,21 @@ public override void Serialize(ref char value)
...
         public override void Serialize(ref byte value)
         {
-            var result = UnderlyingStream.ReadByte();
-            if (result == -1)
+            Span<byte> buffer = MemoryMarshal.CreateSpan(ref value, 1);
+            var read = UnderlyingStream.Read(buffer);
+            if (read != sizeof(byte))
                 throw new EndOfStreamException();
-            value = (byte)result;
+            value = buffer[0];
```
manio143, line 136: Same here - `value` should already be populated

sources/core/Stride.Core/Serialization/Binary/BinarySerializationWriter.cs
```diff
@@ -31,55 +32,109 @@ public override void Serialize(ref bool value)
...
         }
 
-#pragma warning disable CS0618 // Type or member is obsolete
         /// <inheritdoc />
         public override unsafe void Serialize(ref float value)
         {
-            UnderlyingStream.Write(Unsafe.As<float, uint>(ref value));
+            unsafe
+            {
+                fixed (float* ptr = &value)
```
manio143, line 40: You can use the same MemoryMarshal here as with the reads to avoid `fixed`.
Maybe it makes sense to create a helper generic method `Span<byte> MemoryMarshalAsBytes<T>(ref T value)` to save on space in all those methods.

manio143 (changes requested):
Left additional comments - let me know if I need to clarify things further.

sources/core/Stride.Core/Serialization/Binary/BinarySerializationWriter.cs
Eideren: `Writer.Write` has support for float though, so you can just do 
```cs
public override unsafe void Serialize(ref float value)
{
    Writer.Write(value);
}
```

IXLLEGACYIXL: cant we then just use a binarywriter

Eideren: Right, sorry, I haven't looked through your PR. I don't know the whole context but if it's not a breaking change yeah. Though it's likely that you'll have to adapt anything that depends on this, if you're up for the task go ahead :)

IXLLEGACYIXL: Do you know why they had the native stream at all back then ? why didnt they use the binarywriter

manio143: If you look at #1469 - I think the whole Native stream was to allow having a memory stream over unmanaged memory, which became a thing in .NET 5.
If we can fully replace this code with something that comes in the framework out of the box, let's do it.

PR #1675, opened 2023-06-03
[WIP] Remove obsolete code

IXLLEGACYIXL:
# PR Details

Removing unused Obsolete stuff that isnt used.

## Description

The Core module has alot of Obsolete code or code that could be replaced with more modern methods.

## Motivation and Context

The Core has many obsolete methods that can be removed and arent used at all.
Maintenance would be easier when 

## Types of changes

## Checklist

## Help
I need help in testing this..im not sure if it didnt destroy anything.
The Tests passed in VS.


sources/core/Stride.Core/Utilities.cs
```diff
@@ -111,37 +111,6 @@ public static unsafe void FreeMemory(nint alignedBuffer)
             }
         }
 
-        /// <summary>
-        ///   Read stream to a byte[] buffer
-        /// </summary>
-        /// <param name = "stream">input stream</param>
-        /// <returns>a byte[] buffer</returns>
-        [Obsolete("Allocates. Read into the destination.")]
```
Ethereal77, line 119: The reason this method was marked obsolete was that it allocates a buffer for copying the stream data into and it was preferable a variant that copies directly to a provided buffer.
Your alternative using `MemoryStream` does the same.
Would not be better to provide the alternative that can read a `Stream` directly to a `Span<byte>`?

sources/engine/Stride/Graphics/Image.cs
```diff
@@ -557,9 +557,9 @@ public static unsafe Image Load(byte[] buffer, bool loadAsSRGB = false)
         /// <remarks>This method support the following format: <c>dds, bmp, jpg, png, gif, tiff, wmp, tga</c>.</remarks>
         public static Image Load(Stream imageStream, bool loadAsSRGB = false)
         {
-            if (imageStream == null) throw new ArgumentNullException("imageStream");
+            if (imageStream == null) throw new ArgumentNullException("imageStream is null");
```
Jklawreszuk, line 560: Adding "is null" is not necessary. You will receive following message : 
```
System.ArgumentNullException: Value cannot be null. (Parameter 'imageStream is null')
```

sources/core/Stride.Core/Utilities.cs
IXLLEGACYIXL: i will check that out

sources/engine/Stride/Graphics/Image.cs
Jklawreszuk, line 560: I think adding "is null" is not necessary. You will receive following message : 
```bash
System.ArgumentNullException: Value cannot be null. (Parameter 'imageStream is null')
```

IXLLEGACYIXL: will remove it.

sources/core/Stride.Core/Utilities.cs
IXLLEGACYIXL: what about that? i dont really get how i can convert a span to a byte[] else
```csharp
    public class DataContainer 
    {
        public byte[] Data;

        public static DataContainer Load(Stream stream)
        {
            byte[] dataArray = new byte[stream.Length];
            stream.Read(dataArray);
            return new DataContainer { Data = dataArray };
        }
```
if i read directly to the span it would look like that ( i dont have that much expierence with spans )
```csharp
        public byte[] Data;

        public static DataContainer Load(Stream stream)
        {
            Span<byte> buffer = new Span<byte>();
            stream.Read(buffer);
            return new DataContainer { Data = buffer.ToArray() };
        }
```
this would again including copying the content according to ToArray docs of span

manio143: Using spans only makes sense if you're working within the stack for the operations. Here the class uses an array in memory for long term access which means using the array is the valid approach.

sources/editor/Stride.Core.Assets.Editor/ViewModel/AssetCollectionViewModel.cs
```diff
@@ -244,9 +244,6 @@ public AssetCollectionViewModel([NotNull] IViewModelServiceProvider serviceProvi
 
         public int AssetCount => assets.Count;
 
-        // FIXME: this property was added because for some reason DataGridEx has a lot of issues when displaying both assets and folders
```
manio143, line 247: I'm a little worried about this comment here, but my understanding is that this method was no longer referenced for you to remove it, so maybe the comment is obsolete.

manio143 (approved):

sources/editor/Stride.Core.Assets.Editor/ViewModel/AssetCollectionViewModel.cs
IXLLEGACYIXL: what should i do?

sources/core/Stride.Core/Utilities.cs
IXLLEGACYIXL: will apply that

sources/editor/Stride.Core.Assets.Editor/ViewModel/AssetCollectionViewModel.cs
IXLLEGACYIXL: it was unused and DataGridEx is empty.. for "future developement".

sources/engine/Stride/Graphics/Image.cs
IXLLEGACYIXL: done

sources/core/Stride.Core/Utilities.cs
IXLLEGACYIXL: https://github.com/stride3d/stride/pull/1675/commits/6df7f021f3fb9ad07a3841462361354eee931de5
done with that

sources/editor/Stride.Core.Assets.Editor/ViewModel/AssetCollectionViewModel.cs
Eideren: Isn't that property still used over here ? https://github.com/stride3d/stride/blob/a19a8fb9a2af3e79b61c4d3185efcf5b3e3cbd6a/sources/editor/Stride.Core.Assets.Editor/View/AssetViewUserControl.xaml#L181-L187
Or is that stuff also obsolete ?

PR #1677, opened 2023-06-08
Remove VisualStudio reattach deadcode

Jklawreszuk:
# PR Details

## Description

PR focuses on removing an unnecessary piece of code. The code attaches the Visual Studio debugger after every program reset to the GameStudio process in the **NET Framework target only**, making this code basically dead. I recommend attaching through VS directly. See article about [reattaching to process](https://devblogs.microsoft.com/devops/reattach-to-process-in-visual-studio-2017) or use [ReAttach ](https://marketplace.visualstudio.com/items?itemName=ErlandR.ReAttach) plugin instead.

## Types of changes

- [x] Refactoring / dependency upgrade

## Checklist


Eideren:
I think this was made specifically to catch issues that occur early in the application lifetime, going through the motion to reattach the debugger to the process may take too much time.
I'll merge this in since as you said the gamestudio already lost that feature but I think this could still be nice to bring back.

Issue #1671, opened 2023-05-27
Move VisualStudio related projects to another solution file


Jklawreszuk:
**First problem**

As many may be noticed, current build system is based on **.NET Framework**.
As a developer I want to build Stride engine using built-in NET5+ msbuild equivalent. Take a look an example :
```cmd 
dotnet msbuild /t:Build Stride.sln
```
Very simple.

However, this will fail because of following error:
```
error MSB4801: task factory "CodeTaskFactory" is not supported on the .NET Core version of MSBuild - Stride.VisualStudio.Package.csproj
```
And the line causing an error:
```xml
<Import Project="$(VSToolsPath)\VSSDK\Microsoft.VsSDK.targets" .../>
```
The most obvious option would be to migrate the entire VisualStudio plug-in project to NET Core, but I don't really believe it would be that easy, and I also don't believe we have enough contributors to take care of it.
I would like to mention that I am not 100% sure if only VisualStudio plugin causes an errors, but It would be nice to separate this from NET Core projs.

**Second problem**

What if a developer wants to build the whole engine, but doesn't want to use VisualStudio or just doesn't want to install any plug-in?
I think this should be **optional**, but it's not, because when you execute `dotnet msbuild /t:Stride.sln`, the Stride.sln file contains VisualStudio projects. 
Moreover, because of this, any non-Windows developer can't build either, making the whole process dependent on Windows. 

**Todo (potential solution)**

- [x] - Make Stride Stride.VisualStudio.sln file
- [x] - Figure out if the change affects the engine

manio143:
I believe that wouldn't be enough to drop the dependency on Visual Studio MSBuild - the C++ projects (not sure if just C++/CLI or also win-x64 native) need to be built from VS Build Tools environment.
But it's definitely a good goal to have at making Stride build system modernized.

PR #1684, opened 2023-06-10
Separate VisualStudio projects from main solution

Jklawreszuk:
# PR Details

## Related Issue

Solves issue #1671 

## Types of changes

### DIFF
--- a/build/Stride.VisualStudio.sln
+++ b/build/Stride.VisualStudio.sln
@@ -1,0 +1,49 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 17
+VisualStudioVersion = 17.0.31903.59
+MinimumVisualStudioVersion = 10.0.40219.1
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Stride.VisualStudio.Commands", "..\sources\tools\Stride.VisualStudio.Commands\Stride.VisualStudio.Commands.csproj", "{EF920B86-36CA-4457-8D6C-1F300F97C355}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Stride.VisualStudio.Commands.Interfaces", "..\sources\tools\Stride.VisualStudio.Commands.Interfaces\Stride.VisualStudio.Commands.Interfaces.csproj", "{C2E8CBA7-04A8-44F9-ABA2-7D245AFF88E7}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Stride.VisualStudio.Package", "..\sources\tools\Stride.VisualStudio.Package\Stride.VisualStudio.Package.csproj", "{740ECD08-3EF4-40D9-B0D3-B37BC354F79D}"
+EndProject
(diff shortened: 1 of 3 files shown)


manio143:
Would be good to have a look at how the TeamCity builds the VS Plugin for publishing.
Another idea is to create a solution filter for the editor which would exclude the plugin for those interested in the editor only (with appropriate docs change).

Jklawreszuk:
@manio143 To be honest, I do not have any experience using TeamCity, but here is (propably) Stride's TeamCity pipeline diagram:

image

First, It runs "Package" target
```bash
MSBuild.exe Stride.build -t:Package @C:\BuildAgent\temp\agentTmp\1.rsp /m /nr:false /p:StridePlatforms="Windows;Android;iOS;Linux;macOS" /p:StrideGraphicsApiDependentBuildAll=true /p:StrideEditorTargetFrameworks="net472;net5.0-windows" /p:StrideSign="true" /p:StrideSignUser="sc-stride@dotnetfoundation.org" /p:StrideSignSecret="*******"
```
And If we take a look into stride.buid file, "Package" target executes other targets i.e. changed VSIXPlugin target:
```xml
<Target Name="Package" DependsOnTargets="PreparePackageBuild;$(_StridePackageTargetPrerequisitesInstaller);Build;SignPackage;VSIXPlugin"/>
```

Finally, TC is deploying vsix packages 
```bash
MSBuild.exe Stride.build /t:PublishVSIX @C:\BuildAgent\temp\agentTmp\1.rsp /p:StrideStoreUrl=https://api.nuget.org/v3/index.json /p:StrideStoreApiKey=*******
```
and definition:
```xml
  <Target Name="PublishVSIX">
    <Error Condition="'$(StrideStoreApiKey)' == ''" Text="Missing env variable StrideStoreApiKey"/>
    <Exec Command='"$(NuGetPath)" push ..\bin\vsix\*.nupkg -ApiKey $(StrideStoreApiKey) -Source $(StrideStoreUrl) -NoSymbols'/>
  </Target>
  ```
  
**TLDR:** Imho, It should work 😅. Question I have is using solution filter more beneficial than just move projects to solution file?

manio143:
@Jklawreszuk Your explanation is great and I think it's fine to do it like this 😊
Since very few devs would compile the VS Plugin locally I think it makes sense to have its own solution.

Eideren:
Alright, let's merge this in then :)
No objections @manio143 / @Jklawreszuk ?

manio143 (approved):

Jklawreszuk:
@Eideren No, I think it should be merged then :)

xen2:
We can do that as a later step, but I remember checking about .slnf (Solution Filter) and thinking they would be good for such use case: no need to maintain several .sln, and it's easy to load everything when doing refactors.

I remember postponing it because MSBuild/Teamcity didn't support it well at the time, but it might be better now.

Jklawreszuk:
@xen2 But my intention was to logically separate the actual engine compilation process from the plugin compilation for Visual Studio, I'm not sure if the solution filter solves that. So I decided to create another sln project - yes, it makes another solution to maintain, but this approach is clear, and current contributors rather rarely look at this code anyway. I'm not saying the solution filter is a bad idea, but I just don't see it

tebjan:
@Jklawreszuk this PR appears to have led to the Stride.VisuallStudio nugets are not being released anymore. but @xen2 just pushed a quick fix...

PR #1688, opened 2023-06-13
Remove native code from stride.core project

Jklawreszuk:
# PR Details

This PR tries to remove native files from the entire Stride.Core project

## Checklist


sources/engine/Stride.Native/NativeInvoke.cs
```diff
@@ -43,7 +43,7 @@ internal class Module
         [ModuleInitializer]
         public static void Initialize()
         {
-            Core.Native.NativeInvoke.Setup();
+            
         }
```
Eideren, line 47: Probably fine to remove it entirely, no ?

Eideren (commented):
Looks good. Did you have the time to check that linux allocator in the same folder ?

Jklawreszuk:
> Looks good. Did you have the time to check that linux allocator in the same folder ?

@Eideren As far as I can see, the LinuxAllocator.cpp file is currently being used by Stride.Audio and Stride.Navigation, so currently we can't get rid of this file.

sources/engine/Stride.Native/NativeInvoke.cs
Jklawreszuk: Good idea. I removed it and it seems to work anyway

Eideren:
@Jklawreszuk it sounds like we could probably replace it with `Marshal.AllocHGlobal` or something similar, though I haven't looked at how those are used
Thanks !

PR #1704, opened 2023-06-21
fix #1703 Changing the comparison project related and not UPath related

### LINKED ISSUE #1703: Game Studio doesnt reload sub projects after changes
4.1
Any
**Platform(s)**: Does the problem occur on Windows, Android...?
Game Studio
**Describe the bug**
Game Studio doesnt reload other assemblies besides the main one for refreshing the game studio.
Which means:
You have a second Project, it inherits from a interface and is a Datacontract then Gamestudio will not update it unless you restart it as sub projects are only loaded once.
As far as i could track it down:
https://github.com/stride3d/stride/blob/master/sources/editor/Stride.GameStudio/DebuggingViewModel.cs#L176C1-L181
This never Receives from the Buffer any DLLs besides the main one.

This is caused by
https://github.com/stride3d/stride/blob/master/sources/editor/Stride.Assets.Presentation/AssetEditors/ProjectWatcher.cs#L200
```cs
                if (needProjectReload == false
                    && (e.ChangeType == FileEventChangeType.Deleted || e.ChangeType == FileEventChangeType.Renamed || e.ChangeType == FileEventChangeType.Created)
                    && Path.GetExtension(changedFile)?.ToLowerInvariant() == ".cs"
                    && !UPath.Combine(new UFile(trackedAssembly.Project.FilePath).GetFullDirectory(), new UDirectory("obj")).Contains(new UFile(changedFile)))
                {
                    needProjectReload = true;
                }
```
this will be always true for the main project, so it will trigger the reload of the project here
https://github.com/stride3d/stride/blob/master/sources/editor/Stride.Assets.Presentation/AssetEditors/ProjectWatcher.cs#L229
So it will never reach the other dlls

Also swapping the order of the dlls wont help as it will create a endless loop
1. second project rebuilds
2. main project changes so it gets build
3. second project rebuilds
4. ...

**Expected behavior**
It should behave like in the main project where i add a type in code and it properly reloads and i dont have to restart game studio

**Additional context**
The Assembly Registry and TypeDescriptorExtension work.
I managed to get it "working" by skipping hardcoded dlls to check if the other stuff works and it does.
the skip caused 3 times reloading the dlls as:
1. main project changed and gets rebuild
2. second project changed and gets rebuild
3. both projects get rebuild ( which would be expected )

IXLLEGACYIXL:
# PR Details
The Comparison was wrong as the UPath always fits the main Project.
It will never reach the other Projects
So i changed the Comparison to be Project related and not UPath related

## Motivation and Context

that bug was annoying.
Each change in a sub project didnt get reloaded.

## Types of changes

## Checklist

### DIFF
--- a/sources/editor/Stride.Assets.Presentation/AssetEditors/ProjectWatcher.cs
+++ b/sources/editor/Stride.Assets.Presentation/AssetEditors/ProjectWatcher.cs
@@ -216,7 +216,7 @@ namespace Stride.Assets.Presentation.AssetEditors
                 if (needProjectReload == false
                     && (e.ChangeType == FileEventChangeType.Deleted || e.ChangeType == FileEventChangeType.Renamed || e.ChangeType == FileEventChangeType.Created)
                     && Path.GetExtension(changedFile)?.ToLowerInvariant() == ".cs"
-                    && !UPath.Combine(new UFile(trackedAssembly.Project.FilePath).GetFullDirectory(), new UDirectory("obj")).Contains(new UFile(changedFile)))
+                    && trackedAssembly.Project.Documents.Any(x => x.FilePath.Equals(changedFile,StringComparison.OrdinalIgnoreCase)))
                 {
                     needProjectReload = true;
                 }
--- a/sources/editor/Stride.Assets.Presentation/AssetEditors/ProjectWatcher.cs
+++ b/sources/editor/Stride.Assets.Presentation/AssetEditors/ProjectWatcher.cs
@@ -202,7 +202,7 @@ namespace Stride.Assets.Presentation.AssetEditors
             string changedFile;
             var renameEvent = e as FileRenameEvent;
             changedFile = renameEvent?.OldFullPath ?? e.FullPath;
-
+            
             foreach (var trackedAssembly in trackedAssemblies)
             {
                 // Report change of the assembly binary
@@ -214,9 +214,9 @@ namespace Stride.Assets.Presentation.AssetEditors
                 // Also check for .cs file changes (DefaultItems auto import *.cs, with some excludes such as obj subfolder)
                 // TODO: Check actual unevaluated .csproj to get the auto includes/excludes?
                 if (needProjectReload == false
-                    && (e.ChangeType == FileEventChangeType.Deleted || e.ChangeType == FileEventChangeType.Renamed || e.ChangeType == FileEventChangeType.Created)
+                    && ((e.ChangeType == FileEventChangeType.Deleted ||e.ChangeType == FileEventChangeType.Renamed ||  e.ChangeType == FileEventChangeType.Created)
                     && Path.GetExtension(changedFile)?.ToLowerInvariant() == ".cs"
-                    && trackedAssembly.Project.Documents.Any(x => x.FilePath.Equals(changedFile,StringComparison.OrdinalIgnoreCase)))
+                    && changedFile.StartsWith(Path.GetDirectoryName(trackedAssembly.Project.FilePath),StringComparison.OrdinalIgnoreCase)))
                 {
                     needProjectReload = true;
                 }


Ethereal77:
The check for when the modified file was in the `obj` directory was to not reload the project in case a generated file was modified (what other `.cs` files are there in `obj`?).
Why that check? And why was this check making the reload to fail?

Also, this hardcoded `obj` directory name is brittle. Maybe we should (in other PR) make the project system use the MSBuild variables like `$(IntermediateOutputDir)`.

IXLLEGACYIXL:
`Why that check? And why was this check making the reload to fail?`
it failed because the obj path is always in the main project as it includes all others
so if you change your sub project. it detects that the obj path has changed, so it will report that assembly to be reloaded and at that point it doesnt care anymore which file changed.

so it will always report your mian assembly to relaod, it cant ever possible reload the sub project

Ethereal77:
I see. Then the changed files should be tracked at a solution level, then decide to which project it belongs, and reload it.

IXLLEGACYIXL:
you can detect the Project which it belongs to by not comparing with the obj path

change incomming => obj path change => project file change 
thats the order of changes
with my comparison i skip the obj path so i only watch project file changes and these i can find with the documents of the assembly

and all in all it works EXCEPT renaming.. which im currently tracking down
create,
alter,
delete works

IXLLEGACYIXL:
uhhh to be fair... renaming doesnt work in the main project either

hmmm.. so my thing fulfills what is currently "working" in the main project.. which doesnt include renaming

IXLLEGACYIXL:
@Ethereal77 thanks for mentioning that i search through the documents and not the root path of the dir.

Now renaming and deleting and creating works in sub projects.
you get a popup after renaming "files have changed" but it reloads then correctly

i tested it for now .. would be nice if someone else could test it

IXLLEGACYIXL:
what doesnt work is deleting through game studio a file and get a reload as it doesnt trigger the projectwatcher

because it doesnt delete the file at all, it just excludes it so that is intentional i assume and wanted to be.
If you create a script asset in a sub project it should reload right after adding it and you should have the new class available

IXLLEGACYIXL:
from my tests this PR works...  if someone else could check it out it would be nice as i doubt that i can cover everything

there are no unit tests for this and i doubt that im able to write them as i barely understand broadcasting with async

IXLLEGACYIXL:
it doesnt matter which source does it.
If game studio really deletes it then it will trigger the project watcher

but the studio doesnt do it so it wont trigger the watcher
i dont know how this is integrated and i didnt watch into that, i first wanted to fix the project watcher that it handles the developement in VS correctly

trying to delete for this test from gamestudio was the first time i used that functionality

IXLLEGACYIXL:
used it now for a few days and couldnt find any side effects, would be still good if someone else could test it additionally

Eideren:
I'm getting a couple of issues after creating a new script from the asset view,
A popup shows up with:
```
Game Studio can't auto-reload the project file MyGame3.csproj because you have local changes such as new or deleted scripts.
Click OK to keep reloading or Cancel to keep the current version.
```
And the build output shows the following:
```
[CSC(0,0)]: Error: Cannot open 'D:\Documents\Stride Projects\MyGame3\MyGame3\obj\Debug\net6.0-windows\MyGame3.pdb' for writing -- 'The process cannot access the file 'D:\Documents\Stride Projects\MyGame3\MyGame3\obj\Debug\net6.0-windows\MyGame3.pdb' because it is being used by another process.'
```
Pressing the reload button after this does work though.

IXLLEGACYIXL:
i cant recreate that issue.

how did you open stride and VS?
1. i opened stride
2. launched through solution explorer VS
3. create async script in asset view

and this didnt give me an error. How did you do it?

Eideren:
It works fine if I don't launch in debug mode through VS, so probably my VS messing with it. Do you have time to try this out @manio143 ?

Eideren:
Nah, I don't have it enabled either

manio143:
I can confirm that when GameStudio is launched in Debug mode through VS there's something broken with the reloading and locking the pdb file. That happens on save inside GameStudio.

If you afterwards click manually the assembly reload button it works without issue, which suggests the problem is at the point where recompiling is triggered.

When running GameStudio from the launcher I was getting a crash on loading freetype DLL. Not caused by this ofc, but something to look at later.

Running GameStudio from debug bin folder directly (rather than through VS) worked fine.

---

Another thought I had though - in the code we now check that the file path starts with the project folder path. This can be a problem for files explicitly included in the project with `<Compile Include="../Shared.cs" />`, right?

IXLLEGACYIXL:
```
Another thought I had though - in the code we now check that the file path starts with the project folder path. This can be a problem for files explicitly included in the project with <Compile Include="../Shared.cs" />, right?
```
ive tried it

here the problem is that the normal stride and my PR dont cover that case for reloading as the project watcher doesnt cover such a file

you dont get the reload assembly notification in neither stride versions... so there is a problem with the project watcher which is a problem for another PR

IXLLEGACYIXL:
so the summary is:

the locked pdb file also appears in master 
project watcher doesnt cover it in master or my PR

Issue #1707, opened 2023-07-04
Box 2D collider does not work properly when colliding with another Box 2D collider


Jklawreszuk:
**Release Type**: Official Release/GitHub
**Version**: master 
**Platform(s)**: Windows (any?)

**To Reproduce**
1. Create new empty Stride project
2. Add ~Rigidbody component or~ Character component to Sphere entity
	- Add Box Collider Shape and check Is2D flag 
3. Add StaticBody to Ground entity
	- Add Box Collider Shape and check Is2D flag 

**Expected behavior**
Collision should stop entity from moving

**Test cases for collision detection I made**
![image](https://github.com/stride3d/stride/assets/31008367/13f00b69-d282-494d-abd8-de6e30f22d44)

**Video**

https://github.com/stride3d/stride/assets/31008367/5582127e-c6de-4f62-a764-c30c06d7365f

PR #1708, opened 2023-07-05
Fix BoxColliderShape by creating Box2D shape when is2D

Jklawreszuk:
# PR Details

## Description

This PR fixes incorrect BoxCollider detection when the is2D checkbox is selected, simply by creating a 2D BoxShape version.

## Related Issue

#1707 

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Physics/Shapes/BoxColliderShape.cs
+++ b/sources/engine/Stride.Physics/Shapes/BoxColliderShape.cs
@@ -37,7 +37,7 @@ namespace Stride.Physics
 
             if (Is2D)
             {
-                InternalShape = new BulletSharp.Convex2DShape(shape) { LocalScaling = cachedScaling };
+                InternalShape = new BulletSharp.Box2DShape(size / 2) { LocalScaling = cachedScaling };
             }
             else
             {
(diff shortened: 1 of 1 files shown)


Eideren:
So, looks like convex shapes are broken then ?

Jklawreszuk:
@Eideren I don't think so, but since Bullet's documentation is poor, I can only speculate why it is wrong only for the Box shape. 
Let me outline my reasoning process:
In our engine, Convex2DShape is instantiated only with the following shapes: Box (rectangle in 2D), Sphere (circle in 2D) and Capsule (rectangle + 2 circles? - 2D). Let's try to analyze each shape one by one: 
1. **rectangle** - A rectangle in the 2D world is not convex. A rectangle does not meet the criteria for convexity, because if we take two points on its boundary (e.g. two adjacent corners), the line segment connecting them will pass outside the rectangle.
2. **circle** - a circle in the 2D world is convex. For any two points inside the circle, the line segment connecting them will always lie entirely inside the circle. This property is true for any pair of points chosen inside the circle, making it a convex shape.
3. **"Flat" capsule** (rectangle with two circles): If the ends of the capsule are convex curves (e.g., semicircles) and the body connecting them is a straight line segment, then the shape of the capsule can be considered convex. Our capsule meets these criteria and so it is also convex.

**TDLR:** I think because the rectangle is not convex, so putting BoxShape as an argument is not correct / does not make sense.

Eideren:
Thanks :)

Issue #1024, opened 2021-02-19
Fullscreen with Vulkan appears to be Broken


herocrab:
@Eideren: [This issue has a bounty](https://github.com/stride3d/stride/issues/1024#issuecomment-1065571362)
---
**Release Type**: Official Release

**Version**: 4.0.1.1400

**Platform(s)**: Windows

**Describe the bug**
I am not able to go full screen using Vulkan. My project is setup using SDL context, but I tested with the default and found the same results. I am only making calls to Window:

            var window = Game.Window;
            _windowPosition = window.Position;
            window.IsBorderLess = true;
            window.Position = Int2.Zero;
            window.SetSize(resolution); // Breaks Vulkan

            // window.PreferredFullscreenSize = resolution;
            // window.IsFullscreen = true; // Breaks Vulkan

With both approaches the game window goes full screen size but the draw call ceases to run. There is no crash or exceptions thrown within my IDE.

**To Reproduce**
1. Create a new project.
2. Add <StrideGraphicsApi>Vulkan</StrideGraphicsApi> to .csproj.
3. Attempt any of the methods above.

**Expected behavior**
I expect Vulkan to be able to go full screen.

I will create a repo of this and upload when I sit down to code again.

herocrab:
I was able to reproduce this with a default project, targeting Vulkan and then setting Game.Window.Fullscreen to true. Unfortunately I can't upload the repo as it's larger than 10MB.

herocrab:
I found this issue on Focus engine (Stride Fork):

https://github.com/phr00t/FocusEngine/issues/70#issuecomment-782376470

A workaround is to set the screen resolution and fullscreen flag in a configuration file and use it on launch. Apparently resizing Stride when using Vulkan is very complex.

Eideren:
### This issue has a [$200,-](https://opencollective.com/stride3d/projects/bug-bounty-vulkan-fullscreen) bounty, if you are interested please follow the steps described [here](https://github.com/stride3d/stride/wiki/Bounty)

tamamutu:
@stride3d Willing to give this a look if this hasn't been fixed.

tebjan:
sounds great and would be really appreciated. To work on an issue, create a branch and make a draft PR without code so everyone can see that you are working on it. Then start committing to that branch and convert to regular PR when you are ready for review. Thanks!

Eideren:
Fixed in PR #1709

Aggror:
Hi @tamamutu,

[...]

PR #1654, opened 2023-05-01
[Core] Remove some obsolete code

Jklawreszuk:
# PR Details

## Description
I've managed to remove some obsoletes and unused code in the Core module that did not require much refactoring to remove.

## Types of changes

- [x] Docs change / refactoring / dependency upgrade
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] My change requires a change to the documentation.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #1709, opened 2023-07-07
Vulkan Fullscreen Fix

tamamutu:
# PR Details
Fix Stride projects using the Vulkan backend to properly support fullscreen.

## Description
Enabling swapchain recreation in SwapChainGraphicsPresenter.Vulkan.cs so that changes in resolution at runtime don't stop the game from rendering to the screen.

## Related Issue
#1024

## Motivation and Context
Makes window creation more convenient to the end user and makes window behavior consistent across graphics backends.

## Types of changes

## Checklist


NicusorN5:
Shouldn't you mark this as a bug fix in the "types of changes"?

Eideren:
That's fine, we can infer it from the title, and the pr description is likely wip as well

tamamutu:
Update: Issue seems to be that the Vulkan backend doesn't handle swapchain recreation and instead just sets the backbuffer to the screen resolution and renders to the visible window area. When going fullscreen at a resolution different to the screen's current resolution, swapchain recreation is needed, but since the renderer doesn't handle it, the program just breaks and nothing is rendered to the screen. From what I can tell, resizing works just fine if you call OnRecreated() in the Present() method of SwapchainGraphicsPresenter when Vulkan reports that the swapchain is out of date. I'll add some commits to this PR in a sec just so I can make sure this fix doesn't only work on my end.

Eideren:
@tamamutu great, let me know if/when you want me to try it out

tamamutu:
I'd really appreciate it if you could try and get it working on your end, if anything weird comes up or there's any issues with it just let me know and I'll see if I can debug it.

Eideren:
Yep, works as expected, remove the wip whenever you're ready and I'll merge it in.
Just a note for other users, Vulkan can't really be built with this PR as is, it was broken in #1654 and will be fixed in #1697 - you would need to either pull from before pr 1654 or cherry pick what's missing.

tamamutu:
Sweet as, the PR should be good to go now.

PR #1712, opened 2023-07-12
[WIP]Remove unnecessary code

IXLLEGACYIXL:
# PR Details

Aim for removing unused or obsolete code

## Description

Removing Blobstream.
The Methods were never used and EVEN IF they would be used it would not help as you cant Write to the BlobStream as the methods never got implemented
Blob is immutable so writing shouldnt work at all i assume.
But the blobstream didnt get used at all in the engine. 

Im not sure if the Blob is used either at all. Blob is only used in the ObjectDatabase and nowwhere else, but these methods are also unused.So Blob could be also completely deleted.

it may be that users need it? then i would add it back in

Removing LoadStream
it was deprecated as it allocated memory, replaced in the 2 spots were it was used with normal stream methods
Removing MDM_MAIN 
it was not used and obsolete

## Motivation and Context

Small Code improvement

## Types of changes

## Checklist


IXLLEGACYIXL:
the blob is used if the define of STRIDE_PLATFORM_DESKTOP is set.. so its probably not "not used"

manio143:
I'm not sure removing members from Blob and ObjectDatabase because they're unused is a good move - if a certain part of the system is not needed it may be removed, but here there may be use case for those. Let's make sure we understand the functionality those are supposed to provide and only remove them if that functionality is preserved through different means.

IXLLEGACYIXL:
ehh i will let it be as is , as i dont know how to rewrite load image.. maybe someone else can make ;)

PR #1744, opened 2023-09-02
Adds sRGB and Gamma conversions to ColorUtility.sdsl

tebjan:
# PR Details

Adds the basic conversion methods for sRGB, and screen gamma.

Stride already has `ToLinear` methods, which are there same as `SRgbToLinear`, I could not remove them because it could potentially break existing shaders. Stride also has additional methods for single `float` and `float3`, which I have not provided, but I am open to adding this if requested.

## Related Issue

We had this in a temporary file in the vvvv code, but it would be better to put this into the official ColorUtility of Stride:
https://github.com/vvvv/VL.Stride/pull/395#issuecomment-760253956

## Motivation and Context

Getting color issues right when working with designers.

## Types of changes

### DIFF
--- a/sources/engine/Stride.Graphics/Shaders/ColorUtility.sdsl
+++ b/sources/engine/Stride.Graphics/Shaders/ColorUtility.sdsl
@@ -21,4 +21,62 @@ shader ColorUtility
         float3 sRGB = sRGBa.rgb;
         return float4(sRGB * (sRGB * (sRGB * 0.305306011 + 0.682171111) + 0.012522878), sRGBa.a);
     }
+
+    // simple screen gamma conversion
+    float4 GammaToLinear (float4 RGBa, float Gamma = 2.2)
+    {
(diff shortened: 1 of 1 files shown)


VaclavElias:
Should we consider adding this note below to our PR template and perform for both new and existing code? 

If this PR is approved, could we include the following sections for enhanced readability in our documentation: **summary**, **param**, **returns**, and **remarks**?

For reference, you can see an example at the top of this page: https://dotnet.github.io/docfx/docs/basic-concepts.html

Our new documentation will be released soon, and incorporating these changes could greatly improve its quality. 🙂

tebjan:
> **summary**, **param**, **returns**, and **remarks**?

I am not sure if these work in SDSL shader code. In C# code I would say definitely yes!

VaclavElias:
Of course!🤣🤦‍♂️ It's not C# like C# especially if it is not C# 🤣. I will reuse my comment in the C# PRs then..

PR #1747, opened 2023-09-04
[deps] - Remove unused WindowsAPICodePack

Jklawreszuk:
# PR Details

## Description

This PR removes unused dependency from stride project. 

## Related Issue

#1394


manio143:
Do we know what WindowsAPICodePack is?

manio143 (approved):
I looked into history of that csproj and it looks like it was added at the time when  that DLL was manually referenced. When it was converted to NuGet the packing was just forgotten to be removed.

PR #1757, opened 2023-09-14
[WIP] Crossplatform runtime

Jklawreszuk:
# PR Details

As far as I know, it is impossible to create all of Stride's assets outside of the Windows platform. For this reason, developers are still forced to create assets using a program (called AssetCompiler) on Windows just to be able to run the game on other desktop platforms, such as macOS or Linux. 

If you currently want to run your game on Linux/Mac, I highly recommend Manio's tutorial: https://youtube.com/watch?v=BoHEASoQvu8

My work primarly focus on bring AssetCompiler on those OS's by refactoring and replacing references strictly tied to Win platform. In my PR, CompilerApp itself can now run on Linux since its a console application, but there is a lot code to port.

I believe that it's quite a difficult and very long task to work smoothly on all desktop platforms, so this PR is also intended to encourage other contributors (both old and new) to get involved in the process. As always, I will be happy to receive any feedback from you, and if you want to help me out, please write a comment or you can send a PR to my fork on this branch. 

## Current issues / Todo
- [x] - Load asset fonts using SixLabors.Fonts (partialy)
- [x]  -  getting VHACD  to work outside Windows
- [x] -  getting DxtLibTex to work on Linux or find replacement (works!) 
- [x] - Failing to compile SkyBox asset, because of incorrectly generated GLSL shader
- [ ] - testing, commits and code cleanup

## Related Issue

https://github.com/stride3d/stride/issues/360

**Edit : Please, treat this branch as unstable, use it as you like if you want Linux support as soon as possible without building anything on Windows. When I consider that the goal is very close to being achieved, I will try to carefully pick changes into other, smaller PR's (see linux-build branch)**

**Status 17.03.2024:** You can compile following assets : 3D Models except FBX, Sprites, Audio, Fonts (Vulkan only)


jazzay:
Thanks for kicking this off! Perhaps we can collaborate on this in a proper Stride branch at some point. 

Couple points of initial feedback:

- We should avoid hardcoding TargetFramework in all the csproj as those changes will conflict with the net8 migration occurring now. That is meant to be configured centrally via the StrideVariables
- While it's nice to clean up code as you go, maybe avoid that for now as it adds a lot of noise to the PR, and could lead to merge conflicts down the road. Not sure the Stride team philosophy here.

Jklawreszuk:
@jazzay Extra pair of hands is always welcome ! 🙏
Regarding your feedback: 
- Good point, we should avoid hard-coding the net version, but I did so since I believe it requires another PR - which'd automate the upgrading process with, for example, one variable. 
- As for the second suggestion regarding merge conflicts, this is also a good point, I'll try to leave it for last :)

sources/assets/Stride.Core.Packages/Stride.Core.Packages.csproj
```diff
@@ -2,7 +2,7 @@
   <Import Project="..\..\targets\Stride.props" />
   <PropertyGroup>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
-    <TargetFramework>$(StrideEditorTargetFramework)</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
```
Kryptos-FR, line 5: You might want undo those changes and instead cherry-pick the commit 76000fc2 from #1908 .

JeromyWalsh:
This should be tracked in #1922.

xen2:
Concerning VHACD (and other C/C++ library we use), unless there's a good alternative in C# that ticks all the box, I wouldn't try to restrict our search only to C# libraries (since demand for such library in C# is smaller, I suspect on average those library are more likely to have less active developers, possibly lower performance and more bugs due to not enough people working on it, and a higher chance to be discontinued resulting in us having to maintain it, etc.).

However, I agree we don't want to keep it how it is right now, with old binaries and hard to replicate build.

The replacement (VHACD or something else) should be something easy for us to build & keep up to date with upstream (possibly with the new nativepath support @jazzay was working on in #1810).

Jklawreszuk:
@xen2 Yes, I agree, the idea is to have easy to maintain dependencies, so finding similar projects in C# would be best, but not mandatory. Btw I still considered building the missing VHACD.so binary file for Linux, but I failed to do so myself. If anyone would like to try it themselves, here is the VHACD repo https://github.com/stride3d/BulletSharpPInvoke

ViktorJannicke:
Im currently looking into the relation between VHACD and physics. I dont fully understand the need for VHACD and wondered if switching to jolt physics would help getting rid of the dependency. 
But maybe I'm completly off the dependency chain.

sources/assets/Stride.Core.Assets/Compiler/PackageCompiler.cs
```diff
@@ -21,8 +19,12 @@ static PackageCompiler()
         {
             // Compute StrideSdkDir from this assembly
             // TODO Move this code to a reusable method
+
+            var isUnix = Environment.OSVersion.Platform == PlatformID.Unix;
             var codeBase = typeof(PackageCompiler).Assembly.Location;
-            var uri = new UriBuilder(codeBase);
+            // Uri paths starts with // on Unix systems 
+            codeBase = isUnix ? "/" + codeBase : codeBase;
+            var uri = new UriBuilder(new Uri(codeBase));
```
Kryptos-FR, line 27: Absolute paths start with a `/`. Relative paths don't. Are we sure about this change?

Instead of the constant, I would rather use the `Path.DirectorySeparatorChar` static field.

Edit: why are we even using a Uri here since we are only interested with a file path?

sources/core/Stride.Core.Design/IO/UPath.cs
```diff
@@ -292,7 +292,7 @@ public override string ToString()
         [NotNull]
         public string ToWindowsPath()
         {
-            return FullPath.Replace('/', '\\');
+            return OperatingSystem.IsWindows() ? FullPath.Replace('/', '\\') : FullPath;
```
Kryptos-FR, line 295: The method is explicitly named `ToWindowsPath` so it should always return a windows path on any platform.

sources/engine/Stride.Assets/SpriteFont/Compiler/SignedDistanceFieldFontImporter.cs
```diff
@@ -8,226 +8,227 @@
 using Stride.Core.Assets;
 using Stride.Core.IO;
 
-namespace Stride.Assets.SpriteFont.Compiler
+namespace Stride.Assets.SpriteFont.Compiler;
+
+using System.Drawing;
```
Kryptos-FR, line 13: Move usings before namespace declaration.

sources/Directory.Packages.props
```diff
@@ -1,136 +1,137 @@
...
+    <PackageVersion Include="SharpDX.MediaFoundation" Version="4.2.0" />
+    <PackageVersion Include="SharpDX.RawInput" Version="4.2.0" />
+    <PackageVersion Include="SharpDX.XInput" Version="4.2.0" />
+    <PackageVersion Include="Silk.NET.OpenGL" Version="2.20.0" />
+    <PackageVersion Include="Silk.NET.OpenGLES" Version="2.20.0" />
+    <PackageVersion Include="Silk.NET.OpenGLES.Extensions.EXT" Version="2.20.0" />
+    <PackageVersion Include="Silk.NET.OpenXR" Version="2.20.0" />
+    <PackageVersion Include="Silk.NET.Sdl" Version="2.20.0" />
+    <PackageVersion Include="Silk.NET.Windowing.Sdl" Version="2.19.0" />
+    <PackageVersion Include="sixlabors.imagesharp" Version="3.1.2" />
```
Kryptos-FR, line 28: The package name might be case-sensitive on Unix platform.

Kryptos-FR: In addition it seems the file was saved using a different encoding causing the whole content to show as a diff.

Issue #1758, opened 2023-09-14
DX11: Trying to use an external shared texture with NT Handle but no Keyed mutex throws an exception


arturoc:
**Release Type**: Github 

**Version**: master / 636342b207786f5236c1e54db59fdaf3b2e60972

**Platform(s)**:  Windows DX11

**Describe the bug**
When trying to use an external texture shared with an NT handle but no keyed mutex an exception is thrown cause the code only accepts textures that have both shared NT handle and keyed mutex.

**To Reproduce**
1. Create a texture using an NT shared handle but no keyed mutex
2. Try to use it from stride using `SharpDXInterop.CreateTextureFromNative`

**Expected behavior**
The shared texture should be created without problem.

**Log and callstacks**
The application throws an exception at the end of `InitializeFromImpl` in `Texture.Direct3D.cs` while checking the `textureDescription.Options` as the combination is considered valid and the switch case ends up in the default branch throwing an exception: https://github.com/stride3d/stride/blob/636342b207786f5236c1e54db59fdaf3b2e60972/sources/engine/Stride.Graphics/Direct3D/Texture.Direct3D.cs#L178-L198

tebjan:
Closed by #1759

PR #1759, opened 2023-09-14
Fix DX11 shared textures

arturoc:
# PR Details

Right now creating a texture from a native texture with D3D11_RESOURCE_MISC_SHARED_NTHANDLE  will fail unless it is also D3D11_RESOURCE_MISC_SHARED_KEYEDMUTEX. This combination is recomended but not mandatory acording to the specification: https://learn.microsoft.com/en-us/windows/win32/api/d3d11/ne-d3d11-d3d11_resource_misc_flag

## Description

The PR changes the code that checks for shared handles from a switch/case to an if/else where the check is done for flags using a bitwise and instead of comparing the full flag to an specific value. Shared NT is checked first as Shared & SharedNT can be set at the same time.

## Related Issue

https://github.com/stride3d/stride/issues/1758

## Motivation and Context

Allows to use DX11 textures with an NT handle but no keyed mutex

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Graphics/Direct3D/Texture.Direct3D.cs
+++ b/sources/engine/Stride.Graphics/Direct3D/Texture.Direct3D.cs
@@ -175,25 +175,26 @@ namespace Stride.Graphics
             NativeRenderTargetView = GetRenderTargetView(ViewType, ArraySlice, MipLevel);
             NativeDepthStencilView = GetDepthStencilView(out HasStencil);
 
-            switch (textureDescription.Options)
+            if (textureDescription.Options == TextureOptions.None)
             {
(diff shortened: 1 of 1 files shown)


Ethereal77:
I'm in the process of replacing SharpDX with Silk in the PR #1715. I write here to remind myself (or others) this change must be ported to that PR or it will be lost when merged, as that PR is modifying a lot of the internals of `Stride.Graphics` when using the Direct3D APIs.

manio143 (approved):

tebjan:
Good find, I'll merge this, looks sensible that one flag doesn't require the other one.

PR #1764, opened 2023-09-15
[System] Fix binary path location when dlls run from an unexpected executable

Eideren:
# PR Details
A user had an issue running tests from Rider, the folder returned from that function pointed next to the test runner executable instead of the dlls. Pinging @tebjan since this is his area.

## Related Issue
None logged

## Types of changes

## Checklist

### DIFF
--- a/sources/core/Stride.Core/PlatformFolders.cs
+++ b/sources/core/Stride.Core/PlatformFolders.cs
@@ -184,11 +184,10 @@ namespace Stride.Core
         [NotNull]
         private static string GetApplicationBinaryDirectory()
         {
-            var executablePath = GetApplicationExecutableDirectory();
 #if STRIDE_PLATFORM_ANDROID
-            return executablePath;
+            return GetApplicationExecutableDirectory();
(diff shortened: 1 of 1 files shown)


tebjan:
This change assumes that the exe and the binaries are in the same folder, the code you deleted made it possible to provide a clean output directory.

```
/MessyBinaryFolder
Entry.exe
```

We use that in vvvv to have everything in a different folder. Stride.Core.dll is one level up from the exe, and also the asset database and so on...

Eideren:
> This change assumes that the exe and the binaries are in the same folder

I think it's the exact opposite, the previous version assumes the binaries are next to or one level deeper than the executable, my change doesn't assume anything, it will return the directory wherever those binaries may be ?

Unless we're talking about android here, in which case, nothing changed, it already returned the directory where the executable was.

tebjan:
You are right, it takes the folder of `Stride.Core.dll` because that type is in that dll, didn't grasp that on first look. Looks good to me.

PR #1769, opened 2023-09-16
[Math] Add a couple of helpers for Vectors

ch3mbot:
# PR Details

Made a few simple changes to Vector3 and Int3 to add some helpful functions found in other game engines.

## Description

Added the following: 
a Vec3 constructor taking an Int3 (maybe not necessary with the implicit cast)
a function in Vec3 to round it and return an Int3 
an uncast length calculation to Int3 (returns a float similar to Vec3.Length)
an implicit cast form Int3 to Vec3

## Types of changes

### DIFF
--- a/sources/core/Stride.Core.Mathematics/Int3.cs
+++ b/sources/core/Stride.Core.Mathematics/Int3.cs
@@ -144,6 +144,16 @@ namespace Stride.Core.Mathematics
         }
 
         /// <summary>
+        /// Creates a new instance of the <see cref="Int3"/> struct by rounding the X, Y, and Z components of a <see cref="Vector3"/> struct to the nearest integer.
+        /// </summary>
+        public Int3 RoundToInt3(Vector3 value)
+        {
+            X = (int)Math.Round(value.X);
+            Y = (int)Math.Round(value.Y);
+            Z = (int)Math.Round(value.Z);
+        }
+
+        /// <summary>
         /// Gets or sets the component at the specified index.
         /// </summary>
         /// <value>The value of the X, Y, or Z component, depending on the index.</value>
--- a/sources/core/Stride.Core.Mathematics/Int3.cs
+++ b/sources/core/Stride.Core.Mathematics/Int3.cs
@@ -791,6 +791,14 @@ namespace Stride.Core.Mathematics
             z = Z;
         }
 
+        /// <summary>
+        /// Preforms an implicit conversion from <see cref="Stride.Core.Mathematics.Int3"/> to <see cref="Stride.Core.Mathematics.Vector3"/>.
+        /// </summary>
+        public static implicit operator Vector3(Int3 value)
+        {
+            return new Vector3(value);
+        }
+        
 #if WPFInterop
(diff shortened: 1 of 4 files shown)


sources/core/Stride.Core.Mathematics/Int3.cs
```diff
@@ -143,6 +143,16 @@ public Int3(int[] values)
             Z = values[2];
         }
 
+        /// <summary>
+        /// Creates a new instance of the <see cref="Int3"/> struct by rounding the X, Y, and Z components of a <see cref="Vector3"/> struct to the nearest integer.
+        /// </summary>
+        public Int3 RoundToInt3(Vector3 value)
```
manio143, line 149: Can you make this a constructor or an explicit cast instead? And instead of using `Math.Round` I think it would be better to use explicit cast on it like the constructor with `Vector2`
```
public static explicit operator Int3(Vector3 value)
{
    return new Int3((int)value.X, (int)value.Y, (int)value.Z);
}
```

sources/core/Stride.Core.Mathematics/Int3.cs
```diff
@@ -189,6 +199,15 @@ public int Length()
             return (int)Math.Sqrt((X * X) + (Y * Y) + (Z * Z));
         }
 
+        /// <summary>
+        /// Calculates the unrounded length of the vector.
+        /// </summary>
+        /// <returns>The length of the vector unrounded.</returns>
+        public float LengthUnrounded()
```
manio143, line 206: Can you see about coming up a different name here? Like `PreciseLength` maybe?

sources/core/Stride.Core.Mathematics/Int3.cs
Eideren: Casting to int is not the same as rounding, casting to int removes the fractional part of the value (be it positive or negative numbers), while rounding, well rounds up or down depending on the fractional part.
If the method is called `Round` it should be rounding, and that's a very particular operation, I do think it should be a specific function instead of a constructor.

sources/core/Stride.Core.Mathematics/Int3.cs
Eideren: I actually prefer having `Length` first, so either `LengthPrecise` or keeping `LengthUnrounded` is fine too, helps discoverability with intellisense. His name is very explicit in regard to what the function actually does so I actually lean a bit more towards the author's version right now.
Let's see if anyone else can chime in on this.

ch3mbot: `LengthUnrounded` might me somewhat misleading since the `Length` function doesn't actually round the value, but I'm unsure of what the best alternative is.

Eideren: Sure, it truncates the value so you could change its name to `LengthUntruncated` if you want

Eideren:
@ch3mbot should we merge this in or do you still want to change the naming scheme ?

Eideren:
Thanks !

PR #1774, opened 2023-09-17
Refactor Profiler and submit metrics to .NET Diagnostics API

manio143:
# PR Details

## Description

The Stride Profiler was a bit bloated with overloads and logic spread across different places. This PR addresses some of those issues by refactoring the base code. It also adds support for `System.Diagnostics.Metrics` to allow users to consume performance metrics outside of the game.

* Simplification of overloads by using nullable optional parameters
* Moving from longs to TimeSpans for improved typing of time data
* Addition of `ProfilingEventMessage` and modification of `ProfilingMessage` to make the logic of message formatting easier and separated from other concerns of the `ProfilingEvent`
* Move from `Dictionary<object, object>` to `TagList` which should be more efficient and easier on allocations
* Addition of `Meter("Stride.Profiler")` for external metrics consumption

Breaking changes in API
```diff
-profile.SetAttribute("MyAttribute", 5);
+profile.Attributes.Add("MyAttribute", 5);
```

## Related Issue

N/A

## Motivation and Context
I was exploring how to efficiently emit profiling data from a Stride game to existing tools for .NET that allow to measure performance. Initially I was exploring `EventSource` and `EventCounter` (the slightly older APIs) but they seemed less efficient to use directly, so in the end I used the [Metrics](https://learn.microsoft.com/en-us/dotnet/core/diagnostics/metrics).

As I was adding the support for those I saw a lot of complexity in the profiler code and decided to try to refactor it a bit.

We can now inspect the metrics using standard .NET tools
### dotnet-counters
```
dotnet counters monitor -p <pid> --counters Stride.Profiler
```
image

### Open Telemetry
```
var provider = Sdk.CreateMeterProviderBuilder()
  .AddMeter("Stride.Profiler")
  .AddAzureMonitorMetricExporter(o => o.ConnectionString = "")
  .Build();
```
image

## Types of changes

## Checklist


sources/core/Stride.Core/Diagnostics/ProfilingKey.cs
```diff
@@ -39,20 +44,13 @@ public ProfilingKey([NotNull] string name, ProfilingKeyFlags flags = ProfilingKe
         /// <param name="name">The name.</param>
         /// <exception cref="System.ArgumentNullException">parent</exception>
         public ProfilingKey([NotNull] ProfilingKey parent, [NotNull] string name, ProfilingKeyFlags flags = ProfilingKeyFlags.None)
+            : this($"{parent ?? throw new ArgumentNullException(nameof(parent))}.{ValidateNameNotEmpty(name)}", flags)
         {
-            if (parent == null) throw new ArgumentNullException(nameof(parent));
-            if (name == null) throw new ArgumentNullException(nameof(name));
-            Children = new List<ProfilingKey>();
```
Eideren, line 45: I know it's tempting to have a one liner, but it's just not very readable I don't think.
I would rather you send over `$"{parent}.{name}"` regardless of how valid those two values are, and throw in this constructor. In both cases it will throw from this constructor anyway.
Is there a particular reason or use case why `parent` can be whitespace but not `name` ? You don't check parent but you do check name for whitespace.

sources/core/Stride.Core/Diagnostics/ProfilingMessage.cs
```diff
@@ -56,7 +57,25 @@ public ProfilingMessage(int profileId, [NotNull] ProfilingKey profilingKey, Prof
...
+        /// </summary>
+        public override string Text
+        {
+            get => Message?.ToString();
+            set
+            {
+                // ignore null set in base constructor call, otherwise throw
+                if (value != null) throw new NotSupportedException("Set Message instead.");
+            }
+        }
```
Eideren, line 77: This isn't great, it could create a new message with the content of the string provided instead of throwing ? I know you would prefer users to set the message instead but it does make the api more fragile/obtuse.

sources/core/Stride.Core/Diagnostics/ProfilingState.cs
```diff
@@ -142,27 +87,22 @@ public void Begin(string textFormat, params object[] textFormatArguments)
...
+        internal void BeginGpu(long timeStamp, long tickFrequency)
         {
-            EmitEvent(ProfilingMessageType.Begin, null, value0, value1, value2, value3);
-        }
+            // Perform event only if the profiling is running (to save on time calculations)
+            if (!isEnabled) return;
 
-        internal void Begin(long timeStamp)
-        {
+            this.tickFrequency = tickFrequency > 0 ? tickFrequency : Stopwatch.Frequency;
```
Eideren, line 103: Shouldn't this throw instead ? If the frequency tends to be invalid, I would think it's the caller's responsibility to ensure that the frequency is valid ?

Eideren:
LGTM otherwise, nice job @manio143 !

sources/core/Stride.Core/Diagnostics/ProfilingState.cs
manio143: It's confusing because I didn't use `this.` on the right side - basically if the `tickFrequency` has been set we keep it and if not we fall back to the Stopwatch frequency. I might use a nullable field instead ??= assignment here.

sources/core/Stride.Core/Diagnostics/ProfilingMessage.cs
manio143: I didn't think of just setting the text message - sounds good

sources/core/Stride.Core/Diagnostics/ProfilingState.cs
Eideren: > basically if the tickFrequency has been set we keep it

Okay, maybe I'm confused indeed, but afaik, this is what this code does:
``thisFreq = paramFreq > 0 ? paramFreq : Stopwatch.Frequency``
From what you've just said, this is what I think you meant to do:
``thisFreq = thisFreq > 0 ? paramFreq : Stopwatch.Frequency``
Did I understand correctly ?

sources/core/Stride.Core/Diagnostics/ProfilingKey.cs
Eideren: Oh, right, parent is a `ProfilingKey`, my bad

sources/core/Stride.Core/Diagnostics/ProfilingState.cs
manio143: Hmm, no, my mistake (was responding based on diff and missed the param to that method).
Now even I got a little confused. I'll try to make this a bit more clear.
The basic issue is that TickFrequency is Stopwatch based everywhere except for GPU events where it's set on BeginGpu but needs to be available for EndGpu.
I decided to move this set into a property explicitly.

Eideren:
Thanks @manio143 !

@tebjan not sure if you guys are on latest or still 4.0, but this change may require some work on your part if you're using this API. Just notifying you.

Issue #1753, opened 2023-09-13
Stride Launcher Directs Users to Runtime Install When SDK is Missing


MeharDT:
**Release Type**: Official Release

**Version**:
Stride Launcher 5.0.6
Stride 4.1.0.1838

**Describe the bug**
In a development environment without the .NET 6.0 SDK installed, the user is prompted to install the SDK when launching Stride 4.1.x. The prompt will direct them to the download page for the .NET 6.0 Desktop Runtime and not the SDK.

Subsequent launches will no longer warn that the SDK isn't installed, the user will only be notified when attempt to build their project.

**To Reproduce**
1. Launch Stride Launcher 5.0.6 on a PC without the .NET 6.0 SDK installed
2. When the Launcher prompts about a required .NET install or update, press "Yes"
3. Your default browser should launch and direct you to the download page for the Runtime

**Expected behavior**
The Launcher prompt should direct users to the latest version of the .NET 6.0 SDK

NicusorN5:
I tried to reproduce this issue, but Stride opened a link to download the SDK.
![image](https://github.com/stride3d/stride/assets/20599225/21e36611-9510-44f0-acd1-abdb45f5b1a3)
The opened URL was `https://dotnet.microsoft.com/en-us/download`.

MeharDT:
Hmm, that's a different error.

![stride_error](https://github.com/stride3d/stride/assets/397608/7a13c558-0c2c-496b-9823-88ca1249702b)

Clicking Yes will take you here,

`https://dotnet.microsoft.com/en-us/download/dotnet/thank-you/runtime-desktop-6.0.22-windows-x64-installer?cid=getdotnetcore`

If you install the .NET 6.0 SDK separately, the above prompt no longer appears in subsequent launches.

NicusorN5:
This error is hardcoded into NET 6 applications.

Doprez:
ahhh I dont think that's managed by Stride I think thats just the default runtime exception for .NET apps, in this case the GameStudio app.

MeharDT:
Ah, interesting! Is there any reason why @NicusorN5 and I would receive different warnings for not having .NET 6 installed?

Doprez:
There may be a silly overlooked component in the engine lol 
https://github.com/stride3d/stride/blob/896eedbb55d0cb0a4251e5bdf3d4edba83983568/sources/editor/Stride.GameStudio/Program.cs#L234
[...]

PR #1809, opened 2023-09-20
Check to see if the .NET 6.0 SDK is Installed When Launching Stride

MeharDT:
# PR Details
Stride 4.1 requires the .NET 6.0 SDK. The existing FindAndSetMSBuildVersion() method can be bypassed if an older version of the SDK is installed (i.e., 5.0) because the method is checking for .NET Core 3 or above.

After bypassing this prerequisite check, Stride will function normally until the user attempts to build their project.

## Description

The _x.Version.Major_ requirement was raised in Stride.Core.Assets\PackageSessionPublicHelper.cs, the check is performed when the user opens Stride 4.1.x. If the SDK is missing the user is prompted to install it, the detected SDK version is provided to help with troubleshooting.

## Related Issue

Prompted by discussion in issue https://github.com/stride3d/stride/issues/1753

## Types of changes

## Checklist

### DIFF
--- a/sources/editor/Stride.GameStudio/Program.cs
+++ b/sources/editor/Stride.GameStudio/Program.cs
@@ -225,6 +225,19 @@ namespace Stride.GameStudio
                 InitializeLanguageSettings();
                 var serviceProvider = InitializeServiceProvider();
 
+                //Check to see that the required .NET SDK is available, otherwise Stride will launch with an older version of the SDK
+                if (StrideGameStudio.EditorVersionMajor == "4.1" && !RuntimeInformation.FrameworkDescription.ToString().StartsWith(".NET 6"))
+                {
+                    var message = "Could not find required .NET 6.0 SDK.\r\n\r\n" +
+                                  "Check that you have a valid installation with the required workloads, or go to [https://dotnet.microsoft.com/en-us/download](https://dotnet.microsoft.com/en-us/download) and install.\r\n\r\n" +
+                                  "Detected SDK: " + RuntimeInformation.FrameworkDescription.ToString() + "\r\n";
+
+                    await serviceProvider.Get<IEditorDialogService>().MessageBox(message, Core.Presentation.Services.MessageBoxButton.OK, Core.Presentation.Services.MessageBoxImage.Error);
+                    app.Shutdown();
+                    return;
+
+                }
+
                 try
                 {
                     PackageSessionPublicHelper.FindAndSetMSBuildVersion();
--- a/sources/editor/Stride.GameStudio/Program.cs
+++ b/sources/editor/Stride.GameStudio/Program.cs
@@ -48,7 +48,6 @@ using EditorSettings = Stride.Core.Assets.Editor.Settings.EditorSettings;
 using MessageBox = System.Windows.MessageBox;
 using MessageBoxButton = System.Windows.MessageBoxButton;
 using MessageBoxImage = System.Windows.MessageBoxImage;
-using Microsoft.Build.Locator;
 
 namespace Stride.GameStudio
 {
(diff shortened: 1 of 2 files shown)


Eideren:
Given the fact that we'll be upgrading to .Net 8 shortly, we'll have to remember to change this scope as well. How about we future proof it. Checking for an available SDK of the .net version the game studio is built in instead of an explicit one ?

MeharDT:
That's a good idea! Is there a way to programmatically check which version of .net an application was built with?

Eideren:
Well, I thought that your `RuntimeInformation.FrameworkDescription` did, but that seems to only report the version of the runtime, not SDK. So I'm not sure that this PR would even work if the user had an older SDK, or no SDK. Just having the runtime installed is enough to bypass the check afaik.

MeharDT:
Hmm, the .NET SDK now includes the Runtime so that shouldn't be an issue moving forward, but I wasn't thinking of the Runtime as a standalone install. I guess we need some additional checks. 🤔

MeharDT:
I was thinking FindAndSetMSBuildVersion() is deprecated now that only the .NET SDK is supported but we can simply raise x.Version.Major value here for a similar result,

https://github.com/stride3d/stride/blob/c6fd21b1bc974fd3fed0f6089df31bb356b3e574/sources/assets/Stride.Core.Assets/PackageSessionPublicHelper.cs#L41C58-L41C58

I'll update the PR if there are no objections, this method still relies on RuntimeInformation.FrameworkDescription so @Eideren's point is valid.

Eideren:
So, checking that version against [Environment.Version](https://learn.microsoft.com/en-us/dotnet/api/system.environment.version?view=net-7.0#system-environment-version) then ?

MeharDT:
> So, checking that version against [Environment.Version](https://learn.microsoft.com/en-us/dotnet/api/system.environment.version?view=net-7.0#system-environment-version) then ?

From my tests so far I think we can leave it as is because _x.Version.Major_ needs to be updated manually regardless. We also need an exact version match with whatever that version is.

MeharDT:
Updated PR and post.

Now that the .NET SDK includes the Runtime I think this will still help many users avoid issues with out of date SDKs moving forward (and I assume _x.Version.Major_ was meant to remain current anyway).

sources/editor/Stride.GameStudio/Program.cs
```diff
@@ -234,6 +234,7 @@ private static async void Startup(UFile initialSessionPath)
                     var message = "Could not find a compatible version of MSBuild.\r\n\r\n" +
                                   "Check that you have a valid installation with the required workloads, or go to [www.visualstudio.com/downloads](https://www.visualstudio.com/downloads) to install a new one.\r\n" +
                                   "Also make sure you have the latest [.NET 6 SDK](https://dotnet.microsoft.com/) \r\n\r\n" +
+                                  "Detected SDK: " + RuntimeInformation.FrameworkDescription.ToString() + "\r\n\r\n" +
```
Eideren, line 237: This is still the runtime version though, not the sdk, right ? And I think it might be the runtime that the application is under, not the one installed. So if the gamestudio is self contained, which might be the case (?), this'll always report net6 ?

MeharDT: Ah you're right, good catch! I've removed that addition, the change in FindAndSetMSBuildVersion() still works as intended. You can test this by uninstalling the .NET 6 SDK and running a previously built version of this PR.

Eideren:
Thanks !

sources/editor/Stride.GameStudio/Program.cs
manio143: > gamestudio is self contained

I'm not sure that is the case. Looking at the contents of `Stride.GameStudio` NuGet package it is built as runtime dependent.

MeharDT: > > gamestudio is self contained
> 
> I'm not sure that is the case. Looking at the contents of `Stride.GameStudio` NuGet package it is built as runtime dependent.

Hmm, when I tested before the final PR update the "Detected SDK" was still reporting .NET 6.0 despite the SDK/runtime being uninstalled. MSBuildLocator on the other hand did pick up that the SDK was missing.

PR #1826, opened 2023-09-22
Add euler rotation conversion methods for Quaternion struct

Sardelka9515:
# PR Details

Add euler rotation conversion methods to Quaternion struct

## Description

Added `Quaternion.ToEulerXYZ`, `Quaternion.RotationEulerXYZ` and a few overloads

## Motivation and Context

`RigidBodyComponent.ApplyTorque` only accepts an euler rotation, but there doesn't seem to be an offical to convert a quaternion to euler rotation.

I tried to use the code from `Transform.RotationEulerXYZ` to perform the conversion first, but it gave weird results when passed into the physics engine continously. I think it might need some attention?

## Types of changes

## Checklist

### DIFF
--- a/sources/core/Stride.Core.Mathematics/Quaternion.cs
+++ b/sources/core/Stride.Core.Mathematics/Quaternion.cs
@@ -42,6 +42,10 @@ namespace Stride.Core.Mathematics
     [StructLayout(LayoutKind.Sequential, Pack = 4)]
     public struct Quaternion : IEquatable<Quaternion>, IFormattable
     {
+        const float SINGULARITY_THRESHOLD = 0.4999995f;
+        const float RAD_2_DEG = (float)(180.0 / Math.PI);
+        const float DEG_2_RAD = (float)(Math.PI / 180.0);
+
         /// <summary>
         /// The size of the <see cref="Stride.Core.Mathematics.Quaternion"/> type, in bytes.
         /// </summary>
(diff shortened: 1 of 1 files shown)


Eideren:
Hey @Sardelka9515 , thanks for the PR ! At a cursory glance, those new functions seem very close to the variants of `RotationYawPitchRoll` did you try that one out, how do they differ ? Is it just a case of the output of them being closer to what the physics engine expects ?

Sardelka9515:
Hmm, I think I found something. 
The RotationEulerXYZ essentially does the same thing as RotationYawPitchRoll, but the getter `Quaternion.YawPitchRoll` seems to be returning wrong result

Sardelka9515:
@Eideren  Is `YawPitchRoll` intended to be used as the euler rotation? because the [comment here](https://github.com/stride3d/stride/blob/2596083d98755370103493e7d2c2aaf0cde014b3/sources/engine/Stride.Engine/Engine/TransformComponent.cs#L127) says it has inversed z axis.

tebjan:
> The RotationEulerXYZ essentially does the same thing as RotationYawPitchRoll, but the getter `Quaternion.YawPitchRoll` seems to be returning wrong result

Since it uses Decompose and Euler angles are over-defined, it could well be correct. Did you create a quaternion from them and compare the result visually? The numbers might not match, the it could be the same rotation.

Sardelka9515:
Sorry, it seems that it does return the correct result, but cannot be used to apply torque to the physics engine because the rotation direction is not guranteed to be retained? I think an approach to apply quaternion as torque might still be needed.

Issue #1633, opened 2023-03-20
Drag & Drop to the Resource folder (with prompt)


Obsdark:
**Is your feature request related to a problem? Please describe.**
Is an improvement to the usability of the system, you can say it helps to improve velocity of development and fix issues with migrations and order.

**Describe the solution you'd like**
When you drag & drop or in some other manner add a file (i.e. png, 3dmodel, some audio file, etc) to the project through the editor, make the editor show a prompt than suggest to copy such file to the Resource folder of the project, this folder came with the project in VS and his purpose is to use to save this kind of things.

If accepted, the file is copied to the Resource folder of the project and the copy inside the folder is referenced by the asset, instead of simply reference the file on his original position. The prompt could also have a checkbox so to "remind my desition", both this values (the fact than make the copy to the Resource folder first and the checkbox value) are both values in the config file.

**Additional context**
This will help to improve the work of teams making the project more concise and self-contained, which in time improve the chance of share it between different parts of the team, this offer some improvement in quality of life than, although being a minor addition, would result on an improved experience for the application users besides fixing problems related to asset order with an elegant solution than allow for an optional, but visible prevention of problems caused by disorder of the assets.

Eideren:
Right now on the master branch the editor shows a message box to the user when the asset is outside of the resource folder, here are the lines related to this check:
https://github.com/stride3d/stride/blob/3eac07770518617c8c047821f32428aa863f8c8b/sources/editor/Stride.Core.Assets.Editor/ViewModel/AssetCollectionViewModel.cs#L544-L565
We could ask the user whether they want to move the source file instead, and move the file accordingly.
If anyone wants to tackle this, feel free to do so :)

SeleDreams:
> Right now on the master branch the editor shows a message box to the user when the asset is outside of the resource folder, here are the lines related to this check:
> 
> https://github.com/stride3d/stride/blob/3eac07770518617c8c047821f32428aa863f8c8b/sources/editor/Stride.Core.Assets.Editor/ViewModel/AssetCollectionViewModel.cs#L544-L565
> 
> We could ask the user whether they want to move the source file instead, and move the file accordingly.
> If anyone wants to tackle this, feel free to do so :)


I am personally thinking of implementing an option to automatically copy it in a way where if the user places the file in assets/audio, then the resource would be placed in resources/audio.
[...]

PR #1781, opened 2023-09-17
[Asset] Show warning message when importing assets that are not inside of the resource folder

Eideren:
# PR Details
image

## Related Issue
Could extend to cover issue #1633 depending on what we want ?

## Types of changes
- [ ] Docs change / refactoring / dependency upgrade
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist
- [ ] My change requires a change to the documentation.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #1827, opened 2023-09-22
Added the ability to copy imported assets automatically to the Resources dir

SeleDreams:
# PR Details

This pull request adds a prompt to copy imported assets to the Resources directory directly from the editor.

## Description
- I added a prompt asking if the user wishes to copy the file they imported to the Resources directory.
- If they say yes, another prompt asks them if they want to copy it to the default directory, if they say yes, it will copy it to the default Resources directory following the structure of the Assets dir (if they import the file TheSong.mp3 in Assets/Music the resource would get copied to Resources/Music/TheSong.mp3), else, a save file dialog will appear allowing them to choose where in the resources directory they want to place the file.
- If they say no, the file will get imported as is.
- If the resource already exists, a prompt will appear allowing them to either override the file, either cancel the operation.

## Related Issue
#1633 

## Motivation and Context

This change will avoid new users breaking their projects by incorrectly importing their assets. This improvement makes the workflow more intuitive.

## Types of changes

## Checklist

### DIFF
--- a/sources/editor/Stride.Core.Assets.Editor/ViewModel/AssetCollectionViewModel.cs
+++ b/sources/editor/Stride.Core.Assets.Editor/ViewModel/AssetCollectionViewModel.cs
@@ -471,7 +471,7 @@ namespace Stride.Core.Assets.Editor.ViewModel
             base.Destroy();
         }
 
-        public async Task<List<AssetViewModel>> RunAssetTemplate(ITemplateDescriptionViewModel template, IList<UFile> files, PropertyContainer? customParameters = null)
+        public async Task<List<AssetViewModel>> RunAssetTemplate(ITemplateDescriptionViewModel template, [CanBeNull] IList<UFile> files, PropertyContainer? customParameters = null)
         {
             if (template == null)
                 return new List<AssetViewModel>();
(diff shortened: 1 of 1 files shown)


Aggror:
Is this related to this one?
https://github.com/stride3d/stride/pull/1781

SeleDreams:
> Is this related to this one? #1781

It's also related to this one as it builds upon it yes

Aggror:
Nice work!

sources/editor/Stride.Core.Assets.Editor/ViewModel/AssetCollectionViewModel.cs
```diff
@@ -537,31 +535,99 @@ private string ComputeNamespace(DirectoryBaseViewModel directory)
...
-        private async Task<List<AssetViewModel>> InvokeAddAssetTemplate(LoggerResult logger, string name, DirectoryBaseViewModel directory, TemplateAssetDescription templateDescription, IEnumerable<UFile> files, PropertyContainer? customParameters)
+        private async Task<string> GetAssetCopyDirectory(DirectoryBaseViewModel directory, UFile file)
         {
-            List<AssetViewModel> newAssets = new List<AssetViewModel>();
-            if (files is not null)
-            {
-                foreach (var file in files)
+            string path = directory.Path;
+            var message = Tr._p("Message", "Do you want to place the resource in the default location ?").ToFormat(file.FullPath);
+            string finalPath = directory.Package.Package.ResourceFolders[0] + "/" + path + file.GetFileName();
```
Eideren, line 542: Could use the `Path` API, `Path.Combine` for example

sources/editor/Stride.Core.Assets.Editor/ViewModel/AssetCollectionViewModel.cs
```diff
@@ -537,31 +535,99 @@ private string ComputeNamespace(DirectoryBaseViewModel directory)
...
+            if (pathResult == MessageBoxResult.No)
+            {
+                bool done = false;
+                do
                 {
-                    bool inResourceFolder = false;
-                    foreach (var resourceFolder in directory.Package.Package.ResourceFolders)
+                    var fileDialog = Dialogs.CreateFileSaveModalDialog();
+                    fileDialog.Filters = new List<FileDialogFilter>() { new FileDialogFilter("", file.GetFileExtension()) };
+                    fileDialog.InitialDirectory = directory.Package.Package.ResourceFolders.FirstOrDefault().FullPath.Replace("/", "\\");
```
Eideren, line 551: While the editor doesn't run on anything else but windows for now, might as well reduce issues we might have when migrating.
https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/8.0/file-path-backslash
> Backslash ( \ ) characters are valid in directory and file names on Unix

You can use `Path.GetFullPath(myPath)` to convert into the OS's directory separator.

sources/editor/Stride.Core.Assets.Editor/ViewModel/AssetCollectionViewModel.cs
```diff
@@ -537,31 +535,99 @@ private string ComputeNamespace(DirectoryBaseViewModel directory)
...
+
+                    // If the user closes the dialog, assume that they want to use the default directory
+                    if (result != DialogResult.Ok)
+                    {
+                        done = true;
+                    }
+                    else
                     {
-                        if (file.FullPath.StartsWith(resourceFolder.FullPath))
+                        var fullPath = Path.GetFullPath(fileDialog.FilePath).Replace("\\", "/");
```
Eideren, line 562: Should skip the replace under non-windows, you can test against `Path.DirectorySeparatorChar` or os directly. The output path should already be using forward slash in those OS for dir separator.
Without that check, if the directory name or file name contains a backslash in those platforms it'll break.

sources/editor/Stride.Core.Assets.Editor/ViewModel/AssetCollectionViewModel.cs
```diff
@@ -537,31 +535,99 @@ private string ComputeNamespace(DirectoryBaseViewModel directory)
...
+                                    
+                                    // Abort if the user says no or closes the prompt
+                                    if (copyResult != MessageBoxResult.Yes)
+                                    {
+                                        return newAssets;
+                                    }
+                                    File.Delete(finalPath);
+                                }                                
+                                
+                                File.Copy(file.FullPath, finalPath);
```
Eideren, line 619: Maybe copy with overwrite [Copy(String, String, Boolean)](https://learn.microsoft.com/en-us/dotnet/api/system.io.file.copy?view=net-7.0#system-io-file-copy(system-string-system-string-system-boolean)) when it exists, and use normal copy if the file doesn't.
That way it's an atomic operation, if a file somehow moves there between the delete and copy it won't fail. 
Very minor though, feel free to ignore.

Eideren (changes requested):
Thanks a bunch @SeleDreams ! There's a couple of minor things I would like to go over, but otherwise it looks great !

SeleDreams:
I changed some stuff based on the suggestions from @Eideren

Eideren:
Thanks a bunch @SeleDreams !

Issue #1153, opened 2021-09-29
[Physics] Simulation / Fixed Update tick


Eideren:
**Is your feature request related to a problem? Please describe.**
Logic related to physics in many case has to run during physics tick to properly affect and synchronize with the physics engine, right now the engine does not provide a way to do so.

**Describe the solution you'd like**
Here's a bare-bone way to provide this feature for users: https://github.com/Eideren/xenko/commit/e80b188cf6a73c441002cb89a8f020e1a2901e76 
This can't be integrated as is though, we need to provide users with an easy way to interact with it, like ``SyncScript`` does for ``Update``.

**Describe alternatives you've considered**
Running that sort of logic during ``Update`` can work in most cases but not when you need predictable/deterministic results.

PR #1831, opened 2023-09-22
[Physics] Pre and Post simulation tick event

Eideren:
# PR Details
Add events to Simulation, users can subscribe to those to have their logic run before or after a physics tick occur.

## Related Issue
#1153 - changing that one to cover the implementation of a processor over those new events

## Motivation and Context
See #1153

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Physics/Simulation.cs
+++ b/sources/engine/Stride.Physics/Simulation.cs
@@ -1081,7 +1081,7 @@ namespace Stride.Physics
             add
             {
                 if (preSimulationTick is null)
-                    discreteDynamicsWorld.SetInternalTickCallback((world, step) => preSimulationTick?.Invoke(this, step), new object(), isPreTick:true);
+                    discreteDynamicsWorld.SetInternalTickCallback((world, step) => preSimulationTick?.Invoke(this, step), isPreTick:true);
                 preSimulationTick += value;
             }
             remove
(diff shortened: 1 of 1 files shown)


sources/engine/Stride.Physics/Simulation.cs
```diff
@@ -1066,12 +1053,67 @@ internal void Simulate(float deltaTime)
...
+        /// <summary>
+        /// Called right before processing a tick of the physics simulation,
+        /// this may occur from zero up to <see cref="MaxSubSteps"/> time per update.
+        /// </summary>
+        public event SimulationTickEvent PreTick
+        {
+            add
+            {
+                if (preSimulationTick is null)
+                    discreteDynamicsWorld.SetInternalTickCallback((world, step) => preSimulationTick?.Invoke(this, step), new object(), isPreTick:true);
```
manio143, line 1084: For this `new object()` - is that required to be an instance?
Could `null` or some static object instance be used instead?

manio143 (approved):

sources/engine/Stride.Physics/Simulation.cs
Eideren: Thankfully this is just a single allocation for the lifetime of `Simulation`, but I'll take a look tomorrow anyway

PR #1842, opened 2023-09-24
[Serialization][Editor] Reference type support on interface fields and properties

### LINKED ISSUE #764: Interface member that is a component doesn't get serialized properly

**Version**: beta03

**Platform(s)**: Any

**Describe the bug**
I have a public member on a script that accepts an interface implementation.
If the implementing class inherits from ScriptComponent it is serialized as a reference and not a new instance. Therefore it's not instantiated, because it refers to something that is defined nowhere.

**To Reproduce**
1. Create an interface

        public interface ITest { }
2. Create a script implementing it

        public class Test : ScriptComponent, ITest { }
3. Create a script that uses the interface as a public member

        public class Runner : StartupScript
        {
            public ITest test;
            public override void Start()
            {
                if (test == null) throw new Exception();
            }
        }
4. In Game Studio create an entity with Runner script. From dropdown of field test choose Test.
5. Run game, get an exception

**Expected behavior**
The script `Test` is serialized as a new object, thus the field is properly deserialized and not null.

**Attached example project**
[IFaceBugRepro.zip](https://github.com/stride3d/stride/files/4807990/IFaceBugRepro.zip)

Eideren:
# PR Details
Previously, interface fields and properties were treated as value types in the editor, having such a field would show a combo box where the user could select the type of field they wanted to create.
Now components marked with an interface can be assigned to a field of that interface type on another component, enabling more complex gameplay architecture.

https://github.com/stride3d/stride/assets/5742236/b3081938-e9d7-42df-9dc9-e3c1ad85d464

The serialization already mostly supported this feature, just had to change how `IdentifiableObjectSerializer` operates slightly.

## Related Issue
Fixes #764

## Types of changes

## Checklist

### DIFF
--- a/sources/editor/Stride.Assets.Presentation/TemplateProviders/EntityComponentReferenceTemplateProvider.cs
+++ b/sources/editor/Stride.Assets.Presentation/TemplateProviders/EntityComponentReferenceTemplateProvider.cs
@@ -1,5 +1,7 @@
 // Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net) and Silicon Studio Corp. (https://www.siliconstudio.co.jp)
 // Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using Stride.Core.Extensions;
 using Stride.Engine;
 using Stride.Core.Presentation.Quantum;
 using Stride.Core.Presentation.Quantum.View;
@@ -13,7 +15,13 @@ namespace Stride.Assets.Presentation.TemplateProviders
 
         public override bool MatchNode(NodeViewModel node)
         {
-            return typeof(EntityComponent).IsAssignableFrom(node.Type) && node.Parent?.Type != typeof(EntityComponentCollection);
+            if (node.Parent?.Type == typeof(EntityComponentCollection))
+                return false;
+
+            if (typeof(EntityComponent).IsAssignableFrom(node.Type))
+                return true;
+
+            return node.Type.IsInterface && node.Type.IsImplementedOnAny<EntityComponent>();
         }
     }
 }
--- a/sources/assets/Stride.Core.Assets/Serializers/IdentifiableObjectSerializer.cs
+++ b/sources/assets/Stride.Core.Assets/Serializers/IdentifiableObjectSerializer.cs
@@ -97,8 +97,10 @@ namespace Stride.Core.Assets.Serializers
                 // Return default(T)
                 //return !context.Descriptor.Type.IsValueType ? null : Activator.CreateInstance(context.Descriptor.Type);
                 // Return temporary proxy instance
-                var proxy = (IIdentifiable)AbstractObjectInstantiator.CreateConcreteInstance(context.Descriptor.Type);
-                proxy.Id = identifier;
+                var proxy = AbstractObjectInstantiator.CreateConcreteInstance(context.Descriptor.Type);
+                // Filtering out interface and abstracts here as they're using a proxy type which doesn't have Id implemented
+                if (context.Descriptor.Type.IsInterface == false && context.Descriptor.Type.IsAbstract == false && proxy is IIdentifiable identifiable)
+                    identifiable.Id = identifier;
                 return proxy;
             }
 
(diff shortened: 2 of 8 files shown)


Eideren:
As a note, right now the interface in this context doesn't look great - I tried to improve it, but I'm just not good enough with xaml to get anywhere.
If @SVNMLR or @cNoNim wanna look into this, feel free :)

manio143 (approved):
Looks great! And I would say the UX is fine as well.

sources/editor/Stride.Assets.Presentation/TemplateProviders/EntityComponentReferenceTemplateProvider.cs
```diff
@@ -13,7 +15,13 @@ public class EntityComponentReferenceTemplateProvider : NodeViewModelTemplatePro
...
         public override bool MatchNode(NodeViewModel node)
         {
-            return typeof(EntityComponent).IsAssignableFrom(node.Type) && node.Parent?.Type != typeof(EntityComponentCollection);
+            if (node.Parent?.Type == typeof(EntityComponentCollection))
+                return false;
+
+            if (typeof(EntityComponent).IsAssignableFrom(node.Type))
+                return true;
+
+            return node.Type.IsInterface && node.Type.IsImplementedOnAny<EntityComponent>();
```
manio143, line 24: Although, I just noticed - in your demo you show that when a non-component type is chosen it still shows the entity picker - can you see if it's possible to filter it out? 
Maybe we could check something more about `node`'s value type here?

Eideren: Initially the idea was to remove reference types from the combo box and only use the picker to select them.

Issue is that would prevent users from creating new instances of those component types inline to the field (i.e.: like a value type instead of ref) - this probably wouldn't have been a very common way to use this feature, but might as well account for it. Now it worked before but it looks like I ended up breaking it somewhere along the line.

Anyway, even without that particular feature working, the picker icon is still a shortcut to select any instance of the interface, it's not constrained to the type that was selected in the combo box, so it's one less click for users if they want to swap from an inlined instance to a ref.

manio143: Oh, I see. I get that when we're asking for the interface instance we don't exactly care about the implementing type. But if we're anyways making the user choose some type I'd expect the entity picked to be then constrained to the type I have chosen (especially with bigger scenes). Would it be possible to delegate the picker to act on the chosen type instead - this way we don't have to add all that interface logic to entity picker.

Eideren: Right, I don't think the picker should by default constrain based on the selected type because this would be very easy to miss from the user's perspective. We don't have any UX setup which would make this idea obvious every time a user would use the picker. It would also be kind of awful for users to go back and forth between the combo box and the picker if they selected the wrong type by mistake.

Filtering is a good point though, but I think this should be part of the object picker window itself, that way users can benefit from those filters for direct component properties/fields, and there's so much dead space in that window it would be a shame not to use it ?

Eideren: Sounds good ! I'm really bad at xaml though, so let's get this in and see if anyone can take over ?

Ethereal77: Go with it. Hopefully someone knowledgeable in XAML can implement the filtering. I'm not very proficient in XAML but if no one looks at this, I'll take a look later.

manio143: Ok, let's merge it. And I'll open an enhancement issue on UX of abstract types in lists in general - I think when it's a single interface property the drop-down holds the selected type name.

PR #1846, opened 2023-09-25
[Core] Fix StringComparison across the project

Eideren:
# PR Details
A user had the editor fail to create projects because of this `IndexOf` call:
https://github.com/stride3d/stride/blob/91017546503e82eb18cdb31c97e44e833819c5fc/sources/core/Stride.Core/Reflection/AssemblyRegistry.cs#L75-L83
His culture somehow detected this `,` at index `0` of the string `Stride.Graphics.SpriteFont, Stride.Graphics, Version=4.1.0.1898, Culture=neutral, PublicKeyToken=null` leading to the `AssemblyName` `ctor` throwing.

All string methods that take `char` as input always perform operations through ordinal comparison. 
When they take strings as inputs, the comparison depends on the method, `IndexOf`, `LastIndexOf`, `StartsWith` and `EndsWith` use the current culture.
This PR ensures that every call to those methods uses ordinal wherever it makes sense (which ended up being all of them) to ensure we do not have other related issues hidden in our source.

## Related Issue
The issue mentioned was reported on discord, it hasn't been logged as far as I know.

## Types of changes

## Checklist

### DIFF
--- a/sources/editor/Stride.Core.Assets.Editor/Services/VisualStudioService.cs
+++ b/sources/editor/Stride.Core.Assets.Editor/Services/VisualStudioService.cs
@@ -112,7 +112,7 @@ namespace Stride.Core.Assets.Editor.Services
             foreach (var process in processes)
             {
                 // Get instances that have a solution with the same name currently open (The solution name is displayed in the title bar).
-                if (process.MainWindowTitle.ToLowerInvariant().StartsWith(solutionPath.GetFileNameWithoutExtension().ToLowerInvariant()))
+                if (process.MainWindowTitle.ToLowerInvariant().StartsWith(solutionPath.GetFileNameWithoutExtension().ToLowerInvariant(), StringComparison.Ordinal))
                 {
                     // If there is a matching instance, get its command line.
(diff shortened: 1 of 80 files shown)


sources/editor/Stride.Core.Assets.Editor/Services/VisualStudioService.cs
```diff
@@ -112,15 +112,15 @@ private static Process FindVisualStudioInstance(UFile solutionPath)
             foreach (var process in processes)
             {
                 // Get instances that have a solution with the same name currently open (The solution name is displayed in the title bar).
-                if (process.MainWindowTitle.ToLowerInvariant().StartsWith(solutionPath.GetFileNameWithoutExtension().ToLowerInvariant()))
+                if (process.MainWindowTitle.ToLowerInvariant().StartsWith(solutionPath.GetFileNameWithoutExtension().ToLowerInvariant(), StringComparison.Ordinal))
```
manio143, line 115: You can use OrdinalIgnorecase and skip the to lower call

manio143 (approved):
Looks good. One note I have is that with extension compare a lot of those should be ignore case

Eideren:
Updated with ignore case

manio143 (approved):

Issue #1835, opened 2023-09-24
Roslyn analyzers and source generators (tracking)


manio143:
Creating an issue to support cross-PR discussion for #1778 and #1609

The purpose is to slowly replace `AssemblyProcessor` with Roslyn for the benefit of easier maintenance and improved usability by allowing debugging into the serializers.
Furthermore analyzers would provide compile time warnings to users if they do something not supported by the serialization system providing more clarity over why things are not working.

manio143:
Some notes from looking at code

## Member visibility
✅ `public` - serializers can access it easily
✅ `internal` - serializers are generated in the same assembly as the class and can access it easily
⚠ `private`/`protected` - unavailable to binary serializer (pre .NET 8 UnsafeAccessor), but available to reflection-based YAML serializer

### Read-only/write-only
Binary serialization doesn't support write only members - has to have a getter.
Read only members must be of a reference type (and not a `string`). Readonly fields are supported, as well as properties without a setter.

### Virtual members
Virtual properties are serialized given the class declaring them is serializable.
Overrides can be serialized specifically only if annotated with `[DataMember]`.

## DataMember attribute

### Mode
* Assign - serializer will assign a value to the member (e.g. create a new object and assign its value to the property).
    When the member is writable this is the default.
* Content - serializer will deserialize members of the value in place, only applies to reference types.
    When the member is not writable this is the default.
* Binary - for arrays of blittable types, they will be stored as Base64 string in YAML.
* Never - not serialized - prefer `[DataMemberIgnore]`

The default mode for all members can be also set on `[DataContract]`.

### Order
Used in Editor to decide order of properties. Possibly also in YAML serialization.

### Name
The `Name` is used to override member name in YAML serialization and for the Editor. Has no effect on binary serialization.

### Mask
Used in YAML serialization to filter out members not covered by the bit mask.
One use case in repo in dead code.

## Delegates
Delegates cannot be serialized.

## DataMemberUpdatable Attribute
The UpdateEngine (dynamic object updater based on string paths for complex animations over components) somehow uses serialization metadata and if a member has `[DataMemberIgnore]` but also `[DataMemberUpdatable]` then it is binary serializable but not visible in YAML/Editor.

## Fixed buffers
Currently AssemblyProcessor generated binary serializers do not support `fixed` buffer fields ([msdn](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/unsafe-code#fixed-size-buffers)).

xen2:
[...]

PR #1778, opened 2023-09-17
[WIP][OLD] Adding Stride Diagnostics 

IXLLEGACYIXL:
# PR Details

Adding a Source Generator that Analyzes Projects for Violations of how Stride would expect it to be

## Description

The Analyzer Covers:
1. Access of Collections Properties
2. Access of Array Properties
3. Acces of Normal Properties
4. cant Datamember something that is private or a DataMemberIgnore
5. cant have a DataMember on a Dictionary with a object as key

## Related Issue

n/a

## Motivation and Context

It solves the issue that you get surprised when you open Game Studio and you don't see anything Exported
Also it offers validation for other generators so they dont have to check the same errors

## Types of changes

- [ ] Docs change / refactoring / dependency upgrade
- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [X] My change requires a change to the documentation. ( Updating the docs + making sites for the diagnostics )
- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.

PR #1609, opened 2023-02-13
[WIP] Roslyn based serialization source generator

manio143:
# PR Details

Opening draft PR in order to be able to reference it in documentation, once I make some progress on this I'll update the description.

## Description

**TODO**

## Related Issue

**TODO**

## Motivation and Context

**TODO**

## Types of changes

- [ ] Docs change / refactoring / dependency upgrade
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [ ] My change requires a change to the documentation.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #1847, opened 2023-09-26
[Roslyn] Add base setup for a Roslyn integration into Stride.Core

manio143:
# PR Details

## Description

Should serve as a base for #1609 and #1778.
Adds `Stride.Core.CompilerServices` project which will contain analyzers, code fixes and generators.
Runs over every project in the Stride.sln and is packed with Stride.Core in NuGet.

Verified to be running on user projects as well as Stride projects in the same solution.

## Related Issue

#1835

## Types of changes

## Checklist


sources/core/Stride.Core.CompilerServices.Tests/Stride.Core.CompilerServices.Tests.csproj
```diff
@@ -0,0 +1,29 @@
...
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.4.3" />
+  </ItemGroup>
+  <ItemGroup>
+    <Compile Include="..\..\shared\SharedAssemblyInfo.cs">
+      <Link>Properties\SharedAssemblyInfo.cs</Link>
+    </Compile>
+  </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\..\tests\xunit.runner.stride\xunit.runner.stride.csproj" />
+    <ProjectReference Include="..\Stride.Core\Stride.Core.csproj" ReferenceOutputAssembly="false" OutputItemType="Content" CopyToOutputDirectory="Always" />
```
manio143, line 22: For reference
https://learn.microsoft.com/en-us/archive/blogs/kirillosenkov/how-to-have-a-project-reference-without-referencing-the-actual-binary

While not necessarily needed in this base, what this does is it allows us to reference Stride.Core in the code samples under test for analysis without referencing Stride.Core directly in case it causes complications with the reference to the analyzer itself.

sources/core/Stride.Core.CompilerServices/Analyzers/TestAnalyzer.cs
```diff
@@ -0,0 +1,30 @@
+using System.Collections.Immutable;
+using System.Diagnostics;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Stride.Core.CompilerServices.Analyzers
+{
+    // PLACEHOLDER ANALYZER
```
IXLLEGACYIXL, line 8: i would remove it as my PR has allready analyzers in it to reference for others and its proven that the base PR works.
So an example wouldnt be necessary

im fine with the rest of the PR

IXLLEGACYIXL (changes requested):

sources/core/Stride.Core.CompilerServices/Analyzers/TestAnalyzer.cs
manio143: Sure, I'll remove it

Issue #1845, opened 2023-09-25
Physics of the CharacterComponent seems to be bound to the framerate


SeleDreams:
**Release Type**: Github 

**Version**: master, commit 9101754

**Platform(s)**: Windows

**Describe the bug**
It seems like the speed at which CharacterComponent moves is tied to the framerate the game is running at.
for instance, I made a test project of a ball moving back and forth with a constant speed value, the velocity is set in a fixed update that always runs just before the physics tick using Simulation.PreTick, but when the fps are set lower, for instance at 24fps, the ball will start moving slower.
If the value sent to SetVelocity is multiplied by deltatime, it does seem to have the same speed between 30 and 60fps, however it will start having issues at higher fps like 144fps

**To Reproduce**
1. Create a project
2. Use a CharacterComponent
3. Move by setting a velocity to the CharacterComponent
4. The speed is tied to the FPS

**Expected behavior**
The speed should be independent from framerate

**Screenshots**
https://github.com/stride3d/stride/assets/16335601/7dfe9d79-7917-4929-bfce-5e3cdc607767

**Additional context**
Simple reproduction sample
[MyGame6.zip](https://github.com/stride3d/stride/files/12719581/MyGame6.zip)
.

SeleDreams:
It seems this was caused by the physics max sub steps being set to 1, increasing it fixed the issue

PR #1849, opened 2023-09-26
[Physics] Change simulation substeping behavior

Eideren:
# PR Details
Right now VSync and other update rate control may affect the simulation speed when under 30 fps;
Physics updates at 60hz, by default there is one additional substep, so if the game runs between 60fps (one tick per update) and 30fps (two ticks per update) the substep will continue to make the simulation run at the expected speed, but under that point the simulation will do the additional substep and ignore the rest of the delta time, which will make it look like the simulation 'slows down'.

To fix this we could naively increase the amount of substeps that the simulation can do, to a point where it could cover from 60 to 5fps let's say. This has the side-effect of creating a negative feedback loop where the reason we have such low FPS is because of the simulation, and the added substeps are making things worse.

To palliate that we would have to detect whether it is an artificial throttling, and enable additional substeps in those case. So checking VSync and the game's frame cap when running in the background. Which is both very awkward to setup since we're in the Physics solution, but also hard/impossible to detect since the user might have created their own throttling mechanism.

This fix instead sidesteps the issue by deprecating `MaxSubSteps` in favor of a `MaxTickDuration`, the system will process as many substeps as it needs as long as physics doesn't take longer than that value. If the game runs slow, but physics is actually pretty fast, it'll process all the ticks required and the simulation will continue to update in real time while the rest struggles.

## Related Issue
#1845

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Physics/Simulation.cs
+++ b/sources/engine/Stride.Physics/Simulation.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using Stride.Core.Diagnostics;
 using Stride.Core.Mathematics;
 using Stride.Core.MicroThreading;
@@ -34,6 +35,8 @@ namespace Stride.Physics
         private SimulationTickEvent preSimulationTick, postSimulationTick;
 
         internal readonly bool CanCcd;
+
+        private float carriedDelta;
 
 #if DEBUG
         private static readonly Logger Log = GlobalLogger.GetLogger(typeof(Simulation).FullName);
@@ -90,7 +93,7 @@ namespace Stride.Physics
                 configuration.Flags = OnSimulationCreation?.Invoke() ?? configuration.Flags;
             }
 
-            MaxSubSteps = configuration.MaxSubSteps;
+            MaxTickDuration = configuration.MaxTickDuration;
             FixedTimeStep = configuration.FixedTimeStep;
 
             collisionConfiguration = new BulletSharp.DefaultCollisionConfiguration();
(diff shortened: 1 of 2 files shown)


SeleDreams:
I confirm that this does seem to work without causing visible issues at least in my project

sources/engine/Stride.Physics/Simulation.cs
```diff
@@ -1045,8 +1057,25 @@ internal void Simulate(float deltaTime)
 
             SimulationProfiler = Profiler.Begin(PhysicsProfilingKeys.SimulationProfilingKey);
 
-            if (discreteDynamicsWorld != null) discreteDynamicsWorld.StepSimulation(deltaTime, MaxSubSteps, FixedTimeStep);
-            else collisionWorld.PerformDiscreteCollisionDetection();
+            if (discreteDynamicsWorld != null)
+            {
+                carriedDelta += deltaTime;
+                for (; carriedDelta > FixedTimeStep; )
```
manio143, line 1063: nit: you can use a `while` loop here for better readability

sources/engine/Stride.Physics/Simulation.cs
```diff
@@ -1045,8 +1057,25 @@ internal void Simulate(float deltaTime)
...
+                {
+                    var ts = Stopwatch.GetTimestamp();
+                    discreteDynamicsWorld.StepSimulation(FixedTimeStep, 0, FixedTimeStep);
+                    carriedDelta -= FixedTimeStep;
+                    if (carriedDelta > FixedTimeStep && (Stopwatch.GetTimestamp() - ts) / (double)Stopwatch.Frequency > MaxTickDuration)
```
manio143, line 1068: Since Stopwatch.Frequency is constant, is it possible to replace this division with multiplication of static field = 1.0 /  Stopwatch.Frequency computed once? Or is it not really worth it?

manio143 (approved):
Took me a moment to internalize this, but I think it's good. In high framerate scenarios this week keep a consistent fixed update and in low framerate scenarios it will skip running simulation multiple ticks a frame if a tick is too slow.

sources/engine/Stride.Physics/Simulation.cs
```diff
@@ -1045,8 +1057,25 @@ internal void Simulate(float deltaTime)
...
             SimulationProfiler = Profiler.Begin(PhysicsProfilingKeys.SimulationProfilingKey);
 
-            if (discreteDynamicsWorld != null) discreteDynamicsWorld.StepSimulation(deltaTime, MaxSubSteps, FixedTimeStep);
-            else collisionWorld.PerformDiscreteCollisionDetection();
+            if (discreteDynamicsWorld != null)
+            {
+                carriedDelta += deltaTime;
+                for (; carriedDelta > FixedTimeStep; )
+                {
+                    var ts = Stopwatch.GetTimestamp();
```
froce, line 1065: If I understand things correctly this should be outside the loop, to limit the entire update duration, not each substep

Eideren: Oof, good catch, thanks !

sources/engine/Stride.Physics/Simulation.cs
Eideren: Yep, done

sources/engine/Stride.Physics/Simulation.cs
Eideren: That was the idea initially, but forgot about it while testing, ended up extending it into a cleaner for loop instead

PR #1850, opened 2023-09-26
[Physics] Add ray test flags

Eideren:
# PR Details
Adds flags to change how ray tests are performed, mainly allowing users to ignore back-faces.

## Related Issue
None logged.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Physics/Elements/CollisionFilterGroups.cs
+++ b/sources/engine/Stride.Physics/Elements/CollisionFilterGroups.cs
@@ -79,4 +79,31 @@ namespace Stride.Physics
 
         AllFilter = 0xFFFF,
     }
+
+    /// <summary>
+    /// Flags that control how ray tests are performed
+    /// </summary>
+    [Flags]
+    public enum EFlags : uint
+    {
+        None = 0,
+        /// <summary>
+        /// Do not return a hit when when a ray traverses a back-facing triangle
+        /// </summary>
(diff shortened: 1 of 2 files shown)


sources/engine/Stride.Physics/Elements/CollisionFilterGroups.cs
```diff
@@ -79,4 +79,31 @@ public enum CollisionFilterGroupFlags
 
         AllFilter = 0xFFFF,
     }
+
+    /// <summary>
+    /// Flags that control how ray tests are performed
+    /// </summary>
+    [Flags]
+    public enum EFlags : uint
```
manio143, line 87: Is the name EFlags appearing in Bullet? If not, can we rename it to RayCastFlags or something?

Eideren: Yes, see the source over here https://github.com/bulletphysics/bullet3/blob/39b8de74df93721add193e5b3d9ebee579faebf8/src/BulletCollision/NarrowPhaseCollision/btRaycastCallback.h#L32
I would have also preferred to change the name, but it would be far easier for our users to find resources around this if we keep the same name. What do you think ?

manio143: Agreed - we want to keep Bullet naming where possible so users can use Bullet forums and stuff with Stride.

PR #1860, opened 2023-09-28
Added a new PropertyGroup to Stride.Core.targets...

JeromyWalsh:
# PR Details

...to support setting environment variables which are native to Windows but which might not exist in other OS's.

The first property being added is $(TEMP) which normally comes from %TEMP% environment variable in Windows.

See https://learn.microsoft.com/en-us/dotnet/api/system.io.path.gettemppath?view=net-6.0&tabs=windows for more information.

## Description

Modified the Stride.Core.targets file by adding a new PropertyGroup and a new TEMP element that overrides the environment variable %TEMP% on Windows and provides a value for other OS's that do not support the %TEMP% environment variable.

The value is pulled from the System.IO.Path.GetTempPath() method which points to the following locations:

On Windows: C:\Users\{UserName}\AppData\Local\Temp\
On Linux: /tmp/

## Motivation and Context

The Stride.Core.targets currently leverages the fact that MSBuild is capable of pulling in and utilizing environment variables directly. However, not all OS's support the same environment variables natively. To prevent additional build steps, such as the manual or automated creation of environment variables, this change seeks to unify the environment where possible.

## Types of changes

## Checklist

### DIFF
--- a/sources/core/Stride.Core.AssemblyProcessor/Stride.Core.AssemblyProcessor.csproj
+++ b/sources/core/Stride.Core.AssemblyProcessor/Stride.Core.AssemblyProcessor.csproj
@@ -24,6 +24,25 @@
   <PropertyGroup>
     <StartupObject />
   </PropertyGroup>
+  
+  <!-- Add support for building this project from the Project Directory -->
+  <PropertyGroup>
+    <SolutionDir Condition="'$(SolutionDir)' == ''">$(ProjectDir)..\..\..\build\</SolutionDir>
+  </PropertyGroup>
+
+  <!-- ======================================================================================== -->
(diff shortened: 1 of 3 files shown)


Jklawreszuk (approved):

Jklawreszuk:
Looks good to me

manio143:
Can you do the same here?
https://github.com/stride3d/stride/blob/master/sources/core/Stride.Core/build/Stride.Core.targets#L52-L54

JeromyWalsh:
@manio143 Sure. Give me a few. Also, there's some other changes that accidentally got merged in here. Take a look at those as well if you can.

manio143 (approved):

Issue #909, opened 2020-11-13
Directional Light in the Game Studio


antongit:
**Is your feature request related to a problem? Please describe.**
It's not possible to understand the direction of the Directional Light until one selects its Gizmo AND rotates the View so, that the Gizmo doesn't overlap (hide) the vector.

Here are some pictures.

![image](https://user-images.githubusercontent.com/1189037/99087993-b6c6c200-25cb-11eb-8e63-a52567cf5583.png)

![image](https://user-images.githubusercontent.com/1189037/99088032-c219ed80-25cb-11eb-909e-7719e1f94f57.png)

**Describe the solution you'd like**
Would it be possible to have another Gizmo for the Directional Light (instead of the default Sun Icon) that shows the direction?

This is how they do it in Maya:

![image](https://user-images.githubusercontent.com/1189037/99088058-cc3bec00-25cb-11eb-88ed-89e18e316e65.png)

parhamgholami:
I'm interested in tackling this issue. There are a couple of different ways we could communicate the direction.

One way would be to adopt the Maya/Unity approach (as mentioned in the original post), where there are lines denoting the direction:
![image](https://github.com/stride3d/stride/assets/7075456/fd25db69-7bf1-42ce-8479-a5aea6344e51)

Alternatively, we could do something similar to Unreal:
![image](https://github.com/stride3d/stride/assets/7075456/98e1b61d-6e18-487d-9bf7-5d7987bcfd38)

Based on my understanding of the gizmos, the latter would be trivial to implement. Jack already did the necessary work, but I don't know why the branch was deleted.

With all that in mind, is there any preference for how we show the direction of the light?

PR #1865, opened 2023-09-29
Always show directional light arrow

parhamgholami:
# PR Details

The logic for hiding/showing the directional light arrow based on whether it has been selected was removed. The arrow is now always visible in the editor. This current implementation is identical to JackPilley's, but that branch was quietly closed before it was merged into the main branch. As was mentioned in the related issue, this is the simplest approach to solving the issue, but we can also look at alternative methods if they're preferred.

I've marked this PR as a work-in-progress in case we want to explore alternative solutions.

This is how the directional light looks with the change, regardless of whether it's been selected:
image

## Related Issue

This is in follow-up to the issue requesting the arrow for directional lights always be visible:
https://github.com/stride3d/stride/issues/909

## Motivation and Context

This change is helpful for seeing the direction of the light source at a glance without needing to select it.

## Types of changes

## Checklist


Eideren:
I think this is fine for directional lights since users likely won't have many of them in a scene, for spotlights it may become visual noise when they're in a cluster.

If no one has any objection, I'll merge it in tomorrow.

parhamgholami:
That sounds good! In that case, I'll remove the WIP tag.

Eideren:
Thanks !

PR #1868, opened 2023-09-29
Fix readonly serializing on structs/string

IXLLEGACYIXL:
# PR Details
Fixes Serialization of readonly members

## Description

1. readonly structs/strings dont get serialized anymore
2. get only properties of structs/strings dont get serialized anymore
3. HasSet fix in fielddescriptor

## Related Issue

https://discord.com/channels/500285081265635328/1156630820430303373
First part of fixing the serialization on get only members.
Fixing non backing fields is unaffected by this PR

## Types of changes

## Checklist

- [x] All new and existing tests passed. ( The Image tests and rasterized fonts test fail i was never able to get them working, but they fail without the changes too )

### DIFF
--- a/sources/core/Stride.Core.Reflection/TypeDescriptors/ObjectDescriptor.cs
+++ b/sources/core/Stride.Core.Reflection/TypeDescriptors/ObjectDescriptor.cs
@@ -265,14 +265,12 @@ namespace Stride.Core.Reflection
             if (memberAttribute != null)
             {
                 ((IMemberDescriptor)member).Mask = memberAttribute.Mask;
+                member.Mode = memberAttribute.Mode;
                 if (!member.HasSet)
                 {
-                    if (memberAttribute.Mode == DataMemberMode.Assign ||
-                        (memberType.IsValueType && member.Mode == DataMemberMode.Content))
-                        throw new ArgumentException($"{memberType.FullName} {member.OriginalName} is not writeable by {memberAttribute.Mode.ToString()}.");
+                    if (memberAttribute.Mode == DataMemberMode.Assign || memberType.IsValueType || memberType == typeof(string))
+                        member.Mode = DataMemberMode.Never;
                 }
-
-                member.Mode = memberAttribute.Mode;
                 member.Order = memberAttribute.Order;
             }
 
--- a/sources/core/Stride.Core.Reflection/MemberDescriptors/FieldDescriptor.cs
+++ b/sources/core/Stride.Core.Reflection/MemberDescriptors/FieldDescriptor.cs
@@ -30,7 +30,7 @@ namespace Stride.Core.Reflection
 
         public override bool IsPublic => FieldInfo.IsPublic;
 
-        public override bool HasSet => true;
+        public override bool HasSet => !FieldInfo.IsInitOnly;
 
         public override object Get(object thisObject)
         {


sources/core/Stride.Core.Reflection/TypeDescriptors/ObjectDescriptor.cs
```diff
@@ -265,14 +265,12 @@ protected virtual bool PrepareMember(MemberDescriptorBase member, MemberInfo met
...
             {
                 ((IMemberDescriptor)member).Mask = memberAttribute.Mask;
+                member.Mode = memberAttribute.Mode;
                 if (!member.HasSet)
                 {
-                    if (memberAttribute.Mode == DataMemberMode.Assign ||
-                        (memberType.IsValueType && member.Mode == DataMemberMode.Content))
-                        throw new ArgumentException($"{memberType.FullName} {member.OriginalName} is not writeable by {memberAttribute.Mode.ToString()}.");
+                    if (memberAttribute.Mode == DataMemberMode.Assign || memberType.IsValueType || memberType == typeof(string))
+                        member.Mode = DataMemberMode.Never;
```
manio143, line 272: I think throwing the exception is more correct here - we're saying that to user has explicitly provided an attribute that won't work. We can't silently ignore it and should report error instead.
That exception needs to be caught in the editor and shown to the user in the scene window (like it does currently for some exceptions).

IXLLEGACYIXL: the diagnostics analyzer would tell the user that its invalid

you get a warning

manio143: Ok, fair enough. Preventing editor crashes is more important. I can't find right now how the exception handling works in the editor for this case.

IXLLEGACYIXL: what would you suggest now on how to continue?
should this PR wait until diagnostics is there or should it allready be merged?

should it throw? 
i just dont like exceptions overall, User Expierence will be better without exception in my opinion
none of the serializers should ever fail

manio143: > none of the serializers should ever fail

Well that's not really possible. It's fine for the serializer to throw provided the exception contains enough context for the user to be able to understand the reason and fix it.

In code only environment exceptions are the only way to communicate at runtime back to the user that they have made a mistake. Sure, analyzers will help mitigate it but we can't fully rely on them.
In an editor/asset pipeline environment we have the option to issue warnings to the build log and that is a preferred way of communicating. Warn and ignore member to have something working (where exception fully terminates the program).

For now I think we should merge to make people's life easier - a crashing editor is worse than a missing member.
And in the future as we try to homogenize this layer between the serializers Stride uses we can look to implement an abstract error reporting pipeline which can throw or warn depending on which environment it's being executed in.

IXLLEGACYIXL: allright

Eideren:
Can I close this one @IXLLEGACYIXL ?

IXLLEGACYIXL:
yes

PR #1875, opened 2023-10-01
[Serialization] Fix diverging rules for editor and runtime serialization of fields and properties

Eideren:
# PR Details
### Prevent serializing get-only properties without backing field. 
The idea here is that they aren't safe enough to serialize by default since they are functions, getting the value may throw, and since they do not provide a setter to create a value if the getter ever throws, we decided to only include them if the user decorates them with a `[DataMember]`

### Diverging rules between runtime and editor serialization
Yaml can now serialize internal fields if decorated with `DataMember`, reflecting how serialization and filtering works for properties.
Binary includes properties with internal getters, as long as they are decorated with `DataMember` as well.

Those two changes were required to ensure that both serializer behaved in the same way.

## Related Issue
Didn't find any ?

## Types of changes

## Checklist

Pinging @IXLLEGACYIXL and @manio143 to take a look into this.


sources/core/Stride.Core.AssemblyProcessor/ComplexSerializerRegistry.cs
```diff
@@ -382,6 +384,25 @@ public static IEnumerable<SerializableItem> GetSerializableItems(TypeDefinition
...
+            if (get == null)
+                return false;
+
+            bool forced = property.CustomAttributes.Any(a => a.AttributeType.FullName == "Stride.Core.DataMemberAttribute");
+
+            if (forced && (get.IsPublic || get.IsAssembly))
+                return true;
+
+            if (get.IsPublic)
+                return set?.IsPublic == true || set == null && property.DeclaringType.Fields.Any(x => x.Name == $"<{property.Name}>k__BackingField");
```
manio143, line 401: Would be good to add a comment here with an example of when a backing field is present.

sources/core/Stride.Core.Reflection/TypeDescriptors/ObjectDescriptor.cs
```diff
@@ -237,6 +263,31 @@ where PrepareMember(member, metadataClassMemberInfos?.FirstOrDefault(x => x.Memb
...
+            if (get == null)
+                return false;
+
+            bool forced = property.GetCustomAttribute<DataMemberAttribute>() is not null;
+
+            if (forced && (get.IsPublic || get.IsAssembly))
+                return true;
+
+            if (get.IsPublic)
+                return set?.IsPublic == true || set == null && TryGetBackingField(property, out _);
```
manio143, line 280: This line is what requires attribute on `public { get; internal set; }` if get is public and set isn't we require attribute. Only if set is missing we check for backing fields.

sources/engine/Stride.Graphics/Sprite.cs
```diff
@@ -149,6 +149,7 @@ public Vector4 Borders
         /// <summary>
         /// Gets the value indicating if the image has unstretchable borders.
         /// </summary>
+        [DataMember]
         public bool HasBorders { get; private set; }
```
manio143, line 153: Shouldn't this one be ignored? It's set by `Borders` setter.

manio143: Same for `SizeInPixels` below which is set by PixelsPerUnit->UpdateSizes()

manio143 (commented):
Note that anything in Stride.Core.Assets and Design wasn't running with binary serialization.

sources/engine/Stride.Graphics/Sprite.cs
IXLLEGACYIXL: > Shouldn't this one be ignored? It's set by `Borders` setter.

datamembereing it now lets the editor serialize it but its an invlaid case, and is fixed here https://github.com/stride3d/stride/pull/1868

sources/core/Stride.Core.Reflection/TypeDescriptors/ObjectDescriptor.cs
IXLLEGACYIXL: but internal setter is also valid?

IXLLEGACYIXL: i dont think that a internal setter should be treated differently to a public setter , atleats in the future serializers

Eideren: Right, I wasn't very clear - I do know that this line is filtering out those kinds of properties. What I was wondering was why those are filtered out now, but weren't before.
Now, I believe that this is because the previous serializer is too lenient with properties that have a public getter.

sources/engine/Stride.Graphics/Sprite.cs
Eideren: Yep, thanks !

sources/core/Stride.Core.AssemblyProcessor/ComplexSerializerRegistry.cs
Eideren: Will do

sources/core/Stride.Core.Reflection/TypeDescriptors/ObjectDescriptor.cs
manio143: My guess would be that it's based in how reflection accessibility works on properties. In assembly processor we directly and only inspect getters and setters for accessibility. But here in the ObjectDescriptor we were looking for public properties. So a property can be public and have any accessibility modifier on its setter. We weren't checking it here. And later when checking if it's read only it would only check if a setter is present and not how it looks, because for reflection it didn't matter.

sources/core/Stride.Core.Yaml.Tests/Serialization/SerializationTests2.cs
```diff
@@ -1506,21 +1510,6 @@ public class ObjectWithMask
...
-            var text = @"!ClassWithImplicitMemberType
-Test:
-  String: test
-";
-
-            settings.RegisterTagMapping("ClassWithImplicitMemberType", typeof(ClassWithImplicitMemberType));
-            settings.RegisterTagMapping("ClassWithImplicitMemberTypeInner", typeof(ClassWithImplicitMemberTypeInner));
-            SerialRoundTrip(settings, text);
-        }
-
```
Eideren, line 1523: This test has been removed as omitting object type when they match the type at initialization but diverge from the member's return type has been deprecated in a previous pr.

manio143 (approved):
I'm surprised there's so many places which use private setters and are serializable, nice work! - Being explicit about it seems better.

Eideren:
There's a fair amount members in the engine that don't conform with the binary's rules, i.e.: a bunch of 
```cs
[DataMember]
public float MyVal { get; private set; }
```
But if they did end up at runtime they would have been broken before this change anyway since binary never supported private or protected members. I've looked at a couple and they seem to be used for either the `UpdateEngine` or for editor-only serialization.

IXLLEGACYIXL:
> There's a fair amount members in the engine that don't conform with the binary's rules, i.e.: a bunch of
> 
> ```cs
> [DataMember]
> public float MyVal { get; private set; }
> ```
> 
> But if they did end up at runtime they would have been broken before this change anyway since binary never supported private or protected members. I've looked at a couple and they seem to be used for either the `UpdateEngine` or for editor-only serialization.

this is an invalid case and is not allowed.. atleast it should be and my diagnostics count that as error

but i only found around 8 cases with invalid combinations

Eideren:
Yes, as I said, those are only in our source and used only in the editor, right now yaml still allows them through to avoid breaking everything until we take care of them. If you have an easy way to detect them feel free to share, what I was using for internal doesn't scale to private/protected.

IXLLEGACYIXL:
> Yes, as I said, those are only in our source and used only in the editor, right now yaml still allows them through to avoid breaking everything until we take care of them. If you have an easy way to detect them feel free to share, what I was using for internal doesn't scale to private/protected.

https://gist.github.com/IXLLEGACYIXL/b311896965b67af082eafde723619b54
i cant find more , mybe they dont ref stride core in their packages?

Eideren:
We've discussed a bit more on discord and ended up agreeing on a less restrictive rule, [there's a document covering it over here](https://github.com/stride3d/stride-docs/blob/39226f0c58a669744b3f476c20f568cf5010145c/en/manual/scripts/serialization.md). 
My comment there https://github.com/stride3d/stride/pull/1875#issuecomment-1747253823 wasn't very explicit in what I meant, in that, at the time I wanted to prevent any and all usage of `{ get; private/protected set; }` from being serialized while @IXLLEGACYIXL / Joreyk thought this was specifically about value types not being assignable when using a `private/protected` setter.

Last PR fixed the issue Joreyk logged.

sources/core/Stride.Core.AssemblyProcessor/ComplexSerializerRegistry.cs
```diff
@@ -282,25 +281,33 @@ public static IEnumerable<SerializableItem> GetSerializableItems(TypeDefinition
...
-            // If there is a explicit or sequential layout, sort by offset
-            if (type.IsSequentialLayout || type.IsExplicitLayout)
-                fieldEnum = fieldEnum.OrderBy(x => x.Offset);
+                if (ignoredMembers.Contains(field))
+                    continue;
 
-            foreach (var field in fieldEnum)
-            {
-                fields.Add(field);
+                if (field.IsPublic || (field.IsAssembly && field.CustomAttributes.Any(a => a.AttributeType.FullName == "Stride.Core.DataMemberAttribute")))
```
IXLLEGACYIXL, line 292: this would break user code that has a inherited version of the datamemberattribute

lets say i make a class `public class FancyDataMember : DataMemberAttribute`
then that would be ignored even though its still a datamember attribute


in roslyn i would get from the compilation the type and check if its original definition is from datamemberattribute and if so it counts ie
```cs
    public static bool HasAttribute(this ISymbol symbol, INamedTypeSymbol attribute)
    {
        if (symbol.GetAttributes().Any(attr => attr.AttributeClass?.OriginalDefinition.Equals(attribute, SymbolEqualityComparer.Default) ?? false))
        {
            return true;
        }
        return false;
    }
```
i dont know if something like that is possible in cecil

but from looking at the current implementation that case isnt handled either.. so it would be a bug fix too in line 311

sources/core/Stride.Core.AssemblyProcessor/ComplexSerializerRegistry.cs
```diff
@@ -382,6 +389,27 @@ public static IEnumerable<SerializableItem> GetSerializableItems(TypeDefinition
...
 
+        static bool IsAccessibleThroughAccessModifiers(PropertyDefinition property)
+        {
+            var get = property.GetMethod;
+            var set = property.SetMethod;
+
+            if (get == null)
+                return false;
+
+            bool forced = property.CustomAttributes.Any(a => a.AttributeType.FullName == "Stride.Core.DataMemberAttribute");
```
IXLLEGACYIXL, line 400: same problem with fullname comparison as mentioned earlier

sources/assets/Stride.Core.Assets/SolutionPlatform.cs
```diff
@@ -276,14 +267,13 @@ public SolutionConfiguration(string name)
         {
             if (name == null) throw new ArgumentNullException("name");
```
IXLLEGACYIXL, line 268: as youve added later on also nameof it would be nice to have here also a nameof `throw new ArgumentNullException(nameof(name))`

IXLLEGACYIXL (changes requested):

sources/core/Stride.Core.AssemblyProcessor/ComplexSerializerRegistry.cs
IXLLEGACYIXL: line 311 of the old code

Eideren: It's actually a fairly involved change, types have to be resolved which likely incur a non-trivial perf' cost if used within this scope. There should be an initial phase to process all types to look for these, resolve only when necessary, and finally, pass those type refs to this method to filter.
And given how niche this issue would be, best to leave this for when we rework the assembly processor.

sources/core/Stride.Core.Reflection/TypeDescriptors/ObjectDescriptor.cs
```diff
@@ -318,8 +320,10 @@ protected virtual bool PrepareMember(MemberDescriptorBase member, MemberInfo met
                 member.Mode = memberAttribute.Mode;
                 if (!member.HasSet)
                 {
-                    if (memberAttribute.Mode == DataMemberMode.Assign || memberType.IsValueType || memberType == typeof(string))
-                        member.Mode = DataMemberMode.Never;
+                    if (memberAttribute.Mode == DataMemberMode.Assign)
+                        throw new ArgumentException($"{memberType.FullName} {member.OriginalName} is not writeable by {memberAttribute.Mode.ToString()}, its {nameof(DataMemberMode)} must not be {memberAttribute.Mode}.");
```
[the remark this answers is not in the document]
Eideren: You referenced an outdated change in your review, the current one doesn't throw.
I'll just leave it as is and let you guys figure this one out, it's not really related to this PR and we've had more than enough back and forth I feel :P

sources/core/Stride.Core.AssemblyProcessor/ComplexSerializerRegistry.cs
manio143: I don't think we can support custom subclasses of attributes. That's not a common practice and most analyzers would not care as they resolve symbols explicitly.

IXLLEGACYIXL (approved):

sources/core/Stride.Core.AssemblyProcessor/ComplexSerializerRegistry.cs
IXLLEGACYIXL: allright, then we can leave it as is

IXLLEGACYIXL (approved):

manio143 (approved):

Eideren:
Will merge this one once https://github.com/stride3d/stride/pull/1694 is merged to avoid more merge conflicts over there.
Might introduce conflicts on this one but we'll see.

IXLLEGACYIXL:
i have one last thing.
in object descriptor is this code 
```cs
            // Member is not displayed if there is a YamlIgnore attribute on it
            if (AttributeRegistry.GetAttribute<DataMemberIgnoreAttribute>(memberInfo) != null)
                return false;
```

the comment is wrong, probably its a leftover from yamldotnet migration like manio said in discord
could you maybe remove it/alter it?

PR #1878, opened 2023-10-02
Fixed Exception Caused By Privacy Policy URL in Crash Reporter

MeharDT:
# PR Details

Clicking the Privacy Policy link in the Crash Reporter will generate an exception.

## Description

The Process requires shell execute to be enabled, this was the default setting in .NET Framework but is no longer the case with .NET Core.

This PR can be tested by adding the following to Main() in Stride.GameStudio/Program.cs,

```
            try
            {
                throw new Exception("Test");
            }
            catch (Exception ex)
            {
                HandleException(ex, 0);
            }
```

## Types of changes

## Checklist

### DIFF
--- a/sources/editor/Stride.Editor.CrashReport/CrashReportForm.cs
+++ b/sources/editor/Stride.Editor.CrashReport/CrashReportForm.cs
@@ -107,7 +107,11 @@ namespace Stride.Editor.CrashReport
         {
             try
             {
-                Process.Start(PrivacyPolicyUrl);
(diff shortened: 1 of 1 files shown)


MeharDT:
Unrelated but the Crash Reporter form still uses Xenko images, I wasn't able to find the correct .xaml file in the solution to fix it.

manio143 (approved):

Eideren:
Thanks !

PR #1881, opened 2023-10-03
fix: Excludes Stride.CommunityToolkit from packages upgrade

VaclavElias:
# PR Details

This update excludes any packages starting with the name `Stride.CommunityToolkit` from the package upgrade.

## Description

The current problem is that the package

 `<PackageReference Include="Stride.CommunityToolkit" Version="1.0.0-preview.11" />`

is being updated to 

`<PackageReference Include="Stride.CommunityToolkit" Version="4.1.0.1898" />`

which is not desired. 

## Motivation and Context

The intention behind `Stride.CommunityToolkit` is to utilize the same namespace structure as the Stride engine, where possible, with the addition of the term `CommunityToolkit` to differentiate it from the main repository. The overarching goal is that once certain functionality is integrated into Stride, the `CommunityToolkit` prefix could be dropped, provided the remaining namespace remains accurate.

## Types of changes

Just additional condition was added.

## Checklist

### DIFF
--- a/sources/engine/Stride.Assets/StridePackageUpgrader.cs
+++ b/sources/engine/Stride.Assets/StridePackageUpgrader.cs
@@ -164,7 +164,7 @@ namespace Stride.Assets
 
                     foreach (var packageReference in packageReferences)
                     {
-                        if (packageReference.EvaluatedInclude.StartsWith("Stride.", StringComparison.Ordinal) && packageReference.GetMetadataValue("Version") != CurrentVersion)
+                        if (packageReference.EvaluatedInclude.StartsWith("Stride.", StringComparison.Ordinal) && !packageReference.EvaluatedInclude.StartsWith("Stride.CommunityToolkit", StringComparison.Ordinal) && packageReference.GetMetadataValue("Version") != CurrentVersion)
                         {
                             packageReference.SetMetadataValue("Version", CurrentVersion).Xml.ExpressedAsAttribute = true;
                             foreach (var metadata in packageReference.Metadata)


Eideren (approved):

manio143:
I'm not sure how I feel about this. I'm not sure if we have the Stride* prefix reserved on NuGet.org so others may create unofficial packages that will get into the same problem. I think a better option would be to explicitly list the packages which need updating, even though there's quite a few of them.

VaclavElias:
I thought about that briefly, that I am adding some exclusion which is fixing my specific case, but what about other such cases. Stride* prefix is not reserved on NuGet.

There are other packages:
https://www.nuget.org/packages/Stride.Awesome.Shaders/1.0.1-Test from Youness
This one is interesting 🙂 https://www.nuget.org/packages/Stride3d.Microsoft.Extensions.Hosting

Shall I create a HashSet with all our NuGet names and place it here as you suggested to explicitly list them and then I can update the logic accordingly https://github.com/stride3d/stride/blob/master/sources/engine/Stride.Assets/StrideConfig.cs

We would just need to remember to update the set if the new package is added to Stride.

Eideren:
> We would just need to remember to update the set if the new package is added to Stride.

If we want to go that route it has to be automated

Eideren:
Moving this to draft until we've figured a proper way to implement this in. Feel free to start a discussion about this on discord @VaclavElias

Doprez:
Would an API call to Nuget to see what versions exist be too much? I assume we can check if the version exists and if not then just skip the package upgrade.

Eideren:
Sounds good to me as long as it handles offline mode properly, a fallback to local nugget cache when connection is unavailable, That way users working on the engine without internet aren't locked out of it.

Doprez:
Would it need the offline usage if this is for package upgrading?

Eideren:
Yes, for users working on the engine without internet. When building the engine, the engine packages are deployed to the local nuget cache. If the user doesn't have a connection, the upgrader should skip packages that don't have a corresponding version in the local cache.

Doprez:
Can this be closed now due to https://github.com/stride3d/stride/pull/2093 ?

PR #1894, opened 2023-10-05
Removed all references to $(SolutionDir) from build artifacts

JeromyWalsh:
# PR Details

Removed all references to SolutionDir or replaced them with references to $(MSBuildThisFileDirectory).

## Description

The $(SolutionDir) property is not generally regarded as an MSBuild-defined property and is instead understood to be a Visual Studio property. Microsoft has made it clear by dropping support for the -r flag from "dotnet build" on solution files, among other things, that the unit of work for command-line and automated builds should be the *.csproj/*.vcxproj files, not the .sln file.

This change replaces any references to $(SolutionDir) with the appropriate alternatives, or removes it outright when unnecessary.

## Motivation and Context

This is a future-forward change that not only untangles an unnecessary, tool-dependent property, but lowers maintenance costs and moves us one step forward towards building on other platforms and outside of Visual Studio or other tools that require a .sln file.

## Types of changes

## Checklist


sources/core/Stride.Core.AssemblyProcessor/Stride.Core.AssemblyProcessor.csproj
```diff
@@ -27,7 +27,7 @@
   
   <!-- Add support for building this project from the Project Directory -->
   <PropertyGroup>
-    <SolutionDir Condition="'$(SolutionDir)' == ''">$(ProjectDir)..\..\..\build\</SolutionDir>
+    <DependencyDir>$(ProjectDir)..\..\..\..\deps</DependencyDir>
```
Ethereal77, line 30: Just a little nitpick:
Shouldn't it be `DependenciesDir` instead? It's the directory where **all** Stride dependencies are located.

JeromyWalsh: Hi @Ethereal77, grammatically speaking you're not wrong, and if people have a strong preference it's an easy change to make.

That being said, historically the preference has been for readability/laziness rather than grammatical accuracy. ;-) As examples, there used to be the "src" and "source", folders. Still remaining are the "bin", and "obj" folders. In our own project hierarchy there's the "build" folder, though technically it's the "builds" folder. Those are all cases where the folder name acted as a sort/filter rather than as a description of the conents, ie. "this is where a source file goes," or "this is where a build file goes."

I also just found DependecyDir easier to read and say than DependenciesDir. But again, that's likely just a dialectical preference or laziness. If people prefer the latter, I'm happy to change it. Thoughts?

Ethereal77: Not a strong preference. I'm not a native english speaker, so it doesn't matter much to me. Anyway I agree with you on the simplicity point.
**Readability > all else** 😁

Eideren:
Will merge this one once https://github.com/stride3d/stride/pull/1694 is merged to avoid more merge conflicts over there.
Might introduce conflicts on this one but we'll see.

JeromyWalsh:
> Will merge this one once #1694 is merged to avoid more merge conflicts over there. Might introduce conflicts on this one but we'll see.

Dude, you don't need to merge my PR for me. :-) You do enough already. I'll merge it and handle resolving the merge conflicts.

Eideren:
Ahah, thanks. This was also to let other maintainers know that they shouldn't merge it yet if they happen to pass by :)

xen2 (approved):

Eideren:
Thanks !

PR #1895, opened 2023-10-05
Remove MSBuild.Extras from project

Jklawreszuk:
# PR Details

## Details

Msbuild.Extras probably is no longer needed, since we are targeting net6.0+ so I removed all related files and lines of code, but ofc would be proper to verify this ~(not tested yet, if anyone has time I would appreciate)~

Edit: I did not noticed any difference


Eideren:
And the last one for today, thanks again @Jklawreszuk, you're doing gods work around here :)

PR #1897, opened 2023-10-06
[Editor] Re-introduce workaround for missing input while navigating

Eideren:
# PR Details
Looked into this issue for the past two weeks, and still no clue as to what the heck is going on here.
Right now the workaround is only active while right click is being held in the scene view, keeping drag and drop working as intended.

Microsoft Spy++ does show all key message as being received by the window, but the different `WndProc` and `PeekMessage` do not receive some of the key up and key down.

Also included a very small change to improve how fast the view is brought back when swapping between tabs. Detach occurs every time tabs are changed, there's no need to unparent the window at that time.

## Related Issue
#94 

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Input/Windows/InputSourceWinforms.cs
+++ b/sources/engine/Stride.Input/Windows/InputSourceWinforms.cs
@@ -48,7 +48,6 @@ namespace Stride.Input
             input = inputManager;
 
             uiControl.LostFocus += UIControlOnLostFocus;
-            MissingInputHack();
 
             // Hook window proc
             defaultWndProc = Win32Native.GetWindowLong(uiControl.Handle, Win32Native.WindowLongType.WndProc);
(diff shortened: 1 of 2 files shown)


Ethereal77:
As GameStudio is registering correctly all the input, maybe a possible solution is to create an editor service that registers all input that should be directed to a game or preview instance, and then passes those inputs "manually" (as custom messages, for example) via one of the hidden "editor game systems".

Eideren:
That's what I intended to do initially, but if we were to implement something this involved I would much rather we have someone more knowledgeable about windows' internals fix this issue altogether as the bug here is likely not just affecting inputs. Or replace the entire windowing and input system in editor with a wpf surface or something similar.
It would also introduce some added input latency, and it feels like we already have more than enough of that as is on the editor side.

Eideren:
This is still making it really hard for users to navigate around the scene view.
Will merge this in soon if there are no objections.

PR #1926, opened 2023-10-11
[Native] - Implement IsXInputDevice method in C#

Jklawreszuk:
# PR Details

## Description

My PR implements IsXInputDevice method in C#

## Related Issue

#1394 

## Todo

- [ ] - Need to check for regression


sources/engine/Stride.Input/Windows/InputSourceWindowsDirectInput.cs
```diff
@@ -4,9 +4,11 @@
 #if (STRIDE_UI_WINFORMS || STRIDE_UI_WPF)
 using System;
 using System.Collections.Generic;
+using System.IO;
+using System.Management;
 using SharpDX;
 using SharpDX.DirectInput;
-using Stride.Native.DirectInput;
+using Microsoft.Win32;
```
IXLLEGACYIXL, line 11: would that make a problem with linux migration?

Jklawreszuk: Note that this file compiles only when STRIDE_UI_WINFORMS or STRIDE_UI_WPF and are set only when StrideFramework is Windows

Jklawreszuk: so propably no

manio143:
While it's amazing to see so much C++ code replaced by so little C#, I'm worried we might not be able to accept code copied from a GPL licensed project into an MIT licensed project, unless with the explicit permission of the copyright holder to re-license the code.

Jklawreszuk:
@manio143 Oh my. Good point, I forgot about the fact that GNU is not a completely "free" license. I'll try to investigate my own implementation then

tebjan:
You could also ask the author if he allows us to use that part of the code. The author always has full control over his work, no matter the license the code is published with.

IXLLEGACYIXL:
The problem is that the GPL code doesnt work with sharpdx

Sharpdx takes the HID path from a device and somehow builds an product ID ( GUID ) from it 
what the GPL code does is just splitting the HID and takes the values and compares them which will never fit to the sharpdx guid

`$devices = Get-WmiObject -Query 'SELECT * FROM Win32_PNPEntity WHERE DeviceID LIKE "%VID_%&PID_%&IG_%"'`
this is how you can select the xinput devices
accordign to microsoft if it has IG_ in it then its an xinput device
https://learn.microsoft.com/en-us/windows/win32/xinput/xinput-and-directinput

to convert a sharpdx guid to a hid i only found that 
https://stackoverflow.com/questions/67059441/how-to-map-directinput-sharpdx-device-to-its-corresponding-hid-c
but that mentioned class is internal 

one option would be to accurately query with wql to get all xdevices and comparing the device path instead of the guid

the gpl code is just doing alot of fancy things but in the end it just converts this example hid path
`HID\VID_045E&PID_028E&IG_00\3&8968588&0&0000`
to that
`045E028E`
( not sure anymore aobut the leading zeros ) 
and this number is far away from a sharpdx guid

archanox:
I haven't checked if this code works or not, but chatgpt says you can do this,

```
using SharpDX.XInput;

// Create a controller object for the first player
var controller = new Controller(UserIndex.One);

// Check if the controller is connected
if (controller.IsConnected)
{
    // Get the controller capabilities
    var capabilities = controller.GetCapabilities(DeviceQueryType.Any);

    // Check if the controller is an XInput device
    if (capabilities.Type == DeviceType.Gamepad)
    {
        // Do something with the controller
        Console.WriteLine("XInput device detected");
    }
    else
    {
        // The controller is not an XInput device
        Console.WriteLine("Non-XInput device detected");
    }
}
else
{
    // No controller is connected
    Console.WriteLine("No controller detected");
}
```
And
```
using SharpDX.DirectInput;

// Create a DirectInput object
var directInput = new DirectInput();

// Create a device instance from the product GUID
var deviceInstance = directInput.GetDevice(productGuid);

// Check if the device is an XInput device
if (deviceInstance.Type == DeviceType.Gamepad)
{
    // The device is an XInput device
    Console.WriteLine("XInput device detected");
}
else
{
    // The device is not an XInput device
    Console.WriteLine("Non-XInput device detected");
}
```

IXLLEGACYIXL:
anything can be a xinput device, even a mouse, it depends how windows tries to use it

sources/engine/Stride.Input/Windows/InputSourceWindowsDirectInput.cs
```diff
@@ -116,6 +118,44 @@ public void OpenDevice(DeviceInstance deviceInstance)
...
+        private bool IsXInputDevice(Guid productGuid)
+        {
+            // Set security level to IMPERSONATE
+
+            DComSessionOptions DComOptions = new DComSessionOptions();
+            DComOptions.Impersonation = ImpersonationType.Impersonate;
+
+            var mySession = CimSession.Create(null, DComOptions);
+
+            IEnumerable<CimInstance> allDevices = mySession.QueryInstances(@"root\cimv2", "WQL", "SELECT * FROM Win32_PNPEntity");
```
manio143, line 131: I imagine this query is not super cheap and results won't be different if called more than once a frame. Consider calling the query and transforming the IEnumerable with regex once, caching the result in the `Scan()` method, so that `IsXInputDevice` can be a simpler lookup.

archanox:
> anything can be a xinput device, even a mouse, it depends how windows tries to use it

Sorry, I might have misunderstood the purpose of the function being replaced.

Jklawreszuk:
Sorry @archanox, I didn't have time to reply. I want to thank you for your help, however, both examples do not compile , because API doesn't even exist. So, that's I gave you thumbs down, nothing personal.. 😅

sources/engine/Stride.Input/Windows/InputSourceWindowsDirectInput.cs
```diff
@@ -116,6 +118,43 @@ public void OpenDevice(DeviceInstance deviceInstance)
...
+            // Set security level to IMPERSONATE
+
+            DComSessionOptions DComOptions = new DComSessionOptions();
+            DComOptions.Impersonation = ImpersonationType.Impersonate;
+
+            var mySession = CimSession.Create(null, DComOptions);
+
+            // Select all device IDs that contain "IG_".  If so, it's an XInput device
+            // This information can not be found from DirectInput 
+            IEnumerable<CimInstance> allDevices = mySession.QueryInstances(@"root\cimv2", "WQL", "SELECT DeviceID FROM Win32_PNPEntity WHERE DeviceId LIKE '&IG_%'");
```
manio143, line 133: Are you sure this is optimization was correct?
`LIKE '&IG_%'` would match strings that start with `&IG_`, but in the regex you previously had that at the end of the string.

Jklawreszuk: Oh, yea my bad little typo, I fixed now

sources/engine/Stride.Input/Windows/InputSourceWindowsDirectInput.cs
manio143: @Jklawreszuk - do you think your WQL optimization is enough here? The `Scan()` method goes in a loop over the devices and checks for each one if it's XInput - are we confident to say there usually wouldn't be many devices returned by DirectInput and thus this isn't a big perf issue? Otherwise we make a Windows call a few times in a tight loop and I imagine those are usually slower. However, at the same time Scan is called on initialization and then only upon manual request it seems.

Jklawreszuk: @manio143 Yea, Very good suggestion 👍. I guess, It is unlikely that something will change under the scan() method call

sources/engine/Stride.Input/Windows/InputSourceWindowsDirectInput.cs
```diff
@@ -18,11 +20,15 @@ internal class InputSourceWindowsDirectInput : InputSourceBase
         private readonly HashSet<Guid> devicesToRemove = new HashSet<Guid>();
         private InputManager inputManager;
         private DirectInput directInput;
+        private IEnumerable<CimInstance> allDevices;
+        private Regex regex;
```
manio143, line 24: Consider renaming it to `xInputDeviceIdRegex`

sources/engine/Stride.Input/Windows/InputSourceWindowsDirectInput.cs
```diff
@@ -116,6 +134,37 @@ public void OpenDevice(DeviceInstance deviceInstance)
...
         }
+
+        private bool IsXInputDevice(Guid productGuid)
+        {
+            // Set security level to IMPERSONATE
+
+            DComSessionOptions DComOptions = new DComSessionOptions();
+            DComOptions.Impersonation = ImpersonationType.Impersonate;
+
+            var mySession = CimSession.Create(null, DComOptions);
```
manio143, line 145: I believe all above and including this line has to be moved to `GetAllXInputDevices()`

sources/engine/Stride.Input/Windows/InputSourceWindowsDirectInput.cs
```diff
@@ -54,12 +60,24 @@ public override void Update()
...
         }
 
+        /// <summary>
+        /// Select all device IDs that contain "IG_".  If so, it's an XInput device
+        /// This information can not be found from DirectInput 
+        /// </summary>
+        private IEnumerable<CimInstance> GetAllXInputDevices()
+        {
+            return mySession.QueryInstances(@"root\cimv2", "WQL", "SELECT DeviceID FROM Win32_PNPEntity WHERE DeviceID LIKE '%&IG_%'");
+        }
```
manio143, line 70: Consider returning a list of strings here:
```csharp
var query = mySession.QueryInstances(...);
var deviceIdPrefixes = query.Select(device => regex.Match(device.CimInstanceProperties["DeviceID"].Value.ToString()))
    .Where(match => match.Success)
    .Select(match => (match.Groups[1].ToString() + match.Groups[2].ToString()).ToLower())
    .ToList();
```
Then in `IsXInputDevice(Guid productGuid)` all you need to do is loop over prefixes and check
```csharp
if (productGuid.ToString().StartsWith(guidPart))
{
     return true;
}
```

sources/engine/Stride.Input/Windows/InputSourceWindowsDirectInput.cs
Jklawreszuk: done

sources/engine/Stride.Input/Windows/InputSourceWindowsDirectInput.cs
Jklawreszuk: done! Also I think ToList() is unnecessary.

manio143: Without the ToList you will repeat regex matching on each enumeration (if the query returns a materialized enumerable).

Jklawreszuk: right! then I'll add it too

manio143 (approved):
Code wise it's good. If you can confirm it got tested I'll merge it.

IXLLEGACYIXL:
> Code wise it's good. If you can confirm it got tested I'll merge it.

i can test it with a playstation controller and emulate that it should be treated as xinput, dunnno if thats sufficient

manio143:
There seems to be an issue with the build about no input files to clang - if there's no more cop files, we may need to turn off the property which triggers native lib compilation https://teamcity.stride3d.net/buildConfiguration/Engine_BuildWindowsOpenGL/23137?hideProblemsFromDependencies=false&hideTestsFromDependencies=false

Jklawreszuk:
@manio143 I've managed to fix building my PR by simply checking if the input (c files collection) is empty. 
I personally don't have a gamepad that supports xinput so it returned false oposed to @IXLLEGACYIXL, so it seems to work 👀

IXLLEGACYIXL:
> @manio143 I've managed to fix building my PR by simply checking if the input (c files collection) is empty. I personally don't have a gamepad that supports xinput so it returned false oposed to @IXLLEGACYIXL when testing on XInput device, so it seems to work 👀

i tried your query and it works, but i cant tell if my xinput is even valid, as the playstation controller is emulated and my mouse too .. .i get with your query ( your test script ) these two devices... i think a real xinput device is needed which i dont have , only the emulated one so dont count on my test

manio143:
Build looks good. I'll see if this works with my Xbox controller.

manio143:
Ok, so this doesn't fully work for me.
The device's `productGuid` is `02ff045e-0000-0000-0000-504944564944`, but the deviceId received from the Cim query is `"045e02ff"` so it didn't recognize it correctly.
Seems the bug is in the string concatenation.
Instead of
```
match.Groups[1].ToString() + match.Groups[2].ToString()
```
should be
```
match.Groups[2].ToString() + match.Groups[1].ToString()
```

@IXLLEGACYIXL you mentioned this was working for you, can you provide what was the device Id in your case?

Also, there were duplicates in the deviceId list from Cim query - maybe we can run a `.Distinct()` before `ToList()`?

Jklawreszuk:
Alright, good to hear 👍. Anyway, yes, let's filter duplicates. I hope that string concatenation is deterministic. Otherwise method needs to exctract PID and VID separately(??)

IXLLEGACYIXL:
> Ok, so this doesn't fully work for me.
> The device's `productGuid` is `02ff045e-0000-0000-0000-504944564944`, but the deviceId received from the Cim query is `"045e02ff"` so it didn't recognize it correctly.
> Seems the bug is in the string concatenation.
> Instead of
> ```
> match.Groups[1].ToString() + match.Groups[2].ToString()
> ```
> should be
> ```
> match.Groups[2].ToString() + match.Groups[1].ToString()
> ```
> 
> @IXLLEGACYIXL you mentioned this was working for you, can you provide what was the device Id in your case?
> 
> Also, there were duplicates in the deviceId list from Cim query - maybe we can run a `.Distinct()` before `ToList()`?

ive only tested the query itself, not the entire thing

the guid is built but hidden and internal thats what ive written earlier

manio143:
>  I hope that string concatenation is deterministic (i.e. PID + VID)

The way your regex is written guarantees determinism here.

IXLLEGACYIXL:
> Ok, so this doesn't fully work for me. The device's `productGuid` is `02ff045e-0000-0000-0000-504944564944`, but the deviceId received from the Cim query is `"045e02ff"` so it didn't recognize it correctly. Seems the bug is in the string concatenation. Instead of
> 
> ```
> match.Groups[1].ToString() + match.Groups[2].ToString()
> ```
> 
> should be
> 
> ```
> match.Groups[2].ToString() + match.Groups[1].ToString()
> ```
> 
> @IXLLEGACYIXL you mentioned this was working for you, can you provide what was the device Id in your case?
> 
> Also, there were duplicates in the deviceId list from Cim query - maybe we can run a `.Distinct()` before `ToList()`?

so no for me its not working and i cant try as i lose my mouse with the emulator
ive only tried the wql

PR #1927, opened 2023-10-11
Add editor settings for the camera speed increase/decrease hotkeys

adrsch:
# PR Details

Just lets you change the keybinding for this in the editor settings. Defaults to the old hardcoded bindings so it should be unnoticed by most. 

## Types of changes

## Checklist


Eideren:
Kind of sad that multiple keys or key combinations are not supported by that class, but that's something for another PR. Thanks !

PR #1951, opened 2023-10-16
Update SSH.NET to 2023.0.0

WojciechNagorski:
Update SSH.NET to the latest version https://www.nuget.org/packages/SSH.NET/2023.0.0

## New features:
* Support for .NET 6, 7, and .NET Standard 2.0/2.1
* Support for RSA-SHA256/512 signature algorithms
* Support for parsing OpenSSH keys with ECDSA 256/384/521 and RSA
* Support for SHA256 and MD5 fingerprints for host key validation
* Added async support to SftpClient and SftpFileStream
* Added ISftpFile interface to SftpFile
* Removed support for old target frameworks
* Improved performance and stability
* Added the ability to set the last write and access time for Sftp files


Eideren:
Thanks for the PR !
Looks like your lib is used to start remote play sessions on linux and mac from windows. The project builds with this change, but testing it is something else. I don't know the setup we require to run this particular feature, and I doubt xen will have the time to check this one out.

There's not a whole lot, but here's more context as for why the author was contributing on another repo https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/pull/2057#issuecomment-1760737871
If you want to clarify anything @WojciechNagorski feel free.

PR #1955, opened 2023-10-16
[Presentation] Reduce allocations when parsing number in NumericTextBox

Kryptos-FR:
Following #1811, remove some string allocations.


Eideren:
Thanks !

PR #1959, opened 2023-10-17
[Breaking] Scoping generic extension methods

Fydar:
> _"3rd PR and I'm already suggesting breaking changes; I hope this isn't too intrusive, but I considered this quite high priority."_

Adding a generic scope to unscoped generic extension methods.

# PR Details

## Description

This means that these extension methods no longer work with `IntPtr` and `IReferencable` types, however this feature was unused internally and the engine continues to compile like normal.

Should we be interested in using a similar extension method, separate extension methods for `IReferencable` types could be created, potentially without "Dispose" in the name, and instead use the "Release" naming associated with `IReferencable` types.

## Related Issue

- None

## Motivation and Context

This was causing every object to have additional methods appearing in the IDE and on the documentation, which can be very confusing for new developers and irritating for experienced ones.

Everywhere in the documentation are notes about this extension method being usable, despite it being meaningless in the majority of use cases.

image

And it is also appears on every object, regardless of type.

image

## Types of changes

## Checklist

### DIFF
--- a/sources/core/Stride.Core/ComponentBaseExtensions.cs
+++ b/sources/core/Stride.Core/ComponentBaseExtensions.cs
@@ -18,6 +18,7 @@ namespace Stride.Core
         /// <param name="container">The container that will keep a reference to the component.</param>
         /// <returns>The same component instance</returns>
         public static T DisposeBy<T>(this T thisArg, ICollectorHolder container)
+            where T : IDisposable
         {
             if (ReferenceEquals(thisArg, null))
                 return default(T);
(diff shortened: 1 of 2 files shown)


manio143:
> This means that these extension methods no longer work with `IntPtr` and `IReferencable` types

I got a little confused as to why this is mentioned if no code changes were necessary in relation to those types.
I see it's based on this code
https://github.com/stride3d/stride/blob/51f60ac529593379d3954999faf86ef8abf00cec/sources/core/Stride.Core/ObjectCollector.cs#L56-L59

The reason why those methods are generic over `T` is because they seemed to be used in an initialization pattern
```csharp
var smth = new Thing().DisposeBy(collector);
```

* Adding an overload for `IntPtr` makes sense.

But there's no way to overload on `T` where it's either `IDisposable` or `IReferencable`.
* I would suggest adding
    ```csharp
    T ReleaseBy<T>(this T reference, ICollectionHolder holder) where T : IReferencable;
    ```

manio143 (approved):

Eideren:
Did you test this change out in editor/in game ? I don't really have time right now to check

Fydar:
Building the solution works, building and running the editor works.

Eideren:
Thanks !

PR #1964, opened 2023-10-18
Fix #1769 and introduce an optional argument to specify a different r…

Kryptos-FR:
…ounding strategy if needed.

# PR Details

## Types of changes

## Checklist

### DIFF
--- a/sources/core/Stride.Core.Mathematics/Int3.cs
+++ b/sources/core/Stride.Core.Mathematics/Int3.cs
@@ -186,7 +186,7 @@ namespace Stride.Core.Mathematics
         /// </remarks>
         public int Length()
         {
-            return (int)Math.Sqrt((X * X) + (Y * Y) + (Z * Z));
+            return (int)MathF.Sqrt((X * X) + (Y * Y) + (Z * Z));
         }
 
         /// <summary>
@@ -195,7 +195,7 @@ namespace Stride.Core.Mathematics
         /// <returns>The length of the vector untruncated.</returns>
         public float LengthUntruncated() 
         {
-            return (float)Math.Sqrt((X * X) + (Y * Y) + (Z * Z));
+            return (float)MathF.Sqrt((X * X) + (Y * Y) + (Z * Z));
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


Kryptos-FR:
I removed the new constructors because we already have conversion operators, so it's not needed. If someone wants to convert a `Vector3` to an `Int3` (or the opposite) they can just do:
```csharp
Vector3 vec3 = new Vector3(1.0f, 1.0f, 1.0f);
Int3 int3 = (Int3)vec3;
vec3 = (Vactor3)int3;
```

sources/core/Stride.Core.Mathematics/Int3.cs
```diff
@@ -200,12 +190,12 @@ public int Length()
...
         /// <summary>
-        /// Calculates the unrounded length of the vector.
+        /// Calculates the untruncated length of the vector.
         /// </summary>
-        /// <returns>The length of the vector unrounded.</returns>
+        /// <returns>The length of the vector untruncated.</returns>
         public float LengthUntruncated() 
         {
-            return Math.Sqrt((X * X) + (Y * Y) + (Z * Z));
+            return (float)Math.Sqrt(Length());
```
froce, line 198: should be `LengthSquared()`

Kryptos-FR: oops

Kryptos-FR: And actually given that Length also does the inlining manually, I shouldn't call a method but also inline.

sources/core/Stride.Core.Mathematics/Int3.cs
```diff
@@ -551,7 +541,32 @@ public static Int3 Min(Int3 left, Int3 right)
...
+        /// </summary>
+        /// <param name="value">The source vector.</param>
+        /// <param name="result">When the method completes, contains an new vector composed of the rounded values of the specified vector.</param>
+        /// <param name="rounding">The rounding strategy to use.</param>
+        public static void Round(ref Vector3 value, out Int3 result, MidpointRounding rounding = default)
+        {
+            result.X = (int)Math.Round(value.X, rounding);
+            result.Y = (int)Math.Round(value.Y, rounding);
+            result.Z = (int)Math.Round(value.Z, rounding);
+        }
```
Eideren, line 556: Replace `ref Vector3 value` with `in Vector3 value`
`MathF` has an overload for midpoint rounding see https://learn.microsoft.com/en-us/dotnet/api/system.mathf.round?view=net-7.0

Kryptos-FR: > Replace ref Vector3 value with in Vector3 value

Those should be wider changes throughout the whole codebase. A lot of methods don't use the `in` keyword. For me that's outside this PR which is mostly a build fix.

> MathF has an overload for midpoint rounding

Should I use MathF for the other places in that file (e.g. Sqrt and Pow)?

Eideren: > For me that's outside this PR
Fixing it later would be a breaking change.

>  A lot of methods don't use the in keyword.
I took care of a bunch of them a while ago but not all of them, no. So there's a bit of a mix right now.

> Should I use MathF for the other places in that file (e.g. Sqrt and Pow)?
Yes please, as long as you touched those lines in this PR, we'll leave the rest of the file for another PR.

Kryptos-FR: Seems reasonable.

Kryptos-FR: For Round I will use `MathF`. However for `Length()` and `LengthUntruncated()` I think we should keep using `Math`. The rationale is that it is doing a square root of a multiplication of `int`. An `int` * `int` fits in a `long` and the equivalent floating point type is `double`. So to avoid overflows, we should only convert the result to `float` at the end.

Eideren:
Thanks !

PR #1967, opened 2023-10-18
[Shaders] Fix negative values in shader parser to be treated as literals.

### LINKED ISSUE #1963: SpaceEscape sample TransformationBendWorld shader broken

**Version**: 4.0.1.1400

**Platform(s)**: All

**Describe the bug**
`DeformFactorX` & `DeformFactorY` in `TransformationBendWorld.sdsl` gets reset to zero

**To Reproduce**
1. Run the SpaceEscape sample.

**Expected behavior**
`DeformFactorX` & `DeformFactorY` should retain their default values as written in the shader.

**Additional context**
Bug introduced in #765 caused by
https://github.com/vvvv/stride/blob/7fa2275c3f4d392bb9837d46029f46d45f8c03cb/sources/engine/Stride.Graphics/Effects/Effect.cs#L310-L318

**Workaround**
In `BendFogRenderFeature.cs` go to method `PrepareEffectPermutations` and add the lines
```
// Generate shader permuatations
renderEffect.EffectValidator.ValidateParameter(GameParameters.EnableBend, renderMesh.Mesh.Parameters.Get(GameParameters.EnableBend));
// <----- START OF THE WORKAROUND
if (renderMesh.Mesh.Parameters.Get(GameParameters.EnableBend))
{
    var eval1 = renderMesh.Mesh.Parameters.Get(TransformationBendWorldKeys.DeformFactorX);
    var eval2 = renderMesh.Mesh.Parameters.Get(TransformationBendWorldKeys.DeformFactorY);
}
// <----- END OF THE WORKAROUND
renderEffect.EffectValidator.ValidateParameter(GameParameters.EnableFog, renderMesh.Mesh.Parameters.Get(GameParameters.EnableFog));
renderEffect.EffectValidator.ValidateParameter(GameParameters.EnableOnflyTextureUVChange, renderMesh.Mesh.Parameters.Get(GameParameters.EnableOnflyTextureUVChange));
```

This will immediately evaluate the default values preventing the shader using zero'd values.

Basewq:
# PR Details

## Description

Fixes shaders with negative default values by adding parser rule to handle negative values.

## Related Issue

Bug introduced in #765
Fixes #1963

## Motivation and Context

SpaceEscape sample project's `TransformationBendWorld.sdsl` shader was broken due to #765 adding overriding default values.
However that change caused a problem from this line
https://github.com/stride3d/stride/blob/51f60ac529593379d3954999faf86ef8abf00cec/sources/engine/Stride.Shaders.Parser/ShaderLinker.cs#L699
leading to this method:
https://github.com/stride3d/stride/blob/51f60ac529593379d3954999faf86ef8abf00cec/sources/engine/Stride.Shaders.Parser/ShaderLinker.cs#L984-L990
The default value would only accept `LiteralExpression` otherwise it would return default (which is 0).
For some reason, a negative value would be a `UnaryExpression`, so this pull changes things to force the negative value into a `LiteralExpression`.

Unfortunately, this does mean there's another problem (not fixed in this pull), where a default value being something like:
`stage float DeformFactorX = 1 + 2;`
will still cause an issue because the default value (1+2) is not a literal.
Not really sure what the solution is here. Either do a evaluation on the expression or maybe change the `ToScalar` method calls to return null so it does not override the default value?

Also, while I tested this directly against the SpaceEscape project and tried running against the existing unit tests (which seems to have passed), I am unsure about the larger implication of where this parser operates. Also don't really understand how the parser grammar actually works, so if anyone actually knows how this works, feel free to make the proper fix.
### Other notes
Probably a future thing, but it seems Irony is still alive, and has a `NumberLiteral`
https://github.com/IronyProject/Irony
Not sure if that would be the proper long term fix.

## Types of changes

## Checklist

### DIFF
--- a/sources/shaders/Stride.Core.Shaders/Grammar/ShaderGrammar.Ast.cs
+++ b/sources/shaders/Stride.Core.Shaders/Grammar/ShaderGrammar.Ast.cs
@@ -1056,6 +1056,21 @@ namespace Stride.Core.Shaders.Grammar
             }
         }
 
+        protected void CreateNegativeFloatLiteral(ParsingContext context, ParseTreeNode node)
+        {
+            var literalFloat = Ast<Literal>(node);
+            //// negative_float_literal.Rule =
+            ////    [0]       [1]
+            ////    "-" + float_literal;
+            var postiveValueNode = node.ChildNodes[1];
+            var positiveValueLiteral = (Literal)postiveValueNode.AstNode;
+            if (positiveValueLiteral.Value is float floatValue)
+            {
+                literalFloat.Value = -floatValue;
+            }
+            literalFloat.Text = "-" + postiveValueNode.Token.Text;
+        }
+
         /// <summary>
         /// The create integer literal.
         /// </summary>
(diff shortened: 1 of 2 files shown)


manio143 (approved):

azeno:
I think `UnaryExpression` is correct. Changing the grammar seems off - it's rather my PR which doesn't handle it correctly. Wasn't aware of this issue at the time. Couldn't the `ToScalar` method check more cases, like `UnaryExpression` or even `BinaryExpression` and evaluate accordingly?

azeno:
For example look at the C# expression `-1`, it also reports it as a `UnaryExpression` (https://roslynquoter.azurewebsites.net/)

azeno:
Let me open a new PR with my proposals.

Basewq:
Closed in favor of #1990

PR #1908, opened 2023-10-08
[Editor] Remove some windows dependencies in editor libraries

Kryptos-FR:
# PR Details

Target .NET 6 (cross-platform) for editor libraries that don't have any Windows dependencies.

## Description

Introduce `StrideEditorAppTargetFramework` in `sources\targets\Stride.Core.TargetFrameworks.Editor.props` and change existing `StrideEditorTargetFramework` to target `net6.0` only.

## Related Issue

N/A

## Motivation and Context

Some editor libraries should be cross-platform (esp. `Stride.Core.Assets.Quantum`). This will allow to reuse these libraries for other tools that don't necessarily target Windows.

On the other hand, editor libraries with `Presentation` in their name have a strong dependency to either Windows Form or WPF.

## Types of changes

- [x] Build change

## Checklist

- [x] All new and existing tests should pass.

PR #1968, opened 2023-10-18
[WIP]changed path to compiler app

Doprez:
# PR Details

Currently in master, projects can't find the compiler app because the path has changed with the windows dependency removal changes. 

## Description

This removes the windows tag from the path and projects are now able to find compiler app.

## Related Issue

only reported in Discrod atm https://discord.com/channels/500285081265635328/707563402482024488/1164325964776493198

## Types of changes

- [ ] Docs change / refactoring / dependency upgrade
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [ ] My change requires a change to the documentation.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.
- [x] **I have built and run the editor to try this change out.**

PR #1969, opened 2023-10-19
[CompilerApp] Use StrideEditorTargetFramework

Kryptos-FR:
# PR Details

`Stride.Core.Assets.CompilerApp.targets`  hadn't been updated in PR #1908

## Types of changes

## Checklist
- [x] Solution on a clean clone builds.

### DIFF
--- a/sources/assets/Stride.Core.Assets.CompilerApp/build/Stride.Core.Assets.CompilerApp.targets
+++ b/sources/assets/Stride.Core.Assets.CompilerApp/build/Stride.Core.Assets.CompilerApp.targets
@@ -118,8 +118,8 @@
   <Target Name="_StridePrepareAssetCompiler">
     <PropertyGroup>
       <!-- First try NuGet layout, then git checkout -->
-      <StrideCompileAssetCommand Condition="'$(StrideCompileAssetCommand)' == '' And Exists('$(MSBuildThisFileDirectory)..\lib\net6.0-windows7.0\')">$(MSBuildThisFileDirectory)..\tools\net6.0-windows7.0\Stride.Core.Assets.CompilerApp.exe</StrideCompileAssetCommand>
-      <StrideCompileAssetCommand Condition="'$(StrideCompileAssetCommand)' == '' And Exists('$(MSBuildThisFileDirectory)..\Stride.Core.Assets.CompilerApp.csproj')">$(MSBuildThisFileDirectory)..\bin\$(Configuration)\net6.0-windows7.0\Stride.Core.Assets.CompilerApp.exe</StrideCompileAssetCommand>
+      <StrideCompileAssetCommand Condition="'$(StrideCompileAssetCommand)' == '' And Exists('$(MSBuildThisFileDirectory)..\lib\net6.0\')">$(MSBuildThisFileDirectory)..\tools\net6.0\Stride.Core.Assets.CompilerApp.exe</StrideCompileAssetCommand>
+      <StrideCompileAssetCommand Condition="'$(StrideCompileAssetCommand)' == '' And Exists('$(MSBuildThisFileDirectory)..\Stride.Core.Assets.CompilerApp.csproj')">$(MSBuildThisFileDirectory)..\bin\$(Configuration)\net6.0\Stride.Core.Assets.CompilerApp.exe</StrideCompileAssetCommand>
     </PropertyGroup>
 
     <Error Condition="!Exists('$(StrideCompileAssetCommand)')" Text="Stride AssetCompiler could not be found (Command: &quot;$(StrideCompileAssetCommand)&quot;)"/>


manio143:
See my comment on #1968 about Stride solution targets vs user targets

Kryptos-FR:
Good point

> Also, I don't think you can use that property in the `build` folder which is shipped to the user - either has to be defined in the same file or use net6.0 explicitly.

Good point. I didn't realize it was shipped.

sources/assets/Stride.Core.Assets.CompilerApp/build/Stride.Core.Assets.CompilerApp.targets
```diff
@@ -118,8 +118,8 @@
   <Target Name="_StridePrepareAssetCompiler">
     <PropertyGroup>
       <!-- First try NuGet layout, then git checkout -->
-      <StrideCompileAssetCommand Condition="'$(StrideCompileAssetCommand)' == '' And Exists('$(MSBuildThisFileDirectory)..\lib\net6.0-windows7.0\')">$(MSBuildThisFileDirectory)..\tools\net6.0-windows7.0\Stride.Core.Assets.CompilerApp.exe</StrideCompileAssetCommand>
-      <StrideCompileAssetCommand Condition="'$(StrideCompileAssetCommand)' == '' And Exists('$(MSBuildThisFileDirectory)..\Stride.Core.Assets.CompilerApp.csproj')">$(MSBuildThisFileDirectory)..\bin\$(Configuration)\net6.0-windows7.0\Stride.Core.Assets.CompilerApp.exe</StrideCompileAssetCommand>
+      <StrideCompileAssetCommand Condition="'$(StrideCompileAssetCommand)' == '' And Exists('$(MSBuildThisFileDirectory)..\lib\net6.0\')">$(MSBuildThisFileDirectory)..\tools\net6.0\Stride.Core.Assets.CompilerApp.exe</StrideCompileAssetCommand>
```
Kryptos-FR, line 121: That path doesn't look correct by the way. I can't find that `tools` folder in either the nuget package or the installed stride.

ly3027929699: i can't find the .exe too

Doprez: When building with net6.0-windows7.0 I get the tools folder ith the exe in it.
image
 But not with just net6.0 and that was part of the problem it seemed.

ly3027929699: can you gen exe with the .net8.0? @Doprez

Doprez: it's the same issue, with net8.0-windows it works but with just net8.0 it doesn't.

Basewq: https://learn.microsoft.com/en-us/dotnet/core/deploying/
'exe' is a Windows thing so only appears when the platform is specified, otherwise you only get the 'cross-platform dll'

Doprez: That makes more sense. I guess the next question is why is it an exe instead of a loadable dll/nuget like other libraries to remove this dependancy. Converting the CompilerApp cross platform would probably be a fairly large task I am guessing.

manio143: The way this is worked around for AssemblyProcessor is that it is loaded as an MSBuild task, but here it wouldn't be possible (VS would require it to target netstandard2.0).
I suggest reverting asset compiler to target windows for the moment to unblock compilation, and file an issue to discuss how to make it run cross platform correctly (linked with the Linux issue made by Caspian).

manio143: Would it be possible instead to invoke `dotnet /path/to/dll`?

Doprez:
While this gets worked on should the original PR be reverted? it breaks GameStudio in master currently.

manio143:
Agreed, let's revert asset compiler TFM change and keep the targets as they were for the moment to unblock people.

IXLLEGACYIXL:
after running into issues again, when the assetcompiler compiled once with net6.0-windows and you change the assemblyprocessor to net6 it still works but it uses the old assemblyprocessor

Kryptos-FR:
Not relevant any more (we moved to .NET 8). Will be done differently on the other xplat PRs. Closing.

Issue #1971, opened 2023-10-20
Support ScissorRectangle in image effects


dotlogix:
**Is your feature request related to a problem? Please describe.**
I want to apply image effects only to specific regions in a render texture as far as I understand this should be possible with scissor rectangles. However the image effect abse class resets the current pipeline so setting it before the image effect is applied has no effect. 

**Describe the solution you'd like**
Would be nice to have a method to set the scissor rectangle in the effect itself same as SetViewport.

**Describe alternatives you've considered**
Rewriting a custom image effect class and copy pasting the whole code but with additional scissor support

**Additional context**
Currently trying to create an ImageProcessing helper class so I can easily draw/composite/resize/extract texture data at runtime.

PR #1972, opened 2023-10-20
Add support for multiple viewports and scissors

dotlogix:
# PR Details

Adding support for multiple viewports as well as scissors for image effects

## Description

Adding support for multiple viewports as well as scissors for image effects

## Related Issue

#1971 

## Motivation and Context
I want to apply image effects only to specific regions in a render texture as far as I understand this should be possible with scissor rectangles. However the image effect base class resets the current pipeline so setting it before the image effect is applied has no effect.

## Types of changes

## Checklist

Unfortunately I can not run the tests locally at least a lot of them fail during the build but doesn't seem to be related to my changes so probably sth is missing in my setup?

- [x] My change requires a change to the documentation. (Might make sense to add this to the docs?)

### DIFF
--- a/sources/engine/Stride.Rendering/Rendering/Images/ImageEffect.cs
+++ b/sources/engine/Stride.Rendering/Rendering/Images/ImageEffect.cs
@@ -2,8 +2,9 @@
 // Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
 
 using System;
-
+using System.Runtime.InteropServices;
 using Stride.Core;
+using Stride.Core.Mathematics;
 using Stride.Graphics;
 
 namespace Stride.Rendering.Images
@@ -22,7 +23,11 @@ namespace Stride.Rendering.Images
         private Texture[] outputRenderTargetViews;
         private Texture[] createdOutputRenderTargetViews;
 
-        private Viewport? viewport;
+        private int boundViewportCount;
+        private readonly Viewport[] viewports = new Viewport[CommandList.MaxViewportAndScissorRectangleCount];
+
+        private int boundScissorCount;
+        private readonly Rectangle[] scissors = new Rectangle[CommandList.MaxViewportAndScissorRectangleCount];
 
         /// <summary>
         /// Initializes a new instance of the <see cref="ImageEffect" /> class.
@@ -142,13 +147,71 @@ namespace Stride.Rendering.Images
             outputRenderTargetViews = renderTargetViews;
         }
 
+
         /// <summary>
-        /// Sets the viewport to use .
+        /// Binds a single viewport to the rasterizer stage.
         /// </summary>
-        /// <param name="viewport">The viewport.</param>
-        public void SetViewport(Viewport? viewport)
+        /// <param name="value">The viewport.</param>
+        public void SetViewport(Viewport? value)
         {
-            this.viewport = viewport; // TODO: support multiple viewport?
+            if (value.HasValue)
+            {
+                var viewport = value.Value;
+                SetViewports(MemoryMarshal.CreateReadOnlySpan(ref viewport, 1));
+            }
+            else
+            {
+                SetViewports(ReadOnlySpan<Viewport>.Empty);
+            }
+        }
+
+        /// <summary>
+        /// Binds a set of viewports to the rasterizer stage.
+        /// </summary>
+        /// <param name="values">The set of viewports to bind.</param>
+        /// <remarks>Only <see cref="CommandList.MaxViewportAndScissorRectangleCount"/> scissors can be used simultaneously</remarks>
+        public void SetViewports(ReadOnlySpan<Viewport> values)
+        {
+            if (values.Length > CommandList.MaxViewportAndScissorRectangleCount)
+            {
+                throw new ArgumentException($"Can not use more than {CommandList.MaxViewportAndScissorRectangleCount} viewports simultaneously");
+            }
+
+            boundViewportCount = values.Length;
+            values.CopyTo(viewports);
+        }
+
+        /// <summary>
+        /// Binds a single scissor rectangle to the rasterizer stage.
+        /// </summary>
+        /// <param name="value">The scissor rectangle.</param>
+        public void SetScissorRectangle(Rectangle? value)
+        {
+            if (value.HasValue)
+            {
+                var scissor = value.Value;
+                SetScissorRectangles(MemoryMarshal.CreateReadOnlySpan(ref scissor, 1));
+            }
+            else
+            {
+                SetScissorRectangles(ReadOnlySpan<Rectangle>.Empty);
+            }
+        }
+
+        /// <summary>
+        /// Binds a set of scissor rectangles to the rasterizer stage.
+        /// </summary>
+        /// <param name="values">The set of scissor rectangles to bind.</param>
+        /// <remarks>Only <see cref="CommandList.MaxViewportAndScissorRectangleCount"/> viewports can be used simultaneously</remarks>
+        public void SetScissorRectangles(ReadOnlySpan<Rectangle> values)
+        {
+            if (values.Length > CommandList.MaxViewportAndScissorRectangleCount)
+            {
+                throw new ArgumentException($"Can not use more than {CommandList.MaxViewportAndScissorRectangleCount} scissors simultaneously");
+            }
+
+            boundScissorCount = values.Length;
+            values.CopyTo(scissors);
         }
 
         protected override void PreDrawCore(RenderDrawContext context)
@@ -167,7 +230,7 @@ namespace Stride.Rendering.Images
         /// <param name="context"></param>
         protected virtual void SetRenderTargets(RenderDrawContext context)
         {
-            // Transtion inputs to read sources
+            // Transition inputs to read sources
             for (int i = 0; i <= maxInputTextureIndex; ++i)
             {
                 if (inputTextures[i] != null)
@@ -210,9 +273,14 @@ namespace Stride.Rendering.Images
                 context.CommandList.SetRenderTargetsAndViewport(outputDepthStencilView, null);
             }
 
-            if (viewport.HasValue)
+            if (boundViewportCount > 0)
             {
-                context.CommandList.SetViewport(viewport.Value);
+                context.CommandList.SetViewports(boundViewportCount, viewports);
+            }
+
+            if (boundScissorCount > 0)
+            {
+                context.CommandList.SetScissorRectangles(boundScissorCount, scissors);
             }
         }
 
--- a/sources/engine/Stride.Graphics/CommandList.cs
+++ b/sources/engine/Stride.Graphics/CommandList.cs
@@ -12,7 +12,7 @@ namespace Stride.Graphics
     public partial class CommandList : GraphicsResourceBase
     {
         private const int MaxRenderTargetCount = 8;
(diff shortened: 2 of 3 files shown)


dotlogix:
Seems like the scissors have to be enabled in the rasterizer stage.
Currently there is no logic around that in the ImageEffect but the ImageEffectShader overrides the current pipeline state not sure if this should be moved down and instead be a virtual method as well?

My current workaround in testing is to expose a property of type RasterizerStateDescription similar to how it's done for the DepthStencilState and BlendState not sure if this is the desired version?

Any suggestions?

dotlogix:
@manio143, @tebjan, @xen2 could I get some feedback if this could happen soon-ish?
I created a utility class for the community toolit which needs this functionality. Would be nice if this or sth similar could be merged?

sources/engine/Stride.Rendering/Rendering/Images/ImageEffect.cs
```diff
@@ -143,12 +148,79 @@ public void SetDepthOutput(Texture depthStencilView, params Texture[] renderTarg
...
+        /// <param name="values">The viewports.</param>
+        public void SetViewports(int viewportCount, Viewport[] values)
+        {
+            boundViewportCount = viewportCount;
+            if(viewportCount > 0)
+                Array.Copy(values, 0, this.viewports, 0, boundViewportCount);
+        }
+
+        /// <summary>
+        /// Binds a single scissor rectangle to the rasterizer stage. See <see cref="Render+states"/> to learn how to use it.
```
manio143, line 189: Is this a valid `cref` reference?

sources/engine/Stride.Rendering/Rendering/Images/ImageEffect.cs
```diff
@@ -143,12 +148,79 @@ public void SetDepthOutput(Texture depthStencilView, params Texture[] renderTarg
...
+        /// <summary>
+        /// Sets the viewport to use.
+        /// </summary>
+        /// <param name="viewportCount">The viewports.</param>
+        /// <param name="values">The viewports.</param>
+        public void SetViewports(int viewportCount, Viewport[] values)
+        {
+            boundViewportCount = viewportCount;
+            if(viewportCount > 0)
+                Array.Copy(values, 0, this.viewports, 0, boundViewportCount);
```
manio143, line 185: I think it would be preferable to do bounds check before calling copy so that you can throw a more meaningful ArgumentOutOfRangeException

manio143 (commented):
Leaving some code quality comments. I think it's fine, but I'll leave it to someone with more graphics pipeline experience to approve the changes.

sources/engine/Stride.Rendering/Rendering/Images/ImageEffect.cs
```diff
@@ -143,12 +148,79 @@ public void SetDepthOutput(Texture depthStencilView, params Texture[] renderTarg
...
-            this.viewport = viewport; // TODO: support multiple viewport?
+            SetViewports(values.Length, values);
+        }
+
+        /// <summary>
+        /// Sets the viewport to use.
+        /// </summary>
+        /// <param name="viewportCount">The viewports.</param>
+        /// <param name="values">The viewports.</param>
+        public void SetViewports(int viewportCount, Viewport[] values)
```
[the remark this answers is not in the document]
dotlogix: It is public in the command list and there was a todo comment about this. I think this is mostly done for convenience.
Like you can hold your own viewport buffer with 12 elements but only 2 of them are active.
This method would allow you to avoid unneccesary array allocations.
Could also use a Span<ViewPort> as an alternative but this is true for most of the methods in Stride so I though I will avoid this here to not mess up the code style.

sources/engine/Stride.Rendering/Rendering/Images/ImageEffect.cs
dotlogix: Nope these comments are copied from CommandList. Basically most of this logic was copied from there :)

sources/engine/Stride.Rendering/Rendering/Images/ImageEffect.cs
dotlogix: Array.Copy does the bound check internally and tells you that the arrays are not the same size.
If you take a look at CommandList we actually only have a for loop and get an IndexOutOfRange exception when you pass more than 16 elements.

sources/engine/Stride.Rendering/Rendering/Images/ImageEffect.cs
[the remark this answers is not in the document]
manio143: At least update the comment to describe that this is what this method does - takes first count elements from the array.

sources/engine/Stride.Rendering/Rendering/Images/ImageEffect.cs
manio143: Ok, let's leave it for the moment

sources/engine/Stride.Rendering/Rendering/Images/ImageEffect.cs
manio143: I know Array.Copy has bound checks, but they will be confusing to the user of this API unless they look into the source code when an exception is thrown. You need to check at least
* Count is not greater than array length
* Count is not greater than destination length

dotlogix:
So I adjusted the API to use spans instead of arrays where possible to allow zero allocation viewport updates without the confusing count parameter. Also adjusted the API in command list slightly to support the same.

I also changed how comparisons are done in command list to update [...]Dirty.
I would actually like to also change inputs and outputs to support the same but I guess this would be too much for one proposal.

Also made viewport and scissor in command list nullable to allow to reset the viewport easily (same as image effect)
Not sure about the stackalloc in the single item overload. We can also use the same we had before but I thought it might be beneficial to only maintain one main method and just pass all calls to this one instead.

stackalloc Type[]{ value } can be similified with collection spans in the future it would then just be [value] imho it is easier to understand but pls let me know what you think.

If this is sth we want to do I would also like tot create a proposal to allow Spans in more part's of the API

manio143:
Looks good. Small nitpick - please format the code to give a space after `if`s.

dotlogix:
Thx for the second review round will update the code style as soon as I can.

There are still a few questions open however.
1. Do we want to mark the array methods as obsolete? There were some comments in discord that we shouldn't force ppl to use spans however I would prefer to keep it this way.
2. Do we accept the small extra cost of using Spans for single elements. It is definetly slower by a few ns.
3. This method signature does not exist anymore and the logic of the array overload does not match this one exactly.
`public void SetRenderTargetAndViewport(Texture depthStencilView, Texture renderTargetView, Texture secondRenderTarget)`
Imho I think the previous version actually contained a bug so I think it wasn't used anyway but it is a braking change.
4. What about the array signaturs should I add them again? This version is compile-compatible but not binary compatible.
Also without these signature everything which used reflection to call these methods will not work anymore.
Better to keep all the existing methods and just redirect to the new version and mark them as obsolete?

manio143:
> Do we want to mark the array methods as obsolete? There were some comments in discord that we shouldn't force ppl to use spans however I would prefer to keep it this way.

I think that's perfectly fine to mark them obsolete. Isn't there implicit casting from array to Span? In that case for most users the transition will be seamless.

> Do we accept the small extra cost of using Spans for single elements. It is definetly slower by a few ns.

I think it would be good to understand how often those methods are called per frame. If the overhead is actually impactful it would make sense to do special casing for those instances. Otherwise I would keep it simple.

> This method signature does not exist anymore and the logic of the array overload does not match this one exactly.
> 
> ---
> What about the array signaturs should I add them again?

Unless I'm missing something I see you still implementing those methods - is that comment not applicable any more? - I see it's slightly different in the matter of null checks, but I think we can assume it's fine and people won't be passing nulls willy nilly here. And frankly I got a little lost in reading this diff, so please explain what is the change in behavior.

Overall it all looks great - good job :)

dotlogix:
**TLDR:**
As you mentioned the null behavior is different and also some properties now return ReadOnlySpan<T> instead of T[] to prevent untracked mutations.
So there are actually a bunch of potentially breaking but imho correct changes.

**Method modifications:**
I added the array overloads again because of the comments in our discord channel about my related proposal.
```
public void SetRenderTargetAndViewport(Texture depthStencilView, Texture renderTargetView, Texture secondRenderTarget)
```
This signature had a bug in the current code if I understand it correctly. When passing null for renderTargetView and non-null for secondRenderTarget it would set count = 1 and uses the first element instead of the second one even though only the second index is actually set. I think this is definetly not intended.

```
public void SetRenderTargetAndViewport(Texture depthStencilView, Texture renderTargetView)
```
This signature is not handling null values anymore which we might want to add or throw if null is passed (I would prefer this version). Maybe throwing for any null value is the better way as it prevent's strange edge cases to happen but this is definetly a breaking change.

**Properties:**
Viewport & Scissor
Now return Empty if the count is 0 as we don't reset the values when changing the viewports to a count=0 would also be a valid option. The previous implementation always returns the first element regardless of the count which could be filled with non-valid data.

RenderTargets, Viewports and Scissors
Now return a sliced ReadOnlySpan<T> instead of T[] with all 16 elements.
This was done because atm the Array access allows directly modifing the values in the array whitout notifing the underlying implementation.

**Performance:**
According to a few small benchmarks converting a Span with one element is definetly slower (factor of 6x actually) but we are talking about ~12ns per call. Imho the removal of special cases is still worth the change but I can also do it the most performant way if we prefer this.

I am just not sure if we can just force such changes or if this needs to happen in a major breaking-release.
Just a bit scared after the comments in our discord about breaking changes.

manio143:
On the null checks - I think it would be good to throw.

On the breaking changes - as with everything "it depends". There are APIs which we can modify and they are unlikely impacting many users and others that may.
I had a quick look on NuGet for Stride related packages and saw for example Myra using the CommandList APIs. That package is compiled using >= version constraint, which means users can reference the same version while updating Stride. Binary braking changes in that case would force the author of Myra to release a new version.

Maybe have a look on GitHub for any usages of the API you're braking? Myra might be safe as I saw single item Scissors, but I only had a very quick look.

tebjan:
The standard libs of VL might be the repo with the most heavy use of the Stride graphics API. Specifically `VL.Stride.Runtime` and `VL.Stride.Windows`: https://github.com/vvvv/VL.StandardLibs

dotlogix:
@tebjan , @manio143 I decided to split the span support and the scissor support as this PR does too much imho and the span support needs more and better planning/preparation.
I removed all unnceccesary work and moved it to a different banch on my machine.

tebjan:
looks good, thanks!

dotlogix:
@tebjan T[] and Span<T> are implicitly convertible so using an array works just fine.
Don't think we should add these APIs especially with the Plan to move to a more Span<T> focused API.

dotlogix:
So when can we merge this? Would like to finish my community toolkit PR as well 😄

tebjan:
I'll try to have a loot at it this weekend. do you have a sample image effect code (snippet) to test it with?

manio143 (approved):

dotlogix:
@tebjan you can take a look here:
https://github.com/stride3d/stride-community-toolkit/pull/57

This whole class uses this feature extensively and also contains a bunch of examples

tebjan:
Very good and minimal code, thanks!

PR #1977, opened 2023-10-22
Let the user pick which animation stack to import in an fbx

adrsch:
# PR Details

If multiple animation stacks are present in an fbx, let the user choose the index to import.

## Description

This only allows it to be set after the asset is initially created and imported. Also, the index is clamped so it doesn't break if you set something out of bounds.

Tested with a mixamo animation, since they set everything in the 2nd stack and a t-pose in the first.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Assets.Models/AnimationAsset.cs
+++ b/sources/engine/Stride.Assets.Models/AnimationAsset.cs
@@ -60,10 +60,10 @@ namespace Stride.Assets.Models
         public override UFile MainSource => Source;
 
         /// <summary>
-        /// Index for FbxAnimStack.
+        /// Index of imported animation in the source file.
         /// </summary>
         /// <userdoc>
-        /// Index of animation to import if multiple are present in the FBX file
+        /// Index of animation to import if multiple are present in the source file.
         /// </userdoc>
         [DataMember(-10)]
         [Display("Animation Index")]
--- a/sources/engine/Stride.Assets.Models/AnimationAsset.cs
+++ b/sources/engine/Stride.Assets.Models/AnimationAsset.cs
@@ -60,6 +60,16 @@ namespace Stride.Assets.Models
         public override UFile MainSource => Source;
 
         /// <summary>
+        /// Index for FbxAnimStack.
+        /// </summary>
+        /// <userdoc>
+        /// Index of animation to import if multiple are present in the FBX file
+        /// </userdoc>
+        [DataMember(-10)]
+        [Display("Animation Index")]
+        public int AnimationStack { get; set; }
+
+        /// <summary>
         /// Enable clipping of the animation duration
         /// </summary>
         /// <userdoc>
--- a/sources/tools/Stride.Importer.Assimp/MeshConverter.cs
+++ b/sources/tools/Stride.Importer.Assimp/MeshConverter.cs
@@ -142,14 +142,14 @@ namespace Stride.Importer.Assimp
             return ConvertAssimpScene(scene);
         }
 
-        public unsafe AnimationInfo ConvertAnimation(string inputFilename, string outputFilename)
+        public unsafe AnimationInfo ConvertAnimation(string inputFilename, string outputFilename, int animationIndex)
         {
             uint importFlags = 0;
             var postProcessFlags = PostProcessSteps.None;
 
             var scene = Initialize(inputFilename, outputFilename, importFlags, (uint)postProcessFlags);
 
-            return ProcessAnimations(scene);
+            return ProcessAnimations(scene, animationIndex);
         }
 
         public unsafe Rendering.Skeleton ConvertSkeleton(string inputFilename, string outputFilename)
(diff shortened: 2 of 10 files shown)


manio143:
Code changes look fine. Since this is modifying the public API of the animation asset, does this scale to any other format than FBX? If so, we probably need to file an issue to get this implemented for other model importers. If not, we need to add to the property description that this is a special case for FBX and the value is ignored for other file types. Maybe even it would make sense to call it `FbxAnimationStack` if it's a feature specific to that file format. However, I imagine it's the case that other file formats would also support multiple animations. But would other formats also use an integer index or maybe a string identifier?

adrsch:
> Code changes look fine. Since this is modifying the public API of the animation asset, does this scale to any other format than FBX? If so, we probably need to file an issue to get this implemented for other model importers. If not, we need to add to the property description that this is a special case for FBX and the value is ignored for other file types. Maybe even it would make sense to call it `FbxAnimationStack` if it's a feature specific to that file format. However, I imagine it's the case that other file formats would also support multiple animations. But would other formats also use an integer index or maybe a string identifier?

I totally forgot about other formats! I'll get that added.

Ideally I feel like the importer would have a nice little dropdown with the names of all the animations, or at least limited to valid indices. Not sure how I'd implement that with the UI framework, the only dropdowns I've seen have been bound to the names of a hardcoded enum and that would probably be more annoying than an integer field.

sources/tools/Stride.Importer.Assimp/MeshConverter.cs
```diff
@@ -250,24 +250,32 @@ private unsafe Rendering.Skeleton ProcessSkeleton(Scene* scene)
...
             var meshIndexToNodeIndex = new Dictionary<int, List<int>>();
             RegisterNodes(scene->MRootNode, -1, nodeNames, meshIndexToNodeIndex);
 
-            for (uint i = 0; i < Math.Min(1, scene->MNumAnimations); ++i)
+            if (scene->MNumAnimations > 0)
             {
-                var aiAnim = scene->MAnimations[i];
+                if (animationIndex < 0)
+                {
+                    animationIndex = 0;
```
manio143, line 269: This also needs a warning to notify the user we're using a different value

sources/tools/Stride.Importer.Assimp/MeshConverter.cs
```diff
@@ -250,24 +250,32 @@ private unsafe Rendering.Skeleton ProcessSkeleton(Scene* scene)
...
-                Logger.Warning($"There are {scene->MNumAnimations} animations in this file, using only the first one.");
-
             var nodeNames = new Dictionary<IntPtr, string>();
             GenerateNodeNames(scene, nodeNames);
 
             // register the nodes and fill hierarchy
             var meshIndexToNodeIndex = new Dictionary<int, List<int>>();
             RegisterNodes(scene->MRootNode, -1, nodeNames, meshIndexToNodeIndex);
 
-            for (uint i = 0; i < Math.Min(1, scene->MNumAnimations); ++i)
```
manio143, line 268: Was this supporting multiple animations before? Are we not reducing functionality then?

tebjan:
Isn't it possible to just import all and create the appropriate number of animation assets according to the names of the animations?

sources/tools/Stride.Importer.Assimp/MeshConverter.cs
adrsch: That's the min of 1 and the animation count, so it was a weird way of saying skip that code if there's zero animations. Maybe at one point it was kind of supported and then the support was removed? Not sure why else they'd write it that weirdly.

adrsch:
Yeah, I think that would be possible. Maybe with a way to skip certain animations if you want, since my use case is just to use the 2nd animation and ignore the first, lol. Not sure if the names have uniqueness enforced. If they don't the index may need to be used to keep it consistent across imports.

sources/engine/Stride.Assets.Models/AnimationAsset.cs
```diff
@@ -59,6 +59,16 @@ public AnimationAsset()
         [DataMemberIgnore]
         public override UFile MainSource => Source;
 
+        /// <summary>
+        /// Index for FbxAnimStack.
+        /// </summary>
+        /// <userdoc>
+        /// Index of animation to import if multiple are present in the FBX file
```
manio143, line 66: Can you adjust those comments, now that they're not FBX specific?

manio143 (approved):
Looks good. One more comment about XML comment.
Hopefully this is a simple first step towards importing multiple animation assets.

sources/engine/Stride.Assets.Models/AnimationAsset.cs
adrsch: Done! Yeah, I think something like the way material import from fbx is handled is very feasible.

manio143: @adrsch Did you not push after saying "Done"?

adrsch: Oh... yeah I did forget to push. Oops...

PR #1990, opened 2023-10-25
[Shaders] Fixes `EffectValueDescription.DefaultValue` for negative values

azeno:
## Description

Negative values are represented as unary expressions. This commit adds support for unary expressions when reading the default value of a parameter. In case an initializer expression is not supported / can't be evaluated (for example `float x = 3 + 4;`) the `DefaultValue` property is set to `null`, leaving it up to the consumer to deal with it.

## Related Issue

Fixes issue #1963
Replacement for #1967 

## Motivation and Context

As outlined in #1967 the handling of negative values didn't work at all. Instead of changing the shader grammar this PR fixes the offending code itself.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Shaders.Parser/ShaderLinker.cs
+++ b/sources/engine/Stride.Shaders.Parser/ShaderLinker.cs
@@ -837,10 +837,17 @@ namespace Stride.Shaders.Parser
             }
 
             static T[] ToArray<T>(List<Expression> v)
+                where T : unmanaged
             {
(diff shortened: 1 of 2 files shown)


tebjan:
Thanks for the info and the quick fix! @Basewq if your issue and PR are solved by this, let us know.

Basewq:
Yep, tested on the SpaceEscape issue and appears fixed.

tebjan:
Thanks a lot!

PR #1992, opened 2023-10-25
Remove ILRepack dependency

Jklawreszuk:
# PR Details

## Description

This PR removes ILRepack tool from Stride.Core.AssemblyProcessor.csproj 
ILRepack is currently working correctly only on Windows platforms allowing merging multiple assemblies into single one.
I'm not sure but it looks like this tool is dedicated mainly for NET Framework platform.

## Motivation and Context

As an developer I wanted to build Stride engine on non-Windows platforms (at least partly).
After some working on that matter, I found "GenerateHash" target in Stride.Core.AssemblyProcessor.csproj which bundles 
dlls into single one using ILRepack. Project Wiki recommends using `mono ilrepack` to run on Linux for example, but it's throwing exception instead.

My hypothesis is that it doesn't matter if it's packaged or not, and I've checked this by running the editor and haven't noticed any differences so far. If anyone has sometime to that out I'll be very happy.

Here is a difference of dll's using ILSpy (mine vs current)
image

## Types of changes

## Checklist

    - [x] **I have built and run FirstPersonShooter sample**


xen2 (commented):
I remember we had several issues in the past:
- We had our own custom changes to Cecil (which are now merged in official Cecil library)
- Xamarin had its own custom Cecil which was conflicting with our Cecil. I think it might have been loaded by MSBuild but not sure, it's been a while. I think it's not a problem anymore (let's double-check teamcity build for iOS/Android if there is no error).

I think it's probably fine now to keep them as separate files.

Eideren:
Thanks !

Kryptos-FR:
Any reason why the new binary files have been added as binary and not through git LFS? This has increased the size of the repository considerably (and they will stay forever in the history, unless we do a rewrite to exclude them).

PR #1995, opened 2023-10-28
Fix failing to load data/db/index file on non-Win desktop platforms

Jklawreszuk:
# PR Details

## Description

On Linux, on creating simple empty window app by simply calling `Game.Run()`, Stride will throw `FileNotFoundException: Could not find file myproj/net6.0/data/db/index`. This bug should also appear on macOS in the same way.

I have now extended setting `isReadOnly` variable to `false` for all desktop platforms, so that the index file doesn't have to be created manually. Tested on Fedora 38, Windows 10.

## Types of changes

## Checklist

- [x] **I have built and run app on Linux and Windows**

### DIFF
--- a/sources/core/Stride.Core.Serialization/Storage/ObjectDatabase.cs
+++ b/sources/core/Stride.Core.Serialization/Storage/ObjectDatabase.cs
@@ -37,8 +37,8 @@ namespace Stride.Core.Storage
             ContentIndexMap = new ObjectDatabaseContentIndexMap();
 
             // Try to open file backends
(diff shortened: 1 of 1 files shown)


VaclavElias:
I had the same issue with Linux code only. I had to copy that file manually from Windows 🤣

Jklawreszuk:
I always created the file manually and it worked too. This isn't invasive change and will make running the game less annoying 😃

tebjan:
Nice one, thanks!

PR #1911, opened 2023-10-08
Fixes OpenXR

MaximilianEmel:
# PR Details

- Fixes OpenXR crash
- Fixes OpenXR mirroring
- Removes Remnants of the deprecated Fove & GoogleVR APIs

## Description

- Upgrades Silk.NET.OpenXR from 2.15.0 to 2.17.1

## Related Issue

https://github.com/stride3d/stride/issues/1909

## Motivation and Context

It fixes OpenXR, and cleans things up.

## Types of changes

- [ ] Docs change / refactoring / dependency upgrade
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [ ] My change requires a change to the documentation.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #2023, opened 2023-11-02
Fixes OpenGL

MaximilianEmel:
# PR Details

Updates Stride.Graphics' OpenGL to match the new version of the API that's used in Silk.NET-2.17.1

## Description

- SearchPathContainer is now an array
- Replaced specific enums with the generic GLEnum

## Related Issue

#1911

## Motivation and Context

When updating OpenXR to the latest version, Silk's OpenGL bindings also got updated, which updated the version of OpenGL, which had breaking changes.

## Types of changes

## Checklist


MaximilianEmel:
The editor and a project also built with OpenGLES, but the game didn't launch. It is likely [OpenGLES] is generally not setup to work with Windows/Desktop in Stride, or my graphics driver doesn't support OpenGLES.

MaximilianEmel:
I believe there's still work to be done (some visible differences in the rendering between D3D11 and OpenGL), but I haven't checked if that's been introduced from the Silk upgrade, or was always like that. There's also quite a lot of Deprecated things, that while still continue to function, ideally should be updated. Might it be a good idea to keep Silk's OpenXR and OpenGL versions unsynchronized (until this full upgrade [now maybe to Silk.NET-2.18.0] is fully formed)? There were no issues from [being unsynchronized], as far as I could tell (more investigation needed). Especially considering there shouldn't be much or any overlap between OpenXR and OpenGL (more investigation needed).

MaximilianEmel:
Results from testing against 83fdc0455ae2d310162ebae19c2dabc1db000f62 (last working version of OpenGL on Windows):
- The rendering differences are the same.
- The same error still occurs apon launching the game with OpenGLES on Windows.

Conclusions:
- No rendering issues are originating as a result of the Silk.NET upgrade.
- OpenGLES still needs to be tested for its target platforms, though I believe it should work.

sources/engine/Stride.Graphics/OpenGL/GLCoreLibraryNameContainer.cs
```diff
@@ -12,22 +12,22 @@ namespace Silk.NET.OpenGL
     internal class GLCoreLibraryNameContainer : SearchPathContainer
     {
         /// <inheritdoc />
-        public override string Linux => "libGL.so.1";
+        public override string[] Linux => new[] { "libGL.so.1" };
```
manio143, line 15: Can you make those properties `{ get; } =` instead of `=>` to avoid per call allocations of the array?

MaximilianEmel: I'll look into this as soon as I can.

MaximilianEmel: I would like to mention I modeled how I did it based on how Silk.NET internally does the same thing. e.g. https://github.com/dotnet/Silk.NET/blob/da7cae57614f91d0387fa51af8a52c99b7c43d98/src/Windowing/Silk.NET.SDL/SDLLibraryNameContainer.cs

MaximilianEmel: In fact, does it even make sense in the first place to have this file here, instead of just the one already in Silk.NET?

MaximilianEmel:
According to me, It is ready to merge.

manio143:
I see that just as there was `GLCoreLibraryNameContainer ` there's also `OpenGLESLibraryNameContainer` in Stride code base. Those were introduced by @xen2 in f982b42cc1f

MaximilianEmel:
I have been modifying both.

tebjan:
Great, thanks a lot!

PR #2044, opened 2023-11-16
[Math] Relax constraints on arguments passed as ref across math types

Eideren:
# PR Details
See title, this enables arguments passed as `in` to be passed into those functions without copying.
Has a side effect of allowing users to pass their variables without specifying `in` or `ref` modifier, dotnet will produce a warning instead of syntax error.
Example:
```cs
void F0(ref readonly int i){}

F0(124); // warning: Argument 1 should be a variable because it is passed to a 'ref readonly' parameter
int i = 124;
F0(in i); // OK
F0(ref i); // OK

void F1(in int i) => F0(in i); // OK
void F2(ref int i) => F0(in i); // OK
void F3(ref readonly int i) => F2(ref i); // Error, F2 is not readonly
void F3(in int i) => F2(ref i); // Error, F2 is not readonly
```

See `ref readonly` feature specification https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-12.0/ref-readonly-parameters

## Related Issue
None

## Motivation and Context
Prevents a forced copy for passing structs that are received as `in`

## Types of changes

## Checklist

### DIFF
--- a/sources/core/Stride.Core.Mathematics/BoundingBox.cs
+++ b/sources/core/Stride.Core.Mathematics/BoundingBox.cs
@@ -104,10 +104,10 @@ namespace Stride.Core.Mathematics
         /// </summary>
         /// <param name="ray">The ray to test.</param>
         /// <returns>Whether the two objects intersected.</returns>
-        public bool Intersects(ref Ray ray)
+        public bool Intersects(ref readonly Ray ray)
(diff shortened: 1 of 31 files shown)


manio143 (approved):
Looks good

PR #2045, opened 2023-11-17
Use central package management

Kryptos-FR:
# PR Details

https://learn.microsoft.com/en-us/nuget/consume-packages/central-package-management

## Description

Adds `Directory.Packages.props`  and references all package versions there.

## Related Issue

Discussed on Discord.

## Motivation and Context

Prevent out-of-sync dependencies between projects.

## Types of changes

- [ ] Docs change / refactoring / dependency upgrade
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [ ] My change requires a change to the documentation.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.
- [x] **I have built and run the editor to try this change out.**

PR #2052, opened 2023-11-20
[Build] Use central package management  - missed a couple of projects

Kryptos-FR:
# PR Details

Continuation of #2045, where a few projects were missed.


tebjan:
Thanks for the aftercare!

PR #2055, opened 2023-11-20
[Build] Move Directory.Packages.props

Kryptos-FR:
# PR Details

Follows #2045 and #2052.

## Description

Move the file under `sources/` to avoid impact with `deps/` and `samples/`.


tebjan:
That makes sense, thanks!

Eideren:
There is a new warning that showed up with your previous related PR, `There are 2 package sources defined in your configuration. When using central package management, please map your package sources with package source mapping (https://aka.ms/nuget-package-source-mapping) or specify a single package source. The following sources are defined: nuget.org, Stride` - would be nice to address those in this PR if you have time

VaclavElias:
You signature wasn't trusted 😂. Make sure you don't sign with your left hand, if you write with your right hand 🤣.

xen2 (approved):

PR #2056, opened 2023-11-20
WIP POC: Object Descriptor in Roslyn

IXLLEGACYIXL:
# PR Details

Instead of scanning with reflection in runtime multiple times, get the info through roslyn.
Should be used to unify Yaml, Object Descriptor and Assembly Processor towards Roslyn

## Types of changes

## Checklist


archanox:
I came here from the now deleted CoreAoT issue. I'm curious if there is a short list of where we are using reflection where we can supercede it with Source Generators?

IXLLEGACYIXL:
> I came here from the now deleted CoreAoT issue. I'm curious if there is a short list of where we are using reflection where we can supercede it with Source Generators?

Reflection on its own isnt that big of a problem as it works with AOT too, it just works differently

the big problems that i know
https://github.com/stride3d/stride/pull/1609 => assemblyprocessor has to die and this is the roslyn approach
to kill the assemblyprocessor there needs to be a unified way to detect the fields which have to be serialized
as there are 3 serializers using almost the same rules for serialization and to start with assemblyprocessor is almost impossible

so i started with the first serializer, the objectdescriptor here ( editor serializer ) i wanted to remove the detection of what should be serialized and not the entire reflection

when this is fixed we can continue with yaml https://github.com/NexStandard/StrideSourceGenerator/tree/IncrementalApproach
and assemblyprocessor

then we have the major problems source generated

you could start with a "peace meal" conversion of the core components like Stride.Core and Stride.Core.Design to convert them to AOT , i doubt that we will ever reach that with trimming

probably stride.core.mathematics is a good place to start to convert to AOT which wouldnt require roslyn

i dont know much about other areas like shaders and so forth, it would be good to have atleast some of the components AOT compiled, there should be something to allow a mix of aot and non aot but idk

xen2:

PR #2063, opened 2023-11-22
[Build] Fix misc warnings

Eideren:
# PR Details

## Related Issue
None

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Assets.Tests2/Stride.Assets.Tests2.csproj
+++ b/sources/engine/Stride.Assets.Tests2/Stride.Assets.Tests2.csproj
@@ -12,7 +12,6 @@
   <ItemGroup>
     <PackageReference Include="xunit" />
     <PackageReference Include="xunit.runner.visualstudio" />
-    <PackageReference Include="TeamCity.VSTest.TestAdapter" />
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\..\tests\xunit.runner.stride\xunit.runner.stride.csproj" />
--- a/sources/engine/Stride.Graphics/SDL/Window.cs
+++ b/sources/engine/Stride.Graphics/SDL/Window.cs
@@ -39,16 +39,16 @@ namespace Stride.Graphics.SDL
         /// <param name="title">Title of the window, see Text property.</param>
         public unsafe Window(string title)
         {
-            WindowFlags flags = WindowFlags.WindowAllowHighdpi;
+            WindowFlags flags = WindowFlags.AllowHighdpi;
 #if STRIDE_GRAPHICS_API_OPENGL
-            flags |= WindowFlags.WindowOpengl;
+            flags |= WindowFlags.Opengl;
 #elif STRIDE_GRAPHICS_API_VULKAN
-            flags |= WindowFlags.WindowVulkan;
+            flags |= WindowFlags.Vulkan;
 #endif
 #if STRIDE_PLATFORM_ANDROID || STRIDE_PLATFORM_IOS
-            flags |= WindowFlags.WindowBorderless | WindowFlags.WindowFullscreen | WindowFlags.WindowShown;
+            flags |= WindowFlags.Borderless | WindowFlags.Fullscreen | WindowFlags.Shown;
 #else
-            flags |= WindowFlags.WindowHidden | WindowFlags.WindowResizable;
+            flags |= WindowFlags.Hidden | WindowFlags.Resizable;
 #endif
             // Create the SDL window and then extract the native handle.
             sdlHandle = SDL.CreateWindow(title, Sdl.WindowposUndefined, Sdl.WindowposUndefined, 640, 480, (uint)flags);
@@ -191,12 +191,12 @@ namespace Stride.Graphics.SDL
 
         private WindowFlags GetFullscreenFlag()
         {
-            return FullscreenIsBorderlessWindow ? WindowFlags.WindowFullscreenDesktop : WindowFlags.WindowFullscreen;
+            return FullscreenIsBorderlessWindow ? WindowFlags.FullscreenDesktop : WindowFlags.Fullscreen;
         }
 
         private static bool CheckFullscreenFlag(uint flags)
         {
-            return ((flags & (uint)WindowFlags.WindowFullscreen) != 0) || ((flags & (uint)WindowFlags.WindowFullscreenDesktop) != 0);
+            return ((flags & (uint)WindowFlags.Fullscreen) != 0) || ((flags & (uint)WindowFlags.FullscreenDesktop) != 0);
         }
 
         /// <summary>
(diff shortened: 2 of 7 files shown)


sources/engine/Stride.Assets.Tests2/Stride.Assets.Tests2.csproj
```diff
@@ -12,7 +12,6 @@
   <ItemGroup>
     <PackageReference Include="xunit" />
     <PackageReference Include="xunit.runner.visualstudio" />
-    <PackageReference Include="TeamCity.VSTest.TestAdapter" />
```
Eideren, line 15: Duplicate, already referenced in props

sources/engine/Stride.UI.Tests/Layering/EditTextTests.cs
```diff
@@ -17,7 +17,7 @@ namespace Stride.UI.Tests.Layering
     [System.ComponentModel.Description("Tests for EditText layering")]
     public class EditTextTests
     {
-        class DummyFont : SpriteFont { }
+        public class DummyFont : SpriteFont { }
```
Eideren, line 20: `AssemblyScanProcessor: Can't register type [Stride.UI.Tests.Layering.EditTextTests/DummyFont] for scan type [Stride.Core.Serialization.Contents.ReferenceSerializerAttribute] because it is a nested private type`

sources/tools/Stride.ProjectGenerator/Stride.ProjectGenerator.csproj
```diff
@@ -8,12 +8,6 @@
     <StrideBuildTags>WindowsTools</StrideBuildTags>
     <StrideCompilerTargetsEnable Condition="'$(StridePackageBuild)' == 'true'">false</StrideCompilerTargetsEnable>
   </PropertyGroup>
-  <PropertyGroup Condition="'$(Configuration)' == 'Debug'">
-    <Prefer32Bit>true</Prefer32Bit>
-  </PropertyGroup>
-  <PropertyGroup Condition="'$(Configuration)' == 'Release'">
-    <Prefer32Bit>true</Prefer32Bit>
-  </PropertyGroup>
```
Eideren, line 16: `C:\Program Files\dotnet\sdk\8.0.100\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.RuntimeIdentifierInference.targets(275,5): warning NETSDK1189: Prefer32Bit is not supported and has no effect for netcoreapp target.`

sources/engine/Stride.VirtualReality/OpenVR/OpenVR.cs
```diff
@@ -16,18 +16,6 @@ namespace Stride.VirtualReality
...
-        /// <typeparam name="T">The type of the reference.</typeparam>
-        /// <param name="value">The reference whose initialization should be skipped.</param>
-        /// <returns>The reference to <paramref name="value"/>.</returns>
-        /// <remarks>Take care to ensure that the struct has been initialized appropriately, otherwise the struct's fields could contain uninitialized data from the stack.</remarks>
-        private static unsafe ref T SkipInit<T>(out T value)
-        {
-            Unsafe.SkipInit(out value);
-            return ref value;
-        }
-
```
Eideren, line 30: `warning CS9088: This returns a parameter by reference 'value' but it is scoped to the current method`

sources/engine/Stride.Graphics/SDL/Window.cs
```diff
@@ -39,16 +39,16 @@ static Window()
         /// <param name="title">Title of the window, see Text property.</param>
         public unsafe Window(string title)
         {
-            WindowFlags flags = WindowFlags.WindowAllowHighdpi;
+            WindowFlags flags = WindowFlags.AllowHighdpi;
```
Eideren, line 42: Obsolete, Silk renaming stuff afaict.

sources/engine/Stride.VirtualReality/OpenVR/OpenVR.cs
Kryptos-FR: Would be good to know the reason why this was used. @xen2 any idea?

Eideren: Right, I didn't specify, but this is fairly obvious when you look at the callsites.
The issue this method was solving was unsafe as cast with an uninitialized out value, i.e. this, is illegal in c#:
```cs
void MyF(out int value) => Unsafe.As<float, int>(ref value) = 0.1f;
```
The out value has to be assigned before usage as ref parameters.
So, this method is used to make the compiler think it was initialized to avoid the added cost of actually initializing it when the value would be initialized through other means (in this case, through the ref returned by `Unsafe.As`).

Now, the warning this was generating is related to the scopes of ref return values, the ref may leak depending on how it is used by the callsite, pointing at some undefined address location. In this case it couldn't leak since everything happened in the same scope, but it is trivial to solve the issue without using this fairly unsafe method - we can just convert the value you would assign to the out parameter instead of converting the out parameter itself, like I have done below in the same file.

Kryptos-FR: Interestingly, the JIT is different (albeit equivalent) for both versions. And surprisingly, the version that used SkipInit did actually do a memset (because it is vectorized for some reason). [See on SharpLab](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgBEBLbAcwDsJcGDmFwBuGvSasASgFdeQ/DBYBhCPgAOHADYwoAZT0A3YTDE0JjfQAtsUDQBlswFnIUcl46hIDMTUgwqDADeNAzhDGERglCyYBgMAGrY2lHhodQRWUx+AGbaENgJuV5ZAL5pkZkRGlAcRkUwSSkMRgwAvAy8MADuzdohDLkdDHQsAKzDZaURlcR+xCgMAPLaACYAFBCyCRwKrSmyMACUlRnZEQCqvLjYucoAgrgAPHsYaEMFRQB8G7DD+gA1hwNABJXgcDBbHYHbRHY7HEZGFglSoVarhOYLJarNakaG7fYNOEnM6VLLXW73FhAkHgyEE2HwmYXSl3R4vN4ffKFDC/f5Mk5IlEshjorJYphLAByvUZb0FpwxIXJEWJRxGbOpT2ePKKHze/Jgw2RuWOovFs2VAG0AFKQgDiMG6dTAGwwAE8NDAILkNm8ER97RgnS7hO6vT6/XqMAiALqVWr1RpSJAMeRUpoCgAqDFpYIhGGe2d+2wSufVpOV5wuDC1ynz9KhZcVoqyxAA7AwBZWLTQykA==).

Issue #2067, opened 2023-11-26
Multiple instantiation of prefabs causes ambiguous IDs


laske185:
**Release Type**: Official Release

**Version**: 4.2.0.2042

**Platform(s)**: Windows

**Describe the bug**
When a scene contains two or more entity components with an entity processor that instantiates the same prefab and attaches the instantiated entities to the entity of the component, then the editors crashes when changing the hierarchy.

This is caused because the instantiated entities and components got the same IDs as the corresponding elements in the prefab. When this instantiation happens multiple times, the hierarchy has different elements with the same ID that bring forth errors.

**To Reproduce**
1. Create an entity component
2. Create an entity processor for that component that instantiates the prefab and attaches the entities to the entity of the component
3. Assign the component to at least two entities in the scene in the editor
4. Assign a prefab to the components
5. Change the hierarchy, f.e. by adding a new empty entity.

Example implementation:
Component:
```cs
[DefaultEntityComponentProcessor(typeof(MyProcessor))]
[DataContract]
public class MyComponent : EntityComponent
{
    public Prefab Prefab { get;  set; }
}
```

Processor:
```cs
internal class MyProcessor : EntityProcessor<MyComponent, MyProcessor.MyData>
{
    private class MyData
    {
        public required Entity Entity { get; init; }
        public List<Entity> Entities { get; } = [];
    }

    public override void Update(GameTime time)
    {
        base.Update(time);

        foreach (var (component, data) in ComponentDatas)
        {
            if (data.Entities.Count == 0 && component.Prefab != null)
            {
                var entities = component.Prefab.Instantiate();
                foreach (var entity in entities)
                {
                    data.Entity.AddChild(entity);
                }
                data.Entities.AddRange(entities);
            }
        }
    }

    protected override MyData GenerateComponentData([NotNull] Entity entity, [NotNull] MyComponent component) => new()
    {
        Entity = entity
    };
}
```

**Expected behavior**
All entities and components instantiated by a prefab should get a unique ID.

**Log and callstacks**

The `EntityHierarchyEditorController` throws an `ArgumentException` here:
```cs
protected override Dictionary<Guid, IIdentifiable> CollectIdentifiableObjects()
{
    var allEntities = Game.ContentScene.Yield().BreadthFirst(x => x.Children).SelectMany(x => x.Entities).BreadthFirst(x => x.Transform.Children.Select(y => y.Entity));
[...]

PR #2068, opened 2023-11-26
create unique IDs for from prefab instantiated entities and components

laske185:
# PR Details

After instantiation of a prefab, all entities and components get a new ID.

Closes: #2067 

## Description

Fixes the crash of the Game Studio described in #2067 caused by ambiguous IDs. This could also fix some other consequential errors happening through the ambiguous IDs.

## Related Issue

#2067

## Motivation and Context

The Game Studio crashed each time I changed the hierarchy in my scene.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Engine/Engine/Prefab.cs
+++ b/sources/engine/Stride.Engine/Engine/Prefab.cs
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net) and Silicon Studio Corp. (https://www.siliconstudio.co.jp)
 // Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
 
+using System;
 using System.Collections.Generic;
 using Stride.Core;
 using Stride.Core.Collections;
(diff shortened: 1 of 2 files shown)


sources/engine/Stride.Engine/Engine/Prefab.cs
```diff
@@ -30,7 +31,33 @@ public sealed class Prefab
         public List<Entity> Instantiate()
         {
             var newPrefab = EntityCloner.Clone(this);
+            CreateNewIds(newPrefab.Entities);
```
Ethereal77, line 34: Shouldn't this be something for `EntityCloner` to do? Seems more like a responsability for it, as its goal is to clone a hierarchy of `Entity`s and `EntityComponent`s so they can be used afterwards.
Thoughts?

Kryptos-FR: No, you don't want to duplicate everything. Otherwise, you are never going to be able to reuse stuff.

Kryptos-FR:
I'm surprised by this PR because I recall we should already have code that "fixup" IDs in case it is needed. Did you check for any regression elsewhere?

sources/engine/Stride.Engine/Engine/Prefab.cs
Ethereal77: I didn't mean to duplicate it, but to **move** it to `EntityCloner`, so it can clone an entity tree with or without unique ids.

laske185:
There are at least two ways in which prefabs are instantiated, depending on whether it happens in the Game Studio when the hierarchy is changed or at runtime. The Game Studio uses the `AssetCloner` for instantiation, as the prefabs are held here as a 
`PrefabAsset`.
When instantiating a prefab or cloning entities at runtime, however, the `EntityCloner` is used. The `EntityCloner` is only used within the engine for the public functions for instantiating prefabs and cloning entities. These themselves are not used within the engine.
Cloning is implemented through serialization. In contrast to the `EntityCloner`, the `AssetCloner` generates new IDs during deserialization. 
From my standpoint, it would also make sense if the `EntityCloner`, like the `AssetCloner`, generates a new ID when deserializing an entity or component. I would then change the implementation analogous to the `AssetCloner`.

Eideren:
Yep, sounds alright to me as well, tried it out on my end just to validate and instantiation does copy the `Id` from the source entity instead of replacing it with another one.

laske185:
I no longer need this change.

PR #1190, opened 2021-11-17
[Threading] Threadpool, lifo semaphore

Eideren:
# PR Details
Now that .Net6 uses their fully managed threadpool by default we can hook into their lifo semaphore which improves performances slightly.
I did benchmark their ``System.Threading.ThreadPool.UnsafeQueueUserWorkItem`` again under net6 and ours still performed better, so still no reason to migrate to dotnet's threadpool.

## Types of changes
- [ ] Docs change / refactoring / dependency upgrade
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist
- [ ] My change requires a change to the documentation.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #2083, opened 2023-12-16
[Dispatcher] improve api, reduce overhead, improve performances for items > 1k

Eideren:
# PR Details
Threadpool now accepts generic work items, allowing the dispatcher and other callers to compose work items out of structures, in turn enabling the JIT to inline the composed call tree into just one function call from the thread's scope up to the actual work the caller dispatched.

Arguments can now be passed by ref to the dispatcher, which is also passed by ref to threads resolving a lot of issues that would require painful workarounds and allocations, like allowing jobs to read and write to a shared structure, interlocked operations per dispatch, and outputting from a job.

Added an interface for processing items in batches, allowing for improved throughput for jobs that have thousands of items to process.

I only changed a couple of engine calls to be batched, others may benefit in more demanding games.

## Types of changes

## Checklist

### DIFF
--- a/sources/core/Stride.Core/Threading/Dispatcher.cs
+++ b/sources/core/Stride.Core/Threading/Dispatcher.cs
@@ -93,7 +93,8 @@ namespace Stride.Core.Threading
 
         private static void TypeAdapter<TJob>(object obj) where TJob : IBatchJob
         {
-            var batch = (BatchState<TJob>)obj;
+            var batch = obj as BatchState<TJob>; // 'as' and assert instead of direct cast to improve performance
+            Debug.Assert(batch is not null);
             try
             {
                 ProcessBatch(batch.Job, batch);
@@ -488,9 +489,9 @@ namespace Stride.Core.Threading
 
         /// <summary>
         /// An implementation of a job running in batches.
-        /// This object is shared across all threads scheduled for the job;
-        /// Making changes to it in one thread will be reflected in others.
-        /// Prefer implementing this over struct if possible to inline the call appropriately.
+        /// Implementing this as a struct improves performance as the JIT would have an easier time inlining the call.
+        /// Implementing this as a class would provide more utility as this object would be shared across all threads,
+        /// allowing for interlocked operations and other communication between threads.
         /// </summary>
         public interface IBatchJob
         {
@@ -502,7 +503,7 @@ namespace Stride.Core.Threading
             void Process(int start, int endExclusive);
         }
 
-        private class BatchState<TJob> where TJob : IBatchJob
+        private sealed class BatchState<TJob> where TJob : IBatchJob
         {
             private static readonly ConcurrentStack<BatchState<TJob>> Pool = new();
 
--- a/sources/core/Stride.Core/Threading/Dispatcher.cs
+++ b/sources/core/Stride.Core/Threading/Dispatcher.cs
@@ -29,11 +29,6 @@ namespace Stride.Core.Threading
 
         public delegate void ValueAction<T>(ref T obj);
 
-        static class Pool<T>
-        {
-            public static ConcurrentPool<T[]> Arrays = new(() => new T[1]);
-        }
-
         /// <summary>
         /// The call producing the least amount of overhead, other methods are built on top of this one.
         /// </summary>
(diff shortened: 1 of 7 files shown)


Kryptos-FR:
Any benchmark?

Eideren:
I'll send it over later this week

Eideren:
Here's a benchmark running on latest changes:

| Method  | ItemsToProcess | Mean       | Error     | StdDev     |
|-------- |--------------- |------------|-----------|------------|
| New     | 10             |   1.093 us | 0.0215 us |  0.0393 us |
| Old     | 10             |   5.357 us | 0.1058 us |  0.2038 us |
| NewLong | 10             |   1.383 us | 0.0277 us |  0.0477 us |
| OldLong | 10             |   5.525 us | 0.0521 us |  0.0435 us |
| New     | 100            |   1.453 us | 0.0207 us |  0.0162 us |
| Old     | 100            |   2.735 us | 0.0528 us |  0.0494 us |
| NewLong | 100            |   4.312 us | 0.0857 us |  0.0880 us |
| OldLong | 100            |   7.431 us | 0.1068 us |  0.0892 us |
| New     | 1000           |   2.084 us | 0.0414 us |  0.0960 us |
| Old     | 1000           |   7.699 us | 0.1528 us |  0.2510 us |
| NewLong | 1000           |   8.039 us | 0.0244 us |  0.0228 us |
| OldLong | 1000           |   9.355 us | 0.1729 us |  0.2123 us |
| New     | 10000          |   6.890 us | 0.1352 us |  0.2761 us |
| Old     | 10000          |   8.564 us | 0.0888 us |  0.0693 us |
| NewLong | 10000          |  23.778 us | 0.1935 us |  0.1715 us |
| OldLong | 10000          |  23.918 us | 0.1049 us |  0.0876 us |
| New     | 100000         |  29.414 us | 0.2818 us |  0.2636 us |
| Old     | 100000         |  33.156 us | 0.6590 us |  1.3162 us |
| NewLong | 100000         | 193.900 us | 1.9274 us |  1.5048 us |
| OldLong | 100000         | 232.131 us | 7.2847 us | 21.4791 us |

Difference between `New` and `NewLong` is that more operations are run per item for the latter. It doesn't matter too much in this case since both the old and new use the same job splitting tactic but it does give some insight into how much overhead there is to just schedule a job.
[Benchmark.zip](https://github.com/stride3d/stride/files/13732288/Benchmark.zip)

sources/core/Stride.Core/Threading/ThreadPool.cs
```diff
@@ -47,8 +51,30 @@ public sealed partial class ThreadPool : IDisposable
...
+            }
+            try
+            {
+                semaphore = new DotnetLifoSemaphore(spinCount);
+            }
+            catch
+            {
+                // For net6+ this should not happen, logging instead of throwing as this is just a performance regression
+                if(Environment.Version.Major >= 6)
+                    Console.Out?.WriteLine($"{typeof(ThreadPool).FullName}: Falling back to suboptimal semaphore");
```
manio143, line 73: Should we log errors to Stride's logging system instead of Console? A lot of the time on Windows you're not going to have a console attached.

sources/core/Stride.Core/Threading/ThreadPool.cs
```diff
@@ -189,5 +256,36 @@ public void Dispose()
...
+
+        private sealed class DotnetLifoSemaphore : ISemaphore
+        {
+            private readonly IDisposable semaphore;
+            private readonly Func<int, bool, bool> wait;
+            private readonly Action<int> release;
+
+            public DotnetLifoSemaphore(int spinCount)
+            {
+                Type lifoType = Type.GetType("System.Threading.LowLevelLifoSemaphore");
```
manio143, line 280: Would be good to add a comment here that we use reflection to access an internal type

manio143 (commented):
Nice work

sources/core/Stride.Core/Threading/Dispatcher.cs
```diff
@@ -25,497 +25,318 @@ public class Dispatcher
...
+            {
+                batch.Release();
             }
         }
 
-        public static void For<TLocal>(int fromInclusive, int toExclusive, [Pooled] Func<TLocal> initializeLocal, [Pooled] Action<int, TLocal> action, [Pooled] Action<TLocal> finalizeLocal = null)
+        private static void TypeAdapter<TJob>(object obj) where TJob : IBatchJob
         {
-            using (Profile(action))
+            var batch = (BatchState<TJob>)obj;
```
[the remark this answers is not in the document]
Eideren: Yep, it's guaranteed;
```cs
ThreadPool.Instance.QueueUnsafeWorkItem(batch, &TypeAdapter<TJob>, batchCount - 1);
```
This call schedules the function `void TypeAdapter<TJob>(object)` to run on other threads with `batch` passed as the `obj` parameter. Both `TypeAdapter<TJob>` and `BatchState<TJob>`'s generic types match as they both come from the `ForBatched<TJob>` generic.
It would be a fairly serious bug if there wasn't any guarantee.
But as you said, `as` is faster than direct cast, it won't move the needle that much compared to how long those jobs tend to take, but anything helps. I'll change it to `as` shortly.

sources/core/Stride.Core/Threading/Dispatcher.cs
```diff
@@ -665,51 +486,84 @@ private static void Swap<T>([NotNull] T[] collection, int i, int j)
             collection[j] = temp;
         }
 
-        private class BatchState
+        /// <summary>
+        /// An implementation of a job running in batches.
+        /// This object is shared across all threads scheduled for the job;
```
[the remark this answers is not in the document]
Eideren: Right, forgot about that one when changing things over, good catch !

sources/core/Stride.Core/Threading/ThreadPool.cs
```diff
@@ -189,5 +256,36 @@ public void Dispose()
...
+        private interface ISemaphore : IDisposable
+        {
+            public void Release( int Count );
+            public void Wait( int timeout = - 1 );
+        }
+
+        private sealed class DotnetLifoSemaphore : ISemaphore
+        {
+            private readonly IDisposable semaphore;
+            private readonly Func<int, bool, bool> wait;
```
xen2, line 275: Curious, can C# 9.0 function pointers (and [GetFunctionPointer()](https://learn.microsoft.com/en-us/dotnet/api/system.runtimemethodhandle.getfunctionpointer?view=net-8.0) be useful in this scenario?

(not sure it would make an actual perf difference though, but curious as if usable enough for this use case, as this would mean I can probably use it in some other places)

manio143 (approved):

sources/core/Stride.Core/Threading/ThreadPool.cs
Eideren: From my limited testing, yes, although I do not know the implications this has for JIT and such; I do remember that the address static function pointers lay on when taking its address is not fixed. If the method is 'moved' after JIT took care of the method, then we might have to retrieve the function pointer from its runtime method handle on every call to make sure we run the optimal version ...

Kryptos-FR:
#1190 can probably be closed now?

Eideren:
Yep, thanks for the reminder @Kryptos-FR

Issue #2053, opened 2023-11-20
Stride.CommunityToolkit Updates when it shouldn't


Doprez:
**Release Type**: Official Release and GitHub

**Version**: 4.2.0.2043

**Platform(s)**: Windows

**Describe the bug**
When opening Stride Stride.CommunityToolkit gets updated to the latest version, in this case 4.2.0.2043. 

**To Reproduce**
1. add the Stride.CommunityToolkit  nuget
2. open Strides editor

**Expected behavior**
since Stride.CommunityToolkit  does not follow the same release schedule it should be ignored from this upgrade step.

**Screenshots**
![image](https://github.com/stride3d/stride/assets/73259914/2a03e124-62b3-46e8-bff4-33e8fe19dd24)

**Log and callstacks**
```
[C:\dev\stride-dev\DotRecast-Stride\DotRecast.Stride\DotRecast.Stride\DotRecast.Stride.csproj(0,0)]: Error: Unable to find package Stride.CommunityToolkit with version (>= 4.2.0.2043)
  - Found 10 version(s) in nuget.org [ Nearest version: 1.0.0-preview.14 ]
  - Found 0 version(s) in Local nugets
  - Found 0 version(s) in Microsoft Visual Studio Offline Packages
  - Found 0 version(s) in C:\Program Files\dotnet\library-packs
  - Found 0 version(s) in Stride Dev
  - Found 0 version(s) in vvvv
[C:\dev\stride-dev\DotRecast-Stride\DotRecast.Stride\DotRecast.Stride.Windows\DotRecast.Stride.Windows.csproj(0,0)]: Error: Unable to build assembly reference [C:\dev\stride-dev\DotRecast-Stride\DotRecast.Stride\DotRecast.Stride\bin\Debug\net8.0\DotRecast.Stride.dll]
```

VaclavElias:
I don't know anymore where we were discussing it but xen2 was suggesting, that only Stride NuGet packages should be updated, not the non Stride NuGet packages. So the toolkit, shouldn't be upgraded obviously, as anyone can name the NuGet package `Stride.Something.` and the engine shouldn't just upgrade it.

The question was how to get list of all NuGet packages, without listing them manually in the code, so when a new package is added, so we don't have to manually upload also the list..

VaclavElias:
Discussed here actually https://github.com/stride3d/stride/pull/1881

Doprez:
[...]

PR #2093, opened 2024-01-04
Add package upgrade list

Doprez:
# PR Details

There are some cases where nugets contain "Stride." but do not follow the same release version. This adds a list that can be used to check if the package should be upgraded.

This is a replacement for https://github.com/stride3d/stride/pull/1881

below is a test I did excluding `Stride.Physics` and `Stride.Navigation` showing that no other packages are effected but the ones not in the array.
image
image

## Description

Adds a static array of valid projects and adds a couple checks for initial upgrades and sequential upgrades.

## Related Issue

https://github.com/stride3d/stride/issues/2053

## Types of changes

## Checklist


Doprez:
Once this gets reviewed I would like to do a demo video using the community toolkit with stride but this error is mildly annoying for me and Im sure if I advertise it many average users would deal with the same issue.

sources/assets/Stride.Core.Assets/StridePackagesToUpgrade.cs
```diff
@@ -0,0 +1,79 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Stride.Core.Assets;
+public static class StridePackagesToUpgrade
+{
+    public static string[] PackageNames =
```
Kryptos-FR, line 10: `Contains()` in a `List<T>` or `Array` is a *O(n)* operation. Better use a `HashSet<string>` for speed.

Kryptos-FR: And if possible, sort them by name to make it easier to maintain.

Doprez: good calls, changed to `HashSet` and alphabetized the list per region and cascading.

VaclavElias:
Maybe we should go with this solution "temporarily" (long term) till the Stride reservation is resolved and taken care of.

Doprez:
I think this list will still be useful for some edge cases but you are right that Stride reservation should be put in place. There were still some packages that had Stride but were not on the same version that I found in nuget.org.

https://www.nuget.org/packages/Stride.GraphX.WPF.Controls
https://www.nuget.org/packages/Stride.GNU.Gettext
https://www.nuget.org/packages/Stride.OpenTK
https://www.nuget.org/packages/Stride.Metrics

and a few more but these are some of the examples that may cause issues for adventerous nuget users.

sources/assets/Stride.Core.Assets/StridePackagesToUpgrade.cs
```diff
@@ -0,0 +1,79 @@
+using System;
```
Kryptos-FR, line 1: Nit-pick: add the header.

```csharp
// Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net)
// Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
```

sources/assets/Stride.Core.Assets/StridePackagesToUpgrade.cs
```diff
@@ -0,0 +1,79 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Stride.Core.Assets;
+public static class StridePackagesToUpgrade
```
Kryptos-FR, line 8: Nit-pick: separate by an empty line.

```suggestion
namespace Stride.Core.Assets;

public static class StridePackagesToUpgrade
```

Kryptos-FR (changes requested):

Eideren:
Xen2 sent a request for reserving the prefix over on nuget so I think it's safe to change from a whitelist to a blacklist instead; list only the specific repo under the prefix that should not be updated instead of listing those that should be. So that would be `Stride.Community.*`, `Stride.Awesome.Shaders` and the other you listed above

Doprez:
> Xen2 sent a request for reserving the prefix over on nuget so I think it's safe to change from a whitelist to a blacklist instead; list only the specific repo under the prefix that should not be updated instead of listing those that should be. So that would be `Stride.Community.*`, `Stride.Awesome.Shaders` and the other you listed above

I changed over to an exclusion list like you said.

here is the quick test I did with the changes
exclusion list

image

result

image

Doprez:
@Eideren I addded a for loop to check the `starts with` validation as well. I couldn't quickly think of a better way than using `ElementAt()` since its a hashset and not a list I cant just index normally so I feel like this defeats the purpose of a hashset but I could be wrong.

Eideren:
Sorry about the delay. Yep, just swap to an array, do an ordinal string equality and we're done :)

Doprez:
Not a problem at all, I think Im done with this PR now.

Eideren:
Thanks !

PR #2094, opened 2024-01-06
Remove binary DataMemberMode

IXLLEGACYIXL:
# PR Details

Removing unused Binary DataMemberMode as its not evaluated

## Description

neither Stride or AssemblyProcessor actually do something with the binary mode, its nowhere used
the only thing it does is throwing an exception if its not applied to an array

## Types of changes

## Checklist

- [x] My change requires a change to the documentation. ( mode is gone, update to stride docs to remove it from the generated files )
- [x] **I have built and run the editor to try this change out.** ( yes but a recompilation of assemblyprocessor was necessary )

## Help
i have no idea if i should push the assemblyprocessor dlls too to the PR


Kryptos-FR:
We will have to check the history of the file (from an old archive of Xenko). I don't think it was added without reason, but it's also possible it was a workaround that has since been fixed and is no longer valid.

IXLLEGACYIXL:
> We will have to check the history of the file (from an old archive of Xenko). I don't think it was added without reason, but it's also possible it was a workaround that has since been fixed and is no longer valid.

i cant find references or strings that would use it, i can run studio with no problem

IXLLEGACYIXL:
i cant find any section that is actively doing something with it

manio143:
I would say first thing is to mark this value as [Obsolete] if not used anywhere and keep it like that for a few package releases.

Then when we do remove it from the enum, we need to preserve the other values. So if `Binary` has a value `3` and then `Never` has a value `4`, when removing `Binary`, you have to also change `Never = 4` to preserve the binary value of this enum.
This is especially important in the context of someone having a project compiled with Stride before this change, who then references the old assembly in a newer project.

IXLLEGACYIXL:
> I would say first thing is to mark this value as [Obsolete] if not used anywhere and keep it like that for a few package releases.
> 
> Then when we do remove it from the enum, we need to preserve the other values. So if `Binary` has a value `3` and then `Never` has a value `4`, when removing `Binary`, you have to also change `Never = 4` to preserve the binary value of this enum. This is especially important in the context of someone having a project compiled with Stride before this change, who then references the old assembly in a newer project.

allright then i will add the numbers and make a separate PR with obsoleting it
and keep this one open

Kryptos-FR:
> > We will have to check the history of the file (from an old archive of Xenko). I don't think it was added without reason, but it's also possible it was a workaround that has since been fixed and is no longer valid.
> 
> i cant find references or strings that would use it, i can run studio with no problem

So I looked at some old code from 8-9 years ago. It was there but already not used. It might be even older than before the code was put in a git repository :).  It seems safe to remove it.

I don't think we need the obsolete extra-step as long as we put the numbers as @manio143 suggested to prevent breaking changes.

IXLLEGACYIXL:
> > > We will have to check the history of the file (from an old archive of Xenko). I don't think it was added without reason, but it's also possible it was a workaround that has since been fixed and is no longer valid.
> > 
> > 
> > i cant find references or strings that would use it, i can run studio with no problem
> 
> So I looked at some old code from 8-9 years ago. It was there but already not used. It might be even older than before the code was put in a git repository :). It seems safe to remove it.
> 
> I don't think we need the obsolete extra-step as long as we put the numbers as @manio143 suggested to prevent breaking changes.

either way i gotta add the interface numbers so the order stays teh same

IXLLEGACYIXL:
added the numbers

IXLLEGACYIXL:
i have to push again assembly processor dlls...

IXLLEGACYIXL:
i had to entirely rebuild the studio but it works

Eideren:
Thanks !

PR #2082, opened 2023-12-16
Make Stride.Core.Presentation cross-platform

Kryptos-FR:
# PR Details

## Description

Make `Stride.Core.Presentation` cross-platform and UI-agnostic (albeit still dependent on the MVVM pattern), and move Windows/WPF-specific code to `Stride.Core.Presentation.Wpf`.

In addition, a few other libraries now target `net-8.0` instead of `.net-8.0-windows7.0`.

Bonus point: the code in `Stride.Core.Presentation` is modernized with implicit usings and nullable reference support.

## Related Issue

Partially based on #2034, which will be rebased on top of this PR once merged.

## Motivation and Context

Mostly code reuse.

Also, the launcher has much less dependencies on WPF than the other editor libraries. With that change, it should be possible to start porting it to Avalonia with minimal effort, and independently from the rewrite.

## Types of changes

- [x] Docs change / refactoring / dependency upgrade
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [ ] My change requires a change to the documentation.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.
  - except the ones already failing 
- [x] **I have built and run the editor to try this change out.**

PR #2034, opened 2023-11-13
Cross-platform editor rewrite - prototype

Kryptos-FR:
# PR Details

This PR contains the initial work on a new cross-platform editor using [Avalonia](https://github.com/AvaloniaUI/Avalonia) for the UI framework.

## Description

In order to keep the changes separate from the existing editor libraries, a new folder `sources/xplat-editor` has been created. In this folder, there are a bunch of new projects, some which have the same naming of existing ones. The idea is that at some point, it should be possible to replace the old projects with the new ones, adapt the existing WPF code and have two versions of the editor (one implemented with WPF, the other with Avalonia) that we can compare to ensure we have all the features ported from the old one to the new one.

## Motivation and Context

*Why not?*

Just kidding. There have been numerous wishes over the years to have an editor working on other platforms than Windows.
There is also a need to have a fully-fledged plugin system, and in order to achieve that similar changes to the one allowing cross-platform experience are needed. Hence, this project aims to achieve both.

## Types of changes

- [x] Docs change / refactoring / dependency upgrade
  - There is a README per project, which surely counts as a doc change.
  - There will be a need to write or rewrite a lot of docs, once we have reached a stable point.
- [x] Bug fix (non-breaking change which fixes an issue)
  - While reworking how the different editor libraries interact with each others, some bugs are likely fixed (but hard to know which ones).
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
  - Likely some breaking changes in the way the editor libraries are working. However, the runtime libraries should be untouched for the most part.

## Checklist

- [x] My change requires a change to the documentation.
- [ ] I have added tests to cover my changes.
  - Nope 😁.
- [ ] All new and existing tests passed.
  - No idea. Not the point of this PR.
- [x] **I have built and run the editor to try this change out.**
  - The new one that is 😉.

PR #2096, opened 2024-01-06
Avalonia Launcher rewrite

Kryptos-FR:
# PR Details

This PR contains the initial work on a porting the Launcher to [Avalonia](https://github.com/AvaloniaUI/Avalonia) .

It is based on #2082, which can either be reviewed/merged separately or directly with this PR.

## Related Issue

- #2034
- #2082

## Motivation and Context

Part of the effort to transition from WPF/Winforms to Avalonia, though it doesn't directly depend on #2034.

## Types of changes

## Checklist

  - Need to rewrite the doc of the Launcher has some prerequisites might have changed. And the visuals are slightly altered. 
- [x] **I have built and run the launcher to try this change out.**


Kryptos-FR:
Merging into `xplat-editor`. If we want to have it for a future official release (and before the whole GameStudio Avalonia port is done), we can just merge that branch again into `master`.

xen2:
Nice job! I tried it and it seems to work well.
Wrote down a few minor cosmetic details (but we can improve on that later, no need to do that now -- I suppose your focus was more on the port itself).
image

Again, great work :)

xen2:
Also, I am curious, how is your experience with Avalonia so far? is it going smoothly, not too many (unexpected) issues?
No specific blockers on the viewmodel side, quantum, stride itself, etc.?

Kryptos-FR:
> Wrote down a few minor cosmetic details for later (but we can improve on that in a second pass later, no need to do that now -- I suppose your focus was more on the port itself than this kind of smaller details).

Good catch. I indeed didn't focus much on visuals, but I think those can be fixed easily. The center alignment issue seems to be a styling difference between Avalonia and WPF though, I'll double check.

> Again, great work :)

Thanks :)

> Also, I am curious, how is your experience with Avalonia so far? is it going smoothly, not too many (unexpected) issues?
No specific blockers on the viewmodel side, quantum, stride itself, etc.?

It's rather smooth. Existing view models work almost as-is, which means I will be able to have at some point both WPF and Avalonia versions side-by-side sharing the same view models. That way, it will be easier to compare and fix any difference in behavior. On the view side, there are a couple of difference in term of how styling works, but the control have for the most part the same name and the same properties. I was pleased to be able to rewrite our own value converters without much trouble (incl. the ability to chain them inline as we are doing in the GameStudio).

PR #2097, opened 2024-01-08
Remove -77

IXLLEGACYIXL:
# PR Details

Changing -77 to 77

## Description

XAML always spammed errors during usage continously that negative values are invalid

## Motivation and Context

Nobody knows why its negative and there was no answer for it and the menuitems are the same as before, just not a spam of bind exceptions

## Types of changes

## Checklist

### DIFF
--- a/sources/core/Stride.Core.Serialization/Serialization/Contents/ContentManager.cs
+++ b/sources/core/Stride.Core.Serialization/Serialization/Contents/ContentManager.cs
@@ -519,7 +519,7 @@ namespace Stride.Core.Serialization.Contents
                     // Find serializer
                     var serializer = Serializer.GetSerializer(headerObjType, objType);
                     if (serializer == null)
-                        throw new InvalidOperationException($"Content serializer for {headerObjType}/{objType} could not be found.");
+                        throw new InvalidOperationException($"Content serializer for {url} could not be found. Was expecting to find type {objType} but the actual type of the resource was {headerObjType}");
                     contentSerializerContext = new ContentSerializerContext(url, ArchiveMode.Deserialize, this)
                     {
                         LoadContentReferences = settings.LoadContentReferences,


manio143:
From what I see this code was trying to set MaxWidth to be ActualWidth - 77. Not sure what would trigger the MaxWidth to cause clipping the width value (maybe higher resolution screen?), but I don't think changing the value to positive would be correct.
SumNum performs addition on two doubles. Maybe we need SubNum which will do subtraction?

Kryptos-FR:
Yeah it could have been a DPI issue (we had quite a lot of them in the past, but I think some were fixed in WPF in later versions). I'll try to find the original commit.

sources/editor/Stride.Core.Assets.Editor/View/DefaultPropertyTemplateProviders.xaml
```diff
@@ -157,9 +157,9 @@
         <!-- Parameters of the SumNum converters used here must be the total width of fixed columns + MinWidths of sizable columns: 8+5+64=77 -->
         <ColumnDefinition Width="8"/>
         <ColumnDefinition Width="{Binding NameColumnSize, RelativeSource={RelativeSource FindAncestor, AncestorType={x:Type sd:PropertyView}}, Mode=TwoWay}"
-                          MaxWidth="{Binding ActualWidth, ElementName=Grid, Converter={sd:SumNum}, ConverterParameter={sd:Double -77}}" MinWidth="64"/>
+                          MaxWidth="{Binding ActualWidth, ElementName=Grid, Converter={sd:SumNum}, ConverterParameter={sd:Double 77}}" MinWidth="64"/>
```
Kryptos-FR, line 160: Setting the value to a positive one doesn't make more sense: we would be expanding the MaxWidth to an ever increasing value. After some thought I believe it might have been a workaround for some constraints not working on columns in an earlier version of WPF. Today, it is pointless. A final fix should be to just remove it altogether.

```suggestion
                          MinWidth="64"/>
```

And apply the same on all other cases (as well as removing the comment above).

PR #2098, opened 2024-01-09
[Serialization] More useful throw when missing parameterless ctor

Eideren:
# PR Details
Serialization throws with a random null ref exception when the type being deserialized does not have a parameterless constructor defined, this PR throws with a more specific exception. Users should be able to figure it out from there until @IXLLEGACYIXL work gets integrated.

## Types of changes
   ^ It would seem like this would be a breaking change, but all calling logic to that method actually expected it to never return null.

## Checklist

### DIFF
--- a/sources/core/Stride.Core.Yaml.Tests/Serialization/SerializationTests.cs
+++ b/sources/core/Stride.Core.Yaml.Tests/Serialization/SerializationTests.cs
@@ -822,6 +822,7 @@ Mother:
             try
             {
                 SerializeThenDeserialize(new ClassWithNonEmptyCtor(default));
+                Assert.Fail("An exception should have been thrown given that the class cannot be instanced because of the missing empty constructor ");
             }
             catch (Exception ex)
             {
--- a/sources/core/Stride.Core.Yaml.Tests/Serialization/SerializationTests.cs
+++ b/sources/core/Stride.Core.Yaml.Tests/Serialization/SerializationTests.cs
@@ -816,6 +816,24 @@ Mother:
             Assert.Equal(22, family.Mother.Age);
         }
 
+        [Fact]
+        public void ThrowWithoutEmptyCtor()
+        {
+            try
+            {
+                SerializeThenDeserialize(new ClassWithNonEmptyCtor(default));
+            }
+            catch (Exception ex)
+            {
+                Assert.True(ex.InnerException is DefaultObjectFactory.InstanceCreationException);
+            }
+        }
+
+        class ClassWithNonEmptyCtor 
+        {
+            public ClassWithNonEmptyCtor(bool parameter) { }
+        }
+
 
         [Fact]
         public void DeserializeEmptyDocument()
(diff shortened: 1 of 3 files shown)


IXLLEGACYIXL:
should i write a diagnostics analyzer for that case?

Eideren:
Yep, that would be best

Kryptos-FR:
Do we have unit tests covering these serialization cases? If so, they might need updating.

IXLLEGACYIXL:
i have it halfway finished but i need polysharp for it, which is currently removed, lets see how it goes when the compilerservices are restored

Eideren:
Added a simple test, it should be enough for this feature ... ?

sources/core/Stride.Core.Yaml.Tests/Serialization/SerializationTests.cs
```diff
@@ -816,6 +816,24 @@ public void DeserializeWithRepeatedSubObjects()
...
+        [Fact]
+        public void ThrowWithoutEmptyCtor()
+        {
+            try
+            {
+                SerializeThenDeserialize(new ClassWithNonEmptyCtor(default));
+            }
+            catch (Exception ex)
+            {
+                Assert.True(ex.InnerException is DefaultObjectFactory.InstanceCreationException);
```
IXLLEGACYIXL, line 829: cant you use here Assert.Throws?

Eideren: I looked at it initially but didn't find an overload for inner exceptions ?

IXLLEGACYIXL: ` Assert.That(ex.InnerException, Is.TypeOf<BadException>() );` mybe that one?

Eideren: Changed to that one instead `Assert.IsType<DefaultObjectFactory.InstanceCreationException>(ex.InnerException);`, no API for `.That()`

sources/core/Stride.Core.Yaml/Serialization/DefaultObjectFactory.cs
```diff
@@ -119,11 +120,12 @@ public object Create(Type type)
                 }
             }
 
-            return null;
+            throw new InstanceCreationException($"Failed to create instance of type '{type}', type does not implement a parameterless constructor.");
```
Kryptos-FR, line 123: Nit: a type *has* or *doesn't have* a parameterless constructor. It doesn't really implement it, the same way an interface or abstract method can be.

```suggestion
            throw new InstanceCreationException($"Failed to create instance of type '{type}', type does not have a parameterless constructor.");
```

Eideren: done

Kryptos-FR (approved):

PR #2104, opened 2024-01-13
feat: Roslyn Diagnostics Analyzer for empty constructor

IXLLEGACYIXL:
# PR Details

Adds a Warning for not having an Empty Constructor

## Description

Diagnostics Analyzer for https://github.com/stride3d/stride/pull/2098 to throw at Compile Time a warning when an empty constructor is missing.

## Types of changes

- [x] Docs change / refactoring / dependency upgrade ( i have to make with vaso a new diagnostics website )

## Checklist

- [x] All new and existing tests passed. ( 66 fail as same with master )


sources/targets/Stride.Core.targets
```diff
@@ -40,20 +40,6 @@
...
-    <ItemGroup>
-      <!-- We want to include Stride analyzers by default -->
-      <AnalyzersNotToRemove Include="@(Analyzer)" Condition="$([System.String]::Copy(%(FullPath)).Contains('Stride'))" />
-      <!-- We want to include source generators by default -->
-      <AnalyzersNotToRemove Include="@(Analyzer)" Condition="$([System.String]::Copy(%(FullPath)).Contains('Generator'))" />
-      <AnalyzersNotToRemove Include="@(Analyzer)" Condition="$([System.String]::Copy(%(FullPath)).Contains('SourceGeneration'))" />
-      <AnalyzersToRemove Include="@(Analyzer)" Exclude="@(AnalyzersNotToRemove)" />
-      <Analyzer Remove="@(AnalyzersToRemove)"/>
-    </ItemGroup>
-  </Target>
```
Kryptos-FR, line 56: This shouldn't be removed.

sources/core/Stride.Core.CompilerServices/Stride.Core.CompilerServices.csproj
```diff
@@ -18,10 +20,15 @@
...
 	</ItemGroup>
-
 	<ItemGroup>
     <PackageReference Include="Microsoft.CodeAnalysis.Analyzers" />
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp" />
+    <PackageReference Include="PolySharp">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageReference>
+
```
Kryptos-FR, line 30: Where is it used? Is it required at runtime?

sources/core/Stride.Core.CompilerServices/Stride.Core.CompilerServices.csproj
```diff
@@ -3,7 +3,9 @@
 	<PropertyGroup>
 		<TargetFramework>netstandard2.0</TargetFramework>
 		<Description>Code generators for Stride.Core and its dependents</Description>
+    <LangVersion>12</LangVersion>
     <LangVersion>preview</LangVersion>
```
Kryptos-FR, line 7: Property collision. Use `latest`.

sources/targets/Stride.Core.targets
IXLLEGACYIXL: it doesnt make a difference, stride compiles fine without it


i dont know why it is there

sources/core/Stride.Core.CompilerServices/Stride.Core.CompilerServices.csproj
IXLLEGACYIXL: https://github.com/stride3d/stride/pull/2104/files#diff-015a4d048579b446685ff8c8dbddb5302ff100cd3fbfc1df1cef57c74a4032b6R66

its used here for pattern matching as i got told that i should do it that way

sources/core/Stride.Core.CompilerServices.Tests/Stride.Core.CompilerServices.Tests.csproj
```diff
@@ -22,7 +22,7 @@
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\..\tests\xunit.runner.stride\xunit.runner.stride.csproj" />
-    <ProjectReference Include="..\Stride.Core\Stride.Core.csproj" ReferenceOutputAssembly="false" OutputItemType="Content" CopyToOutputDirectory="Always" />
+    <ProjectReference Include="..\Stride.Core\Stride.Core.csproj" ReferenceOutputAssembly="true" OutputItemType="Content" CopyToOutputDirectory="Always" />
```
IXLLEGACYIXL, line 25: i had problems here as it searched in shared folder of c# this works too and doesnt make problems with copying stride.core to the output

sources/targets/Stride.Core.targets
Kryptos-FR: Don't make changes that are not needed. It was there for a reason, and unless your changes require it to be removed it shouldn't. A PR should have the minimal changes that are required for a given feature, nothing more.

If there is an argument to be made about removing that part, it should be on its own PR and discussed there.
If as you said "it doesn't make a difference", then leave it.

I believe one of the reason this was added was to speed-up the build by disabling analysis that is not required  (esp. during CI), as well as removing a lot of warnings that would make the logs harder to read in case a certain build needs investigating.

sources/core/Stride.Core.CompilerServices/Stride.Core.CompilerServices.csproj
Kryptos-FR: Patter-matching is a feature of the language, it doesn't require an additional dependency. Your code seems to compile without Polysharp what does it add?

IXLLEGACYIXL: grafik
System.Index and System.Index.GetOffset

ive changed it to a foreach loop now and added the analyzer removal back in

Kryptos-FR: Just to keep a trace of what we discussed on Discord:
1. `netstandard2.0` is a requirement for analyzers/source generators. Nothing we can do about.
2. because of that, certain languages features aren't available (like index/spread operations on collections in C# 12), unless we use a library such as PolySharp to fill the gaps.

sources/targets/Stride.Core.targets
Kryptos-FR: After discussion on Discord, it appears that the Stylecop settings might be some Xenko leftovers. Maybe @xen2 can shine a light on this. But basically:
1. we are adding a package reference for Stylecop assembly for all projects that have `<StrideCodeAnalysis>true</StrideCodeAnalysis>`. This seems to be most core and engine projects (not applied to editor).
2. since these generates a lot of warnings in the build log, those analyzers are disabled by default (unless `StrideEnableCodeAnalysis` is also set to `true`.
3. however, some analyzers/generators still need to be enabled at all time. Hence this "complex" rule with a allow-list of analyzers to not remove.

We have two options:
1. Either we keep the stylecop rule, in which case we can add another exception for PolySharp in that "alow-list".
2. We decide we don't want stylecop anymore and we just remove all of that.
  a. we do keep the first `StrideCodeAnalysis` section to use the proper ruleset

IXLLEGACYIXL: currently stylecop is added to each project, but no project has the StrideEnableCodeAnalysis so it gets removed immediately after

Kryptos-FR: I think it was initially meant as an optional command line parameters when running a build.
Something like: `msbuild -p:StrideEnableCodeAnalysis=true ...`

Anyway, if #2105 is approved and merged, we won't need to discuss it anymore :sweat_smile:

sources/core/Stride.Core.CompilerServices.Tests/Stride.Core.CompilerServices.Tests.csproj
manio143: Ideally we don't want the test project to reference Stride.Core to ensure we don't accidentally introduce subtle bugs in case it affected the analyzer run. But maybe it doesn't matter. If referencing it helps always copy the Stride.Core.dll to output to be available to the test compilation, then I think it's fine.

Kryptos-FR:
Note: make sure to validate and merge #2105 first.

sources/core/Stride.Core.CompilerServices/Analyzers/STRDIAG010InvalidConstructor.cs
```diff
@@ -0,0 +1,88 @@
...
+            {
+                if (datacontractData is { NamedArguments: [.., { Key: "Inherited" } ], NamedArguments: [..,{ Value: TypedConstant  inherited }] })
+                {
+                    isInherited = (bool)inherited.Value!;
+                }
+                break;
+            }
+            type = type.BaseType;
+        }
+        if(isInherited)
```
manio143, line 74: If it's both inherited and present on the class will it be reported twice?

sources/core/Stride.Core.CompilerServices/Analyzers/STRDIAG010InvalidConstructor.cs
```diff
@@ -0,0 +1,88 @@
...
+    }
+    private static void TryReportDiagnostics(INamedTypeSymbol symbol,SymbolAnalysisContext context)
+    {
+        if (HasEmptyConstructor(symbol))
+            return;
+        else
+            Rule.ReportDiagnostics(context, symbol);
+    }
+    private static bool HasEmptyConstructor(INamedTypeSymbol type) 
+        => type.Constructors.Any(x => x.Parameters.Length == 0);
```
manio143, line 87: Does this work fine for structs?
Also you probably have to check that the constructor is public. Generic constraints only work with public `new()` and I believe that is being used in the serialization system.

sources/core/Stride.Core.CompilerServices/Analyzers/STRDIAG010InvalidConstructor.cs
```diff
@@ -0,0 +1,88 @@
...
+using Stride.Core.CompilerServices.Common;
+
+namespace Stride.Core.CompilerServices.Analyzers;
+
+[DiagnosticAnalyzer(LanguageNames.CSharp)]
+public class STRDIAG010InvalidConstructor : DiagnosticAnalyzer
+{
+    public const string DiagnosticId = "STRDIAG010";
+    private const string Title = "Invalid Constructor";
+    private const string MessageFormat = "The Type '{0}' doesn't have an empty constructor, which is needed for Serialization";
```
manio143, line 15: If I'm right about public below, should be updated here. And I think the phrase is `public parameterless constructor`.

sources/core/Stride.Core.CompilerServices/Analyzers/STRDIAG010InvalidConstructor.cs
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        var type = symbol.BaseType;
+        bool isInherited = false;
+        while (type != null)
+        {
+            // Check if the type has the specified DataContractAttribute through inheritance
+            if (type.TryGetAttribute(dataContractAttribute,out var datacontractData) && datacontractData.AttributeConstructor is not null)
+            {
+                if (datacontractData is { NamedArguments: [.., { Key: "Inherited" } ], NamedArguments: [..,{ Value: TypedConstant  inherited }] })
```
manio143, line 66: Is this pattern matching syntax based on index of the argument (second in this case)? I think it would be better to use something like `TryGetValue` pattern if supported by `NamedArguments` if the index can change

sources/core/Stride.Core.CompilerServices/Analyzers/STRDIAG010InvalidConstructor.cs
IXLLEGACYIXL: it has to be public , internal doesnt count, the editor wont find the type if its internal

IXLLEGACYIXL: is updated

sources/core/Stride.Core.CompilerServices/Analyzers/STRDIAG010InvalidConstructor.cs
IXLLEGACYIXL: it works with structs as a struct has always an empty constructor , doesnt matter if you declare a non empty one

sources/core/Stride.Core.CompilerServices/Analyzers/STRDIAG010InvalidConstructor.cs
IXLLEGACYIXL: https://github.com/stride3d/stride/pull/2104/commits/84c1376c00bb0864f49e8293caa59bb1583531de
added a test to confirm it works

the pattern matching looks "loop until Inherited appears, then put it inot a bool" and fills it up so thats fine

sources/core/Stride.Core.CompilerServices/Analyzers/STRDIAG010InvalidConstructor.cs
IXLLEGACYIXL: added an early return if the class itslelf has datacontract

sources/core/Stride.Core.CompilerServices/Analyzers/STRDIAG010InvalidConstructor.cs
manio143: Would it work to inline Value decomposition into the object earlier?
```
{ NamedArguments: [.., { Key: "Inherited", Value: value }] }
```

IXLLEGACYIXL: yes that worked [4e81a8f](https://github.com/stride3d/stride/pull/2104/commits/4e81a8f24f036a3f066858c82b4607276b1afd5d)

IXLLEGACYIXL:
i will reopen when i know whats wrong with git lfs

IXLLEGACYIXL:
i think its ready to merge, should i change the name of the PR to use the new tags

sources/targets/Stride.Core.targets
```diff
@@ -30,7 +30,7 @@
       </ProjectReferenceWithConfiguration>
     </ItemGroup>
   </Target>
-
+
```
Eideren, line 33: Could you remove this unnecessary change ?

sources/core/Stride.Core.CompilerServices/Common/SymbolExtensions.cs
```diff
@@ -13,10 +13,23 @@ public static bool IsVisibleToSerializer(this ISymbol symbol, bool hasDataMember
...
+    /// </summary>
+    /// <param name="symbol">The Symbol to search on</param>
+    /// <param name="attribute">The attribute looking for</param>
+    /// <param name="attributeData">The <see cref="AttributeData"/> of the Attribute if it is found</param>
+    /// <returns>true if the attribute is found, else false</returns>
+    public static bool TryGetAttribute(this ISymbol symbol, INamedTypeSymbol attribute, out AttributeData attributeData)
+    {
+        attributeData = symbol.GetAttributes().FirstOrDefault(attr => attr.AttributeClass?.OriginalDefinition.Equals(attribute, SymbolEqualityComparer.Default) ?? false)!;
+        return attributeData is not null;
+    }
```
Eideren, line 31: I think this can be improved, do you have access to [MaybeNullWhen](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.maybenullwhenattribute?view=net-8.0) in this project ? To decorate `AttributeData attributeData` into `[MaybeNullWhen(false)] AttributeData attributeData`
Then:
```cs
if (symbol.GetAttributes().FirstOrDefault(attr => attr.AttributeClass?.OriginalDefinition.Equals(attribute, SymbolEqualityComparer.Default) ?? false) is {} data)
{
	attributeData = data;
	return true;
}

attributeData = null;
return false;
```
You can also change the signature of the function, something like:
```cs
public static AttributeData? TryGetAttribute(this ISymbol symbol, INamedTypeSymbol attribute)
{
    return symbol.GetAttributes().FirstOrDefault(attr => attr.AttributeClass?.OriginalDefinition.Equals(attribute, SymbolEqualityComparer.Default) ?? false);
}
```
And adapt the call site to this new signature instead.

Eideren (commented):
Looks great, thanks, just two small notes

sources/core/Stride.Core.CompilerServices/Common/SymbolExtensions.cs
IXLLEGACYIXL: https://github.com/stride3d/stride/pull/2104/commits/d8103fc25e74ba1f8aef348b9f693279742e49f0 added

sources/targets/Stride.Core.targets
IXLLEGACYIXL: removed it

IXLLEGACYIXL:
completed the changes

IXLLEGACYIXL:
the name of the PR should be changed to meet the tags for the new system that vaso made but i dont really know what it should be named to meet the categories

Eideren:
Thanks !

IXLLEGACYIXL:
https://github.com/stride3d/stride-docs/pull/295
the docs PR for this analyzer

Issue #2107, opened 2024-01-15
Using undo on an assembly reload causes various problems


adrsch:
**Release Type**: GitHub 

**Version**: Very recent

**Platform(s)**: Editor

**Describe the bug**
You can undo an assembly reload with ctrl+z, and it always causes problems.

I have had:
- Serialization errors requiring restart
- Actual crash of editor
- I want to say it causes a bug that's the bane of my existence, adding multiple copies of scripts, but I can't repro that right now so it may be something else. But I feel like this is what's causing it.

**To Reproduce**
1.  Get "assembly reload" as the top thing on the edit history panel
2. Press ctrl+z

**Expected behavior**
Personally I don't expect this to be something you can undo, and I never am doing this on purpose!

**Additional context**
I **_will_** submit a PR removing the ability to undo this unless anyone has other ideas to fix :laughing:

adrsch:
I see in GameStudioAssemblyReloader, there are various operations pushed. Perhaps different things go wrong for each of the effects. But it still does have a near 100% repro rate, at least in my project, for causing problems.

IXLLEGACYIXL:
can you reproduce it with this branch https://github.com/stride3d/stride/pull/2106 ?

Kryptos-FR:
Please provide a MRE so that:
1. The issue can be investigated.
2. If confirmed as a bug, resolution can be discussed
3. Once a fix is implemented, the same MRE can be used to validate it.

adrsch:
> can you reproduce it with this branch #2106 ?
> 
> i was able to unload properly

Very cool change! I will test that as soon as I can

IXLLEGACYIXL:
> > Please provide a MRE so that:
> > 
> > 1. The issue can be investigated.
> > 2. If confirmed as a bug, resolution can be discussed
> > 3. Once a fix is implemented, the same MRE can be used to validate it.
> 
> I am unfamiliar with the terminology MRE. I will type up a reproduction using a sample project and a standard release, since I just came up with one.
> 
> Release 4.2.0.2067
> 
>     1. Create a new project using the "TopDownRPG" sample.
> 
>     2. Open MainScene in the editor
> 
>     3. Select the first LootBox prefab with the CrateScript component attached. Move this to a different position using the editor.
> 
>     4. Open the code project, and open CrateScript.cs
> 
>     5. Introduce a bug in this class preventing it from compiling.
> 
>     6. Go back to the editor, and reload scripts. It will fail to compile due to the error.
> 
[...]

PR #2108, opened 2024-01-15
Let assembly reloading bypass undo/redo to improve editor stability

adrsch:
# PR Details

Hack to prevent serious editor bugs on undo/redo for assembly reloads, while leaving the code for it all intact. Introduces poor UI behavior on the Edit History tab, but that's never caused anyone to lose work.

## Description

Add a transaction flag for bypassing the undo/redo system, which should not be used except for hacks.

Tested by moving things around and/or editing properties, doing stuff to cause an assembly reload, and then using undo. The things done before the editor reload, successfully undo.

## Related Issue

https://github.com/stride3d/stride/issues/2107

## Motivation and Context

Pain

## Types of changes

## Checklist


adrsch:
@dotnet-policy-service agree

Kryptos-FR:
That's not the way to do it. There is already infrastructure in place to have not-cancellable operations.

With that said, I seem to remember (old story from when I was at SiliconStudio) that there was a reason we still needed that operation to be cancellable. I think it was related to serialization.

Kryptos-FR:
In also added a comment in the related issue. At this stage, unless it can be reproduced and confirmed, a hack should not be merged without measuring its consequences.

adrsch:
> That's not the way to do it. There is already infrastructure in place to have not-cancellable operations.
> 
> With that said, I seem to remember (old story from when I was at SiliconStudio) that there was a reason we still needed that operation to be cancellable. I think it was related to serialization.

I figured this hack was terrible! 

Hopefully the standard reproduction I wrote up on the issue a few minutes ago, using a sample project, can help it get fixed properly, I mostly just figured it would be good to get a hack out there. Since this seems like a thing some people will run into not at all, or all the time, depending on their habits with using undo in the editor.

adrsch:
I got the true cause figured out of duplicate scripts getting added in https://github.com/stride3d/stride/issues/2107#issuecomment-1892788016 

Not sure how to fix, it seems like a problem that has got to have been addressed elsewhere in the engine (an item getting processed twice, once as a member of a scene, and second as a member of a prefab)

xen2:

PR #2111, opened 2024-01-19
added option to add a parent reference for SDL window

Doprez:
# PR Details

I wanted an easy way to embed Stride into other UI frameworks so I decided to make a simple breakdown of https://github.com/stride3d/stride/pull/1474 without the built in demos.

## Description

This PR allows for SDL to take in a parent handle and use that for rendering.

WinForms example:
image

In the future I would like to use this for experimenting with Avalonia as the parent host using the NativeControlHost.

## Related Issue

https://github.com/stride3d/stride/pull/1474

## Motivation and Context

Some people have asked for an easy way to embed Stride into UI frameworks in the Discord this will allow an easy constructor to pass in a parent handle.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Graphics/SDL/Window.cs
+++ b/sources/engine/Stride.Graphics/SDL/Window.cs
@@ -37,7 +37,14 @@ namespace Stride.Graphics.SDL
         /// Initializes a new instance of the <see cref="Window"/> class with <paramref name="title"/> as the title of the Window.
         /// </summary>
         /// <param name="title">Title of the window, see Text property.</param>
-        public unsafe Window(string title)
+        public Window(string title) : this(title, IntPtr.Zero) { }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Window"/> class with <paramref name="title"/> as the title of the Window.
+        /// </summary>
+        /// <param name="title">Title of the window, see Text property.</param>
+        /// <param name="parent">Parent window handle</param>
+        public Window(string title, IntPtr parent)
         {
             WindowFlags flags = WindowFlags.AllowHighdpi;
 #if STRIDE_GRAPHICS_API_OPENGL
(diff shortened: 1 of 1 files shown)


Kryptos-FR:
Thanks for the contribution. A couple of questions:

What's causing the inputs to not work?

Can this is be used already to adapt the GameForm used in the GameStudio? And in that way validate that it might work later with Avalonia.

Doprez:
> What's causing the inputs to not work?

Im not 100% sure honestly. The Winforms inputs seem to be registered but the the results are not consistant. When you use the mouse to look around in the particle scene for example, it will slowly rotate and on release of the mouse it will flick the view in the direction it was moving.

A couple thoughts I have had are:
- wondering if it is using 2 seperate inputs, one for the SDL for and one for the main form?
- Do we need to take over the message loop manually?

but I havent found a solution or specific answer.

> 
> Can this is be used already to adapt the GameForm used in the GameStudio? And in that way validate that it might work later with Avalonia.

Your Avalonia branch was actually the main inspiration for me wanting this. Its why I wanted to test with Avalonia in the future when I have more time.

Doprez:
>  it will slowly rotate and on release of the mouse it will flick the view in the direction it was moving
Forget what I said here, this is apparently normal behaviour for this demo...

Doprez:
@Kryptos-FR so I went back and tested all of the templates apart from the VR one and I had zero issues with using inputs.

the problems I faced:
- the mouse cursor is not hidden but is can be locked in place.
- framerate needs to be changed for the minimized window for some reason. When using SDL as a child control it seems to think it is minimized until fullscreened so it will run at a lower FPS unless you set `MinimizedMinimumUpdateRate.MinimumElapsedTime`

tested the following with Winforms:
- FPS template works fine
- Third person template works fine
- Top down template works fine
- Particle demo works fine

Doprez:
image

https://github.com/Doprez/stride-in-avalonia

Decided to mess with Avalonia since everyone at work is just waiting on the weekend lol. And a black screen is better than no screen I think?

Doprez:
Took a step back to try and get the WPF dexample working from the previous PR because Avalonia is being weird to set up.

it works with WPF and Ill add it to the my repo along with my attempts to get Avalonia running:
image

Eideren:
Is this one ready for merge or do you want to figure out the issue you're having with Avalonia first ? I'm guessing if there are changes required to make avalonia work it'll likely be on the client side rather than the engine one but just in case.

Kryptos-FR:
@Eideren if there is any change needed for Avalonia it will be done at that time. I think it's ok to merge this PR as long as WPF is still working.

Kryptos-FR (approved):

Doprez:
Yep, these issues are mostly with understanding avalonia itself, the pr should be good.

Eideren:
Thanks !

PR #2119, opened 2024-01-22
[OpenXR] Adds optional support for the Passthrough extension

joreg:
# PR Details

For OpenXR devices, this PR adds the option to instantiate the [XR_FB_passthrough](https://registry.khronos.org/OpenXR/specs/1.0/html/xrspec.html#XR_FB_passthrough) extension when enabling a VRDevice.

## Description

OpenXR supports a wide range of optional [extensions](https://registry.khronos.org/OpenXR/specs/1.0/html/xrspec.html#extension-appendices-list). Currently the OpenXR implementation in Stride allows to register extensions by adding them to the `OpenXRHmd.extensions` list before the device is started. Like this any exension that does not require any deeper integration with the engine (e.g. the XR_EXT_hand_tracking) can then be dealt with in custom code. 

In order to support the XR_FB_passthrough extension we needed tighter integration with the engine because it requires a composition layer. So this PR "hardcodes" the support for the passthrough extension which still has to be requested for the device in order to be activated. 

In `OpenXRHmd.Enable(..)` we added a new argument "requestPassthrough" which, if true, adds the XR_FB_passthrough to the list of desired extensions and in case the current hardware supports this extension, it will be instantiated in the public `PassthroughExt` property. This public property is then meant to be controlled from custom code by setting its public `Enabled` property.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.VirtualReality/OpenXR/OpenXRHmd.cs
+++ b/sources/engine/Stride.VirtualReality/OpenXR/OpenXRHmd.cs
@@ -634,16 +634,6 @@ namespace Stride.VirtualReality
         }
 #endif
 
-
-        public unsafe void SetPassthroughLayer(IntPtr layer)
-        {
-            if (layer != IntPtr.Zero)
-            {
-                this.compositionLayers[0] = (CompositionLayerBaseHeader*)layer;
-                nextCompositionLayer = 1;
-            }
-        }
-
         public override void Commit(CommandList commandList, Texture renderFrame)
         {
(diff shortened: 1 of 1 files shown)


Eideren:
@joreg how do you want to move forward with this ? @azeno proposed to fix your conflicts in #2121, would you rather take care of that yourself ?

azeno:
We'll rebase this PR and report back. Thanks for merging so quickly!

azeno:
To be continued in #2141

PR #2121, opened 2024-01-23
[OpenXR] Fixes crashes when no device is present

azeno:
# PR Details

The essential part here is that `OpenXRHmd.CanInitialize` will now return false if no device is present. This ensures that the `VRDeviceSystem` sets up a dummy device which in turn prevents many parts from crashing.

## Description

To truly know whether or not an actual device is connected we first need to establish contact with the OpenXR runtime. Therefor code has been moved around a bit to be able to call various parts individually. Tested on two machines, one without any VR related things installed and another one with an actual device.

This commit also contains a bunch of code cleanups:
- Enables C# nullable feature on OpenXR files
- Adds null checks to various places
- Disposes XR API after shutting down the device
- Makes various fields and methods private
- Moves initialization code to a `OpenXRUtils` class to make dependencies more visible
- Turns `OpenXRInput` into a normal class holding a reference to the device
- Turns `CheckResult` into an extension method and simplifies its usage by letting C# infer the called method
- Introduces `OpenXRException` exception class carrying the inferred method and XR error code

## Related Issue

Probably causes conflicts with #2119 which we would offer to fix.

## Motivation and Context

The system shouldn't crash when no VR device is present.

## Types of changes

## Checklist

- [x] I did run a VR patch in vvvv with two machines, one having a device, and the other which doesn't

### DIFF
--- a/sources/engine/Stride.VirtualReality/OpenXR/OpenXRException.cs
+++ b/sources/engine/Stride.VirtualReality/OpenXR/OpenXRException.cs
@@ -1,0 +1,19 @@
+using System;
+using Silk.NET.OpenXR;
+
+namespace Stride.VirtualReality
+{
+    internal sealed class OpenXRException : Exception
+    {
+        public OpenXRException(Result result, string methodName)
+            : base($"{methodName} returned {result}")
+        {
+            Result = result;
+            MethodName = methodName;
+        }
+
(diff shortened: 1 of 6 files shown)


Eideren:
Looks good @azeno - thanks again :) 
We'll continue over in #2119

PR #2131, opened 2024-01-28
feat: Bepu physics integration

Eideren:
# PR Details
Adds support for [Bepu Physics](https://github.com/bepu/bepuphysics2) as an alternative physics engine, we'll keep the current one but will move our focus towards this new one instead.

## Motivation
Unified ecosystem, improved performance, flexibility and debugging experience compared to Bullet's physics engine.

## Types of changes

## Checklist

## Post release
- Query mesh material index from physics tests
- Ragdoll utilities
- Wider support of the different mesh formats (ExtractBepuMesh)
- Wait for https://github.com/bepu/bepuphysics2/issues/325 to add convex support
- Ask if Norbo is available to check some areas that could be improved
- Navigation

A huge thanks goes to @Nicogo1705 and @Doprez for working on this. As well as @RossNordby for Bepu and dealing with our nonsense.


sources/engine/Stride.BepuPhysics/Stride.BepuPhysics/Definitions/BepuAndStrideExtensions.cs
```diff
@@ -4,51 +4,31 @@
...
 using SRigidPose = Stride.BepuPhysics.Definitions.RigidPose;
 
-using NBodyVelocity = BepuPhysics.BodyVelocity;
+using BBodyVelocity = BepuPhysics.BodyVelocity;
 using SBodyVelocity = Stride.BepuPhysics.Definitions.BodyVelocity;
 
 using System.Runtime.CompilerServices;
 
 
 internal static class BepuAndStrideExtensions
```
Doprez, line 16: Since these are dealing with System.Numerics and Stride.Math should this be added directly to the Stride math library? I feel like these would be useful in many places not only Bepu.

Doprez: Some of these seem to already exist in the 3d importer as well https://github.com/stride3d/stride/blob/0e053a3b8873f49e9e0e5450a3dfad368a6ed042/sources/tools/Stride.Importer.3D/Utils.cs#L13

Doprez:
I was trying to use your branch locally but I cant seem to resolve the nugets to the demo project, am I missing a step?
image

Doprez:
I was trying to add configuration and the drop down seems to not show up. 

I also cant build the project due to the previous error of not being able to find the 4.2.0.1 version from the compilerapp.

I am not sure if both are related but it seems like its just some build process that is missing.

adding components seem to work normally compared to the repo from Nicogo.

Eideren:
Forgot to mention that nonsense, there is a ridiculous workaround in place because central package management is still an incomplete often buggy mess.
The issue goes something like this; `Demo` should behave like a normal stride game solution, so instead of referencing the engine projects, like other solutions in the engine, it references the packages of the engine, those packages must be the local ones to ensure that we catch changes the user made to the engine locally.
Previously this worked fine, but under central package management, we must add a package source to reference local packages. Obviously this is just for one specific project, so I shouldn't add it to the global one. Given that, here's what I tried:
- `<ManagePackageVersionsCentrally>false</ManagePackageVersionsCentrally>`
- Adding a `Directory.Packages.props` to a sub-directory to override the global one.
It either only works on clean builds or is just ignored entirely.

So, what you should be doing is build all projects in `build\Stride.sln` then open `sources\engine\Stride.BepuPhysics\Stride.BepuPhysic.Demo.sln` in another VS instance and build/run it there ...

If someone want to take a look at this feel free to do so, I just don't have the patience to deal with nuget's garbage anymore.

sources/engine/Stride.BepuPhysics/Stride.BepuPhysics/Definitions/BepuAndStrideExtensions.cs
Eideren: Sure, feel free to open a PR for that :P

delustra:
Testing the branch, couple of issues surfaced:
1. Collidable Gizmo doesn't update in the editor when I change the Position of the Body Component. 
gizmo
 
 2. I have setup a constraint between two bodies. I set new Mass for the collider of one of the bodies. The next frame the constraint looses Constraint.A and Constraint.B and the Body itself looses it's velocity. 
 My guess is that in CollidableComponent.cs.
 
```
 internal void TryUpdateFeatures()
    {
        if (Simulation is not null)
            ReAttach(Simulation);
    }
```
the items gets detached from the simulation and then gets attached again. That's where it probably looses all its values.

Eideren:
@delustra # 2 should be fixed, let me know if that's not the case :)

Eideren:
# 1 has been dealt with

delustra:
feature request: I have a bodycomponent, which consists of many collider boxes. When I do raytracing to that bodycomponent, I want to get information which exact small collider box was hit within bodycomponent.

delustra:
I think I've found an issue. In CollisionMask.cs it says:
_Collision occurs when 'a' contains all the layers set in 'b' or 'b' contains all the layers of 'a'._
```
    [MethodImpl(MethodImplOptions.AggressiveInlining)]
    public static bool Collide(this CollisionMask a, CollisionMask b)
    {
        CollisionMask and = a & b;
        return and != 0 && (and == a || and == b);
    }
```
in that logic, layers of **a** will always be a subset of layers of **b** (or vice versa). Meaning, that adding a layer of interaction to either **a** or **b** will make them noninteractive.

I'd propose doing something similar to godot. Layer and mask are two different sets of values. Layer is where the object is present and mask is what the object can interact with. They've encountered something similar before https://github.com/godotengine/godot-proposals/issues/2775.

Eideren:
@delustra This is one area I've been meaning to address, but I put it on the back burner as there's one edge case that I have not found a good lightweight solution for yet; both the current solution and unity/godot's implementation still require another system to ignore collisions between specific objects, rag-dolls for example typically do not collide between parts that are directly connected, but do collide with parts of other rag-dolls. Layers and masks would not handle this appropriately. Using some sort of hash table would work of course, but would introduce a significant performance hit I would rather avoid if at all possible.

mihai-ene-public:
Hi, guys. Will this feature be in the next release? Or is there a planned release for this feature?
Thanks

Eideren:
@mihai-ene-public Yes, it's almost ready for release, just a couple of things to iron out. Likely to be released in a minor 4.2 for testing purposes and then 4.3 once we think it is stable enough and we have samples to go along with it.

Doprez:
Is there a proper way to remove bepu entities from the scene? I was trying to do it the same way I usually do with just setting the scene to null but with this branch it hits the Debug.Assert for [bodies](https://github.com/Eideren/xenko/blob/570a12f22aff3f12bed6b9b950749b1674b4dc99/sources/engine/Stride.BepuPhysics/Stride.BepuPhysics/BepuSimulation.cs#L287) and [statics](https://github.com/Eideren/xenko/blob/570a12f22aff3f12bed6b9b950749b1674b4dc99/sources/engine/Stride.BepuPhysics/Stride.BepuPhysics/BepuSimulation.cs#L294). 

The way I am removing now is casting a ray to find the collidable:
```cs
	public bool ScreenRayCastPenetrating(out List<HitInfo> hits)
	{
		hits = [];

		Entity.Transform.GetWorldTransformation(out var _, out var rotation, out var _);
		var worldDir = -Vector3.UnitZ;
		rotation.Rotate(ref worldDir);

		Span<HitInfoStack> buffer = stackalloc HitInfoStack[16];
		int j = 0;
		foreach (var hitInfo in _bepuConfig.BepuSimulations[BepuSimulation].RaycastPenetrating(Camera.Entity.WorldPosition(), worldDir, RayLength, buffer, CollisionMask))
		{
			hits.Add(hitInfo);
		}
		if (j == 0)
		{
			return false;
		}
		return true;
	}
```

and then I just set the scene to null:
```
	public void Interact()
	{
		RaycastHandler.ScreenRayCastPenetrating(out var hits);
		if (hits.Count == 0)
			return;

		var hit = hits[0];

		// AFAIK the hits are sorted by distance, so the first hit is probably the selected object.
		// If that's the case, we'll use the second hit which should be consistant.
		if (hit.Collidable.Entity == _character.Entity)
		{
			if (hits.Count < 2)
				return;
			hit = hits[1];
		}

		// Do something with the hit
		hit.Collidable.Entity.Scene = null;
	}
```

Eideren:
Your logic is correct*, you should not have hit those assertions, I'll look into those asap.

*The comment making the assumption that hits are sorted by distance is incorrect though, and the summary does mention this. The only guarantee you have is that the hits collected are the closest `n` ones; the first one could very well be further away from the origin than the last one. Sorting does require additional processing and may not be required depending on the context, so we're skipping that step, as does unity for example.

Doprez:
So I may have been too hasty in reporting this. I was a couple weeks behind from your branch and after pulling and rebuilding I cant seem to recreate the error. I'll be messing with this a good amount in the coming month due to an inventory system requirement for the player so Ill keep a close eye on this if it comes up again.

>*The comment making the assumption that hits are sorted by distance is incorrect though, and the summary does mention this. The only guarantee you have is that the hits collected are the closest n ones; the first one could very well be further away from the origin than the last one. Sorting does require additional processing and may not be required depending on the context, so we're skipping that step, as does unity for example.

Ah ok, that is good to know. I have a dedicated abstraction class for raycasts so I can filter points before returning in that case. Thank you!

Eideren:
Welp, this PR is done, if anyone has time to look through those changes, even if a small part I would be very grateful :)

sources/engine/Stride.BepuPhysics/Stride.BepuPhysics/Systems/ShapeCacheSystem.cs
Doprez, line 1: This is a hard one for me to leave internal. Since it is required for the Navigation library to be able to create nav meshes, if something is found to be a problem in the near future users cant get the object data from Bepu. I know we discussed this before and you were unsure if we were keeping the shape cache, is that still an issue?

sources/engine/Stride.BepuPhysics/Stride.BepuPhysics/Definitions/UnsortedO1List.cs
Doprez, line 1: Is this required? I didnt check if it was used but if it is, should we use it?

sources/engine/Stride.BepuPhysics/Stride.BepuPhysics/Systems/CollidableProcessor.cs
```diff
@@ -0,0 +1,94 @@
...
+    protected override void OnSystemAdd()
+    {
+        ServicesHelper.LoadBepuServices(Services, out var config, out var shapes, out _);
+        BepuConfiguration = config;
+        ShapeCache = shapes;
+    }
+
+    public override unsafe void Draw(RenderContext context) // While this is not related to drawing, we're doing this in draw as it runs after the TransformProcessor updates WorldMatrix
+    {
+        base.Draw(context);
```
Doprez, line 44: Not needed, these are empty in the base class of a processor

sources/engine/Stride.BepuPhysics/Stride.BepuPhysics/Systems/ConstraintProcessor.cs
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    protected override void OnSystemAdd()
+    {
+        ServicesHelper.LoadBepuServices(Services, out _bepuConfiguration, out _, out _);
+    }
+
+    protected override void OnEntityComponentAdding(Entity entity, ConstraintComponentBase component, ConstraintComponentBase data)
+    {
+        base.OnEntityComponentAdding(entity, component, data);
```
Doprez, line 26: Not needed

sources/engine/Stride.BepuPhysics/Stride.BepuPhysics/Systems/ConstraintProcessor.cs
```diff
@@ -0,0 +1,34 @@
...
+        component.Activate(_bepuConfiguration);
+    }
+    protected override void OnEntityComponentRemoved(Entity entity, ConstraintComponentBase component, ConstraintComponentBase data)
+    {
+        base.OnEntityComponentRemoved(entity, component, data);
```
Doprez, line 31: Not needed

Doprez (commented):
I did a pretty quick overview but I think this looks good to me on first glance. I did have a question about the Shape Cache that would affect me for the navigation stuff in the future.

Doprez:
Ah ok, I got the error again with the updated branch. It seems like the Debug.Assert doesn't always get hit? I dont know if it has to do with order of operations for how Im removing entities from the scene but if that were the case then I would assume it would be more consistant?

Here are all 7 items of the array at the time of break:
image

and the stack trace:
```
>	Stride.BepuPhysics.dll!Stride.BepuPhysics.BepuSimulation.GetComponent(BepuPhysics.BodyHandle handle) Line 298	C#
 	Stride.BepuPhysics.dll!Stride.BepuPhysics.BepuSimulation.GetComponent(BepuPhysics.Collidables.CollidableReference collidable) Line 291	C#
 	Stride.BepuPhysics.dll!Stride.BepuPhysics.CharacterComponent.OnStoppedTouching<Stride.BepuPhysics.Definitions.Contacts.ContactEventsManager.EmptyManifold>(BepuPhysics.Collidables.CollidableReference eventSource, BepuPhysics.Collidables.CollidableReference other, ref Stride.BepuPhysics.Definitions.Contacts.ContactEventsManager.EmptyManifold contactManifold, int contactIndex, Stride.BepuPhysics.BepuSimulation bepuSimulation) Line 158	C#
 	Stride.BepuPhysics.dll!Stride.BepuPhysics.CharacterComponent.OnStoppedTouching<Stride.BepuPhysics.Definitions.Contacts.ContactEventsManager.EmptyManifold>(BepuPhysics.Collidables.CollidableReference eventSource, BepuPhysics.Collidables.CollidableReference other, ref Stride.BepuPhysics.Definitions.Contacts.ContactEventsManager.EmptyManifold contactManifold, int contactIndex, Stride.BepuPhysics.BepuSimulation bepuSimulation) Line 143	C#
 	Stride.BepuPhysics.dll!Stride.BepuPhysics.Definitions.Contacts.ContactEventsManager.Flush() Line 433	C#
 	Stride.BepuPhysics.dll!Stride.BepuPhysics.BepuSimulation.Update(System.TimeSpan elapsed) Line 684	C#
 	Stride.BepuPhysics.dll!Stride.BepuPhysics.Systems.PhysicsGameSystem.Update(Stride.Games.GameTime time) Line 35	C#
 	[External Code]	
 	ImmoApocalypse.Windows.dll!Program.<Main>$(string[] args) Line 4	C#

```

This is actually not a new error either I just forgot the cause. It happens sometimes when the CharacterComponent stops touching an entity (probably because its gone from the scene). We used to have an if statement in the old Nicogo Bepu repo but it looks like this shouldn't cause issues with how the check is being done in the `CharacterComponent.OnStoppedTouching`.

The only thing Im unsure about is why the `Bodies` keeps entities that were removed from the scene? Maybe it makes sense if the Entities are added back to the scene?

Doprez:
I think I fixed the character controller properly. It was just a lingering reference to a disposed component that wasnt even needed anyway. I also cleaned up a few of my comments from the review.
https://github.com/Eideren/xenko/pull/4

The last thing I will mention is I cant successfully build the sln due to the demo project of Bepu. I think we talked about this before with how annoying the package referencing was in Stride? I get the below errors due to "missing" references:
```
Severity	Code	Description	Project	File	Line	Suppression State	Details
Error (active)	NU1101	Unable to find package Stride.BepuPhysics.Debug. No packages exist with this id in source(s): nuget.org. PackageSourceMapping is enabled, the following source(s) were not considered: C:\Program Files\dotnet\library-packs, Local nugets, Microsoft Visual Studio Offline Packages, Stride, Stride Dev, vvvv.	Stride.BepuPhysics.Demo (net8.0)	C:\dev\GitControlledProjects\stride\sources\engine\Stride.BepuPhysics\Stride.BepuPhysics.Demo\Stride.BepuPhysics.Demo.csproj	1		
Error (active)	NU1101	Unable to find package Stride.BepuPhysics._2D. No packages exist with this id in source(s): nuget.org. PackageSourceMapping is enabled, the following source(s) were not considered: C:\Program Files\dotnet\library-packs, Local nugets, Microsoft Visual Studio Offline Packages, Stride, Stride Dev, vvvv.	Stride.BepuPhysics.Demo (net8.0)	C:\dev\GitControlledProjects\stride\sources\engine\Stride.BepuPhysics\Stride.BepuPhysics.Demo\Stride.BepuPhysics.Demo.csproj	1		
Error (active)	NU1101	Unable to find package Stride.BepuPhysics.Debug. No packages exist with this id in source(s): nuget.org. PackageSourceMapping is enabled, the following source(s) were not considered: C:\Program Files\dotnet\library-packs, Local nugets, Microsoft Visual Studio Offline Packages, Stride, Stride Dev, vvvv.	Stride.BepuPhysics.Demo.Windows	C:\dev\GitControlledProjects\stride\sources\engine\Stride.BepuPhysics\Stride.BepuPhysics.Demo.Windows\Stride.BepuPhysics.Demo.Windows.csproj	1		
Error (active)	NU1101	Unable to find package Stride.BepuPhysics._2D. No packages exist with this id in source(s): nuget.org. PackageSourceMapping is enabled, the following source(s) were not considered: C:\Program Files\dotnet\library-packs, Local nugets, Microsoft Visual Studio Offline Packages, Stride, Stride Dev, vvvv.	Stride.BepuPhysics.Demo.Windows	C:\dev\GitControlledProjects\stride\sources\engine\Stride.BepuPhysics\Stride.BepuPhysics.Demo.Windows\Stride.BepuPhysics.Demo.Windows.csproj	1		
```

Maybe we should just remove those projects from the main build sln for now?

sources/editor/Stride.GameStudio/Debugging/AssemblyRecompiler.cs
```diff
@@ -197,6 +197,10 @@ public async Task<UpdateResult> Recompile(Project gameProject, LoggerResult logg
                             {
                                 assemblyProcessorApp.ReferencesToAdd.Add(referencePath.EvaluatedInclude);
                             }
+                            else if (referencePath.EvaluatedInclude.EndsWith("Stride.BepuPhysics.dll")) //todo hard-coded! needs to go when plug in system is in
```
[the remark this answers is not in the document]
Eideren: This runs when reloading the assemblies inside the game studio, when you've made changes to your source while the application is open

sources/engine/Stride.BepuPhysics/.editorconfig
```diff
@@ -0,0 +1,190 @@
+# editorconfig.org
```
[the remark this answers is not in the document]
Eideren: As you know, the project started off as a repo outside of Stride and is now part of it. The style it started off on differs from stride's so I added this editor config in to ensure users contributing to those specific projects adhere to the existing style instead of Stride's.
This is likely fairly controversial, I can remove it if other contributors thinks this is a bad idea, but I would then have to refactor the code to adhere to Stride's style which will pollute the git history.

sources/engine/Stride.BepuPhysics/Stride.BepuPhysics/BepuSimulation.cs
```diff
@@ -0,0 +1,829 @@
...
+
+    /// <summary>
+    /// The scene associated with this simulation.
+    /// </summary>
+    /// <remarks>
+    /// When this is set, entities spawning inside this scene will be automatically associated with
+    /// this simulation as long as their <see cref="CollidableComponent.SimulationSelector"/> is set to <see cref="SceneBasedSimulationSelector"/>.
+    /// See <see cref="SceneBasedSimulationSelector"/> for more info.
+    /// </remarks>
+    public UrlReference<Scene> AssociatedScene { get; init; }
```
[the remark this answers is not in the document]
Eideren: You can have multiple subscenes to the scene specified there and objects inside those subscenes will be assigned to the simulation of their parent scene. 
To me the 1:1 relationship between a scene and the physics simulation makes it easier to reason about; whenever debugging, you would know for a fact that all objects contained in `sceneA` won't interact with `sceneB`. As a user, you can always configure the association however you want by creating a new implementation of `ISimulationSelector` or using the `IndexBasedSimulationSelector` on your bodies.

sources/engine/Stride.BepuPhysics/Stride.BepuPhysics/Systems/ShapeCacheSystem.cs
Eideren: I do want to make this public, but it still needs a lot of work I feel - I could add an `InternalsVisibleTo` for navigation like we've done here https://github.com/stride3d/stride/blob/2a6f6f219b65ff28db220c86f448fab1605c4171/sources/engine/Stride.BepuPhysics/Stride.BepuPhysics/Stride.BepuPhysics.csproj#L19
Although the navigation we have in this PR does not require access to it ?

sources/engine/Stride.BepuPhysics/Stride.BepuPhysics/Definitions/UnsortedO1List.cs
Eideren: This is part of the `CollidableProcessor` for processing statics https://github.com/stride3d/stride/blob/2a6f6f219b65ff28db220c86f448fab1605c4171/sources/engine/Stride.BepuPhysics/Stride.BepuPhysics/Systems/CollidableProcessor.cs#L17-L23
Compared to using a list, removal of item is 01, so significantly faster. Adding is like a list, 01, while being slightly slower.
Compared to using a dictionary, removal and adding is slightly slower, but sequential reads are faster since you can grab its span.
Might be a bit niche though which is why it's internal.

sources/engine/Stride.BepuPhysics/Stride.BepuPhysics/Systems/ShapeCacheSystem.cs
Doprez: https://github.com/Eideren/xenko/blob/2a6f6f219b65ff28db220c86f448fab1605c4171/sources/engine/Stride.BepuPhysics/Stride.BepuPhysics.Navigation/Processors/RecastMeshProcessor.cs#L91

oh but there is a comment saying we dont need not too? hmm I can take a look at this.

Eideren:
@Doprez I removed the demo project from the main solution, it couldn't be built from there anyway. This will take care of the missing packages issue.
As for the character issue it is fixed now.
I also created a test to cover the raycast issue you mentioned earlier but could not repro it in the current state. If you do find it again let me know.

Doprez:
> @Doprez I removed the demo project from the main solution, it couldn't be built from there anyway. This will take care of the missing packages issue. As for the character issue it is fixed now. I also created a test to cover the raycast issue you mentioned earlier but could not repro it in the current state. If you do find it again let me know.

Sounds good! The issue wasnt actually the raycast, it was the fact that the CharacterComponent kept a reference to a disposed object and then tried to look for that reference in the contact removed handler. Ill mess around with your changes and test it out. It was this line that caused it https://github.com/Eideren/xenko/blob/cc05d516d22e0d49b6c1bdc8d29f40520bb4f9eb/sources/engine/Stride.BepuPhysics/Stride.BepuPhysics/CharacterComponent.cs#L158 since the component that was supposed to be returned was removed when the scene is set to null.

I can also create the situation that caused the problem in a repro later it was pretty easy to recreate once I figured out where it was happening.

Doprez:
I just tested the issues I was having with removing objects touching the character component and I cant seem to recreate it anymore!

Doprez:
One thing to mention about the character controller changes. It only jumps while moving, when standing still it does not seem to jump at all for some reason. I can look into it in a bit but maybe you already know why.

Eideren:
@Doprez I can't repro in the characterscene demo, what's your setup ?

Doprez:
Maybe its my PlayerMover class then? here is my jump code:

```cs
	public override void Update(float deltaTime)
	{
		if (_queueJump.TryReceive()) QueueJump();

		Jump();
        }

	private void QueueJump()
	{
		// Jump is queued to allow for a small time frame to press the jump button.
		// This is to allow for a more responsive jump.
		// I may want to base the timer on when the isGrouded state is true so that the player can jump ledges without needing to be grounded.
		_jumpQueued = true;
		_jumpTimeFrame.Start();
	}

	/// <summary>
	/// makes the player jump.
	/// <para>power is determined by the JumpPower variable and uses the CharacterComponent.IsGrounded variable to validate if player can jump.</para>
	/// </summary>
	public void Jump()
	{
		if (!_isCrouched && _jumpQueued && _jumpTimeFrame.ElapsedMilliseconds < _jumpTimeLeniencyMilliseconds)
		{
			_jumpTimeFrame.Reset();
			_jumpQueued = false;
			_character.JumpForce = JumpPower;
			_character.TryJump();
		}
	}
```

And it gets the input for jump like this: 
```cs
			if (Input.IsKeyPressed(key))
			{
				JumpEvent.Broadcast();
			}
```

Ill try out the demo character later this one was mostly taken from the Bullet demo so maybe I mixed something that was wrong.

Doprez:
Sorry, I meant to get back sooner for the CharacterController issue I was having. It does seem to be fine when using the demo code so I must have had something weird with my previous implementation.

PR #2133, opened 2024-01-28
[Tests] Fixes random test failures

azeno:
# PR Details

## Description

Some of the tests in Stride.GameStudio.Tests must not run in parallel as they access a shared class which is not thread safe.

## Types of changes

## Checklist

### DIFF
--- a/sources/assets/Stride.Core.Assets.Tests/TestFileVersionManager.cs
+++ b/sources/assets/Stride.Core.Assets.Tests/TestFileVersionManager.cs
@@ -55,7 +55,7 @@ namespace Stride.Core.Assets.Tests
 
             var ids = new List<Tuple<UFile, ObjectId>>();
             FileVersionManager.Instance.ComputeFileHashAsync(files, (file, id) => ids.Add(new Tuple<UFile, ObjectId>(file, id)));
-            Thread.Sleep(200);
+            Thread.Sleep(500);
 
             Assert.Equal(files.Count, ids.Count);
 
--- a/sources/editor/Stride.GameStudio.Tests/TestCopyPasteComponents.cs
+++ b/sources/editor/Stride.GameStudio.Tests/TestCopyPasteComponents.cs
@@ -15,6 +15,8 @@ using Stride.GameStudio.Tests.Helpers;
 
 namespace Stride.GameStudio.Tests
 {
+    // AssetQuantumRegistry is currently not thread safe. Any unit tests accessing it must not run in parallel.
+    [Collection("TestCopyPaste")]
     public sealed class TestCopyPasteProperties
     {
         private ICopyPasteService service;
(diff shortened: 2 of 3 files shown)


sources/editor/Stride.GameStudio.Tests/XunitAttributes.cs
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net) and Silicon Studio Corp. (https://www.siliconstudio.co.jp)
+// Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using Xunit;
+
+[assembly: CollectionBehavior(DisableTestParallelization = true)]
```
Kryptos-FR, line 6: I would rather not have that for now. It might or might not be needed, but let's isolate fixes. Otherwise, we can't determine what fixed what.

Kryptos-FR (changes requested):

sources/editor/Stride.GameStudio.Tests/XunitAttributes.cs
azeno: How about using the [CollectionAttribute](https://xunit.net/docs/running-tests-in-parallel.html) on `TestCopyPasteProperties` and `TestCopyPasteWithEntities` to mark those three tests which must not run in parallel?

sources/assets/Stride.Core.Assets.Tests/TestFileVersionManager.cs
```diff
@@ -54,8 +55,14 @@ public void TestAsync()
...
             var ids = new List<Tuple<UFile, ObjectId>>();
-            FileVersionManager.Instance.ComputeFileHashAsync(files, (file, id) => ids.Add(new Tuple<UFile, ObjectId>(file, id)));
-            Thread.Sleep(200);
+            var tcs = new TaskCompletionSource();
+            FileVersionManager.Instance.ComputeFileHashAsync(files, (file, id) =>
+            {
+                ids.Add(new Tuple<UFile, ObjectId>(file, id));
+                if (ids.Count == files.Count)
+                    tcs.SetResult();
+            });
```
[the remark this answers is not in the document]
azeno: I see - in that case should I revert it? And simply increase the Thread.Sleep to say 500?

sources/editor/Stride.GameStudio.Tests/XunitAttributes.cs
Kryptos-FR: If that helps passing the tests without changing too much the delay, that might be a good alternative indeed.

azeno:
@Kryptos-FR I've updated the PR.

azeno:
@Kryptos-FR Done

Kryptos-FR:
Looks like the tests passed fine. Thanks.

PR #2141, opened 2024-02-06
[OpenXR] Adds a minimal API to request and control Passthrough

azeno:
# PR Details

## Description

Follow up on #2119 and #2121. It adds the following new public APIs:
- `VRDeviceSystem.RequestPassthrough` - needs to be set in order for the OpenXR extension to get considered.
- `VRRendererSettings.RequestPassthrough` - can be set in GameStudio and get passed to the VR device system by the `ForwardRenderer`
- `VRDevice.SupportsPassthrough` - does the device support it?
- `VRDevice.StartPassthrough` - needs to be called by user code to turn the feature on

The PR is a little bit different to the original one not exposing internal classes to the user. Not entirly sure here what is the best approach. It's only supported by OpenXR and there also by some devices, but having to cast on these types manually in Stride game code seems awkward as well.

## Related Issue

#2119 and #2121 

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.VirtualReality/OpenXR/OpenXRHmd.cs
+++ b/sources/engine/Stride.VirtualReality/OpenXR/OpenXRHmd.cs
@@ -20,7 +20,11 @@ namespace Stride.VirtualReality
         // Public static variable to add extensions to the initialization of the openXR session
         public static List<string> extensions = new List<string>();
 
-        public static OpenXRHmd? New(bool requestPassthrough)
+        /// <summary>
+        /// Creates a VR device using OpenXR.
+        /// </summary>
+        /// <param name="requestPassthrough">Whether or not the XR_FB_passthrough extension should be enabled (if available).</param>
+        internal static OpenXRHmd? New(bool requestPassthrough)
         {
             // Create our API object for OpenXR.
             var xr = XR.GetApi();
@@ -62,6 +66,7 @@ namespace Stride.VirtualReality
 
         // Passthrough
         private OpenXRExt_FB_Passthrough? passthroughExt;
+        private bool passthroughRequested;
 
         private GraphicsDevice? baseDevice;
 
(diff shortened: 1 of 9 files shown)


sources/engine/Stride.VirtualReality/OpenXR/OpenXRHmd.cs
```diff
@@ -19,7 +20,7 @@ public class OpenXRHmd : VRDevice
         // Public static variable to add extensions to the initialization of the openXR session
         public static List<string> extensions = new List<string>();
 
-        public static OpenXRHmd? New()
+        public static OpenXRHmd? New(bool requestPassthrough)
```
manio143, line 23: Since this is public API can you add an XML comment to this method explaining the parameter?

sources/engine/Stride.VirtualReality/OpenXR/OpenXRHmd.cs
```diff
@@ -348,6 +359,22 @@ public override unsafe void Enable(GraphicsDevice device, GraphicsDeviceManager
             Xr.StringToPath(Instance, "/user/hand/left", ref leftHandPath);
         }
 
+        public override IDisposable StartPassthrough()
+        {
+            if (!SupportsPassthrough)
+                throw new NotSupportedException();
```
manio143, line 365: Can you add a message to the exception explaining "The device does not support Passthrough mode"?

manio143:
Overall looks good. The bool property indicating support is great as it clearly indicates to the user if they can enable this passthrough mode on the given device/VR system.

sources/engine/Stride.VirtualReality/OpenXR/OpenXRHmd.cs
azeno: Done, also added a new exception in case passthrough mode wasn't requested at all.

joreg:
@manio143 do you think there is now anything left holding this back from being merged?

PR #2144, opened 2024-02-07
[Assets] Prevent crash of the assets compiler when an assembly cannot be fully loaded.

### LINKED ISSUE #2140: Asset compiler crashes when using `System.Windows.Forms`
**Release Type**: GitHub

**Version**: main branch

**Platform(s)**: Windows

**Describe the bug**
We have troubles building our solution with latest Stride source builds as the asset compiler crashes during the build. This crash wasn't present before and seems to be have introduced by https://github.com/stride3d/stride/pull/2082.

**To Reproduce**
1. Start Game Studio from current main branch
2. Create a new game using Windows as platform
3. Add a public class which inherits from `System.Windows.Forms.Control`, for example 
```
using System.Windows.Forms;

namespace GameUsingWindowsForms
{
    public class SomeControl : Control
    {
    }
}
```
4. Build the game

**Expected behavior**
It shouldn't crash.

**Screenshots**
image

Kryptos-FR:
# PR Details

## Description

Skip looking for asset compilers in an assembly if its types can't be loaded.

## Related Issue

Fixes #2140.

## Types of changes

## Checklist

### DIFF
--- a/sources/assets/Stride.Core.Assets/Compiler/AssetCompilerRegistry.cs
+++ b/sources/assets/Stride.Core.Assets/Compiler/AssetCompilerRegistry.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Reflection;
 using Stride.Core;
@@ -181,7 +182,7 @@ namespace Stride.Core.Assets.Compiler
         private void RegisterCompilersFromAssembly(Assembly assembly)
         {
             // Process Asset types.
-            foreach (var type in assembly.GetTypes())
+            foreach (var type in GetFullyLoadedTypes(assembly))
             {
                 // Only process Asset types
                 if (!typeof(IAssetCompiler).IsAssignableFrom(type) || !type.IsClass)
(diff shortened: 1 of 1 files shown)


azeno:
As an alternative we could also introduce an assembly attribute telling the asset compiler whether or not the assembly even contains such implementations.

We would need to add that attribute to the existing code base (or even better let it get added by the assembly processor).

Not sure about the proper name yet, but for example 
```
[assembly: ContainsAssetCompilers]
```

Kryptos-FR:
> As an alternative we could also introduce an assembly attribute telling the asset compiler whether or not the assembly even contains such implementations.

It's already based on the registry which only takes assembly from the assets category. We don't scan every single assembly and their dependencies, that would be overkill.

But the Game project is such an assembly since it does contain assets. It's just not supposed to have platform-specific dependencies such as Windows Forms. Your case is uncommon as a cleaner architecture would have all Windows dependencies on the Windows-specific project. The Game project is supposed to be cross-platform.

azeno:
Ok thanks. Makes me wonder why many of our projects would even be put in the assets category then. We have a dependency on Stride.Core.Mathematics very deep down. I guess that's the reason why the assembly processor even comes into play. Maybe there're some tweaks to ensure it doesn't run for those projects (setting `StrideAssemblyProcessor` to false for example or setting `ExcludeAssets = "buildTransitive"`), will need to check.
But in any case, at some point we will run into a project with indeed contains Stride assets and has an (unfortunate) dependency on System.Windows.Forms. Therefor I'd be happy with the proposed workaround.

azeno:
Any chance this PR can be merged? It's currently blocking things at our end regarding a project which needs https://github.com/stride3d/stride/pull/2141

PR #2167, opened 2024-02-29
Rework asset editors dependencies

Kryptos-FR:
# PR Details

## Description

Rework the relationship between assets and their editing counter parts.

The idea is to break the dependencies between what can be cross-platform (e.g. view models, previews) and what is UI-specific (editor views, preview views). On top of that, asset view model where directly dependent on their corresponding editor view model. This didn't make sense from an architecture point of view: asset view models should exist outside of an editing context ; they could for instance be reused to write an stand-alone previewer app. On the other hand, the editor view model knows which asset view model it is editing so a dependency in that direction makes sense.

## Related Issue

- #2034
- #2082

## Motivation and Context

As part of the editor rewrite, since it involves touching a lot of files, it is better to merge it into *master* before starting the biggest part of the rewrite. Otherwise, we could have a lot of conflicts as the `xplat-editor` branch will diverge from *master*. With that said, from this point we should only allow minor changes to the editor such as bug fixes, and make sure they are reported periodically to the `xplat-editor` branch.

## Types of changes

  - *Can't be sure that nothing is breaking, needs testing.*
  - Runtime untouched, only editor changes.

## Checklist


Eideren:
Looks like sprite sheets are broken, opening ParticleButtons in the UI particle sample opens the ui for a split second then closes it

Kryptos-FR:
> Looks like sprite sheets are broken, opening ParticleButtons in the UI particle sample opens the ui for a split second then closes it

That's a nice bug actually. For some reason, we were returning `false` in the `Initialize()` method for the `SpriteSheetEditorViewModel`.
https://github.com/stride3d/stride/blob/9cc2e6be443680ea3bf6cf5f6b86ef43d2abdc81/sources/editor/Stride.Assets.Presentation/AssetEditors/SpriteEditor/ViewModels/SpriteSheetEditorViewModel.cs#L180-L185
But so far, we didn't actually care about the boolean value. Returning `true` now obviously fixes it.

Eideren:
There's this screen that shows up after double clicking on a scene, right before it's loaded in and replaced by the scene view
image
In the previous version this screen would be showing up instead:
image
So it sounds like that loading screen never runs in your latest ?

Issue #2024, opened 2023-11-02
Exception when entity with two references to the same sound is assigned a parent


ComputerSmoke:
**Release Type**: Official Release

**Version**: 4.1

**Platform(s)**: Windows

**Describe the bug**
If an entity has an audio component with two keys that reference the same asset, an exception is thrown when that entity (or its parent) is assigned a parent in a script.

**To Reproduce**
1. make entities A and B
2. Add audio emitter component to entity A
3. Import some spatial sound
4. Add two key/value pairs to the audio emitter, each of which reference the imported sound
5. In a script, assign entity A to be a child of entity B.

**Expected behavior**
Not throw an exception, and assign the child as normal.

PR #2176, opened 2024-03-05
Audio emitter multiple references to same asset bugfix

ComputerSmoke:
This fixes the bug where changing the parent of an entity with an audio emitter component referencing the same sound asset multiple times would cause a crash.

# PR Details

Previously when the parent of such an entity was changed, there would be a crash because detach would be called on the same SoundBase object for each corresponding key in Sounds. Now, when an entry is removed from Sounds in AudioEmitterComponent, the corresponding SoundBase is only detached if there are no other keys referencing it in Sounds.

## Description

This is accomplished by maintaining a dictionary mapping each SoundBase to its set of keys in Sounds, and only detaching a SoundBase when its set of Sound keys is empty. This implementation still keeps just one AudioEmitterSoundController per SoundBase.

## Related Issue

https://github.com/stride3d/stride/issues/2024

## Motivation and Context

Users may occasionally want the same sound asset to have multiple references from the same audio emitter (especially during early game development), so Stride should allow it instead of crashing unexpectedly when a parent of the object is set during runtime.

## Types of changes

- [ x ] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [ x ] **I have built and run the editor to try this change out.**

### DIFF
--- a/sources/engine/Stride.Engine/Engine/AudioEmitterComponent.cs
+++ b/sources/engine/Stride.Engine/Engine/AudioEmitterComponent.cs
@@ -40,9 +40,9 @@ namespace Stride.Engine
         /// </summary>
         internal readonly Dictionary<SoundBase, AudioEmitterSoundController> SoundToController = new Dictionary<SoundBase, AudioEmitterSoundController>();
         /// <summary>
-        /// Dictionary associating each SoundBase to its names in Sounds.
+        /// Number of keys in Sounds pointing to each sound
         /// </summary>
-        internal readonly Dictionary<SoundBase, HashSet<string>> SoundKeys = [];
+        internal readonly Dictionary<SoundBase, int> SoundMultiplicities = [];
 
         /// <summary>
         /// Event argument class used to signal the <see cref="AudioEmitterProcessor"/> that a new AudioEmitterSoundController has new added or removed to the component.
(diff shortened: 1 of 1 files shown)


Eideren:
Looks good, thanks again @ComputerSmoke ! Just a small question, it looks like you've used a hashset purposefully here, can you quickly go over why that is compared to a `Dictionary<SoundBase, int>` which would be incremented on add and decremented on remove regardless of the key, and would call detach when zero ?
I can see the behavior of both differing when the same key and sound is used multiple times.

One minor nitpick would be to remove the key after detaching;
```cs
if (SoundKeys[sound].Count == 0)
{
   DetachSound(sound);
   SoundKeys.Remove(sound);
}
```

ComputerSmoke:
@Eideren I thought it was possible for the same key / sound to have two Add events in a row, but on investigating this I see that it is not (setting an existing key calls Remove before Add, and you can't call Add on a key that is already in the set). Counting instead of keeping track of each key should therefore have the same effect as the hash set, so I've changed the implementation to use the counting method you described.

I've also added that key removal for 0 values.

Kryptos-FR:
Wondering if that needs thread-safety guarantees (for instance using a lock). Can this code be called from multiple threads or are all sound-related logic always executed from the same (main?) thread?

ComputerSmoke:
The user could modify Sounds from any thread, but since it's a TrackingDictionary (which isn't thread safe since it's backed by a Dictionary), so anything modifying it in the current code must already be responsible for avoiding concurrent changes to Sounds.

The other place where this code is called is AttachToProcessor and DetachFromProcessor, but both call the Attach and Detach methods which can modify SoundToController, another Dictionary. 

Assuming Attach/Detach for components doesn't collide, the case where something might go wrong is if user code modifies Sounds or calls Attach/Detach while AttachToProcessor is being called. I don't know if such a situation is possible, but if it is then it should be equally unsafe with and without these changes.

Not that we shouldn't add thread safety to anything, but there might already be some more generally implemented for this sort of thing?

Eideren:
Yep it would be nice, but adding thread safety in would be out of scope for this PR I would think, thanks again @ComputerSmoke !

PR #2183, opened 2024-03-10
Overwrite dialog fix

timconner:
# PR Details

Changes asset file import overwrite dialog to show the correct file being replaced.

## Description

There is a bug in the current dialog that prompts if it is okay to overwrite the source file instead of the destination file. This PR corrects that so that it shows the correct file that is about to be replaced.

Additionally, added Path.GetFullPath to use/show the proper directory separator in an OS agnostic way as seen in other places in the project. 

## Related Issue

No issue created, but can easily be reproduced by importing an external resource twice and choosing to use the default location.

## Motivation and Context

The current dialog makes it seem as if Stride is going to modify a file outside of the project, which is not true. This should help eliminate that confusion.

## Types of changes

## Checklist


timconner:
@dotnet-policy-service agree

tebjan:
Very nice, thanks!

PR #2186, opened 2024-03-12
Drop support for DirectX 9 / OpenGL 2.0

Jklawreszuk:
# PR Details

## Description

This is my attempt to remove DirectX 9 / OpenGL 2.0 (called Level 9) from Stride and set Level 10 as default instead.

## Motivation

This is change was inspired by @ykafia (check out his discussion here https://github.com/stride3d/stride/discussions/2185) and based on my personal knowledge.

Stride Engine primarly focus currently on supporting desktop and mobile platforms (but I heard isn't that good tbh), so it would be relevant to check out current state of graphics api distribution among players:

### Android

image

**Source:** https://developer.android.com/about/dashboards/index.html

As we can see, more than 93% of devices support OpenGL 3.0 or later

### Desktop

Since Steam platform is dominant, I think it is also reliable source
 
image

As can be observed, DirectX 9 marketshare is basically irrelevant ( almost 97 % of devices are supporting DX10 or later)
Interestingly, Directx 12 alone supports about 90% of the all desktop computers.

**Source:** https://store.steampowered.com/hwsurvey

## Types of changes

## Checklist


ykafia:
We should probably make sure we have WebGPU native working as a replacement, wgpu-native has been widely tested to work well on android devices and it's using opengl ES and Vulkan under the hood

MeharDT:
My two cents: The DefaultGraphicsProfile changes are fine but I think removing support entirely might be more appropriate for a bigger release (i.e., 4.3) instead of a minor 4.2.x update. It would be easier to communicate and projects that still need feature level 9.x or ES 2.0 can stay on 4.2.x without worry.

Overall it makes sense to streamline supported feature levels. Unity and Unreal dropped support for feature level 9.x about a decade ago, and I believe only the former still supports OpenGL ES 2.

IIRC, feature level 10/OpenGL ES 3.0 work on Android but the Silk implementation on mobile needs improvement. I think we can treat these as two separate and independent issues.

Eideren:
Moved this one to Next Major, but moved it to draft while you figure out those conflicts/whether you still want to work on this

xen2:
👋 About this PR.

Ethereal77:
I think this PR is done already? DX9 were already removed by me in the Silk.NET PR, and OpenGL was entirely removed later by xen2 when he refactored the renderer to better support DX12 & Vulkan. Isn't it?

PR #2188, opened 2024-03-15
fix: Update README.md

Jklawreszuk:
# PR Details

## Description

These projects do not exist, but also I did not found information about those in the past (xenko engine)


Eideren:
Thanks !

PR #2195, opened 2024-03-21
[Presentation] Fix issue with binding quantum nodes when associated name is not found

### LINKED ISSUE #2194: Opening Settings causes Game Studio to crash
**Release Type**: **Official Release**/~~GitHub~~ (please choose appropriate option)

**Version**: Stride 4.2.0.2121 + .NET 8.0.3 + VS 17.9.4 (2022) + Windows 11 10.0.23606.1000 (Insider build, Dev branch) 

**Platform(s)**: Windows, in Game Studio

**Describe the bug**
When using the basic "New game" template (the one with ground and a sphere where you can fly and look around), two issues can be observed:
- The built-in code editor highlights errors at all namespace `using` declarations (both those under `Stride.somethingsomething` and those that are parts of .NET, like `System`). VS 2022 (w/ extension) shows no such errors on the exact same solution, and the project builds properly.
- Opening settings (Edit > Settings) shows an empty settings window and crashes the editor app.

**To Reproduce**
1. Create a new project with the "New game" template.
2. Try to open settings.

**Expected behavior**
Settings window is opened and usable.

**Screenshots**
stridecrash

**Log and callstacks**
Stride-generated crash report: [crashreport.txt](https://github.com/stride3d/stride/files/14692000/crashreport.txt)

**Additional context**
Probably related to #2192 but opened new issue to add more context.

Kryptos-FR:
Fixes #2194

Regression was introduced by #2167 (more precisely commit b389a6f53cabfcd4eb95672d0acb0e861512fdb5).

### DIFF
--- a/sources/editor/Stride.Core.Assets.Editor/Module.cs
+++ b/sources/editor/Stride.Core.Assets.Editor/Module.cs
@@ -17,6 +17,8 @@ namespace Stride.Core.Assets.Editor
             AssetsPlugin.RegisterPlugin(typeof(CoreAssetsEditorPlugin));
             // Initialize translation
             TranslationManager.Instance.RegisterProvider(new GettextTranslationProvider());
+            // HACK
+            Presentation.Quantum.ViewModels.NodeViewModel.UnsetValue = System.Windows.DependencyProperty.UnsetValue;
         }
     }
 }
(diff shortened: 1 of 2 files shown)


Eideren:
I'm guessing this is a specific wpf workaround which will be removed once we move over to Avalonia ?

Kryptos-FR:
Yes and no. Avalonia has its own equivalent [UnsetValue](https://reference.avaloniaui.net/api/Avalonia/AvaloniaProperty/A6EB5667).

I'm not sure if that will be necessary, though. In case it is, I can rework it with a dedicated UnsetValueProvider that will act as a service instead of the current hack. Then WPF or Avalonia can  implement that service (or register the value to it).

PR #2205, opened 2024-03-31
Fix sample projects, remove unused properties

Jklawreszuk:
# PR Details

## Description

Current samples are using incorrect net8 target framework moniker (windows 7 is no longer supported). I also decided to remove additional bloat from project files

## Types of changes

- [x] Refactoring / dependency upgrade

## Checklist

### DIFF
--- a/samples/Audio/SimpleAudio/SimpleAudio.Windows/SimpleAudio.Windows.csproj
+++ b/samples/Audio/SimpleAudio/SimpleAudio.Windows/SimpleAudio.Windows.csproj
@@ -1,58 +1,17 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>net8.0-windows7.0</TargetFramework>
+    <TargetFramework>net8.0-windows</TargetFramework>
     <ApplicationIcon>Resources\GameIcon.ico</ApplicationIcon>
     <OutputType>WinExe</OutputType>
     <RootNamespace>SimpleAudio</RootNamespace>
     <OutputPath>..\Bin\Windows\$(Configuration)\</OutputPath>
     <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
+    <DefineConstants>STRIDE_PLATFORM_DESKTOP</DefineConstants>
   </PropertyGroup>
   <PropertyGroup>
     <StrideCurrentPackagePath>$(MSBuildThisFileDirectory)..\SimpleAudio.sdpkg</StrideCurrentPackagePath>
     <StrideProjectType>Executable</StrideProjectType>
     <StridePlatform>Windows</StridePlatform>
-    <StrideBuildProfile>Windows</StrideBuildProfile>
-  </PropertyGroup>
-  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
-    <DebugSymbols>true</DebugSymbols>
-    <DebugType>full</DebugType>
-    <Optimize>false</Optimize>
-    <DefineConstants>DEBUG;TRACE;STRIDE_PLATFORM_WINDOWS;STRIDE_PLATFORM_WINDOWS_DESKTOP</DefineConstants>
-    <OutputPath>..\Bin\Windows\Debug\</OutputPath>
-    <IntermediateOutputPath>obj\Windows\Debug\</IntermediateOutputPath>
-    <ErrorReport>prompt</ErrorReport>
-    <WarningLevel>4</WarningLevel>
-    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
-  </PropertyGroup>
-  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
-    <DebugType>pdbonly</DebugType>
-    <Optimize>true</Optimize>
-    <DefineConstants>TRACE;STRIDE_PLATFORM_WINDOWS;STRIDE_PLATFORM_WINDOWS_DESKTOP</DefineConstants>
-    <OutputPath>..\Bin\Windows\Release\</OutputPath>
-    <IntermediateOutputPath>obj\Windows\Release\</IntermediateOutputPath>
-    <ErrorReport>prompt</ErrorReport>
-    <WarningLevel>4</WarningLevel>
-    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
-  </PropertyGroup>
-  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Testing|AnyCPU' ">
-    <DebugType>pdbonly</DebugType>
-    <Optimize>true</Optimize>
-    <DefineConstants>TRACE;STRIDE_PLATFORM_WINDOWS;STRIDE_PLATFORM_WINDOWS_DESKTOP</DefineConstants>
-    <OutputPath>..\Bin\Windows\Testing\</OutputPath>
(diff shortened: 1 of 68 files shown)


samples/Audio/SimpleAudio/SimpleAudio.Windows/SimpleAudio.Windows.csproj
```diff
@@ -1,58 +1,14 @@
...
   <PropertyGroup>
-    <TargetFramework>net8.0-windows7.0</TargetFramework>
+    <TargetFramework>net8.0-windows</TargetFramework>
     <ApplicationIcon>Resources\GameIcon.ico</ApplicationIcon>
     <OutputType>WinExe</OutputType>
     <RootNamespace>SimpleAudio</RootNamespace>
     <OutputPath>..\Bin\Windows\$(Configuration)\</OutputPath>
     <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
-  </PropertyGroup>
-  <PropertyGroup>
```
Kryptos-FR, line 10: This helps separate Stride properties from standard properties.

samples/Audio/SimpleAudio/SimpleAudio.Windows/SimpleAudio.Windows.csproj
```diff
@@ -1,58 +1,14 @@
...
     <StrideCurrentPackagePath>$(MSBuildThisFileDirectory)..\SimpleAudio.sdpkg</StrideCurrentPackagePath>
     <StrideProjectType>Executable</StrideProjectType>
     <StridePlatform>Windows</StridePlatform>
-    <StrideBuildProfile>Windows</StrideBuildProfile>
-  </PropertyGroup>
-  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
-    <DebugSymbols>true</DebugSymbols>
-    <DebugType>full</DebugType>
-    <Optimize>false</Optimize>
-    <DefineConstants>DEBUG;TRACE;STRIDE_PLATFORM_WINDOWS;STRIDE_PLATFORM_WINDOWS_DESKTOP</DefineConstants>
```
Kryptos-FR, line 20: Are we sure the Stride constants don't need to be set?

samples/Audio/SimpleAudio/SimpleAudio.Windows/SimpleAudio.Windows.csproj
```diff
@@ -1,58 +1,14 @@
...
     <OutputType>WinExe</OutputType>
     <RootNamespace>SimpleAudio</RootNamespace>
     <OutputPath>..\Bin\Windows\$(Configuration)\</OutputPath>
     <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
-  </PropertyGroup>
-  <PropertyGroup>
     <StrideCurrentPackagePath>$(MSBuildThisFileDirectory)..\SimpleAudio.sdpkg</StrideCurrentPackagePath>
     <StrideProjectType>Executable</StrideProjectType>
     <StridePlatform>Windows</StridePlatform>
-    <StrideBuildProfile>Windows</StrideBuildProfile>
```
Kryptos-FR, line 14: What was the use of this property before? Could it have been used by TeamCity when running the samples as tests?

Kryptos-FR (commented):
Added comments to a single file, but it applies to all.

samples/Audio/SimpleAudio/SimpleAudio.Windows/SimpleAudio.Windows.csproj
Jklawreszuk: It seems that these constants do not exists anywhere in source code except in props file here https://github.com/stride3d/stride/blob/master/sources/targets/Stride.Core.props#L44, but in the end it has no affect on the app behavior at all

samples/Audio/SimpleAudio/SimpleAudio.Windows/SimpleAudio.Windows.csproj
Jklawreszuk: This seems reasonable, done :)

samples/Audio/SimpleAudio/SimpleAudio.Windows/SimpleAudio.Windows.csproj
Jklawreszuk: Yes it was for AssetCompiler as `profile` argument value, but this option was removed

Jklawreszuk: Commit : https://github.com/stride3d/stride/commit/9f8f407d4d5796fc3dbf562849ec4f8ea37c1e47

samples/Audio/SimpleAudio/SimpleAudio.Windows/SimpleAudio.Windows.csproj
Jklawreszuk: @Kryptos-FR  Update: I managed to find that `XENKO_PLATFORM_WINDOWS` and `XENKO_PLATFORM_WINDOWS_DESKTOP` were used for Windows specific code and later merged with `XENKO_PLATFORM_UNIX` constant into one `XENKO_PLATFORM_DESKOP` (https://github.com/stride3d/stride/commit/aa55db5f7a741751fbc0237cb6436d43738ec7a7)

Kryptos-FR: Should we set `STRIDE_PLATFORM_DESKTOP` then?

Jklawreszuk: @Kryptos-FR Yeah keeping `STRIDE_PLATFORM_DESKTOP` seems reasonable, because I tried to use dev packages in custom project and does not detect this constant

samples/Audio/SimpleAudio/SimpleAudio.Windows/SimpleAudio.Windows.csproj
```diff
@@ -1,58 +1,17 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>net8.0-windows7.0</TargetFramework>
+    <TargetFramework>net8.0-windows</TargetFramework>
     <ApplicationIcon>Resources\GameIcon.ico</ApplicationIcon>
     <OutputType>WinExe</OutputType>
     <RootNamespace>SimpleAudio</RootNamespace>
     <OutputPath>..\Bin\Windows\$(Configuration)\</OutputPath>
     <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
+    <DefineConstants>STRIDE_PLATFORM_DESKTOP</DefineConstants>
```
Kryptos-FR, line 9: It looks like the `TRACE` constants was also defined in all configuration. Not sure what it is used for though.

Kryptos-FR: It seems to be useful when we want to use the Trace class and/or print out trace logs. I don't think it is a useful default for game samples. So I'm ok with leaving it out.

Reference: https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.trace

Kryptos-FR (approved):

Eideren:
Thanks for reviewing @Kryptos-FR and @Jklawreszuk for the PR of course

PR #2215, opened 2024-04-05
fix: [Graphics] Fix light component API

Eideren:
# PR Details
`LightComponent.Light.Shadow` is an abstract `LightShadowMap` shared across all light types, the light rendering logic expects specific combination of shadow types per light types. So, users could set `LightPointShadowMapRendererCubeMap` on a Directional Light for example, leading to unexpected rendering issue or exceptions.
This change ensures that each light's shadow can only be set to a supported shadow map type.

## Motivation and Context
Fixes poor usage of inheritance and composition principles. There's more work on the `XShadowMapRenderer` side, but while that logic isn't that well written, it won't crash on us.

## Types of changes
- [x] Bug fix
- [ ] New feature
- [x] Breaking change - users will no longer be able to set unexpected shadowmap types. Or reference `DirectLightBase` as that type has been removed, I could resurrect it as long as it doesn't implement `IDirectLight`, but it wouldn't be very useful then.

## Checklist


Doprez:
I did a quick test to see if I could recreate the issue and it seems better right now.
Directional light has the proper Shadow map assigned
image

And spotlight also seems correct
image

Only thing I didnt test was if I were to add a new shadowmap if the dropdown would appear again. But I dont have a secondary I could test with, I guess I could just make a duplicated shadow map of one of these and see.

Kryptos-FR (approved):

Eideren:
Thanks a bunch for reviewing @Kryptos-FR :)

Issue #2214, opened 2024-04-05
Vulkan/OpenGL error generating vert and frag files


Doprez:
**Release Type**: Official

**Version**: 4.2.0.2122

**Platform(s)**: Windows

**Describe the bug**
I have been playing around with TR.Stride and was trying to export to Vulkan but I keep getting an error about incorrect variable names or conversion errors.

**To Reproduce**
1. pull TR.Stride
2. Remove references to the Ocean project due to compute shader issues
3. follow https://doc.stride3d.net/4.2/en/manual/graphics/graphics-api.html to get the vulkan compilation
4. run the project 

**Expected behavior**
The project should run with the correct generated shader files.

**Log and callstacks**
```
System.AggregateException
  HResult=0x80131500
  Message=One or more errors occurred. (Could not compile shader. See error messages.Info: C:\Users\User\AppData\Local\Temp\tmpdvazal.vert
Info: ERROR: C:\Users\User\AppData\Local\Temp\tmpdvazal.vert:23: 'gl_InstanceID' : undeclared identifier (Did you mean gl_InstanceIndex?)
Info: ERROR: C:\Users\User\AppData\Local\Temp\tmpdvazal.vert:23: 'assign' :  cannot convert from ' temp float' to ' global highp uint'
Info: ERROR: C:\Users\User\AppData\Local\Temp\tmpdvazal.vert:23: '' : compilation terminated 
Info: ERROR: 3 compilation errors.  No code generated.
Info: SPIR-V is not generated for failed compile or link
Error: Failed to generate SPIR-V from GLSL
Info: C:\Users\User\AppData\Local\Temp\tmpxu3ojj.frag
Info: ERROR: C:\Users\User\AppData\Local\Temp\tmpxu3ojj.frag:208: 'DepthStencil_id80' : undeclared identifier 
Info: ERROR: C:\Users\User\AppData\Local\Temp\tmpxu3ojj.frag:208: 'DepthStencil_id80' :  left of '[' is not of type array, matrix, or vector  
Info: ERROR: C:\Users\User\AppData\Local\Temp\tmpxu3ojj.frag:208: '' : compilation terminated 
Info: ERROR: 3 compilation errors.  No code generated.
Info: ERROR: Linking fragment stage: Missing entry point: Each stage requires one entry point
Info: SPIR-V is not generated for failed compile or link
Error: Failed to generate SPIR-V from GLSL
)
  Source=System.Private.CoreLib
  StackTrace:
[...]

PR #2217, opened 2024-04-05
#2214  Vulkan uses gl_InstanceIndex and gl_VertexIndex instead of gl_InstanceId and gl_VertexId as the latter are not supported. 

johang88:
# PR Details

Vulkan uses gl_InstanceIndex and gl_VertexIndex instead of gl_InstanceId and gl_VertexId as the latter are not supported. 

## Related Issue

https://github.com/stride3d/stride/issues/2214

## Motivation and Context

gl_InstanceId and gl_VertexId are removed in https://github.com/KhronosGroup/GLSL/blob/main/extensions/khr/GL_KHR_vulkan_glsl.txt 

## Types of changes

## Checklist


Eideren:
I'm guessing those are strictly nomenclature changes then ? Thanks !

Issue #1947, opened 2023-10-16
Hard time to find exe in game folder for players


OURAHATIM:
**Is your feature request related to a problem? Please describe.**
the problem is the big amount of dll that if i release my game in itch as a zip the player will need to scroll and it's hard to find the exe

**Describe the solution you'd like**
compress the dlls to one dll or 5 or by category because there is too much

Ceebox:
Should the players really be looking around in the game folder for the .exe anyway?

OURAHATIM:
> Should the players really be looking around in the game folder for the .exe anyway?

in itch.io yes since he is downlowding the exe not using a launcher like steam

Doprez:
I was going to suggest `dotnet publish -c Release -r win-x64 /p:PublishSingleFile=true` but it seems like Stride ignores that upon publish? @manio143 would you have an idea why this wouldn't work, would it be the serialization and loading of assets?

manio143:
@Doprez I was able to publish a Stride game fully self contained, single file, with VCRedist included, compiled from GitHub actions pipeline - see the tree here:
https://github.com/manio143/project-step-up/tree/8f4755099b7fe983d5648ed474ce140b5ac78c54

Ethereal77:
I would prefer if Stride, by default on publish, moved all its DLLs to a `runtime` or whatever subdir, and the `.exe` could know where to look for them, so in the base directory there would be just the `.exe` and maybe configs.

The last part (the executable knows where to look for the DLLs) is already done I think. @tebjan Did you not make a PR some time ago for this?

But I think this should be done only for `publish`. When building normally in `Debug` or `Release`, I would avoid the copy to not degrade the compilation times.

SoulRider:
The old school non-code way to do it is to put you code folder inside another folder, and in that outer folder, just have a shortcut to the .exe file.

Fydar:
I recommend checking out this project. It has been actively maintained for many years now.
https://github.com/nulastudio/NetBeauty2

I'm not sure if Stride will work with it. I might give it a test when I'm not so busy and we can create any follow up bugs if Stride is not compatible with NetBeauty2.

But ultimately, it's probably not something that Stride should try and impose on the user. The moving of DLLs can create some incompatibilities with native libraries that expect certain DLLs to be in a certain place; which could result in a lot of bug reports from users directed at Stride.

If NetBeauty2 is something Stride wants to recommend for this use case, we can include NetBeauty2 in the tests.

[...]

Issue #2204, opened 2024-03-31
Game Will Not Launch if PublishSingleFile is Enabled


MeharDT:
**Release Type**: Official Release

**Version**: 4.2.0.2122

**Platform(s)**: Windows

**Describe the bug**
If <PublishSingleFile>true</PublishSingleFile> or "Produce single file" is enabled when creating a build, the resulting executable will fail to launch.

**To Reproduce**
1. Start Stride 4.2.0.2122 and create a New Game project for Windows
2. Once loaded, open Visual Studio and Publish the project to a local folder
3. Ensure File publish options -> Produce single file is selected in Profile settings

**Expected behavior**
Game should launch when the executable is opened.

Doprez:
Can confirm this seems to have been an issue since 4.1, I have tried before.

MeharDT:
Event Viewer log,

```
Application: MyGame.Windows.exe
CoreCLR Version: 8.0.324.11423
.NET Version: 8.0.3
Description: The process was terminated due to an unhandled exception.
Exception Info: System.TypeInitializationException: The type initializer for 'Stride.Core.IO.VirtualFileSystem' threw an exception.
 ---> System.TypeInitializationException: The type initializer for 'Stride.Core.PlatformFolders' threw an exception.
 ---> System.ArgumentNullException: Value cannot be null. (Parameter 'path1')
   at System.ArgumentNullException.Throw(String paramName)
   at System.IO.Path.Combine(String path1, String path2)
   at Stride.Core.PlatformFolders.GetApplicationLocalDirectory() in C:\BuildAgent\work\b5f46e3c4829a09e\sources\core\Stride.Core\PlatformFolders.cs:line 109
   at Stride.Core.PlatformFolders..cctor() in C:\BuildAgent\work\b5f46e3c4829a09e\sources\core\Stride.Core\PlatformFolders.cs:line 31
   --- End of inner exception stack trace ---
   at Stride.Core.PlatformFolders.set_IsVirtualFileSystemInitialized(Boolean value) in C:\BuildAgent\work\b5f46e3c4829a09e\sources\core\Stride.Core\PlatformFolders.cs:line 90
   at Stride.Core.IO.VirtualFileSystem..cctor() in C:\BuildAgent\work\b5f46e3c4829a09e\sources\core\Stride.Core.IO\VirtualFileSystem.cs:line 85
   --- End of inner exception stack trace ---
[...]

PR #2225, opened 2024-04-09
fix: PublishSingleFile Launch Crash

MeharDT:
# PR Details
Stride applications built for Windows will crash when PublishSingleFile is set to true.

## Description
When a Stride application is launched, `Assembly.Location` is used to find the executable directory and create cache folders. This returns a null value when PublishSingleFile is true. It is instead recommended to use `AppContext.BaseDirectory` to check for the executable directory.

Additional Info: https://learn.microsoft.com/en-us/dotnet/core/deploying/single-file/warnings/il3000

## Related Issue
https://github.com/stride3d/stride/issues/2204

## Types of changes

## Checklist

### DIFF
--- a/sources/core/Stride.Core/PlatformFolders.cs
+++ b/sources/core/Stride.Core/PlatformFolders.cs
@@ -186,6 +186,8 @@ namespace Stride.Core
         {
 #if STRIDE_PLATFORM_ANDROID
             return GetApplicationExecutableDirectory();
+#elif STRIDE_PLATFORM_DESKTOP
+            return Path.GetDirectoryName(AppContext.BaseDirectory);
 #else
             return Path.GetDirectoryName(typeof(PlatformFolders).Assembly.Location);
 #endif


Doprez:
Might be able to also close https://github.com/stride3d/stride/issues/1947 with this fix.

MeharDT:
> Might be able to also close #1947 with this fix.

That's a good idea, inline with that the docs should also be updated at some point to include PublishSingleFile instructions.

https://doc.stride3d.net/latest/en/manual/files-and-folders/distribute-a-game.html

sources/core/Stride.Core/PlatformFolders.cs
```diff
@@ -186,6 +186,8 @@ private static string GetApplicationBinaryDirectory()
         {
 #if STRIDE_PLATFORM_ANDROID
             return GetApplicationExecutableDirectory();
+#elif STRIDE_PLATFORM_DESKTOP
+            return Path.GetDirectoryName(AppContext.BaseDirectory);
```
Kryptos-FR, line 190: Is this a correct path if **not** published as single file?

MeharDT: I believe so. My tests included both single file and non single file published builds, all launched without issue.

PR #1256, opened 2021-12-28
[Graphics] Dithered shadows for semi-transparent materials

Eideren:
# PR Details
Produces semi-transparent shadows by poking more and more holes in the shadow map based on the transparency of the object, shadow map filtering will blur those holes with their neighbor which will result in those partially opaque pixels.

image
image
Model credits: [MysteryPancake](https://sketchfab.com/3d-models/glass-of-water-1b6bff8fa38d492dbae3ca66a89fbd83)

## Types of changes
- [ ] Docs change / refactoring / dependency upgrade
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist
- [ ] My change requires a change to the documentation.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #2242, opened 2024-05-12
Fix implicit truncation of vector type shader warning

Basewq:
# PR Details

Explicitly declare the data components used in the modulo operator & the data conversion to take place (ie. float modulo and convert to int2 data type).

## Description

See above.

## Related Issue

Using material Transparency -> Dithered Shadows (enabled by default) on a model would trigger "implicit truncation of vector type" warning at run-time.
Feature was introduced in #1256

## Motivation and Context

`streams.ShadingPosition` is `float4` data type. The original code would do a float modulo operation returning a `float4` and implicit convert to int2. This would trigger a warning due to zw components being dropped.
While the explicit `int2(...)` is probably not needed, it's better to spell it out that we're not the same data type.

## Types of changes

## Checklist

Full disclosure: I did not build/run the editor. I tested the original line & modified line on a local project with a custom shader. The original line triggered the warning on the project, and the modified line did not trigger the warning - the app did not crash.

### DIFF
--- a/sources/engine/Stride.Rendering/Rendering/Shadows/ShadowMapCasterAlphaDithered.sdsl
+++ b/sources/engine/Stride.Rendering/Rendering/Shadows/ShadowMapCasterAlphaDithered.sdsl
@@ -31,7 +31,7 @@ namespace Stride.Rendering.Shadows
         {
             base.PSMain();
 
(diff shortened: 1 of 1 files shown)


Eideren:
Thanks a bunch for fixing my stuff @Basewq !

Issue #2229, opened 2024-04-16
About link for Third Party Software is wrong for OpenTK


DigitalBox98:
**Release Type**: Official Release

**Version**: Version 4.2.0.21.22

**Platform(s)**: Windows

**Describe the bug**
OpenTK link (wrong) : http://www.opentk.com/

**To Reproduce**
Open the About/Third party window

**Expected behavior**
OpenTK link : https://opentk.net/

Kryptos-FR:
Fixed by #2248

PR #2248, opened 2024-05-14
fix opentk link

Doprez:
# PR Details

Fixes the URL of one of the third party links.

## Related Issue

https://github.com/stride3d/stride/issues/2229

## Types of changes

## Checklist


Jklawreszuk (approved):

Doprez:
I think you are correct, I wasnt sure if I should just remove it or keep it for historical reasons.

Jklawreszuk:
I would suggest deleting this line, because information about OpenTK usage will be still available via previous master branch versions like `master-4.1`, `master-4.0` etc, so It is only confusing

Doprez:
Definitely makes more sense. It has now been removed.

Eideren:
Thanks !

Issue #2008, opened 2023-10-30
[Engine] Calling game.Exit() in a game running in Fullscreen mode causes an unhandled exception error on exit.


SoulRider:
**Release Type**: Official Release

**Version**: 4.1.0.1948

**Platform(s)**: Windows

**Describe the bug**
Calling game.Exit() in a game running in Fullscreen mode causes an unhandled exception error on exit.  It appears the render device is resetting back to it's default windowed resolution on exit of fullscreen, and is trying to reload assets which have been unloaded by the game as it exited.

**To Reproduce**
1. Set game.Window.PreferredFullscreenSize to a size different to windowed resolution. I use:

            GraphicsOutput[] array = game.GraphicsDevice.Adapter.Outputs;
            game.Window.PreferredFullscreenSize = new Int2(array[0].CurrentDisplayMode.Height, array[0].CurrentDisplayMode.Width);

2. Set game.Window.IsFullscreen = true;
3. Call game.Exit()

**Expected behavior**
Game to exit silently and successfully.

**Screenshots**
![SameErrorOtherGame](https://github.com/stride3d/stride/assets/1388684/c17b2fba-0974-407a-a93f-aabe61fb9e99)

**Log and callstacks**
N/A

**Additional context**
I have a simple project available which shows this issue here - [SimpleFullscreenCrashTest.zip](https://github.com/stride3d/stride/files/13207256/SimpleFullscreenCrashTest.zip).  

I have confirmed the issue using both normal fullscreen and 'FullscreenIsBorderlessWindow'.  Issue occurs whether you run from editor, Visual Studio or from a build file directly.

I upgraded a game from 4.0 which did not have this issue to the version 4.1.0.1948, and am now experiencing this in that game. I also have a game that has only had this issue since upgrading to 4.1.0.1948 from the previous patch suggesting this is a new issue in 4.1.0.1948

SoulRider:
In trying to write a workaround for this bug, I narrowed the bug down and can confirm it happens whenever you exit fullscreen mode, you do not have to be exiting the game.  Exiting the game is disabling fullscreen as well.  Exactly what in the disabling fullscreen process is causing the issue I haven't yet figured out.

SoulRider:
I have been investigating running the game in different contexts to see if it makes a difference.  Desktop and WPF both work but still have the same fullscreen crash bug, however SDL context doesn't work at all.  I get an error on launch. Attached is the error screenshot.  

[...]

Issue #1493, opened 2022-08-06
Exception when switching window from fullscreen to windowed mode


RedKorshun:
**Release Type**: Official Release 

**Version**: 4.1.0.1734

**Platform**: Windows

**Describe the bug**
Exception during switching to windowed mode.

**To Reproduce**
1. Create a new game project using default settings
2. Add a _SyncScript_ asset with the following code:
```
using Stride.Engine;
using Stride.Input;

public class FullScreenToggler : SyncScript
{
    public override void Update()
    {
        if (Input.Keyboard.IsKeyDown(Keys.F))
        {
            Game.Window.IsFullscreen = !Game.Window.IsFullscreen;
        }
    }
}
```
3. Add a _ScriptComponent_ with the script from **step 2** to the Camera (or any other) entity.
4. Run the game using _**F5**_ hotkey (or in any other way).
5. Toggle fullscreen mode using the _**"F"**_ or _**"Alt+Tab"**_ hotkey.
6. Toggle windowed mode using exactly _**"F"**_ hotkey (handled by the script from **step 2**).

**Expected behavior**
Game window is switched to windowed mode like it was before **step 5** execution.

**Screenshots**
![ContentSerializationExceptions](https://user-images.githubusercontent.com/35571766/183230361-e8ef5eba-5d73-451b-874f-d06d6765d325.png)

**Log and callstacks**
- [ReproducingExceptionInfo.txt](https://github.com/stride3d/stride/files/9274255/ReproducingExceptionInfo.txt)
[ReproducingAfterExceptionConsole.log](https://github.com/stride3d/stride/files/9274277/ReproducingAfterExceptionConsole.log) - console log after you click on continue button (see screenshot above).
- [RealProjectExceptionInfo.txt](https://github.com/stride3d/stride/files/9274256/RealProjectExceptionInfo.txt)

**Additional context**
1. Using the _**"Alt+Tab"**_ hotkey  in **step 6** (for switching back from fullscreen mode to windowed) works as expected.
2. Even an empty scene with the only empty entity on it which has the _ScriptComponent_ with the script from **step 2** works as described.
3. There are also [similar issues](https://github.com/stride3d/stride/issues?q=is%3Aissue+is%3Aopen+label%3Abug+fullscreen), but they all belongs to a different kind of exception.

RedKorshun:
Since "Alt+Tab" is working as expected this may not be a big issue, but there is a common practice to add a fullscreen toggler in game video options and it won't work there.

[...]

PR #2251, opened 2024-05-16
fix: Update Texture reload to prevent crash due to wrong type.

Doprez:
# PR Details

when reloading a Texture it would try to pass in a value of type Image to the content manager for an asset that of type Texture.

## Description

This adds a new method to the Content loader to check the type of an asset from the URL.

This also adds that check to the Texture class to determine how it should be treated when reloading.

## Related Issue

https://github.com/stride3d/stride/issues/1493

## Motivation and Context

Its the start of a fix for a seemingly larger problem with disposing of the graphics device when the game is in fullscreen.

## Types of changes

## Checklist

### DIFF
--- a/sources/core/Stride.Core.Serialization/Serialization/Contents/ContentManager.cs
+++ b/sources/core/Stride.Core.Serialization/Serialization/Contents/ContentManager.cs
@@ -405,15 +405,8 @@ namespace Stride.Core.Serialization.Contents
         
         public Type GetType(string url)
         {
-            if(LoadedAssetUrls.TryGetValue(url, out var reference))
-            {
-                if (reference.Object != null)
-                {
-                    return reference.Object.GetType();
-                }
-            }
-
-            return null;
+            _ = LoadedAssetUrls.TryGetValue(url, out var reference);
+            return reference.Object?.GetType();
         }
 
         internal Reference FindDeserializedObject(string url, Type objType)
--- a/sources/core/Stride.Core.Serialization/Serialization/Contents/ContentManager.cs
+++ b/sources/core/Stride.Core.Serialization/Serialization/Contents/ContentManager.cs
@@ -402,6 +402,19 @@ namespace Stride.Core.Serialization.Contents
 
             return result;
         }
+        
+        public Type GetType(string url)
+        {
+            if(LoadedAssetUrls.TryGetValue(url, out var reference))
+            {
+                if (reference.Object != null)
+                {
+                    return reference.Object.GetType();
+                }
+            }
+
+            return null;
+        }
 
         internal Reference FindDeserializedObject(string url, Type objType)
         {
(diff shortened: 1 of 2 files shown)


Doprez:
I was also debating if there should be the possibility to prevent crashing on failed Asset loading by adding a try catch but I know its best to avoid if possible.
```csharp
        public T Load<T>(string url, ContentManagerLoaderSettings settings = null) where T : class
        {
            try
            {
                return (T)Load(typeof(T), url, settings);
            }
            catch(Exception ex)
            {
                Log.Error($"Error loading asset '{url}': {ex.Message}");
                return null;
            }
        }
```
instead of 
```csharp
        public T Load<T>(string url, ContentManagerLoaderSettings settings = null) where T : class
        {
            return (T)Load(typeof(T), url, settings);
        }
```

sources/core/Stride.Core.Serialization/Serialization/Contents/ContentManager.cs
```diff
@@ -402,6 +402,19 @@ private object DeserializeObject(string initialUrl, string newUrl, Type type, ob
 
             return result;
         }
+        
+        public Type GetType(string url)
+        {
+            if(LoadedAssetUrls.TryGetValue(url, out var reference))
```
Jklawreszuk, line 408: Cool, but prefer null-conditional operators to prevent nested loops 👀. Here is my suggestion
```suggestion
            public Type GetType(string url)
        {
            _ = LoadedAssetUrls.TryGetValue(url, out var reference); 
            return reference.Object?.GetType();
        }
```

Jklawreszuk (changes requested):

sources/core/Stride.Core.Serialization/Serialization/Contents/ContentManager.cs
Doprez: Updated!

Jklawreszuk (approved):
Looks legit :smile:

Eideren:
This is far more efficient:
```cs
var textureDataReloaded = assetManager.Load<object>(url);
if (textureDataReloaded is Texture tex)
   // ...
else if (textureDataReloaded is Image img)
   // ...
```
As for the method signature you introduced in `ContentManager` - if we were to have a method for this purpose, I think it's best to have a `bool TryGetLoadedAsset(string myUrl, out object myAsset)` and then the user would test the asset's type if that's what they're looking for, but they could also just use the asset itself as an added feature. But this would effectively be the same as calling `IsLoaded(url)` followed by `Load<object>(url)` and testing the type like shown above. So all in all I think it's best to remove `public Type GetType(string url)` and we could instead document the fact that `Load<T>` works with a derived types.

Doprez:
Updated based on the suggestion. I agree with the documentation especially, I had no clue you could just use the base type object even though it makes perfect sense thining about it now.

Eideren:
Thanks !

PR #2252, opened 2024-05-17
Fix crash on exit for fullscreen

Doprez:
# PR Details

Recreate was being called even when the game was exiting, causing disposed references to be called.

## Description

This adds a check on recreate if the game is exiting to skip the recreate function.

## Related Issue

## Motivation and Context

crash bad.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Games/Desktop/GameForm.cs
+++ b/sources/engine/Stride.Games/Desktop/GameForm.cs
@@ -156,6 +156,8 @@ namespace Stride.Games
         /// </summary>
         public event EventHandler<EventArgs> FullscreenToggle;
 
+        public event EventHandler<EventArgs> DisableFullScreen;
+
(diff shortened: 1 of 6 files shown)


Eideren:
Does the workaround mentioned here https://github.com/stride3d/stride/issues/2008#issuecomment-2118869130 still make sense, it's probably best to remove or replace that workaround instead of adding another workaround because of that workaround ?

Doprez:
I can remove it but it still happens from your original comment due to the IsFullScreen caught and ignored error.
image

Doprez:
@Eideren I finally think I figured it out. This seems to work for both the issues of fullscreen toggle AND exiting when in fullscreen. You were right it had to do with the Window events.

Im going to close this and make a simpler PR since the solution was just to add an event for specifically disabling the fullscreen

Issue #1577, opened 2022-12-29
Add Margin property to ConvexHullCollider


manio143:
**Is your feature request related to a problem? Please describe.**
Bullet allows specifying a margin for colliders which can be especially useful in case of ConvexHull. Reported as missing by @mathmb1986 in #1574 

Suggested code changes in the thread mentioned above:
https://github.com/stride3d/stride/blob/master/sources/engine/Stride.Assets/Physics/ColliderShapeAssetCompiler.cs
L135
```diff
                    ConvexHullColliderShapeDesc convexHullDescClone = new ConvexHullColliderShapeDesc
                    {
                        Scaling = convexHullDesc.Scaling,
+                       Margin = convexHullDesc.Margin, 
                        LocalOffset = convexHullDesc.LocalOffset,
                        LocalRotation = convexHullDesc.LocalRotation,
                        Decomposition = convexHullDesc.Decomposition,
                    };
```
https://github.com/stride3d/stride/blob/master/sources/engine/Stride.Physics/Shapes/ConvexHullColliderShape.cs
L40
```diff
+        public float Margin
+        {
+            get { return InternalShape.Margin; }
+            set { InternalShape.Margin = value; }
+        }
+
```
https://github.com/stride3d/stride/blob/master/sources/engine/Stride.Physics/Data/ConvexHullColliderShapeDesc.cs
L51
```diff
+        /// <userdoc>
+        /// The Margin of the generated convex hull.
+        /// </userdoc>
+        [DataMember(48)]
+        public float Margin { get; set; }
+
```
L84
```diff
                    shape = new ConvexHullColliderShape(ConvexHulls[0][0], ConvexHullsIndices[0][0], Scaling)
                    {
                        NeedsCustomCollisionCallback = true,
+                       Margin = Margin,
                    };
```
L109
```diff
                    var subHull = new ConvexHullColliderShape(verts, indices, Scaling);
+                   subHull.Margin = Margin;
```
L136
```diff
                    var subHull = new ConvexHullColliderShape(verts[0], indices[0], Scaling);
+                   subHull.Margin = Margin;
```
L151
```diff
                        var subHull = new ConvexHullColliderShape(subVerts, subIndex, Scaling);
+                       subHull.Margin = Margin;
```

A closer look should be placed on application of margin to all `subHull` components to see if this makes sense. Also, maybe it would make more sense to add `Margin` property on the base class `ColliderShape` - which is where Margin is defined in Bullet - this way all collider shapes could benefit.

[...]

PR #2257, opened 2024-05-20
[Physics] Adding margins to convex hull colliders (Issue1577)

dloe:
# PR Details

Allow specification of margin for collider shapes. Shows to be useful for ConvexHull colliders. 

## Description

Margin property now able to be set on ConvexHull collider shape. Added margin getter and setting vars onto ConvexHullColliderShape as well as the viewable field in the properties box for the user to adjust. The base class ColliderShape contains where margin is defined as recommened so all collider shapes can potentially benefit from the new variable.

## Related Issue

Link to issue: 
https://github.com/stride3d/stride/issues/1577

Link to initial discussion: 
https://github.com/stride3d/stride/discussions/1574

## Motivation and Context

Originally reported as a missing field that solves an issue related to a test project not being able to adjust the corresponding collider. This margin not being able to be adjusted was causing unintended collisions and not allowing the user to be able to properly set up their ConvexHullCollider properties to fix the problem in the property box.

For additonal testing, I made a fork of the users orignal project that showcases the issue here: 
https://github.com/dloe/StrideEngineIssue1577_RigidBodyPhysicsBugProject

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Physics/Data/ConvexHullColliderShapeDesc.cs
+++ b/sources/engine/Stride.Physics/Data/ConvexHullColliderShapeDesc.cs
@@ -50,6 +50,12 @@ namespace Stride.Physics
         public Vector3 Scaling = Vector3.One;
 
         /// <userdoc>
+        /// The Margin of the generated convex hull.
+        /// </userdoc>
+        [DataMember(48)]
+        public float Margin { get; set; }
+
+        /// <userdoc>
         /// If this is not checked, the contained parameters are ignored and only a simple convex hull of the model will be generated.
         /// </userdoc>
         [DataMember(50)]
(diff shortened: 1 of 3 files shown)


sources/engine/Stride.Physics/Data/ConvexHullColliderShapeDesc.cs
```diff
@@ -49,6 +49,12 @@ public class ConvexHullColliderShapeDesc : IAssetColliderShapeDesc
         [DataMember(45)]
         public Vector3 Scaling = Vector3.One;
 
+        /// <userdoc>
+        /// The Margin of the generated convex hull.
+        /// </userdoc>
+        [DataMember(48)]
+        public float Margin { get; set; }
```
Eideren, line 56: Just a small note, it looks like margins in bullet by default should be around `0.04` for dynamic and `0.0` for static objects. Convex hulls are likely used as replacement for the lack of support of non-static meshes, so best set it to the default value for dynamics, so `0.04` by default.

dloe: Good call, adding that now

Eideren:
Looks good, thanks a bunch !

PR #2258, opened 2024-05-20
Added IsPressed, IsDown, & IsReleased for virtual buttons to Inputmanager

MechWarrior99:
# PR Details

Added `IsVirtualButtonPressed`/`Down`/`Released` methods to the InputManager and surrounding classes.

## Description

Added `IsVirtualButtonPressed`, `IsVirtualButtonDown` and `IsVirtualButtonReleased` methods to the `InputManager`, mirroring the `GetVirtualButton` method.  Added `IsPressed` .etc methods to `VirtualButtonBinding`, `VirtualButtonConfig` and `VirtualButtonSet`. Again, mirring the `GetValue` methods.

Renamed `GetVirtualButton` to `GetVirtualButtonValue`.

## Motivation and Context

Allows pressed, down, and release states to be used with VirtualButtons from the InputManager.

Renamed  `GetVirtualButton` to `GetVirtualButtonValue` since now that there are more `VirtualButton` methods it was confusing as to what it was doing (not actually getting a VirtualButton, but reading a input value from one). And matches the termonoligy in the rest of the API were it is refered to by `GetValue`.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Input/InputManager.cs
+++ b/sources/engine/Stride.Input/InputManager.cs
@@ -481,7 +481,19 @@ namespace Stride.Input
                 pair.Value.Listeners.Remove(listener);
             }
         }
-        
+
+        /// <summary>
+        /// Gets a binding value for the specified name and the specified config extract from the current <see cref="VirtualButtonConfigSet"/>.
+        /// </summary>
+        /// <param name="configIndex">An index to a <see cref="VirtualButtonConfig"/> stored in the <see cref="VirtualButtonConfigSet"/>.</param>
+        /// <param name="bindingName">Name of the binding.</param>
+        /// <returns>The value of the binding.</returns>
+        [Obsolete("This method is obsolete. Call GetVirtualButtonValue instead.")]
+        public virtual float GetVirtualButton(int configIndex, object bindingName)
+        {
+            return GetVirtualButtonValue(configIndex, bindingName);
+        }
+
         /// <summary>
         /// Gets a binding value for the specified name and the specified config extract from the current <see cref="VirtualButtonConfigSet"/>.
         /// </summary>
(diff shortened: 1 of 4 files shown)


MechWarrior99:
@dotnet-policy-service agree

Eideren:
Might want to keep a method stub with an obsolete attribute that just calls into the new one, no reason to break users code with this PR.

VaclavElias:
Does any manual page need to be updated with these new features?

MechWarrior99:
Yes, at the very least the example should be updated on the [Virtual Buttons page](https://doc.stride3d.net/latest/en/manual/input/virtual-buttons.html) to use `Input.GetVirtualButtonValue()`, though ideally the page would be further updated to show usage of the new methods. Though not sure how needed that really is.

VaclavElias:
Ohh, I see. Well, once it is merged here, it would be nice to do a PR also to docs and add some content at later stage? Could you do that as well? I can assist with formating but I wouldn't be able to create a content in this case.

MechWarrior99:
Yeah that shouldn't be a issue, I can do that

Eideren:
Forgot to mention, can you add a summary above each one of those methods, particularly for `Pressed` and `Down` since they are easily mixed up.

MechWarrior99:
Yeah looks like I forgot some. Added. Probably a conversation for else where, but it is probably a good idea for us to consider changing the name `Down` to something like `Held` or `Pressing`. Even as I was working on this, it took extra mental work to keep `Pressed` and `Down` straight haha.

Eideren:
Yeah, it is kind of a pain, feel free to open an issue about this.

delustra:
The feature is great! Thanks for the PR. We'll need to update that https://doc.stride3d.net/latest/en/tutorials/csharpbeginner/virtual-buttons.html as well (and Youtube too I believe)

Issue #1021, opened 2021-02-16
PhysicsProcessor ParentScene may be initialized to null


manio143:
**Release Type**: GitHub

**Version**: master

**Platform(s)**: N/A

**Describe the bug**

I'm not currently sure why or if always but on the below line the `parentScene` is not set, which results in later inability to display debug colliders in game.
https://github.com/stride3d/stride/blob/74b2b4ad07d7811688e540259cedb8abb95ead9f/sources/engine/Stride.Physics/Engine/PhysicsProcessor.cs#L215

**To Reproduce**
1. Create a new project
2. Add a static collider to the ground
3. Add Debug physics shapes script
4. Launch game and press Ctrl+Shift+P

**Expected behavior**
Debug collider shapes are displayed.

**Additional context**
The issue can be easily mitigated by adding 
```csharp
SceneSystem.SceneInstance.GetProcessor<PhysicsProcessor>().ParentScene = Entity.Scene;
```
to the beginning of the debug script.

kizilsum:
Hi! Is this still an issue? If so, I would like to take this one. @manio143 @Eideren

Eideren:
@kizilsum Sure, go ahead !
Just to clarify, the `help wanted` message removed above is because anyone can now freely contribute to the engine. You can open PRs for any issue you want, right now the engine doesn't have a lot of traffic so you can just do so without notifying us first if you don't want to :)

wrshield:
Hello, I'd like to try working on this issue and was wondering if anyone had some advice or ideas on where to start.  Thanks!

Doprez:
https://doc.stride3d.net/4.2/en/manual/engine/entity-component-system/index.html I assume this will be an issue within the Entity processor? I havent messed with Bullet too much but each of these physics shapes should be getting initialized in the EntityProcessor if it works like other systems.

I assume it should be around [here](https://github.com/stride3d/stride/blob/55bd9529def5c4e2e5daa09a7d7e570745b6968c/sources/engine/Stride.Physics/Engine/PhysicsProcessor.cs#L134) where it gets set up.

Eideren:
Fixed through #2262

PR #2262, opened 2024-05-21
Fix for Issue #1021 - PhysicsProcessor ParentScene may be initialized to null

wrshield:
# PR Details

## Description

This change is to address issue # 1021 - PhysicsProcessor ParentScene may be initialized to null.
The change was made in PhysicsProcessor.cs at line 160, where I added a check to see if ParentScene is null, and if so assign the entity scene to ParentScene.

## Related Issue

https://github.com/stride3d/stride/issues/1021

## Motivation and Context

The issue was raised because when using the debug physics shape script, the colliders were not being displayed after pressing ctrl+shift+p at runtime.  To verify that this change solves the issue:
1. Create a new project
2. Add a static collider to the ground
3. Add Debug physics shapes script
4. Launch game and press Ctrl+Shift+P
5. Colliders are now shown highlighted red

## Types of changes

## Checklist

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,6 +1,6 @@
 ---
 name: Bug report
-about: Create a report to help us improve
+about: Create a bug report to help us improve Stride.
 title: ''
 labels: 'bug'
 assignees: ''
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,14 @@
+blank_issues_enabled: false
+
+contact_links:
+  - name: Stride Docs
+    url: https://github.com/stride3d/stride-docs/issues
+    about: Please report documentation issues in the Stride Docs repository.
+
+  - name: Stride Website
+    url: https://github.com/stride3d/stride-website/issues
+    about: Please report website issues in the Stride Website repository.
+
+  - name: Stride Community Toolkit
+    url: https://github.com/stride3d/stride-community-toolkit/issues
+    about: Please report community toolkit issues in the Stride Community Toolkit repository.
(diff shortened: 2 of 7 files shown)


wrshield:
> @wrshield please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> ```
> 
> > Options:
> > 
> > 
> > ```
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 
> > ```
> > ```
> 

@dotnet-policy-service agree

Eideren:
I think it would be more appropriate to move that logic above the following line: https://github.com/stride3d/stride/blob/master/sources/engine/Stride.Physics/Engine/PhysicsProcessor.cs#L106
That way it is grouped with its usage, and would only run when the debug stuff is toggled on.
To retrieve a valid `parentScene` value from there you can copy the existing one in `OnSystemAdd`, so `parentScene = Services.GetSafeServiceAs<SceneSystem>().SceneInstance.RootScene;` note the null checks being purposefully removed here to ensure we throw instead of silently failing since nothing would be visible otherwise.

wrshield:
> I think it would be more appropriate to move that logic above the following line: https://github.com/stride3d/stride/blob/master/sources/engine/Stride.Physics/Engine/PhysicsProcessor.cs#L106 That way it is grouped with its usage, and would only run when the debug stuff is toggled on. To retrieve a valid `parentScene` value from there you can copy the existing one in `OnSystemAdd`, so `parentScene = Services.GetSafeServiceAs<SceneSystem>().SceneInstance.RootScene;` note the null checks being purposefully removed here to ensure we throw instead of silently failing since nothing would be visible otherwise.

Makes sense, thanks for the suggestion, I did a new push with this change.

Eideren:
Sorry, I should have specified that you should only conditionally assign that parentScene if none was provided by the user before that call, so something like:
```cs
parentScene ??= Services.GetSafeServiceAs<SceneSystem>().SceneInstance.RootScene;
```
Users may have set the `parentScene` to a specific subscene. Tebjan added the `parentScene` support in so I would guess that VVVV are using this feature in their application.

Eideren:
Thanks !

Issue #2156, opened 2024-02-14
Review code annotations


Kryptos-FR:
Following #2155, we use other kind of annotations for instance for serialization.

Some annotations have the same name as ones existing in the .NET standard libraries (such as `[DataContract]`) which can at least be confusing and at most create mistakes where the wrong one is used.

The purpose of this issue is to list all annotations and review whether the standard ones can be used instead. If so, prepare a deprecation path where at first both can be used and later the custom one is removed.

IXLLEGACYIXL:
On DataContractAttribute
`Stride: alias == System: name`
`System: namespace` isnt implemented in stride but it could be
`System: IsReference` doesnt exist in stride
`System: IsInherited` doesnt exist in system
`DefaultDataMemberMode` is only used 3 times and can be removed in obsoletion path
so the datatypes would be there for datacontract.. being generous isreference and isinherited can somehow be interpreted the same way but if you use system datacontract you have to use DataMember of system as well, so you dont clash

in DataMember you have 
`DataMemberMode` = required , cant be removed
`order` = exists on both
`name` = exists on both
`mask` = used for excluding members based on the mask ( dont know exactly ), may be dropped? dont know

with C#13 i think a new feature comes to extend existing classes with properties, this may also be available for annotations, so just extend the datamember attribute of system by mask and datamember mode and it could work

[NotNull] can be yoinked, inconsistenly used accross the engine and is replacable by nullable

Eideren:
> i think a new feature comes to extend existing classes with properties

If you're referring to extension types, this wouldn't help in our case:

> Instance extension types cannot hold state. For example, they can’t include fields. They can access state on the underlying type or in another location.
https://devblogs.microsoft.com/dotnet/dotnet-build-2024-announcements/

Let's say that's fine, I wonder how safe it is to just adopt it like so, would feeding strides' types to microsoft's `DataContractSerializer` even work out of the box ? What about the other way around, random libraries that have some of their types annotated with microsoft's `DataContract`, how would those be serialized through our serialization system ?

Renaming that attribute should be fairly painless though, just have to mark the current one as obsolete and users can just mass replace `DataContract` with whatever name we thought of for the new one.

IXLLEGACYIXL:
goind towards incremental source generator inherited doesnt work anymore as each class needs the attribute as incremental source generators work on syntax not on semantic

sad that the new feature doesnt allow states :(

IXLLEGACYIXL:
[...]

Issue #2155, opened 2024-02-13
Annotate with nullable attributes


Kryptos-FR:
**Is your feature request related to a problem? Please describe.**
Currently the nullable feature is not enabled on most projects. Instead we have some old custom attributes (such as `[NoNull]`) that are implemented in `Stride.Core`. These attributes are compatible with Resharper but are ignored by the actual .NET analyzers.

**Describe the solution you'd like**
Annotate your code with attributes from https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis. Then gradually enable the nullable feature on projects.

**Describe alternatives you've considered**
None. It can be really useful to detect issues and is expected in modern codebase.

**Additional context**
It will likely result in lots of warnings. To reduce overload, we can start with the editor and assets assemblies and then the core assemblies. Later apply to the remaining engine assemblies.

PR #2266, opened 2024-05-22
Nullable and Modernization for Stride.Core.Serialization

IXLLEGACYIXL:
# PR Details

Modernizing Stride.Core.Reflection 

## Description

1. Should enable Nullable instead of Strides NotNull attribute
2. Should reduce the usage of reflection, mainly used in collection Descriptors ( thats now prototyped for Dictionary, if its in interested i can convert set and list and the others too )
3. Using new C# features ie collection initialization, Pattern Matching etc

## Related Issue

#2156 
 #2155
Removes all [NotNull] annotations and uses the new Nullable 
Removes a bug taht System.Collections anyway crash the editor as serializers are missing

## Types of changes

- [x] Docs change / refactoring / dependency upgrade ( some comments changed  )

## Checklist


sources/core/Stride.Core.Reflection/MemberPath.cs
```diff
@@ -379,46 +379,6 @@ public bool TryGetValue(object rootObject, out object value)
             return true;
         }
 
-        ///// <summary>
```
IXLLEGACYIXL, line 382: thats just zombie code, no idea what its for

sources/core/Stride.Core.Reflection/MemberDescriptors/PropertyDescriptor.cs
```diff
@@ -39,17 +39,17 @@ public PropertyDescriptor(ITypeDescriptor typeDescriptor, PropertyInfo propertyI
...
         }
 
-        public override void Set(object thisObject, object value)
+        public override void Set(object thisObject, object? value)
         {
             if (!HasSet)
                 throw new InvalidOperationException($"The property [{Name}] of type [{DeclaringType.Name}] has no setter.");
 
-            setMethod.Invoke(thisObject, new[] {value});
+            setMethod!.Invoke(thisObject, [value]);
```
Eideren, line 52: I think this would be cleaner as 
```cs
if (setMethod is null)
    throw new InvalidOperationException($"The property [{Name}] of type [{DeclaringType.Name}] has no setter.");

setMethod.Invoke(thisObject, [value]);
```
That way, if the code ever changes, which is very unlikely but still, nullability would still be validated for the invoke instead of always ignored

sources/core/Stride.Core.Reflection/MemberPath.cs
```diff
@@ -357,17 +357,17 @@ public object GetValue(object rootObject)
...
+            ArgumentNullException.ThrowIfNull(rootObject);
             value = null;
             try
             {
                 object nextObject = rootObject;
                 for (int i = 0; i < items.Count; i++)
                 {
                     var item = items[i];
-                    nextObject = item.GetValue(nextObject);
+                    nextObject = item.GetValue(nextObject)!;
```
Eideren, line 370: Wouldn't it be valid for this method to return both true *and* output null `value` at the same time ?
In which case, this `!` should be removed I think ?

sources/core/Stride.Core.Reflection/TypeDescriptors/ObjectDescriptor.cs
```diff
@@ -144,7 +142,7 @@ public virtual void Initialize(IComparer<object> keyComparer)
             }
 
             // Free the member list
-            members = memberList.ToArray();
+            members = [.. memberList];
```
Eideren, line 145: The new syntax is cool, but maybe this is a bit too much, let's keep `ToArray()`, it is more readable

sources/core/Stride.Core.Reflection/TypeDescriptors/PrimitiveDescriptor.cs
```diff
@@ -12,14 +12,15 @@ namespace Stride.Core.Reflection
...
-        private readonly Dictionary<string, object> enumRemap;
+        private static readonly List<IMemberDescriptor> EmptyMembers = [];
+        private readonly Dictionary<string, object?> enumRemap;
 
         public PrimitiveDescriptor(ITypeDescriptorFactory factory, Type type, bool emitDefaultValues, IMemberNamingConvention namingConvention)
             : base(factory, type, emitDefaultValues, namingConvention)
         {
             if (!IsPrimitive(type))
                 throw new ArgumentException("Type [{0}] is not a primitive");
+            enumRemap = new Dictionary<string, object?>(StringComparer.OrdinalIgnoreCase);
```
Eideren, line 23: Looking at the rest of the logic, it looks like you can rollback this change and mark `enumRemap` as nullable ?

sources/core/Stride.Core.Reflection/TypeDescriptors/SetDescriptor.cs
```diff
@@ -11,40 +12,50 @@ namespace Stride.Core.Reflection
...
-        private readonly MethodInfo isReadOnlyMethod;
-
+        private static readonly List<string> ListOfMembersToRemove = ["Comparer", "Capacity"];
+
+        private Action<object, object?> AddMethod;
+        private Action<object, object?> RemoveMethod;
+        private Action<object> ClearMethod;
+        private Func<object, object?,bool> ContainsMethod;
+        private Func<object, int> CountMethod;
+        private Func<object, bool> IsReadOnlyMethod;
```
Eideren, line 22: Naming convention, private members start with a lowercase letter, same issue with the other descriptors

sources/core/Stride.Core.Reflection/TypeDescriptors/SetDescriptor.cs
```diff
@@ -199,33 +169,17 @@ public override object GetValue(object set, int index)
...
-                object[] valueParam = new[] { value };
-                if ((bool)containsMethod.Invoke(set, indexParam))
-                {
-                    removeMethod.Invoke(set, indexParam);
-                }
-                if (!(bool)containsMethod.Invoke(set, valueParam))
-                {
-                    addMethod.Invoke(set, valueParam);
-                }
+                Add(set, index);
```
Eideren, line 182: Is using `{ set, index }` instead of `{ set, value }` a bug on the previous version or a typo in your changes ? Or am I not understanding this function ?

sources/core/Stride.Core.Yaml.Tests/DescriptorTests.cs
```diff
@@ -232,7 +224,7 @@ public void TestDictionaryDescriptor()
         {
             var attributeRegistry = new AttributeRegistry();
             var factory = new TypeDescriptorFactory(attributeRegistry);
-            var descriptor = new DictionaryDescriptor(factory, typeof(Dictionary<int, string>), false,
+            DictionaryDescriptor descriptor = new DictionaryDescriptor(factory, typeof(Dictionary<int, string>), false,
```
Eideren, line 227: Unnecessary change, not a big deal, but if you have time

Eideren (commented):
Looks great, thanks a bunch for your work @IXLLEGACYIXL

sources/core/Stride.Core.Reflection/TypeDescriptors/SetDescriptor.cs
IXLLEGACYIXL: i did that as old collection descriptor had also delegates with pascal case, but ye i can change them to camel

sources/core/Stride.Core.Reflection/TypeDescriptorFactory.cs
```diff
@@ -106,7 +109,7 @@ protected virtual ITypeDescriptor Create(Type type)
             }
             else if (type.IsArray)
             {
-                if (type.GetArrayRank() == 1 && !type.GetElementType().IsArray)
+                if (type.GetArrayRank() == 1 && !type.GetElementType()!.IsArray)
```
Jklawreszuk, line 112: Looks great but I noticed you did used null-forgiving operator for Nullable type that might be null. Is it intentional @IXLLEGACYIXL ?

Jklawreszuk (changes requested):

sources/core/Stride.Core.Reflection/TypeDescriptorFactory.cs
Eideren: This shouldn't return null when the type is an array, which it is given the if test above :)

Jklawreszuk: Right, makes sense! 😅

Jklawreszuk (approved):

sources/core/Stride.Core.Reflection/TypeDescriptors/PrimitiveDescriptor.cs
IXLLEGACYIXL: i thought,

if something is found enumRemap is set in any case
if nothing is found its an empty dictionary which is also valid for all the usages in the class

so later on the null checks can be removed which i forgot, 

its like returning an empty list instead of returning null to be valid in any case

sources/core/Stride.Core.Reflection/TypeDescriptors/SetDescriptor.cs
IXLLEGACYIXL: if (simpleSet != null)
is always true as IsSet is true when System.Collection.Generic.ISet is provided
there is no System.Collection.ISet and all the descriptors work like "on instantiation check if type fits and after that trust me"

so the else in the original code is never invoked ever

sources/core/Stride.Core.Reflection/TypeDescriptors/SetDescriptor.cs
```diff
@@ -11,40 +12,50 @@ namespace Stride.Core.Reflection
...
-            ElementType = interfaceType.GetGenericArguments()[0] ?? typeof(object);
-
-            Type[] ArgTypes = { ElementType };
-            addMethod = interfaceType.GetMethod("Add", ArgTypes);
-            removeMethod = type.GetMethod("Remove", ArgTypes);
-            clearMethod = interfaceType.GetMethod("Clear");
-            containsMethod = type.GetMethod("Contains", ArgTypes);
-            countMethod = type.GetProperty("Count").GetGetMethod();
-            isReadOnlyMethod = type.GetInterface(typeof(ICollection<>)).GetProperty("IsReadOnly").GetGetMethod();
+                throw new ArgumentException(@"Expecting a type inheriting from System.Collections.ISet<T>", nameof(type));
```
IXLLEGACYIXL, line 31: this was just copy paste from the other descriptors like dictionary and list, it seems to be added after the others and implemented the same way even though it only works for generic one as the non generic one doesnt exist or isnt handled

sources/core/Stride.Core.Reflection/MemberPath.cs
IXLLEGACYIXL: is removed

sources/core/Stride.Core.Reflection/MemberDescriptors/PropertyDescriptor.cs
IXLLEGACYIXL: now its nullable

sources/core/Stride.Core.Reflection/TypeDescriptors/ObjectDescriptor.cs
IXLLEGACYIXL: reverted

sources/core/Stride.Core.Reflection/TypeDescriptors/SetDescriptor.cs
IXLLEGACYIXL: changed and simplified names to be constant accross the classes

sources/core/Stride.Core.Yaml.Tests/DescriptorTests.cs
IXLLEGACYIXL: reverted that

sources/core/Stride.Core.Reflection/MemberDescriptors/PropertyDescriptor.cs
Eideren: Well, now you just made it more confusing ... I would rather you use your previous one or my suggestion, sorry :(

sources/core/Stride.Core.Reflection/MemberPath.cs
```diff
@@ -357,17 +357,20 @@ public bool Apply(object rootObject, MemberPathAction actionType, object value)
...
                 {
                     var item = items[i];
-                    nextObject = item.GetValue(nextObject)!;
+                    if(nextObject is not null)
+                    {
+                        nextObject = item.GetValue(nextObject);
+                    }
                 }
                 value = nextObject;
             }
```
Eideren: How about:
```cs
try
{
    object nextObject = rootObject;
    for (int i = 0; i < items.Count; i++)
    {
        var item = items[i];
        var newObject = item.GetValue(nextObject);
        if (newObject is null)
        {
        	value = null;
            return true;
        }
        nextObject = newObject;
    }
    value = nextObject;
}
```

sources/core/Stride.Core.Reflection/MemberDescriptors/PropertyDescriptor.cs
IXLLEGACYIXL: oh now i get it

sources/core/Stride.Core.Reflection/TypeDescriptors/SetDescriptor.cs
Eideren: Yep, I get that part, what I was referring to was this section from the previous logic:
```cs
if (simpleSet.Contains(index))
    simpleSet.Remove(index);
if (!simpleSet.Contains(value))
    simpleSet.Add(value);
```
Which, from my understanding, directly maps to your:
```cs
if (Contains(set, index))
    Remove(set, index);
if(!Contains(set, index))
    Add(set, index);
```

Notice the difference between the versions of the second `if` test, the previous version passes the `value` parameter (`!simpleSet.Contains(value)`) while your version uses the `index` parameter `!Contains(set, index)`
Is that intended ?

IXLLEGACYIXL: [8fc1f8b](https://github.com/stride3d/stride/pull/2266/commits/8fc1f8bb5d3534b47e9043106213a6839d0d240f)
made it back to what it was before

sources/core/Stride.Core.Reflection/MemberDescriptors/PropertyDescriptor.cs
IXLLEGACYIXL: [854cbfb](https://github.com/stride3d/stride/pull/2266/commits/854cbfb8f7a74fba9d9ae797c3c523946f9ca6db)
made it now to your suggestion

sources/core/Stride.Core.Reflection/MemberPath.cs
IXLLEGACYIXL: [291b3ea](https://github.com/stride3d/stride/pull/2266/commits/291b3ea507e619f175b1913538f70aff024ad851)
did that

IXLLEGACYIXL:
ran tests and they fail currently, maybe its a compilation issue... lets see..

IXLLEGACYIXL:
WELL  sample tests fail but on all branches with latest main, but the descriptor tests and all other engien tests pass so it seems good

Eideren:
Yep, it's good, TC reported 4 failing tests but those are unrelated, see #2296

Eideren:
Thanks !

PR #2270, opened 2024-05-24
fix: VS component check should only be performed for Win OS

Jklawreszuk:
# PR Details

## Description

IsVSComponentAvailableAnyVersion method currently may throw exception if executed on platform other than Windows - this prevents AssetCompilerApp from running e.g. on Linux.

## Motivation and Context

This PR is a part of making AssetCompiler crossplatform.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Assets/StrideConfig.cs
+++ b/sources/engine/Stride.Assets/StrideConfig.cs
@@ -235,6 +235,8 @@ namespace Stride.Assets
         /// <returns>true if any of the components in the dictionary are available, false otherwise</returns>
         internal static bool IsVSComponentAvailableAnyVersion(IDictionary<Version, string> vsVersionToComponent)
         {
+            if (!OperatingSystem.IsWindows()) 
+                return false;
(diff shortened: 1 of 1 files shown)


sources/engine/Stride.Assets/StrideConfig.cs
```diff
@@ -236,6 +236,8 @@ internal static void RegisterSolutionPlatforms()
         internal static bool IsVSComponentAvailableAnyVersion(IDictionary<Version, string> vsVersionToComponent)
         {
             if (vsVersionToComponent == null) { throw new ArgumentNullException("vsVersionToComponent"); }
+            if (!OperatingSystem.IsWindows())
```
Doprez, line 239: Should this be moved before the initial null check if this is ignored anyway on other platforms?

Jklawreszuk: Yeah I can change that

Eideren:
Thanks !

PR #2271, opened 2024-05-25
fix: Get correct PackageCompiler.SdkDirectory for all platforms

Jklawreszuk:
# PR Details
 
## Description

PR simplifies getting correct SdkDirectory path. The previous way of moving folders two up, used `UriBuilder`, but it threw an exception on Unix systems

## Motivation and Context

This PR is a part of making AssetCompiler crossplatform.

## Types of changes

## Checklist

### DIFF
--- a/sources/assets/Stride.Core.Assets/Compiler/PackageCompiler.cs
+++ b/sources/assets/Stride.Core.Assets/Compiler/PackageCompiler.cs
@@ -19,12 +19,16 @@ namespace Stride.Core.Assets.Compiler
 
         static PackageCompiler()
         {
+            SdkDirectory = GetSdkDirectory();
+        }
+
+        private static string GetSdkDirectory()
+        {
             // Compute StrideSdkDir from this assembly
-            // TODO Move this code to a reusable method
             var codeBase = typeof(PackageCompiler).Assembly.Location;
-            var uri = new UriBuilder(codeBase);
-            var path = Path.GetDirectoryName(Uri.UnescapeDataString(uri.Path));
-            SdkDirectory = Path.GetFullPath(Path.Combine(path, @"..\.."));
+            // from ../bin/Debug/net{version} -> ../bin
+            var path = Path.GetDirectoryName(codeBase);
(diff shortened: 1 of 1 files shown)


sources/assets/Stride.Core.Assets/Compiler/PackageCompiler.cs
```diff
@@ -20,11 +20,9 @@ public class PackageCompiler : IPackageCompiler
         static PackageCompiler()
         {
             // Compute StrideSdkDir from this assembly
-            // TODO Move this code to a reusable method
             var codeBase = typeof(PackageCompiler).Assembly.Location;
-            var uri = new UriBuilder(codeBase);
-            var path = Path.GetDirectoryName(Uri.UnescapeDataString(uri.Path));
-            SdkDirectory = Path.GetFullPath(Path.Combine(path, @"..\.."));
+            // from ../bin/Debug/net{version} -> ../bin
+            SdkDirectory = Directory.GetParent(codeBase)?.Parent?.Parent?.FullName ?? "";
```
Eideren, line 25: `Directory` goes through and reads from the filesystem afair while `Path` just works on the string representation of the path. 
You can chain `Path.GetDirectoryName` calls if you want to do something similar, see examples in 
https://learn.microsoft.com/en-us/dotnet/api/system.io.path.getdirectoryname?view=net-8.0#system-io-path-getdirectoryname(system-string)

Jklawreszuk: @Eideren Good point. I did a simple benchmark and my fix was about 2 times slower. I started all over and I noticed UriBuilder does nothing so I removed it so it wouldn't throw an exception anymore

Jklawreszuk: @Eideren  I get it now why UriBuilder was in the first place! Thats because .NET Framework `Assembly.CodeBase` was returning URI path : https://learn.microsoft.com/en-us/dotnet/api/system.reflection.assembly.codebase?view=net-8.0 which is useless now.

Eideren:
Thanks !

PR #2273, opened 2024-05-25
[Tests] Added new Test Case for ContentManager

dloe:
Tests saved object verification for use of isLoaded and TryGetAssetUrl functions.

# PR Details

Noticed that coverage in ContentManager required more test cases. Aimed at making a test case to increase coverage.

## Description

Added a test case named VerifyLoadedData that creates Test Classes B and A, the normal databaseProvider class as well as 2 assetManagers. We then save our B class with a urlName of "test". We then load this asset and check if the asset is properly loaded. After we run TryGetAssetUrl to lookup the B class object and verify the url output with the resulting urlName.

## Related Issue

None found.

## Motivation and Context

After some discussions in the discord, it was recommended to look in this area as a place to add more test cases for the existing test suite. Test coverage could be improved here.

## Types of changes

## Checklist

### DIFF
--- a/sources/core/Stride.Core.Tests/TestContentManager.cs
+++ b/sources/core/Stride.Core.Tests/TestContentManager.cs
@@ -168,6 +168,27 @@ namespace Stride.Core.Tests
         }
 
         [Fact]
+        public void VerifyLoadedData()
+        {
+            var b1 = new B();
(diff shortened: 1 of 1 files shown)


Eideren:
LGTM, thanks !

PR #2274, opened 2024-05-25
fix: fullscreen and game exit windowing crashes

Doprez:
# PR Details

There were 2 different issues here. One was an issue with the OnToggle event causing the switch from fullscreen to windowed to be reset to true within the same call causing an exception.

The other issue was with the exit game call trying to recreate a disposed graphics device in the GameBase class.

## Related Issue

https://github.com/stride3d/stride/issues/1493

## Motivation and Context

Window changes shouldnt cause a crash.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Games/Desktop/GameForm.cs
+++ b/sources/engine/Stride.Games/Desktop/GameForm.cs
@@ -156,6 +156,8 @@ namespace Stride.Games
         /// </summary>
         public event EventHandler<EventArgs> FullscreenToggle;
 
+        public event EventHandler<EventArgs> DisableFullScreen;
+
         protected bool enableFullscreenToggle = true;
 
         /// <summary>
@@ -290,6 +292,15 @@ namespace Stride.Games
             FullscreenToggle?.Invoke(this, e);
         }
 
+        /// <summary>
+        /// Raises the DisableFullScreen event
+        /// <Summary>
+        /// <param name="e">The <see cref="System.EventArgs"/> instance containing the event data.</param>
+        private void OnDisableFullScreen(EventArgs e)
+        {
+            DisableFullScreen?.Invoke(this, e);
+        }
+
         protected override void OnClientSizeChanged(EventArgs e)
         {
             base.OnClientSizeChanged(e);
@@ -371,7 +382,7 @@ namespace Stride.Games
                         //also remove full screen if this is the case
                         if (IsFullScreen && !isSwitchingFullScreen) //exit full screen on alt-tab if in fullscreen
                         {
-                            OnFullscreenToggle(new EventArgs());
+                            OnDisableFullScreen(new EventArgs());
                         }
                     }
                     break;
--- a/sources/engine/Stride.Games/GameWindow.cs
+++ b/sources/engine/Stride.Games/GameWindow.cs
@@ -304,6 +304,16 @@ namespace Stride.Games
             IsFullscreen = !IsFullscreen;
         }
 
+        protected void OnDisableFullScreen(object source, EventArgs e)
+        {
+            IsFullscreen = false;
+        }
+
+        protected void EnableFullscreen(object source, EventArgs e)
+        {
+            IsFullscreen = true;
+        }
+
         protected void OnClosing(object source, EventArgs e)
         {
             var handler = Closing;
--- a/sources/engine/Stride.Games/GameWindow.cs
+++ b/sources/engine/Stride.Games/GameWindow.cs
@@ -309,11 +309,6 @@ namespace Stride.Games
             IsFullscreen = false;
         }
 
-        protected void EnableFullscreen(object source, EventArgs e)
-        {
-            IsFullscreen = true;
-        }
-
         protected void OnClosing(object source, EventArgs e)
         {
             var handler = Closing;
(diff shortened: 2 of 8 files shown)


sources/engine/Stride.Games/Desktop/GameForm.cs
```diff
@@ -371,7 +382,7 @@ protected override void WndProc(ref Message m)
                         //also remove full screen if this is the case
                         if (IsFullScreen && !isSwitchingFullScreen) //exit full screen on alt-tab if in fullscreen
                         {
-                            OnFullscreenToggle(new EventArgs());
```
manio143, line 374: Could users take a dependency on this event being invoked? If so we shouldn't remove it.

sources/engine/Stride.Games/GameWindow.cs
```diff
@@ -304,6 +304,16 @@ protected void OnFullscreenToggle(object source, EventArgs e)
             IsFullscreen = !IsFullscreen;
         }
 
+        protected void OnDisableFullScreen(object source, EventArgs e)
+        {
+            IsFullscreen = false;
+        }
+
+        protected void EnableFullscreen(object source, EventArgs e)
```
manio143, line 312: Is this method referenced anywhere?

sources/engine/Stride.Games/Desktop/GameForm.cs
Doprez: It was inherintly broken, this should only ever be false when hitting this if statement if I understood correctly. 

This is a mandatory state change when exiting fullscreen which requires it to be false or else it will reset to true due to the toggle logic and cause a crash.

sources/engine/Stride.Games/GameWindow.cs
Doprez: No, I can remove it. I left it there just in case someone were to inherit the class but its currently useless.

sources/engine/Stride.Graphics/Direct3D/SwapChainGraphicsPresenter.Direct3D.cs
```diff
@@ -239,7 +239,7 @@ protected internal override void OnDestroyed()
             backBuffer.OnDestroyed();
             backBuffer.LifetimeState = GraphicsResourceLifetimeState.Destroyed;
 
-            swapChain.Dispose();
+            swapChain?.Dispose();
```
Eideren, line 242: In which case `swapChain` would be null at that point ?

Doprez: Ah, looks like I should have been more curious about this one. It is another victim of being needlessly called twice for some reason causing the null exception.

sources/engine/Stride.Games/Desktop/GameForm.cs
manio143: My point here is that if someone had some custom logic in a listener to the Toggle event, by removing this invocation their logic would never be invoked. If the whole event is to be made obsolete let's make sure to add an attribute to the event so the users start getting a warning there.

manio143: Although I may have misunderstood - if this isn't the main place where the toggle callback was invoked then ignore my comment.

sources/engine/Stride.Games/GameWindow.cs
manio143: It's signature looks like an event handler. I imagine it's easier for a consumer to run `IsFullscreen = true` instead of invoking this method.

Doprez: It was meant for an event, you are right. The communication between the window and form is very oddly done. Instead of having a reference to the GameWindow directly its awkwardly linked though event.

I removed this one for now since it wasnt a relevent event and users can inherit from the Gamewindow class to add their own if needed.

sources/engine/Stride.Games/Desktop/GameForm.cs
Doprez: It is not the main place where it is used. It was meant to change the fullscreen to not fullscreen when the window was being changed but due to how broken the windowing stuff is it just gets called randomly. With this intance being always false it wont be toggled back to true randomly.

There is another case of `OnFullscreenToggle` being used correctly below when alt tabbing. I still think this logic is wrong but its at least functionally correct.
```
                case Win32Native.WM_SYSKEYDOWN: //alt is down
                    if (wparam == VK_RETURN)
                    {
                        isSwitchingFullScreen = true;
                        if (!enableFullscreenToggle) return;
                        OnFullscreenToggle(new EventArgs()); //we handle alt enter manually
                        isSwitchingFullScreen = false;
                    }
                    break;
```

sources/engine/Stride.Graphics/Direct3D/SwapChainGraphicsPresenter.Direct3D.cs
Doprez: Ok, so I did some more testing today and found out it was the `ResumeManager` that was calling the second time. I removed it since the presenter gets destroyed before `ReumeManager` gets called which seems to be why it was there based on the comment. 

I don't see anymore issues but I don't know when the `ResumeManager` is used so I could be missing something here.

Doprez: Looks like the `ResumeManager` is maintained by the `GameBase` which is also where the Graphics device was being disposed so I think its good to go.

Eideren:
Thanks !

PR #2279, opened 2024-05-25
fix: Use CompilerApp crossplatform binary instead of exe

Jklawreszuk:
# PR Details

Since CompilerApp does not contain any Windows-dependent code (at least directly), I suggest using the `dotnet Stride.Core.Assets.CompilerApp.dll` instead of running Stride.Core.Assets.CompilerApp.dll

This ensures working on any platform: 
https://learn.microsoft.com/en-us/dotnet/core/deploying/#publish-framework-dependent

✔️ Tested on Linux it works perfectly. Need to ensure though, there is not regression on Windows

## Types of changes

## Checklist

### DIFF
--- a/sources/assets/Stride.Core.Assets.CompilerApp/build/Stride.Core.Assets.CompilerApp.targets
+++ b/sources/assets/Stride.Core.Assets.CompilerApp/build/Stride.Core.Assets.CompilerApp.targets
@@ -125,8 +125,8 @@
   <Target Name="_StridePrepareAssetCompiler">
     <PropertyGroup>
       <!-- First try NuGet layout, then git checkout -->
-      <StrideCompileAssetCommand Condition="'$(StrideCompileAssetCommand)' == '' And Exists('$(MSBuildThisFileDirectory)..\lib\net8.0\')">$(MSBuildThisFileDirectory)..\lib\net8.0\Stride.Core.Assets.CompilerApp.exe</StrideCompileAssetCommand>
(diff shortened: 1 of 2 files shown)


Eideren:
Works great, thanks !

sources/assets/Stride.Core.Assets.CompilerApp/build/Stride.Core.Assets.CompilerApp.targets
```diff
@@ -125,8 +125,8 @@
   <Target Name="_StridePrepareAssetCompiler">
     <PropertyGroup>
       <!-- First try NuGet layout, then git checkout -->
-      <StrideCompileAssetCommand Condition="'$(StrideCompileAssetCommand)' == '' And Exists('$(MSBuildThisFileDirectory)..\lib\net8.0\')">$(MSBuildThisFileDirectory)..\lib\net8.0\Stride.Core.Assets.CompilerApp.exe</StrideCompileAssetCommand>
-      <StrideCompileAssetCommand Condition="'$(StrideCompileAssetCommand)' == '' And Exists('$(MSBuildThisFileDirectory)..\Stride.Core.Assets.CompilerApp.csproj')">$(MSBuildThisFileDirectory)..\bin\$(Configuration)\net8.0\Stride.Core.Assets.CompilerApp.exe</StrideCompileAssetCommand>
+      <StrideCompileAssetCommand Condition="'$(StrideCompileAssetCommand)' == '' And Exists('$(MSBuildThisFileDirectory)..\lib\net8.0\')">$(MSBuildThisFileDirectory)..\lib\net8.0\Stride.Core.Assets.CompilerApp.dll</StrideCompileAssetCommand>
```
VaclavElias, line 128: I see, that we are referencing here `net8.0` (also in the old version), can we make it somehow dynamic, so when we upgrade to a new version of .NET that we don't have to come back here to update?

PR #2280, opened 2024-05-25
created a basic class to test points

soorMSWE:
# PR Details

I was looking through places to add testing in the game engine and found that there were no tests for the Point class in Stride.Mathematics. In order to add completeness and increase code coverage of the system, I figured I would add a couple simple tests _just_ to make sure that the points are working as they are supposed to.

## Description

I added a couple simple tests for the Point class in Stride.Mathematics. They are very basic and can be added to or modified if desired.

## Related Issue

There is no issue related to this pull request.

## Motivation and Context

This change is not required. I simply added a couple tests to increase code coverage of the project.

## Types of changes

## Checklist

- [x ] All new and existing tests passed.

### DIFF
--- a/sources/core/Stride.Core.Mathematics.Tests/TestPoint.cs
+++ b/sources/core/Stride.Core.Mathematics.Tests/TestPoint.cs
@@ -1,0 +1,24 @@
+
+using Xunit;
+
+namespace Stride.Core.Mathematics.Tests
+{
+    public class TestPoint
+    {
+        Point testPoint1 = new Point(5,5);
+        Point testPoint2 = new Point(10, 10);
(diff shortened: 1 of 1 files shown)


soorMSWE:
@dotnet-policy-service agree

Eideren:
Hey, thanks a bunch for contributing ! Just a note, since those tests expect those 3 points to be constant I think it would be more appropriate to mark them as `readonly` or you could instead move them inside the scope of each function, as you want. That way there is no ambiguity as to their usage.

soorMSWE:
Sure, I will do that right now. Thank you for the feedback!

Eideren:
Looks good, thanks !

PR #2288, opened 2024-05-27
fix: Repair Projectwatcher on Projects referencing each other

### LINKED ISSUE #2200: Multiple Projects require Module class if they reference eachother.
**Release Type**:  official

**Version**: 4.2.0.2122

**Platform(s)**: Windows

**Describe the bug**
When you hae a multiproject structure and the projects reference eachother, they wont show components in the GameStudio.

**To Reproduce**
1. create base project
2. add another project through VS
3. add a StatupScript to the external project
4. Now you should see the Script in GameStudio after a restart
5. create a reference between the external project and the main
6. after retarting GameStudio you will not see the components in that external project anymore

**Expected behavior**
The components should still show in the drop down since it will be very likely the references will exist.

**Additional context**
There is a work around, in each of the external projects you can add a [Module](https://github.com/stride3d/stride/blob/master/sources/engine/Stride.Audio/Module.cs) class and the components should show up normally.

**Repro**: https://github.com/Doprez/MultiProjectIssue/tree/main

IXLLEGACYIXL:
# PR Details

The Issue was if you have the Projects

```
...\\...\\FancyGame\\FancyGame.csproj
...\\...\\FancyGame.Test\\FancyGame.Test.csproj
```
Then the first one to compare the starts with path is always the root project of the game

now if you change the file `...\\...\\FancyGame.Test\\Bob.cs` the path comparison starts with the first project and a normal StartsWith , which results in true the directory path of the file `...\\...\\FancyGame.Test` starts with the root projects directory path `...\\...\\FancyGame` so the root project gets reloaded as lowest level of the dependencies of the solution.. but that isnt true FancyGame.Test is lower and has the file but it isnt even checked as the first trigger yields

## Related Issue

fix #2200

## Types of changes

## Checklist

### DIFF
--- a/sources/editor/Stride.Assets.Presentation/AssetEditors/ProjectWatcher.cs
+++ b/sources/editor/Stride.Assets.Presentation/AssetEditors/ProjectWatcher.cs
@@ -232,12 +232,16 @@ namespace Stride.Assets.Presentation.AssetEditors
 
                 var needProjectReload = string.Equals(trackedAssembly.Project.FilePath, changedFile, StringComparison.OrdinalIgnoreCase);
 
+
+                var directoryName = Path.GetDirectoryName(trackedAssembly.Project.FilePath) + "\\";
+                var changedFileDirectoryName = Path.GetDirectoryName(changedFile) + "\\";
+
                 // Also check for .cs file changes (DefaultItems auto import *.cs, with some excludes such as obj subfolder)
                 // TODO: Check actual unevaluated .csproj to get the auto includes/excludes?
                 if (needProjectReload == false
                     && ((e.ChangeType == FileEventChangeType.Deleted || e.ChangeType == FileEventChangeType.Renamed || e.ChangeType == FileEventChangeType.Created)
                     && Path.GetExtension(changedFile)?.ToLowerInvariant() == ".cs"
-                    && changedFile.StartsWith(Path.GetDirectoryName(trackedAssembly.Project.FilePath), StringComparison.OrdinalIgnoreCase)))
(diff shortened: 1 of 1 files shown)


sources/editor/Stride.Assets.Presentation/AssetEditors/ProjectWatcher.cs
```diff
@@ -232,12 +232,15 @@ private async Task<AssemblyChangedEvent> FileChangeTransformation(FileEvent e)
 
                 var needProjectReload = string.Equals(trackedAssembly.Project.FilePath, changedFile, StringComparison.OrdinalIgnoreCase);
 
+                var directoryName = Path.GetDirectoryName(trackedAssembly.Project.FilePath) + "\\";
+                var changedFileDirectoryName = Path.GetDirectoryName(changedFile) + "\\";
```
Eideren, line 236: That directory separator would break under linux afaict, use `Path.DirectorySeparatorChar` instead https://learn.microsoft.com/en-us/dotnet/api/system.io.path.directoryseparatorchar?view=net-8.0

Eideren (commented):
Looks good, thanks a bunch @IXLLEGACYIXL

sources/editor/Stride.Assets.Presentation/AssetEditors/ProjectWatcher.cs
IXLLEGACYIXL: allright then i try out ToOsPath as well

IXLLEGACYIXL: did it

Eideren:
Thanks !

PR #2303, opened 2024-06-03
feat: Adjust freeimage to all platforms

Jklawreszuk:
# PR Details

This PR adjust FreeImage loading for macos and linux distributions (these operating systems load this dependency from packages).
From now, FreeImage is handling image files instead of DirectXTex library (LoadWICFile method is only available for Windows) .

## Types of changes

- [x] Docs change / refactoring / dependency upgrade - Additional dependency to install

## Checklist


sources/core/Stride.Core/Native/NativeLibraryHelper.cs
```diff
@@ -12,19 +12,22 @@ namespace Stride.Core
...
         /// Try to preload the library.
         /// This is useful when we want to have AnyCPU .NET and CPU-specific native code.
-        /// Only available on Windows for now.
         /// </summary>
         /// <param name="libraryName">Name of the library, without the extension.</param>
         /// <param name="owner">Type whose assembly location is related to the native library (we can't use GetCallingAssembly as it might be wrong due to optimizations).</param>
+        /// <param name="skipLibrary">If set to true, skips library loading<param>
         /// <exception cref="System.InvalidOperationException">Library could not be loaded.</exception>
-        public static void PreloadLibrary(string libraryName, Type owner)
+        public static void PreloadLibrary(string libraryName, Type owner, bool skipLibrary = false)
```
Kryptos-FR, line 26: What's the rationale here? If you don't want to load the library, just don't call the method from the caller site.

sources/tools/Stride.TextureConverter/Backend/Wrappers/FINetWrapper/FreeImageStaticImports.cs
```diff
@@ -192,14 +192,14 @@ internal static partial class FreeImage
...
-
+		
 		/// <summary>
 		/// Returns a string containing the current version of the library.
 		/// </summary>
 		/// <returns>The current version of the library.</returns>
 		public static unsafe string GetVersion() { return PtrToStr(GetVersion_()); }
 		[DllImport(FreeImageLibrary, CharSet = CharSet.Ansi, EntryPoint = "FreeImage_GetVersion")]
-		private static unsafe extern byte* GetVersion_();
+		public static unsafe extern byte* GetVersion_();
```
Kryptos-FR, line 202: Is it necessary?

sources/tools/Stride.TextureConverter/Frontend/TextureTool.cs
```diff
@@ -40,8 +40,9 @@ static TextureTool()
             NativeLibraryHelper.PreloadLibrary("DxtWrapper", type);
             NativeLibraryHelper.PreloadLibrary("PVRTexLib", type);
             NativeLibraryHelper.PreloadLibrary("PvrttWrapper", type);
-            NativeLibraryHelper.PreloadLibrary("FreeImage", type);
-            NativeLibraryHelper.PreloadLibrary("FreeImageNET", type);
+            NativeLibraryHelper.PreloadLibrary("freeimage", type);
+            //TODO: needs to explain why FreeImageNET is loaded as a dll instead of directly referencing the C# project (this does not affect the compilation process on Linux).
+            NativeLibraryHelper.PreloadLibrary("FreeImageNET", type, !OperatingSystem.IsWindows());
```
Kryptos-FR, line 45: If condition here instead of making a useless call that returns immediately.

Jklawreszuk: I think your point is valid, fixed!

sources/core/Stride.Core/Native/NativeLibraryHelper.cs
Jklawreszuk: done

sources/tools/Stride.TextureConverter/Backend/Wrappers/FINetWrapper/FreeImageStaticImports.cs
Jklawreszuk: I do not remember why I changed the method to public, but it can be undone too (I sorted now the keywords correctly).

Eideren:
TC is complaining, might need to adjust those textures you removed to get loaded by freeimage instead ?
https://teamcity.stride3d.net/buildConfiguration/Engine_Tests_WindowsSimple/25548?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildDeploymentsSection=false&expandPull+Request+Details=true&expandBuildChangesSection=true&expandBuildProblemsSection=true

Jklawreszuk:
@Eideren Glad the tests caught the error 😅 According to the documentation, it seems that FreeImage has dedicated both Unicode and Non-Unicode methods that should be used accordingly on Windows or Unix systems.

Jklawreszuk:
I started looking for inspiration and it looks like our friends at Monogame fixed this by using appropriate method depending on the system.  Will try to go in this direction 
Monogame : 
```cs
[DllImport(NativeLibName, CharSet = CharSet.Unicode, EntryPoint = "FreeImage_LoadU")]
      public static extern IntPtr LoadU(FREE_IMAGE_FORMAT fif, string filename, int flags);

      [DllImport(NativeLibName, EntryPoint = "FreeImage_Load")]
      public static extern IntPtr LoadS(FREE_IMAGE_FORMAT fif, string filename, int flags);

      public static IntPtr Load(FREE_IMAGE_FORMAT fif, string filename, int flags)
      {
          if (CurrentPlatform.OS == OS.Windows)
              return LoadU(fif, filename, flags);
          else
              return LoadS(fif, filename, flags);
      }
```

Jklawreszuk:
Done! @Eideren, Could you try to run tests again?

Eideren:
We have a certificate issue right now that xen is looking into, I'll get back to you once it's done :)

Eideren:
Thanks !

Issue #2255, opened 2024-05-20
Duplication of Entity with List<T> variable (where T is something referenceable in a scene like Entities or Components) crashes editor


YerkoAndrei:
**Release Type**: Official Release

**Version**: 4.2.0.2122

**Platform(s)**: Windows (Editor)

**Describe the bug**
When duplicate (ctrl + d / ctrl+c + ctrl+v) an entity with a script with a not null List<T> variable, the editor crashes.
This happens when the List Type is something in the scene like Entities or Components, for example: `List<Entity>` or `List<TransformComponent>` crashes, but `List<Prefab>` or `List<Sound>` doesn't.

**To Reproduce**
1. Create script with a List<T> variable (Type must be something referenceable in a scene)
2. Add script to any Entity
3. Select the Entity
4. Duplicate with ctrl + d

**Expected behavior**
Entity should duplicate with references

**Log and callstacks**
In this Crash i use a script in the Entity "Botón (Amarillo)"
[Crash.txt](https://github.com/stride3d/stride/files/15371453/Crash.txt)

**Additional context**
Example class:
```
using System.Collections.Generic;
using Stride.Engine;

namespace Example;

public class CrashExample : StartupScript
{
    public List<Entity> entities = new List<Entity> { };

    public override void Start()
    {

    }
}
```

Kryptos-FR:
It's not a random List<T> though, it is explicitly a list of entities. If it doesn't happen with other types, please update the title and description.

YerkoAndrei:
updated

PR #2307, opened 2024-06-05
[GameStudio] Prevents a null IMemberNode to update

YerkoAndrei:
# PR Details
Fixed an error when duplicating an Entity with a not null `List<T>` variable crashes the editor. This happen when T is something referenceable in a scene like Entities or Components.

When trying to cast a Node as IMemberNode, it resulted in a null and the editor crashing, this PR just prevent to call Update on a null.

## Related Issue
#2255

## Types of changes

## Checklist

### DIFF
--- a/sources/presentation/Stride.Core.Quantum/NodeAccessor.cs
+++ b/sources/presentation/Stride.Core.Quantum/NodeAccessor.cs
@@ -57,13 +57,13 @@ namespace Stride.Core.Quantum
         /// <param name="value">The new value to set.</param>
         public void UpdateValue(object value)
         {
-            if (Index != NodeIndex.Empty)
+            if (IsItem)
             {
                 ((IObjectNode)Node).Update(value, Index);
             }
-            else
+            else if (IsMember)
             {
-                (Node as IMemberNode)?.Update(value);
+                ((IMemberNode)Node).Update(value);
             }
         }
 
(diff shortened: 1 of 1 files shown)


YerkoAndrei:
@dotnet-policy-service agree

sources/presentation/Stride.Core.Quantum/NodeAccessor.cs
```diff
@@ -63,7 +63,7 @@ public void UpdateValue(object value)
             }
             else
             {
-                ((IMemberNode)Node).Update(value);
```
manio143, line 66: Maybe it would be better to make use of the `IsMember` and `IsItem` properties in this method to ensure the cast is doable before calling update? Also, what is the type of Node when you see the error, or is it null? If there's another type, maybe there should be another case for it. If it's null, would be good to understand why and should the code allow for that.

Kryptos-FR: I would also prefer we investigate the reason of this mistyping.

YerkoAndrei: You are right, those properties have the info already.
The node when it crashes is the List.
```Node = {{Node: Object List`1 = [System.Collections.Generic.List`1[Stride.Engine.Entity]]}}```

manio143 (approved):
I'm fine merging it to fix the crash, but I think a new issue should be opened for an investigation why there can be values this node is not expecting being passed into it.

Eideren:
Thanks for this @YerkoAndrei !
Opened a new issue for this as you can see above

PR #2310, opened 2024-06-06
Fix UI overridden margin in derived controls being reset

### LINKED ISSUE #2048: UI library margins reset to zero

**Version**: 4.2.2043

**Platform(s)**: Windows

**Describe the bug**
UI library margins reset when using in another UI library or a UI page

**To Reproduce**
1. Create a UI library with zero Margin
2. Use that UI library on a UI page or another UI library
3. In the UI page, change the margins on the child UI library
4. Reload proyect
5. Child UI library should reset to all zero, but still highlight the Margin property as it have a change

**Expected behavior**
The parent UI library or UI page, should save the changes on the child UI library

**Screenshots**
Only property affected on a .sduipage and .sduilib
problema ilib en uipage

**Additional context**
Its not necessary to relaod the project, sometime it break when reloading a UI page tab.

Basewq:
# PR Details

As per #2048, Margin is a struct `Thickness`, with itself holding float fields (`Left`, `Right`, etc).
If an object has a `Thickness` struct as a field/property and you override it in a derived asset, it is marked as overridden, however the fields of the `Thickness` are not marked overridden.
When `AssetPropertyGraph` encounters the `Thickness`'s fields it was not aware its parent has been marked overridden so it mistaken itself as not overridden and overwrites itself with the value from the base asset.

This change ensures any encountered member (field/property) that is a value type must check if it is part of another value type that may be marked overridden, and reject taking the base asset's value if this is the case.

2nd commit is to fix some typos.

## Related Issue

Fixes #2048 

## Types of changes

## Checklist

### DIFF
--- a/sources/assets/Stride.Core.Assets.Quantum.Tests/Helpers/Types.cs
+++ b/sources/assets/Stride.Core.Assets.Quantum.Tests/Helpers/Types.cs
@@ -137,11 +137,34 @@ namespace Stride.Core.Assets.Quantum.Tests.Helpers
             public static int MemberCount => 4 + 3; // 4 (number of members in Asset) + 3 (number of members in Types.MyAssetWithRef2)
         }
 
+        [DataContract]
+        [AssetDescription(FileExtension)]
+        public class MyAssetWithStructWithPrimitives : MyAssetBase
+        {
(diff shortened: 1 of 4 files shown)


Eideren:
Looks great, thanks !

PR #2311, opened 2024-06-06
Fix: Fixed a NullReferenceException if printing null text

NicusorN5:
Fixes this exception if using `DebugText.Print(null, ...)` by adding a plain null check inside the `Print` function.
```
System.NullReferenceException
  HResult=0x80004003
  Message=Object reference not set to an instance of an object.
  Source=Stride.Graphics
  StackTrace:
   at Stride.Graphics.FastTextRenderer.DrawString(GraphicsContext graphicsContext, String text, Int32 x, Int32 y)
```

## Related Issue
None, this is just a trivial bug fix. I didn't decide to create a issue for this.

## Types of changes

- N/A Docs change / refactoring / dependency upgrade
- N/A New feature (non-breaking change which adds functionality)
- N/A Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- N/A My change requires a change to the documentation.
- N/A I have added tests to cover my changes.
- N/A All new and existing tests passed.
(by creating a new project, and using this script containing a single `DebugText.Print` call inside the `Update` method).

### DIFF
--- a/sources/engine/Stride.Engine/Profiling/DebugTextSystem.cs
+++ b/sources/engine/Stride.Engine/Profiling/DebugTextSystem.cs
@@ -43,7 +43,7 @@ namespace Stride.Profiling
         {
             var msg = new DebugOverlayMessage 
             { 
-                Message = message,
+                Message = message ?? "",
                 Position = position,
                 TextColor = color ?? TextColor,
                 RemainingTime = timeOnScreen ?? DefaultOnScreenTime,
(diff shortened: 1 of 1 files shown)


sources/engine/Stride.Engine/Profiling/DebugTextSystem.cs
```diff
@@ -43,7 +43,7 @@ public void Print(string message, Int2 position, Color4? color = null, TimeSpan?
         {
             var msg = new DebugOverlayMessage 
             { 
-                Message = message,
+                Message = message ?? "",
```
Kryptos-FR, line 46: nit: I find it clearer to use `string.Empty` instead of a literal `""`.

Kryptos-FR (approved):
LGTM. Thanks

PR #2315, opened 2024-06-07
fix: Rollback one of the TryAdd introduced through #2299, VS Package is still on net472

Eideren:
# PR Details
See title, TryAdd doesn't exist on net472

## Related Issue
#2299

## Types of changes

## Checklist


Kryptos-FR:
Alternatively, it could be implemented as an extension method that only exists when compiling for net472.

PR #2308, opened 2024-06-05
Fixes assembly load failure when building Stride based project on clean machine

azeno:
## Related Issue

#2232 

## Types of changes

- [ ] Docs change / refactoring / dependency upgrade
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [ ] My change requires a change to the documentation.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.
- [x] **I have built and run the editor to try this change out.**

PR #2317, opened 2024-06-08
Revert "fix: assembly load failure when building Stride based project on clean machine"

Eideren:
Reverts stride3d/stride#2308
Visual Studio solutions and Stride.ConnectionRouter fail to pack the dll as they don't have the referenced dll in their output path
https://teamcity.stride3d.net/buildConfiguration/Engine_Tests_WindowsSimple/25642?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildDeploymentsSection=false&expandBuildChangesSection=true&expandBuildProblemsSection=true

Could you look into this @azeno ?

Now naively adding a condition to this `BuildOutputInPackage` works but best to check if those projects actually need this dll.


azeno:
@Eideren See https://github.com/stride3d/stride/pull/2319

PR #2294, opened 2024-05-28
Added: ModelComponent bounds gizmos

kristian15959:
# PR Details

This PR adds bounding box gizmo for ModelComponent in editor, can be disabled like any other gizmo in scene view.

## Motivation and Context

Detecting problems with mesh/models rendering can become a problematic when you do not know if bounds have correct position and extents.
Updating to 64bit coords showed problematic when converting Meshes to use Double BoundingBoxes and could not see if the correct data was there.

## Types of changes

- [ ] Docs change / refactoring / dependency upgrade
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [ ] My change requires a change to the documentation.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.
- [x] **I have built and run the editor to try this change out.**

image

PR #2321, opened 2024-06-12
style: Set initial visibility for ModelComponent gizmos off

Eideren:
# PR Details
Following #2294 - No need to have them visible by default I would think, checking with @kristian15959 just in case ?

## Types of changes

## Checklist


kristian15959:
Hi, sorry for the later reply. It's something I wanted to add aswell just did not know where to find the setting for it. Thanks for this change I definitely approve :)

PR #2328, opened 2024-06-15
feat: Removal of useless GraphicsApi - Null

Jklawreszuk:
# PR Details

I tried to revisit the Stride.Graphics project and among Graphics Apis there is a StrideGraphicsAPI called `Null`, which does nothing and is just a placeholder for other APIs. In addition, from what I understand the value of `Null` for `StrideGraphicsApi` msbuild property is never set, so I propose to remove this completely.

## Types of changes

- [x] Refactoring

## Checklist


Basewq:
As someone who had been thinking of client/server game a while back, I wasn't aware there was a null api option.
Are you sure this can't be used for headless server? Proper headless mode is one feature Stride is actually missing.

Jklawreszuk:
@Basewq After doing this PR, I’m also thinking about it. I found, in Xenko 3.0.0.6 or older there was a file setting Null https://github.com/stride3d/strideide/b/releases/3.0.0.6/build/Xenko.Null.PreSettings.Local.targets. Xen then tried to simplify the api setting and deleted the file, but nowhere sets Null value (commit: https://github.com/stride3d/stride/commit/784b9c988f8970feef94e0fcd74098396d4d169a). @xen2, was that intentional?

Kryptos-FR:
I think we should keep it, especially if we want to have a headless version. It's possible that the work to have a fully functional headless is not complete, and that code would be a good start.

Jklawreszuk:
Since there are no more comments I am closing the PR now

PR #2330, opened 2024-06-15
.NET 6 reference missed

Doprez:
# PR Details

Updated error message to not confuse new users.

Should probably make a central var or something to make this better since there are a few places where .NET X is hard coded.

## Types of changes

## Checklist

### DIFF
--- a/sources/assets/Stride.Core.Assets/PackageSessionPublicHelper.cs
+++ b/sources/assets/Stride.Core.Assets/PackageSessionPublicHelper.cs
@@ -20,7 +20,7 @@ namespace Stride.Core.Assets
         /// <summary>
         /// The current major .NET version that Stride will use.
         /// </summary>
-        public const byte NetMajorVersion = 8;
+        public static int NetMajorVersion => Environment.Version.Major;
 
         private static readonly string[] s_msBuildAssemblies =
         {
--- a/sources/assets/Stride.Core.Assets/PackageSessionPublicHelper.cs
+++ b/sources/assets/Stride.Core.Assets/PackageSessionPublicHelper.cs
@@ -16,6 +16,9 @@ namespace Stride.Core.Assets
     /// </summary>
     public static class PackageSessionPublicHelper
     {
+
+        public const byte NetMajorVersion = 8; //or NET_MAJOR_VERSION
+
         private static readonly string[] s_msBuildAssemblies =
         {
             "Microsoft.Build",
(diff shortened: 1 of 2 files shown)


sources/editor/Stride.GameStudio/Program.cs
```diff
@@ -224,7 +224,7 @@ private static async void Startup(UFile initialSessionPath)
             {
                 var message = "Could not find a compatible version of MSBuild.\r\n\r\n" +
                               "Check that you have a valid installation with the required workloads, or go to [www.visualstudio.com/downloads](https://www.visualstudio.com/downloads) to install a new one.\r\n" +
-                              "Also make sure you have the latest [.NET 6 SDK](https://dotnet.microsoft.com/) \r\n\r\n" +
+                              "Also make sure you have the latest [.NET 8 SDK](https://dotnet.microsoft.com/) \r\n\r\n" +
```
Jklawreszuk, line 227: Maybe separating the version into a constant wouldn't be a bad idea? What do you think?
```cs
$"Also make sure you have the latest [.NET {PackageSessionPublicHelper.netMajorVersion} SDK](https://dotnet.microsoft.com/) \r\n\r\n"
```

```cs
class PackageSessionPublicHelper{
   public const byte netMajorVersion = 8; //or NET_MAJOR_VERSION
   ....
}

**Review by Jklawreszuk — CHANGES_REQUESTED** [review]

File: sources/editor/Stride.GameStudio/Program.cs (hunk 1 above)
Comment (Doprez) [reply sources/editor/Stride.GameStudio/Program.cs:227]: makes sense to me, is there a reason for using a byte over an int?

Comment (Jklawreszuk) [reply sources/editor/Stride.GameStudio/Program.cs:227]: Not much, it's simply the smallest integer data type avaiable (8 bit instead of 32 bits)

Comment (Jklawreszuk) [reply sources/editor/Stride.GameStudio/Program.cs:227]: But probably short or int will be good enough too 😅

Comment (Doprez) [reply sources/editor/Stride.GameStudio/Program.cs]: lol fair enough, I guess we are very far away from .NET 256. Should be added now, thank you for the idea!

Comment (Jklawreszuk) [reply sources/editor/Stride.GameStudio/Program.cs]: Yeah, We will worry about that in about 248 years lol

File: sources/assets/Stride.Core.Assets/PackageSessionPublicHelper.cs [hunk 2]
```diff
@@ -16,6 +16,9 @@ namespace Stride.Core.Assets
     /// </summary>
     public static class PackageSessionPublicHelper
     {
+
+        public const byte NetMajorVersion = 8; //or NET_MAJOR_VERSION
```
Jklawreszuk, line 20: @Doprez This comment was only to suggest the alternative style of variable naming. Please, delete the comment 😅

Doprez: oop done!

Jklawreszuk (approved):

Eideren:
Would using the current runtime's dotnet version instead of a hard-written one work in those cases ?

Doprez:
Good call:
image

PR #2314, opened 2024-06-07
feat: Adjust DirectXTex wrapper for all platforms

Jklawreszuk:
# PR Details

This PR adds a Linux DxtWrapper binary file for Linux and a simple cmake script to replicate on other platforms . Additionally , my wrapper is based on the latest version of DirectXTex thanks to vcpkg - package manager for c++. I also added UTC32StringMarshaller because there is no support for Wide character on Linux.

- [ ] - My change requires testing on Windows and if that is required,, update Windows shared library too.

## Types of changes

- [ ] Docs change / refactoring / dependency upgrade
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (maybe?)

## Checklist

- [ ] My change requires a change to the documentation.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.
- [ ] **I have built and run the editor to try this change out.**

PR #2338, opened 2024-06-20
feat: Adjust DirectXTex to all platforms - second attempt

Jklawreszuk:
# PR Details

This is the second approach to adjust DirectXTex to all platforms. PR contains the same changes as in the previous PR (https://github.com/stride3d/stride/pull/2314) but without a custom Marshaller -  I did this because of the problems with marshalling `wchar_t` type in .NET - on Windows it stores UTF16 text while on Linux (and Mac?) UTF32 and there is no out of the box solution yet . I found easier solution to this by simply changing all arguments of type `const wchar_t` to `const char*` and then converting to wide string in C++.

## Types of changes

## Checklist


Eideren:
Works great, thanks :)

PR #2345, opened 2024-06-22
Update launcher brand icons

minktusk:
# PR Details

Replace old brand icons for new brand icons

## Related Issue

Any, just some icons change

## Types of changes

## Checklist

### DIFF
--- a/sources/launcher/Stride.Launcher/Resources/Urls.Designer.cs
+++ b/sources/launcher/Stride.Launcher/Resources/Urls.Designer.cs
@@ -19,7 +19,7 @@ namespace Stride.LauncherApp.Resources {
     // class via a tool like ResGen or Visual Studio.
     // To add or remove a member, edit your .ResX file then rerun ResGen
     // with the /str option, or rebuild your VS project.
-    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("System.Resources.Tools.StronglyTypedResourceBuilder", "16.0.0.0")]
+    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("System.Resources.Tools.StronglyTypedResourceBuilder", "17.0.0.0")]
     [global::System.Diagnostics.DebuggerNonUserCodeAttribute()]
     [global::System.Runtime.CompilerServices.CompilerGeneratedAttribute()]
     public class Urls {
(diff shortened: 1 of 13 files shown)


Jklawreszuk:
Thanks, It looks like this is your first PR. I have a question, where did you get the social icons from? Unfortunately, some look to be of lesser quality than the old ones

minktusk:
I get them from the official brand pages, then a resize it

VaclavElias:
Could we use SVG, if XAML supports that?

Doprez:
https://uxwing.com/x-social-media-round-icon/
Im not sure if these are better for quality @Jklawreszuk but it seems like the licenses should be good.

They also provide SVG if you want to try that route and see how they are.

Jklawreszuk:
@minktusk Oh, I see. Like @Doprez suggested, try to find SVG images and try to export them into png files (i.e using Inkscape) - this way you can be sure that the images will be rendered correctly :)
Don't forget to add the license to these logos in the https://github.com/stride3d/stride/blob/master/LICENSE.md file 😊

Jklawreszuk:
If WPF supports SVG file format then you can also replace pngs directly if you wish so. It shouldn't cause much lag

minktusk:
Thanks, I will try change the icons for SVG and see if it works properly!

minktusk:
Sorry for taking so long to update this pull request. I wasn't able to make it run SVG files. Honestly, I don't know if it's because XAML doesn't have support for it, or because I just don't know how to use XAML with SVG. So instead of using SVG, I just export them all to PNG using Inkscape ( Recommended by @Jklawreszuk ).

PS: I also changed all mentions in the code from Twitter to XTwitter!

VaclavElias:
I found this (also as NuGet) but have no experience with it, also not sure if there are other alternatives. Probably, we should keep whatever has less gears and is easier to maintain?

https://elinamllc.github.io/SharpVectors/

Eideren:
Not sure about svg but xaml supports embedded vectors, see https://github.com/stride3d/stride/blob/master/sources/editor/Stride.Assets.Presentation/View/ImageDictionary.xaml
There's a couple of tools built to make the conversion, lookup `svg to xaml`
In any case, having them as bitmaps is fine too, I'll merge this one with the rest this week end

Eideren:
Thanks !

PR #2346, opened 2024-06-23
WIP fix: make hue non nullable

IXLLEGACYIXL:
# PR Details

makes property hue non nullable as its a non allowed type for attribute properties.

## Types of changes

## Checklist

There is no usage in stride of that field thats why it was unnoticed when it was made nullable.
I dont know how to test it out. So i need help someone knowing how to test that out in a project.


Fydar:
Technically is a breaking change as `float` is a value type, so the compiler previously used `Nullable<float>` for the property type, and that's now becoming a `float`.

Looks good to me though. I doubt that's going to affect anyone.

IXLLEGACYIXL:
> Technically is a breaking change as `float` is a value type, so the compiler previously used `Nullable<float>` for the property type, and that's now becoming a `float`.
> 
> Looks good to me though. I doubt that's going to affect anyone.

previous version didnt even compile if you wanted to set it
image

as only primitives are supported, nullable isnt a primitive

Doprez:
The good news is this does let me use the CustomHue property but it doesnt seem to change anything in the editor.

image

Doprez:
This somehow breaks the dynamic colour of the Raw file thumbnails.'

uncompressed thumbnail before change
image

compressed thumbnail before change
image

with this change the thumbnail stays a single colour for some reason with or without a proper value:
image

Basewq:
> This somehow breaks the dynamic colour of the Raw file thumbnails.'

That's because `CustomHue`'s function is to affect the thumbnail, not the property header (which kinda makes this feature a bit useless)

It's only used here:
https://github.com/stride3d/stride/blob/b40ad46edc2a3ccb1b31bfb4a81eb762d42221ae/sources/editor/Stride.Editor/Thumbnails/CustomAssetThumbnailCompiler.cs#L50
Which is where the thumbnail asset then uses as the color:
https://github.com/stride3d/stride/blob/b40ad46edc2a3ccb1b31bfb4a81eb762d42221ae/sources/editor/Stride.Editor/Thumbnails/ThumbnailFromTextureCommand.cs#L79

The thumbnail only persists as a single color (for a given asset) because `displayAttribute?.Name.GetHashCode()` or `type.Name.GetHashCode()` are getting hashcodes from strings where are not persistent hashcodes between different sessions (though should persists in a single app session).

As for whether `CustomHue` ever affected the property header...? Not sure it did?
If not, in my opinion a hex string property would be more user friendly instead of using this

IXLLEGACYIXL:
so is that property just useless in the end? seems like a leftover

if oyu set hue color the color stays the same for the thumbnail sooo its anyway pointless as the intention is different?

Basewq:
Very low value since none of the existing assets use it.
I would point out that it is technically possible to inherit `Asset` (or `AssetWithSource`) and set the Display -> ColorHue for your own custom asset, however creating your own custom asset that appears in the editor involves a lot of setup, so I don't think anyone is aware of this or does this.

IXLLEGACYIXL:
currently setting hue value doesnt compile, and nobody complained

xen2:

Issue #2361, opened 2024-06-27
Error when assets in different projects with the same relative path


Basewq:
**Release Type**: Official Release

**Version**: All

**Platform(s)**: Editor/Asset compiler

**Describe the bug**
Not sure to classify this fully as a bug or an engine quirk.
If you have the two assets with the same name + relative path in different projects, the editor will not compile the asset, and will give the error:

`Error: Assets [C:/MyGame_4_2_0_2188/MyGame_4_2_0_2188/Assets/Capsule.sdpromodel] with location [Capsule] from Package [C:/MyGame_4_2_0_2188/MyGame_4_2_0_2188.Windows/MyGame_4_2_0_2188.Windows.sdpkg] is already loaded from package [C:/MyGame_4_2_0_2188/MyGame_4_2_0_2188/MyGame_4_2_0_2188.sdpkg]
`
The cause of this is:
https://github.com/stride3d/stride/blob/dbb125e19a48715dd9fa7cf200a7d3f75f2fbb78/sources/assets/Stride.Core.Assets/Analysis/AssetAnalysis.cs#L79-L83

I believe the rationale for this is because in a game, all assets in multiple referenced projects are flattened to a single 'database'.
You load the asset via `ContentManager.Load(path)` where path is the asset relative path, therefore the path must be effectively considered an ID.

**To Reproduce**
1. Create new My game template and open in Stride Editor
2. Go to Windows project and add new asset, eg. Add asset -> Model -> Capsule
3. Go to Game project and add new asset, eg. Add asset -> Model -> Capsule
4. Try save project/session.

**Expected behavior**
Not sure, explain the error message better?
Allowing it to have the same relative path sounds nice, but may break `ContentManager.Load`, unless you can specify something like a package id to tie break dupe paths.
This will have future problems when if user distributed plugins/assets come around.

**Additional context**
Somewhat related to the effect compiler issue in #301 & #446 (where it's generating a dupe effect log), but this affects more than just that.

PR #2362, opened 2024-06-27
fix: importing effects to the EffectCompileLog 

### LINKED ISSUE #301: EffectCompileLog is compiled into the Windows project
**Release Type**: Official

**Version**: 3.1.0.1-beta01-0313

**Platform(s)**: Windows

**Describe the bug**
After running the game, the imported effects are compiled into `EffectCompileLog.xkeffectlog`. However, the file is not created under the usual Assets folder, but instead under a subfolder of the Windows project.

This causes issues if an effect log file already existed in the usual Assets folder.

**To Reproduce**
1. Create a new project (for example from a template such as TopDownRPG).
2. Build and run the game from the GameStudio.
3. Import the generated effects.
4. Attempt to save

**Expected behavior**
The effect log file is created under the usual Assets folder.

**Screenshots**
image

**Additional Context**
Issue did not occur in 3.0.0.6 version.

Basewq:
# PR Details

Only one fix:
1. When checking for effect log changes, `EffectCompilerServerSession` should try to find any existing EffectCompilerLog asset in the active project (or its referenced projects, eg. the Game project) if the current package it's looking at does not contain one, and reuse that.
This is because there can only be one EffectCompilerLog asset (see #2361 for the related problem)

## Related Issue

Related to #301, but this only fixes it so it reuses the existing EffectCompilerLog asset if it exists, however if none exists it will generate the log asset in the active project's asset folder.

## Types of changes

## Checklist

### DIFF
--- a/sources/editor/Stride.Assets.Presentation/EffectCompilerServerSession.cs
+++ b/sources/editor/Stride.Assets.Presentation/EffectCompilerServerSession.cs
@@ -173,6 +173,14 @@ namespace Stride.Assets.Presentation
         private void CheckEffectLogAsset(PackageViewModel package)
         {
             var newEffectLogViewModel = package.Assets.FirstOrDefault(x => x.Name == EffectLogAsset.DefaultFile) as EffectLogViewModel;
+            if (newEffectLogViewModel is null && session.CurrentProject is not null)
+            {
+                // Check if the EffectLog asset exists in any of the active project's referenced local projects
+                var projectDependencies = session.CurrentProject.PackageContainer.FlattenedDependencies.Where(x => x.Type == DependencyType.Project).ToList();
+                var localPackageDependencies = session.LocalPackages.Where(pkgViewModel => projectDependencies.Any(projDep => projDep.Package == pkgViewModel.Package));
+                var localPackagesAssets = localPackageDependencies.SelectMany(x => x.Assets);
+                newEffectLogViewModel = localPackagesAssets.FirstOrDefault(x => x.Name == EffectLogAsset.DefaultFile) as EffectLogViewModel;
+            }
             var newEffectLogText = newEffectLogViewModel?.Text;
 
             if (newEffectLogText != effectLogText // Asset changed?


Ethereal77:
Shouldn't the default be the executable project?
If I recall correctly, the effect log enumerates the effects and their required parameters during the application's runtime, making each effect log specific to the application and, consequently, to the platform.

This implies that an effect log for a DX+Win application would differ from those for a Vk+Linux or a GLES3+Android application, right?

Basewq:
Damn, you're asking a hard question there.

My limited understanding is that when you build the project you feed the CompilerApp the platform + graphics api, so the CompilerApp can read the effect log and read the actual shader file to produce the proper shader code in the resource/database.

This is a snippet of the content in the effect log:
```
!EffectCompileRequest
EffectName: StrideForwardShadingEffect
UsedParameters:
    Material.PixelStageSurfaceShaders: !ShaderMixinSource
        Mixins:
            - !ShaderClassSource
                ClassName: MaterialSurfaceArray
        Compositions:
            layers: !ShaderArraySource
                Values:
                    - !ShaderMixinSource
                        Mixins:
                            - !ShaderClassSource
                                ClassName: MaterialSurfaceDiffuse
                        Compositions:
                            diffuseMap: !ShaderClassSource
                                ClassName: ComputeColorConstantColorLink
                                GenericArguments: [Material.DiffuseValue]
                    - !ShaderMixinSource
                        Mixins:
                            - !ShaderClassSource
                                ClassName: MaterialSurfaceGlossinessMap
                                GenericArguments: [false]
                        Compositions:
                            glossinessMap: !ShaderClassSource
                                ClassName: ComputeColorConstantFloatLink
                                GenericArguments: [Material.GlossinessValue]
                    - !ShaderMixinSource
                        Mixins:
                            - !ShaderClassSource
                                ClassName: MaterialSurfaceMetalness
                        Compositions:
                            metalnessMap: !ShaderClassSource
                                ClassName: ComputeColorConstantFloatLink
                                GenericArguments: [Material.MetalnessValue]
                    - !ShaderMixinSource
                        Mixins:
                            - !ShaderClassSource
                                ClassName: MaterialSurfaceLightingAndShading
                        Compositions:
                            surfaces: !ShaderArraySource
                                Values:
                                    - !ShaderClassSource
                                        ClassName: MaterialSurfaceShadingDiffuseLambert
                                        GenericArguments: [false]
                                    - !ShaderMixinSource
                                        Mixins:
                                            - !ShaderClassSource
                                                ClassName: MaterialSurfaceShadingSpecularMicrofacet
                                        Compositions:
                                            environmentFunction: !ShaderClassSource
                                                ClassName: MaterialSpecularMicrofacetEnvironmentGGXLUT
                                            fresnelFunction: !ShaderClassSource
                                                ClassName: MaterialSpecularMicrofacetFresnelSchlick
                                            geometricShadowingFunction: !ShaderClassSource
                                                ClassName: MaterialSpecularMicrofacetVisibilitySmithSchlickGGX
                                            normalDistributionFunction: !ShaderClassSource
                                                ClassName: MaterialSpecularMicrofacetNormalDistributionGGX
    Material.PixelStageStreamInitializer: !ShaderMixinSource
        Mixins:
            - !ShaderClassSource
                ClassName: MaterialStream
            - !ShaderClassSource
                ClassName: MaterialPixelShadingStream
    Lighting.DirectLightGroups:
        - !ShaderMixinSource
            Mixins:
                - !ShaderClassSource
                    ClassName: LightDirectionalGroup
                    GenericArguments: [8]
    Lighting.EnvironmentLights:
        - !ShaderClassSource
            ClassName: LightSimpleAmbient
    StrideEffectBase.RenderTargetExtensions: !ShaderMixinSource
        Macros:
            -   Name: STRIDE_RENDER_TARGET_COUNT
                Definition: 1
            -   Name: STRIDE_MULTISAMPLE_COUNT
                Definition: 1
```

Attached is the full file (renamed .txt) [EffectCompileLog.txt](https://github.com/user-attachments/files/16012416/EffectCompileLog.txt)


I ran both an Android build & a Windows build and they gave the same effect log, though that doesn't mean this would always be the case? Is there a way to make a divergent effect on a log level?
Unfortunately, I don't have enough shader/graphics setup knowledge here.
Maybe it would just be treated as another shader variant?

Also, if it has to be on a platform level, the workflow is gonna be really bad because the editor detection doesn't know what platform the effect being compiled for, and you would have to manually set the active project to the correct platform every time before importing...

Ethereal77:
The original design had the effect log generation occur at runtime by default, meaning the effects and their parameters were identified during gameplay, and the Game Studio recorded them via IPC into the effect log.

However, this process was found to be cumbersome and inconvenient for some during development and experimentation. Consequently, Xen (if I recall correctly) opted to set the runtime discovery and recording to disabled by default, though it can be reactivated in the settings if desired.

Essentially, the effect log was created for the application during execution (I think that's why it is name _log_), ensuring that the compiled effects matched those required for the actual application run.

Basewq:
Right, but is this log supposed to be sharable between platforms or separate? ie. should the log be the `.Game` project or one for each platform?
My expectation is that it's meant to be in the common project, the CompilerApp should generate the correct shader at build time from the log, but it seems this isn't the case?

For the most part, I can just swap the default to generate it on the platform if that's the case.
If developers want it on the Game project they can just move the asset and it'll still work.

Ethereal77:
I don't really know. I haven't investigated this much yet. Maybe someone with more experience in this area can give a clear picture of the effects log.

Basewq:
Ok, redone with a more cautious approach.
Updated the main post, but in summary this change will only fix it so it will always reuse the existing EffectCompilerLog asset if it exists (with the project's dependency).
If none exists, then it will retain its current behavior, ie. create a new log asset in the active project.
Whether that is desired behavior or not is beyond the scope of this pull.

Eideren:
LGTM, what says @Ethereal77 ?

Ethereal77:
LGTM also.

I just wanted to verify my previous statements with someone more familiar with the original purpose of the effect log. Additionally, we should enhance the documentation to make these points clearer.

Eideren:
You could ask xen directly, not sure who actually took care of this specific part of the engine in the paradox days.
Feel free to open a new issue/pr for the second point

Eideren:
Thanks @Basewq !

PR #2363, opened 2024-06-28
fix: [DirectoryWatcher] - Folders are case sensitive on Linux

Jklawreszuk:
# PR Details

For some reason Path is transformed to lowercase, but on Linux distros it causes a runtime error - "The directory name .... does not exist" during game runtime.

## Types of changes

## Checklist

### DIFF
--- a/sources/core/Stride.Core.IO/DirectoryWatcher.Desktop.cs
+++ b/sources/core/Stride.Core.IO/DirectoryWatcher.Desktop.cs
@@ -135,7 +135,7 @@ namespace Stride.Core.IO
 
             if (path != null && Directory.Exists(path))
             {
-                info = new DirectoryInfo(path.ToLowerInvariant());
+                info = new DirectoryInfo(OperatingSystem.IsLinux() ? path :  path.ToLowerInvariant());
             }
             else
             {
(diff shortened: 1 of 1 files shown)


Eideren:
Wouldn't it make sense to remove the `ToLowerInvariant` call entirely ? If it works on linux it will work on windows since it is the more lenient of the two ? Or do we have some specific handling per os here that breaks this assumption ?

IXLLEGACYIXL:
```
                if (needProjectReload == false
                    && ((e.ChangeType == FileEventChangeType.Deleted || e.ChangeType == FileEventChangeType.Renamed || e.ChangeType == FileEventChangeType.Created)
                    && Path.GetExtension(changedFile)?.ToLowerInvariant() == ".cs"
                    && changedFileDirectoryName.StartsWith(directoryName, StringComparison.OrdinalIgnoreCase)))
                {
                    needProjectReload = true;
                }
```
would break other parts that have path comparisons

Eideren:
That example would not break on windows afaict though, on linux it may if you have two directory whose name differs only in character casing. What would break on windows is string comparison without `IgnoreCase` specified.

IXLLEGACYIXL:
on linux it needs  a ordinal comparison, on windows it doesnt matter

Jklawreszuk:
What, then, do you finally suggest? I would like to avoid regression thus I added a IsLinux condition to not break anything

Eideren:
Okay, probably best to leave this for later, can you add a small comment above those lines mentioning that it might be safe to remove that check but to investigate that further on windows ?

Jklawreszuk:
Yeah, taking small steps 😅 Alright , I've added a TODO comment as you suggested.

Eideren:
Thanks !

Issue #2172, opened 2024-03-04
SDF Font Kerning and vertical tracking incorrect


thygrrr:
**Release Type**: Official 4.2

**Version**: 4.2.0.2067 on .NET 8.0.2

**Platform(s)**: Windows 11

**Describe the bug**
Font kerning for SDF fonts is incorrect, letters are squished together and dance up and down.

**To Reproduce**
1. download font Baj Jamjuree from Google Fonts
2. create SDF font, size 20 or 24
3. create a text block and zoom in (sometimes easier to see at runtime)

![image](https://github.com/stride3d/stride/assets/8904620/d9bcc015-2b91-41f0-ac0e-bab6ab323180)

**Expected behavior
Kerning is as defined and hinted in the font. Tracking and scanning is as defined and hinted in the font.
![image](https://github.com/stride3d/stride/assets/8904620/60702072-b3fa-4530-8717-afa105e9f76c)

**Additional context**
Related to #2171 (there's a sample project there)

PR #2380, opened 2024-07-09
[Font] Fix glyph bitmap alignment from SDF fonts

Basewq:
# PR Details

Fix horizontal/vertical alignment of glyphs from generated SDF bitmaps.
According to [the original author](https://github.com/Chlumsky/msdfgen/issues/108#issuecomment-657286643) `-autoframe` more meant for previewing rather than actual usage due to it centering the glyph to fit exactly centered on the bitmap rather than how it should be fitted on the bitmap (ie. snapped to the top-right corner).

Other change is debug build, dump the SDF bitmaps in into a sub-folder in the temp directory instead of straight into the temp directory so it's easier to delete. This does not affect the release build.

## Related Issue

#2172 Only fixes the vertical alignment, not the kerning issue (which I have no clue about).

Merely an FYI, but this is the original logic for autoframe:
https://github.com/stride3d/msdfgen/blob/1af188c77822e447fe8e412420fe0fe05b782b38/main.cpp#L684-L709

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Assets/SpriteFont/Compiler/SignedDistanceFieldFontImporter.cs
+++ b/sources/engine/Stride.Assets/SpriteFont/Compiler/SignedDistanceFieldFontImporter.cs
@@ -6,13 +6,11 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using Stride.Core.Assets;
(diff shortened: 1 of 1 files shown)


Eideren:
Thanks !

PR #2385, opened 2024-07-13
feat: Hide default font name for SystemFontProvider, change default font

Jklawreszuk:
# PR Details

My PR tries to hide default default font name for SystemFontProvider. I also changed default font name from `Courier New` to `Arial` for sake of consistency - (Almost) all samples are using `Arial` font anyway.
In addition, `Arial` is a sans-serif font - these types of fonts are more popular and easier to read.

- [x] - This also fixes samples / templates for systems where Arial font is not available (i.e Linux)

## Types of changes

- [x] Breaking change (kinda, because this changes font :P)

## Checklist

### DIFF
--- a/samples/Games/JumpyJet/Assets/Shared/Font.sdfnt
+++ b/samples/Games/JumpyJet/Assets/Shared/Font.sdfnt
@@ -2,7 +2,5 @@
 Id: f10885f3-72a0-469e-9ba2-6ca635fee4e7
 SerializedVersion: {Stride: 2.0.0.0}
 Tags: []
-FontSource: !SystemFontProvider
-    FontName: Arial
-    Style: Regular
+FontSource: !SystemFontProvider {}
 FontType: !RuntimeRasterizedSpriteFontType {}
--- a/samples/Games/SpaceEscape/Assets/Shared/Font.sdfnt
+++ b/samples/Games/SpaceEscape/Assets/Shared/Font.sdfnt
@@ -2,7 +2,5 @@
 Id: aa317d9b-bbd3-4240-9ad6-11efad9c4d4d
 SerializedVersion: {Stride: 2.0.0.0}
 Tags: []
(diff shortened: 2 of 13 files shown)


Eideren:
So, this PR is mostly to ensure that by default samples use a font that is compatible with the system it was created on, right ?
> This also fixes samples / templates for systems where Arial font is not available (i.e Linux)
You meant to say that `Courrier` is not available in Linux, right otherwise your changes to make the default font for font assets `Arial` doesn't make a lot of sense ?

Jklawreszuk:
> So, this PR is mostly to ensure that by default samples use a font that is compatible with the system it was created on, right ?

@Eideren Correct 😅 On most systems it will be the Arial font but on Linux does not. On most distros most common one would be `Liberation` (i.e. LibreOffice use it as default font). Thus, I would prefer to void direct font calling for samples and default font.

> You meant to say that Courrier is not available in Linux, right otherwise your changes to make the default font for font assets Arial doesn't make a lot of sense ?

Both `Arial` and `Courier New` technically are available on any OS (no one stops you from installing MS fonts, [Example: MS Fonts for Alpine Linux](https://pkgs.alpinelinux.org/package/edge/community/x86/msttcorefonts-installer)). But that's too anoying to install another dependency and most distributions offer Microsoft font replacements (the replacement for `Times New Roman` is  `Liberation Serif` , for `Arial` is  `Liberations Sans` etc)

Eideren:
Right, so not specifying any font would automatically pick one of the fonts available on the current OS ?

Jklawreszuk:
I tested it by running Gamestudio and indeed it sets to `Arial` as expected. 
I have one more observation : It is possible to set back to `Arial` font explicitly and but I think that a matter for another PR to detect "default" font in the GameStudio editor

Eideren:
Thanks for the clarifications

PR #2392, opened 2024-07-16
Add back Windows specific CompilerApp to fix libraries with WPF references

### LINKED ISSUE #2244: AssetCompiler error on multi-project configuration

4.2.0.2122

**Platform(s)**: Does the problem occur on Windows, Android...?
Windows/WPF

**Describe the bug**

I have created a pretty simple solution that contains 2 projects. The main game project and a WPF project. I based both of these off of the examples provided here and here.

On a fresh repo pull when compiling the solution for x64 only I get the following error message

image

I then modify the csproj files to compile under AnyCPU everything works. After this I can then switch the projects back to x64 and compile/run with no issues including after full clears of the bin/obj folders.

There was some discussion of this on the Discord starting here

**To Reproduce**
1. Open the repo in this zip file [TowerNxt-master.zip](https://github.com/stride3d/stride/files/15301567/TowerNxt-master.zip)
272)
2. Build the solution to see the error above
3. Change the projects/solution to compile under AnyCPU. I did this by adding the AnyCPU config to the solution, modifying the csproj directly to be <Platforms>x64;AnyCPU</Platforms>, correcting the configs, then removing the <Platforms> tags all together so it only has AnyCPU
4. Build and find no issues
5. Switch projects/solution back to x64. I did this by manually adding <Platforms>x64</Platforms> to the csproj file
6. Build

**Expected behavior**
I would have expected the solution to compile under x64 configuration

**Screenshots**
See above

**Log and callstacks**
Not possible

**Additional context**
Can discuss with Simba (@Doprez) and Vašo on discord for additional information.

Thank you.

Basewq:
# PR Details

May be controversial to bring the windows specific CompilerApp back.
Pinging @Jklawreszuk & @Kryptos-FR since they've spent a lot of time on the cross-platform transition side of things.
This does not remove the cross-platform version, but rather includes the Windows specific version and use that over the cross-platform version.

## Related Issue

Fixes #2244
The problem is referencing any Stride libraries that touch WPF code, the ModuleInitializer will try to load the WPF framework if it's not loaded, however WPF is Windows only, so it can't be loaded in any way except by having the Windows specific version.

## Types of changes

## Checklist

### DIFF
--- a/sources/assets/Stride.Core.Assets.CompilerApp/Stride.Core.Assets.CompilerApp.csproj
+++ b/sources/assets/Stride.Core.Assets.CompilerApp/Stride.Core.Assets.CompilerApp.csproj
@@ -6,7 +6,7 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <StrideAssemblyProcessor>true</StrideAssemblyProcessor>
-    <TargetFramework>$(StrideXplatEditorTargetFramework)</TargetFramework>
+    <TargetFrameworks>$(StrideXplatEditorTargetFramework);$(StrideEditorTargetFramework)</TargetFrameworks>
     <AppendRuntimeIdentifierToOutputPath>false</AppendRuntimeIdentifierToOutputPath>
     <StrideAssemblyProcessorOptions>--auto-module-initializer --serialization</StrideAssemblyProcessorOptions>
   </PropertyGroup>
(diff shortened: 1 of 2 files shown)


Kryptos-FR:
I don't see any problem with double targeting for now.

That said what is the use case? Stride libraries beside the one used exclusively for the editor aren't supposed to have any dependencies with WPF. And the ones for the editor aren't expected to be used for Games.

Basewq:
> That said what is the use case?

I believe there were a few people in discord that attempted to host Stride in their own WPF application, or something like that (I'm not 100% sure).

[My own example project](https://github.com/Basewq/XenkoProofOfConcepts/tree/master/LevelEditorExtensionExample) tried to show a way to extend the editor, which requires referencing the editor specific libraries to access certain objects/methods.
https://github.com/Basewq/XenkoProofOfConcepts/blob/046c2ce82300ad8ae7d5e9d965e620d442dc789c/LevelEditorExtensionExample/LevelEditorExtensionExample/Editor/LevelEditEditorProcessor.cs#L117
While I actually don't use this, it does show that it's possible to do, although it also shows how difficult it really is to try to extend the editor.

Kryptos-FR (approved):

Eideren:
Thanks !

PR #2405, opened 2024-08-04
Fix shader bool generics to appear in the editor

Basewq:
# PR Details

Bug in editor where declaring a shader, eg.
```
shader MyShader<bool ShaderArg>
{
}
```
won't show the bool argument in the editor when used as a material shader, and you won't be able to build the shader due to the missing parameter.

This fix exposes the bool argument to the editor and allows it to save/load properly.

Also added a separate commit to simplify the type checking.

## Types of changes

## Checklist


Eideren:
Thanks !

Issue #760, opened 2020-06-14
Not Adding Physics Shape To The Character Causes Unhandled Exception


ghost:
**Release Type**: Official Release

**Version**: Version 4.0

**Platform(s)**: All

**Describe the bug**
if you don't add a physics shape to the character component and call any physics function for that character you'll get the following error : Unhandled Exception: System.InvalidOperationException: Attempted to call a Physics function that is avaliable only when the Entity has been already added to the Scene.

**To Reproduce**
1. add a characterComponent to an entity
2. create a new script and add it to that entity
3. in the script create a characterComponent variable then reference it through either game studio or code
4. call any function for that characterComponent such as SetVelocity
5. try running the game and you'll get an error

**Expected behavior**
have a clear error that says "character doesn't have physics shape" or something similar

r00terAF18:
Hello, I would like to give this bug a try, any guidance? Thank you

tebjan:
@r00terAF18 hello, thanks for having a look at this, very nice to hear. i think the first step would be to reproduce the bug with visual studio attached and break at the exception.

then have a look in the debugger whether the error case could be handled earlier, e.g. inside the `CharacterComponent` class. for example, by checking whether a physics shape exists.

dloe:
Hello! I was looking and I wasn't sure if this issue was being worked on. If not, I would love to give it a try, Thanks!

PR #2410, opened 2024-08-13
feat: No Physics Shapes On The CharacterComponent Causes Unhandled Exception 

dloe:
# PR Details

Having no Physics shapes on character component causes unhandled exceptions when user tries to use physics methods on corresponding character component. (Note: Originally showed as a bug)
Originally threw: Unhandled Exception: System.InvalidOperationException: Attempted to call a Physics function that is available only when the Entity has been already added to the Scene.

Now gives warning from logger when a character component has no shapes and shows error via logger when physics method is ran on character component (Much earlier than original exception). Also gives good logging info to user as to where the physics method was called for tracing purposes.

## Related Issue

Issue: https://github.com/stride3d/stride/issues/760

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Physics/Elements/CharacterComponent.cs
+++ b/sources/engine/Stride.Physics/Elements/CharacterComponent.cs
@@ -264,7 +264,6 @@ namespace Stride.Physics
                 logger.Error($"Component:[{this}] attempted to call a Physics function that is available only when the Entity has been already added to the Scene. " +
                     $"This may be due to a {this} without any physical shapes.\nLocation: {frame.GetFileName()} at Line Number: {frame.GetFileLineNumber()} from Method: {frame.GetMethod().Name} ");
                 return;
-                //throw new InvalidOperationException("Attempted to call a Physics function that is available only when the Entity has been already added to the Scene.");
             }
 
             KinematicCharacter.SetWalkDirection(velocity * Simulation.FixedTimeStep);
--- a/sources/engine/Stride.Physics/Elements/CharacterComponent.cs
+++ b/sources/engine/Stride.Physics/Elements/CharacterComponent.cs
@@ -27,9 +27,7 @@ namespace Stride.Physics
         {
             if (KinematicCharacter == null)
             {
-                StackFrame frame = new StackTrace(true).GetFrame(1); //for capturing tracing info
-                logger.Error($"Component:[{this}] attempted to call a Physics function that is available only when the Entity has been already added to the Scene. " +
-                    $"This may be due to a {this} without any physical shapes.\nLocation: {frame.GetFileName()} at Line Number: {frame.GetFileLineNumber()} from Method: {frame.GetMethod().Name} ");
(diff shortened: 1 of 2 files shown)


sources/engine/Stride.Physics/Engine/PhysicsComponent.cs
```diff
@@ -645,6 +645,12 @@ internal void Attach(PhysicsProcessor.AssociatedData data)
             //this is mostly required for the game studio gizmos
             if (Simulation.DisableSimulation)
             {
+                //Give warning if CharacterComponent doesnt have collider shapes
+                if(this is CharacterComponent && ColliderShapes.Count == 0 && ColliderShape == null)
+                {
+                    logger.Warning($"Entity {this} has no physical shapes attatched to CharacterComponent.");
+                }
+
```
Eideren, line 653: This makes sense but it is kind of a duplicate of the check further below, might as well just keep one of them and have it above the `if (Simulation.DisableSimulation)` test

Eideren:
Thanks a bunch, I think your stacktrace and return is the right idea but you might want to wrap that inside a method instead. And log `GetFrame(2)`.

sources/engine/Stride.Physics/Elements/CharacterComponent.cs
```diff
@@ -262,6 +279,7 @@ public void SetVelocity(Vector3 velocity)
         [DataMemberIgnore]
         internal BulletSharp.KinematicCharacterController KinematicCharacter;
 
+        //called after collider shapes are attached to character comp?
```
Eideren, line 282: This method is called in PhysicsComponent.Attach() which is called whenever an entity with that component is added to the scene https://github.com/stride3d/stride/blob/master/sources/engine/Stride.Physics/Engine/PhysicsProcessor.cs#L136 
You can remove this comment, and add it as a summary to https://github.com/stride3d/stride/blob/master/sources/engine/Stride.Physics/Engine/PhysicsComponent.cs#L703 instead but it's fine if you don't

dloe: Good idea, this was a leftover comment I had when I was annotating while learning how this area works. Should have been removed and accidently left it in. I can move that to the PhysicsComponent OnAttach 👍 gives decent context.

dloe:
> Thanks a bunch, I think your stacktrace and return is the right idea but you might want to wrap that inside a method instead given the amount of code duplication. And log `GetFrame(2)` instead once that is done of course.

Good idea, I moved it to a dedicated function inside of CharacterComponent, but could it be moved even farther out to maybe the Logger.cs file?

Eideren:
> could it be moved even farther out to maybe the Logger.cs file?

I'm not sure we should make it part of the public api, in this exceptional case it does make sense, but we rarely need this log as it is fairly brittle; the second caller up the stack may not be as important to log as the third, or the fourth, and creating a stacktrace is a fairly slow operation.

Thanks a bunch for taking care of this one :)

Issue #2005, opened 2023-10-30
Can't change CharacterComponent's collider shape properties


dawnmichal:
**Release Type**: Official Release

**Version**: 4.1.0.1948

**Platform(s)**: Windows

**Describe the bug**

Assigning a new ColliderShape into CharacterComponent in code (when game is running) and then calling CharacterComponent.ComposeShape(); causes the application to crash.

**To Reproduce**
1. Create basic character with CharacterComponent.
2. Assign capsule collider shape using Stride Studio.
3. Assign script which perform this code (ex. when key is pressed)

characterComponent.ColliderShapes[0] = new CapsuleColliderShapeDesc() 
{ 
     Length = 0.6f, 
     LocalOffset = new Vector3(0.0f, 0.70f, 0.0f), 
     Orientation = ShapeOrientation.UpY, Radius = 0.4f 
};

characterPhysics.ComposeShape();

4. This will caus the application to crash.

**Expected behavior**
CharacterComponent starts to use new collider shape with new property values...

**Log and callstacks**

- I can see this in Windows Application Events:
dDrive.Windows.exe 
   1.0.0.0 
   64e81669 
   libbulletc.DLL 
   0.0.0.0 
   5ccee150 
   c0000409 
   0000000000166069 
   3b48 
   01da0915ea6c6440 
   C:\.....\Bin\Windows\Debug\win-x64\dDrive.Windows.exe 
   C:\.....\Bin\Windows\Debug\win-x64\libbulletc.DLL 
   6959c4f8-c585-472a-967f-89ccbed61439

**Additional context**
- I need to change collider shape because of different character poses, like crouching and so on...

- It seems trying to use Dispose() on any type of collider shape causes it. `ColliderShape.Dispose();` is called inside `Stride.Engine.PhysicsComponent.ComposeShape()` method. 

- @Manio143 tip from discord: 

> I imagine we don't properly remove some reference after the dispose frees the native handle to the bullet collider and we still attempt to use it afterwards which is what causes the crash.

manio143:
More context - in the `CharacterComponent.OnAttach` we create a new instance of `PairCachingGhostObject` from BulletSharp which is used to instantiate `KinematicCharacterController` from BulletSharp. We provide it a handle to the collider shape.
https://github.com/stride3d/stride/blob/e200843806af650ad45595f87a78374712ecaeea/sources/engine/Stride.Physics/Elements/CharacterComponent.cs#L267-L271
[...]

PR #2428, opened 2024-09-04
bug: Unable to Change CharacterComponent's collider shape properties without causing crash

dloe:
# PR Details

Assigning a new `ColliderShape` into `CharacterComponent` and running `ComposeShape()` causes a UAF related exception to be thrown. Related to an issue with the `KinematicCharacter` reference inside of the `discreteDynamicWorld` in the `Simulation` object. 

Solution would be to add additional setup to a overridden `ComposeShape()` method inside CharacterComponent. This setup would include the complete removal, reconstruction and reintegration of the KinematicCharacterController to the Simulation. Also added an additional function to try and maintain various physics properties onto the `KinematicCharacter` such as Gravity, Fallspeed, etc. 

## Related Issue

https://github.com/stride3d/stride/issues/2005

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Physics/Elements/CharacterComponent.cs
+++ b/sources/engine/Stride.Physics/Elements/CharacterComponent.cs
@@ -301,19 +301,16 @@ namespace Stride.Physics
         }
 
         /// <summary>
-        /// Reconstruct of character controller to avoid UAF crash related to disposed ref to native object
-        /// 
-        /// - NativeCollisionObject shows old refs
-        /// - Therefore KinematicCharacter also holds old refs
-        /// 
-        /// - Tests show nativeCollisionObject being remade does not cause the read while use exception, its just kinematic character
+        /// Reconstruct of character controller to avoid possible UAF issues
+        /// Context: On ColliderShape changes, when ComposeShape is ran to rebuild ColliderShape properties Disposing the old ColliderShape can cause UAF
+        /// issues inside KinematicCharacter and inside Simulation discreteDynamicWorld due to old disposed references to the native object.
         /// </summary>
         public override void ComposeShape()
         {
             //Disposing of ColliderShape should happen before we remove NativeCollisionObject and KinematicCharacter from Simulation
             base.ComposeShape();
 
-            //make PairCachingGhostObject valid, therefore make new instance of it and/or kinematic character controller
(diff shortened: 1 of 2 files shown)


sources/engine/Stride.Physics/Elements/CharacterComponent.cs
```diff
@@ -300,6 +300,54 @@ protected override void OnAttach()
...
+                KinematicCharacter.Dispose();
+                KinematicCharacter = null;
+                BulletSharp.Math.Vector3 unitY = new BulletSharp.Math.Vector3(0f, 1f, 0f);
+                //Make new KinematicCharacter
+                KinematicCharacter = new BulletSharp.KinematicCharacterController((BulletSharp.PairCachingGhostObject)NativeCollisionObject, (BulletSharp.ConvexShape)ColliderShape.InternalShape, StepHeight, ref unitY);
+                OverrideKinematicCharacterValues(kinematicCharacterProperties);
+
+                //now we can add these references BACK in Simulation's discreteDynamicsWorld
+                Simulation.AddCharacter(this, (CollisionFilterGroupFlags)CollisionGroup, CanCollideWith);
+            }
```
Eideren, line 335: It seems like virtualizing `ComposeShape` may not be necessary as that method runs right before `OnAttach`, 
https://github.com/stride3d/stride/blob/4ebbf54839cf762ef8e9d135c78f453cf30a723f/sources/engine/Stride.Physics/Engine/PhysicsComponent.cs#L660-L672
So, it you could move this section inside `OnAttach` instead and have it work the same way on top of being able to reduce logic duplication ? I might have missed something though ... let me know !

Eideren (commented):
Looks pretty good, just a small question

sources/engine/Stride.Physics/Elements/CharacterComponent.cs
dloe: That is a good thing to point out. However due to the fact that users running `ComposeShape` directly in their projects for this scenario can cause the crash, I ended up using the virtual `ComposeShape` approach. Sorry for that delay, I wanted to run some more tests just to be sure and wasnt at my normal computer.

Eideren: Makes sense, good thinking - we're all working on our free time, don't worry about the delay !

Eideren:
Thanks !

Issue #2376, opened 2024-07-07
Grayscale textures are broken in 4.2.0.2188


Doprez:
**Release Type**: Official Release

**Version**: 4.2.0.2188

**Platform(s)**: Windows

**Describe the bug**
When importing a grayscale file into the editor it will be completely broken. 

![image](https://github.com/stride3d/stride/assets/73259914/bff5479c-8c09-4dbc-aafb-28fa036388f6)

you can make it at least shows something if you uncheck all of the options in the feature panel but it just shows white noise.

![image](https://github.com/stride3d/stride/assets/73259914/e073c25f-3bb4-4093-902b-0606dae65baa)

This is how it looks in 4.2.0.2122:

![image](https://github.com/stride3d/stride/assets/73259914/6f69a573-11bf-4130-b547-41f2a45af28b)

**To Reproduce**
1. get a grayscale image 
![BrickWall02_roughness](https://github.com/stride3d/stride/assets/73259914/ad5088b6-ee8a-45ba-9f82-da7a76848374)

2. import it into the editor

**Expected behavior**
Textures should look good.

Doprez:
Some added info this seems to have broken after 2149 as it works in that gamestudio version as expected.
https://github.com/stride3d/stride/compare/releases/4.2.0.2149...releases/4.2.0.2188

Doprez:
`System.DivideByZeroException: 'Attempted to divide by zero.'`
```
This exception was originally thrown at this call stack:
    Stride.Graphics.PixelFormatExtensions.SizeInBits(Stride.Graphics.PixelFormat) in PixelFormatExtensions.cs
    Stride.TextureConverter.TexLibraries.DxtTexLib.ChangeDxtImageType(Stride.TextureConverter.TexLibraries.DxtTextureLibraryData, Stride.TextureConverter.DxtWrapper.DXGI_FORMAT) in DxtTexLib.cs
    Stride.TextureConverter.TexLibraries.DxtTexLib.Load(Stride.TextureConverter.TexImage, Stride.TextureConverter.Requests.LoadingRequest) in DxtTexLib.cs
    Stride.TextureConverter.TexLibraries.DxtTexLib.Execute(Stride.TextureConverter.TexImage, Stride.TextureConverter.IRequest) in DxtTexLib.cs
    Stride.TextureConverter.TextureTool.Load(Stride.TextureConverter.Requests.LoadingRequest) in TextureTool.cs
[...]

PR #2435, opened 2024-09-08
fix: Restore LoadWICFile for Windows due to incomplete cross-platform functionality

Basewq:
# PR Details

Partial revert of #2314 and #2338 for Windows to allow loading non-32-bit pixel formats.

Note that (as far as I'm aware) DxtWrapper.dll for Windows was not rebuilt with the 'dxtLoadWICFile' function removed, so it also means it does not need to be rebuilt for this revert.

## Related Issue

Problem raised in #2376.
Technically it only fixes it for Windows, but the issue will persist for the non-Windows editor (which doesn't exist yet) and will require implementation/alternative solution if the intention is to remove it again.

## Types of changes

## Checklist


Eideren:
Looks good, thanks !

PR #2422, opened 2024-08-27
fix: PhysicsComponent created with GetOrCreate<T> is unusable

dloe:
# PR Details

When using `GetOrCreate<T>` to create physics components such as `Rigidbody` component, certain properties such as `ColliderShape` are not properly setup. Setting other properties afterwards result in a no-op.

Added a `ReAttach` method to `PhysicsComponent` that can run and reconfigure those properties when `ColliderShapes` is added. Also added a Test case to showcase. 

## Related Issue

Issue: https://github.com/stride3d/stride/issues/1504

## Types of changes

- [ ] Docs change / refactoring / dependency upgrade
- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [ ] My change requires a change to the documentation.
- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.
- [X] **I have built and run the editor to try this change out.**

Issue #1504, opened 2022-08-16
PhysicsComponent created with `GetOrCreate<T>` is unusable


manio143:
**Release Type**: Official Release

**Version**: any

**Describe the bug**
When a physics component, such as `RigidbodyComponent` is obtained through `GetOrCreate<T>` method on an entity, it is attached before the user had a chance to configure it. There's checks in the internal `Attach` method on the physics component which checks if a collider shape is set and if not, it exits without performing additional step - this makes the component enter a state where setting any of the properties afterwards results in no-op.

**To Reproduce**

```csharp
var body = Entity.GetOrCreate<RigidbodyComponent>();
body.ColliderShape = new SphereColliderShape(false, 1.0f);
Debug.Assert(body.ColliderShape != null);
```

**Expected behavior**
Either we should allow the user to add a physics component in this way and modify the internals such that if a collider shape is set after attachment it performs work to bring the component back into good shape; Or we need to allow component creators to disallow them being instantiated with `GetOrCreate` and force users to create the component themselves, by throwing an explanatory exception.

**Additional context**
https://github.com/stride3d/stride/blob/b11317c9389b33bce5bc2de550a0e501dc024694/sources/engine/Stride.Physics/Engine/PhysicsComponent.cs#L685-L689
This makes us not call OnAttach a few lines later which means we don't execute
https://github.com/stride3d/stride/blob/b11317c9389b33bce5bc2de550a0e501dc024694/sources/engine/Stride.Physics/Elements/RigidbodyComponent.cs#L305-L309
And thus `InternalRigidBody` is `null`
https://github.com/stride3d/stride/blob/b11317c9389b33bce5bc2de550a0e501dc024694/sources/engine/Stride.Physics/Elements/RigidbodyComponent.cs#L103-L117

dloe:
Hello! If possible I would like to work this issue! Had a quick question, in the reproduction of this bug I'm assuming this is done in just an simple example project and not in the test suites. Is there an expectation that the Debug.Assert displays a message?

Other than that, I understand and see the issue with InternalRigidBody and other code not being set up properly.

In terms of which of the two suggested solutions to go for, maybe a good starting place could be adding some type of verification on the RigidBodyComponent setter that could properly run the missing OnAttach setup?

IXLLEGACYIXL:
https://github.com/stride3d/stride/pull/2131
As physics is getting reworked is it worth to fix bullet implementation?

Doprez:
[...]

PR #2439, opened 2024-09-09
fix: Found and fixed bug pertaining to ReAttach method for PR #2422

dloe:
# PR Details

Test cases inside `Stride.Physics.Tests.SendCollisionEndedWhenEntityIsRemovedTest` suite started failing after introduction of PR https://github.com/stride3d/stride/pull/2422. Further investigation showed that while entities are being setup inside `Attach`, the `ReAttach` method could potentially run and cause duplicate entities to be added to simulation, resulting in possible null ref exceptions and in this case, test failures.

Added flag `attachInProgress` that only allows the ReAttach method to run when Entity has fully competed all `Attach` behaviors. Still maintains same behavior that was tested and verified in `Stride.Physics.Tests.ColliderShapesTest.VerifyColliderShapeSetup`.

## Related Issue

Original Issue: https://github.com/stride3d/stride/issues/1504
Merged PR: https://github.com/stride3d/stride/pull/2422

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Physics.Tests/ColliderShapesTest.cs
+++ b/sources/engine/Stride.Physics.Tests/ColliderShapesTest.cs
@@ -267,6 +267,8 @@ namespace Stride.Physics.Tests
                 
                 var body = cube.GetOrCreate<RigidbodyComponent>();
 
+                await game.Script.NextFrame();
+
                 //verify values not properly set up
                 Assert.Null(body.ColliderShape);
                 Assert.Equal(RigidBodyTypes.Static, body.RigidBodyType);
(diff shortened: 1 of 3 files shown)


Eideren:
Thanks, although this one kind of looks like a band-aid solution, would there not be a way to prevent re-entry in the first place - something like setting up whatever triggers the re-attach at the very last step of attach itself. If you need to make major changes to re-organize how the logic for those functions is structured for that, do go ahead, although I understand if you want to cut this one short

Eideren:
I'll merge this one in to make sure release doesn't throw in the context mentioned in the PR, and opened a new issue for what I mentioned above.

Issue #2406, opened 2024-08-05
*.glb asset is imported with broken mesh


delustra:
**Release Type**: GitHub 

**Version**: 4.2.0.2188

**Platform(s)**: Windows

**Describe the bug**
GLTF asset is imported incorrectly. Geometry transformations and animations are broken for certain assets. At the same time the same assets work fine in other assets viewers. 

**To Reproduce**
1. Open editor
2. Create asset '3D Model'
3. Import this model [this model](https://drive.google.com/file/d/1RmugfviPMvg3joTHwSvEXzth00OFEiZw/view?usp=sharing)
4. Hangar geometry is broken. The gates are inside the hangar.
![image](https://github.com/user-attachments/assets/493fcdb1-40fb-4d44-bad3-d310f462a1fa)

**Expected behavior**
Hangar imports with proper geometry. This how the same assets looks in the regular windows 3d asset viewer.
![image](https://github.com/user-attachments/assets/b4f7c7b1-7be2-444b-933e-8262dc412aaf)

**Screenshots**
If applicable, add screenshots to help explain your problem (use GitHub drag & drop)

**Log and callstacks**
If any, please attach here any log or callstack (preferably in a `.txt` file using GitHub drag and drop)

**Additional context**

Basewq:
I've debugged the importer with your model and it looks like the underlying assimp library isn't loading the node hierarchy properly because of duplicate names making things linked incorrectly to one another.
I imported it into blender and see your mesh objects have the same name as the bones (eg. 'door_under' is the name of both the mesh object and the bone).

As a workaround, when I renamed the bones (eg. add a _bone suffix) then re-exported it, the animations seem to work correctly.

PR #2441, opened 2024-09-11
fix: extracting embedded texture files for materials from glTF file

Basewq:
# PR Details

Documentation on assimp texture behavior is described here:
https://assimp-docs.readthedocs.io/en/latest/API/API-Documentation.html#_CPPv49aiTexture
(Search `struct aiTexture` if the link doesn't jump to the correct section)

glTF materials referencing embedded textures may be denoted by `*` + index of the texture array, and the full file name must be determined by examining the Texture data.

Apparently it is possible to have uncompressed textures in the file (ie. the 'raw' texture is directly embedded).
I have opted to output a log error rather than implement it.

## Related Issue

Used the model from #2406 (this pull does *not* fix the animation issue in that link)

## Types of changes

## Checklist

### DIFF
--- a/sources/tools/Stride.Importer.3D/Material/Materials.cs
+++ b/sources/tools/Stride.Importer.3D/Material/Materials.cs
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net)
 // Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
 
+using System;
+using System.IO;
 using Silk.NET.Assimp;
 using Stride.Core.Diagnostics;
 using Stride.Core.Mathematics;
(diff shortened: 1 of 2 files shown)


sources/tools/Stride.Importer.3D/Material/Materials.cs
```diff
@@ -160,5 +172,51 @@ public static unsafe MaterialStack ConvertAssimpStackCppToCs(Silk.NET.Assimp.Ass
 
             return ret;
         }
+
+        internal static unsafe (bool isOk, string errorMessage) TryGetTextureFileName(string materialTextureFileName, Scene* scene, out string fileName)
```
Eideren, line 176: The tuple return type for a Try[...] seems a bit awkward to work with;
```cs
var (isOk, errorMessage) = TryGetTextureFileName(elTexPath.AsString, scene, out texFileName);
if (!isOk)
{
    logger?.Error(errorMessage);
    continue; // error !
}
```
vs a try with another out var
```cs
if (!TryGetTextureFileName(elTexPath.AsString, scene, out texFileName, out var errorMessage))
{
    logger?.Error(errorMessage);
    continue; // error !
}
```
Not a big deal, but it's not standard in c#, and not sure it is any easier to read ?

Basewq: Yeah, I can never settle on the right way the TryGet pattern, haha.
I'll change it for returning `bool` and `out` the message since it's standard.

Basewq: Done, and swapped the logic in the TryGet methods so it doesn't need nested if statements.

Eideren:
Thanks !

PR #2444, opened 2024-09-12
 Fix Assimp skeleton/animation importer for models with duplicate node names

### LINKED ISSUE #2406: *.glb asset is imported with broken mesh

Basewq:
# PR Details

See model in #2406 (be aware the second animation seems to play strangely, but this is consistent with how it plays in Blender 3.6 - it didn't import properly in a higher version of Blender)

Assimp has a generic 'Node' hierarchy structure and a 'Bone' list (per mesh).
Each Bone has a corresponding 'Node', however the link between the two is only done by name.
Unfortunately, this is a problem when a model has duplicate names, eg. the model's armature and bone, and Stride will more than likely link the 'Bone' to the non-Bone node.
This fix attempts to make an educated guess whenever duplicated names occurs.
If it fails to properly detect what should be linked, then it'll output a warning/error telling the end-user to make the name unique.

While `aiPostProcessSteps.aiProcess_PopulateArmatureData` does exist on the importer which is supposed to link the bone to the node, it does not make any effort to correctly link a bone to the bone node if duplicate names exist (it just searches for the first matching name), so this is not used.

## Related Issue

This is a fundamental flaw in Assimp, https://github.com/assimp/assimp/issues/4064 may help if it gets resolved.

Fixes #2406

## Types of changes

## Checklist

### DIFF
--- a/sources/tools/Stride.Importer.3D/MeshConverter.cs
+++ b/sources/tools/Stride.Importer.3D/MeshConverter.cs
@@ -11,6 +11,7 @@ using Stride.Animations;
 using Stride.Assets.Materials;
 using Stride.Core;
 using Stride.Core.Diagnostics;
+using Stride.Core.Extensions;
 using Stride.Core.IO;
 using Stride.Core.Mathematics;
 using Stride.Core.Serialization;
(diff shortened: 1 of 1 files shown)


sources/tools/Stride.Importer.3D/MeshConverter.cs
```diff
@@ -547,14 +793,33 @@ private unsafe void RegisterNodes(Node* fromNode, int parentIndex, Dictionary<In
...
+            if (duplicateNodeNameToNodeIndices is not null)
+            {
+                string originalNodeName = fromNode->MName.AsString.CleanNodeName();
+                if (!string.IsNullOrWhiteSpace(originalNodeName) && !string.Equals(originalNodeName, modelNodeDefinition.Name))
+                {
+                    if (!duplicateNodeNameToNodeIndices.TryGetValue(originalNodeName, out var nodeIndices))
+                    {
+                        nodeIndices = new();
+                        duplicateNodeNameToNodeIndices[originalNodeName] = nodeIndices;
+                        //Logger.Warning($"Duplicate node name found: {originalNodeName}. Animations may not play correctly if this node is used in an animation.");
```
Eideren, line 809: Unrelated commented out line afaict ?

Eideren (commented):
Looks good, just one nitpick

sources/tools/Stride.Importer.3D/MeshConverter.cs
Basewq: I was hesitant on whether to log a warning here, because technically the whole point of the other code is to actually try to resolve the duplicate nodes.
But now I just realized I can just output this at Info level.

Basewq: Done. Changed to output at Info level, and rebased the commit to the latest.

Eideren:
Thanks !

PR #2450, opened 2024-09-14
Drop support for UWP

Jklawreszuk:
# PR Details

This PR drops support for UWP and it's codebase.

The main reason is that UWP support has been abandoned since .NET 5 and Microsoft recommends to migrate to [MS GDK](https://github.com/microsoft/GDK) for Xbox consoles.

That said, the last version that targeted .NET 5 version was a very long time ago (4.1.0.1459-beta) and this PR is also intended to avoid unnecessary frustration for new developers over a feature that is long gone. 

## Types of changes

## Checklist


VaclavElias:
This test is missing " I have built and run the editor to try this change out." 🤣

Jklawreszuk:
@VaclavElias Welp...Testing always takes more time, but if you want to help speed it up, feel free to test it by yourself 😜

sources/engine/Stride.Shaders.Compiler/EffectCompiler.cs
```diff
@@ -204,8 +204,6 @@ public override TaskOrResult<EffectBytecodeCompilerResult> Compile(ShaderMixinSo
 #if STRIDE_PLATFORM_DESKTOP
                 case GraphicsPlatform.Direct3D11:
                 case GraphicsPlatform.Direct3D12:
-                    if (Platform.Type != PlatformType.Windows && Platform.Type != PlatformType.UWP)
-                        throw new NotSupportedException();
```
Eideren, line 208: This change is wrong, it should still throw when the platform is not windows

sources/engine/Stride.VirtualReality/VRDeviceSystem.cs
```diff
@@ -91,12 +91,6 @@ private void OnEnabledChanged(object sender, EventArgs eventArgs)
                             }
                         case VRApi.WindowsMixedReality:
                         {
-#if STRIDE_GRAPHICS_API_DIRECT3D11 && STRIDE_PLATFORM_UWP
-                            if (Windows.Graphics.Holographic.HolographicSpace.IsAvailable && GraphicsDevice.Presenter is WindowsMixedRealityGraphicsPresenter)
-                            {
-                                Device = new WindowsMixedRealityHmd();
-                            }
-#endif
```
Eideren, line 99: Might want to throw a not supported here instead of silently failing ? Granted that's what it would have done before outside of UWP

Eideren (commented):
Looks good, just a couple of notes

MeharDT:
Should we reconsider this PR in light of .NET 9 adding legacy support for UWP? At the very least it helps us migrate Stride users away from 4.0 and .NET Framework if they want UWP support.

https://devblogs.microsoft.com/ifdef-windows/preview-uwp-support-for-dotnet-9-native-aot/

Eideren:
@MeharDT I was not aware of this one, you're right, we'll revisit this issue whenever UWP support becomes problematic then

Jklawreszuk:
I marked PR  as draft for now. Let's see what happens  next with UWP support

Jklawreszuk:
I decided to reopen PR, because It seems UWP support in net9.0 was only temporary solution for customers to get some time to migrate to other solution.
At the moment, the solution I see is to adapt XBOX GDK with collaboration with other people with .NET Foundation, Monogame or Microsoft directly and hide implementation under separate repository (because of NDA's like Playstation does). 
There are also rumors about WIN32 support for the future console, which would also solve the problem.

Anyways, feel free to review my PR !

sources/engine/Stride.Shaders.Compiler/EffectCompiler.cs
Jklawreszuk: Oh my you're right, Fixed!

sources/engine/Stride.VirtualReality/VRDeviceSystem.cs
Jklawreszuk: Oh yes, It sounds reasonable instead of getting nothing. Fixed!

MeharDT:
> I decided to reopen PR, because It seems UWP support in net9.0 was only temporary solution for customers to get some time to migrate to other solution. At the moment, the solution I see is to adapt XBOX GDK with collaboration with other people with .NET Foundation, Monogame or Microsoft directly and hide implementation under separate repository (because of NDA's like Playstation does). There are also rumors about WIN32 support for the future console, which would also solve the problem.
> 
> Anyways, feel free to review my PR !

The dev blog I linked to was updated to include mention of .NET 10 templates for UWP in VS 2026. Was anyone able to deploy these to an Xbox One or Series console?

If they work on console, I think we should delay removal until GDK support can be implemented so we don't lose another platform. If they only work on PC, we might as well remove it now.

Jklawreszuk:
Oh, I didn't notice that good to know 😅. In that case, we can indeed delay removal until we check whether UWP still works on XBOX, or we can start implementing GDK now, although  to be honest, console support is not a priority for me at the moment.

xen2:

PR #2454, opened 2024-09-17
feat: Port native Audio providers to C# 

Jklawreszuk:
# PR Details

The goal of this PR is to make the Stride.Audio module fully working in C#, without worring about C++ build tools.

## Todo

- [x] Port OpenAL to Silk.NET.OpenAL - complete and tested on SimpleAudio sample
- [ ] Port Celt/Opus to .NET
- [ ] Port XAudio to Silk.NET.XAudio - almost complete, but not tested yet
- [ ] Port OpenSL ES to .NET - Silk.NET does not provide an OpenSLES wrapper, and Android considers it obsolete.

## Types of changes

## Checklist

- [x] **I have built and run the editor to try this change out.** (Only OpenAL)


sources/Directory.Packages.props
Kryptos-FR, line 1: Looks like the whole file got "touched" in the latest commit (maybe the encoding change or the EOL character). Try to keep the encoding on all file to be UTF-8 and EOL to be LF when pushing (but can be CRLF when checking out on Windows). Please check whether the git config option on Windows and Linux are consistent.

Kryptos-FR: With that said, sometimes git does the strangest diff and it's out of our control (could be the case here).

Jklawreszuk:
Funnily enough, my OpenAL audio layer fixes a bug where when opening the audio a second time it crashes the entire game. 
I'll try to get back to porting soon ♥️

sources/engine/Stride.Audio/Layers/XAudio/AudioProvider.cs
```diff
@@ -26,8 +27,12 @@ public AudioProvider()
...
         {
-            var buffer = new AudioBuffer(maxBufferSizeBytes);
-			//buffer.buffer.PContext = buffer;
+            var buffer = new AudioBuffer();
+            AudioBuffer* bufferPtr = &buffer;        
+			buffer.Buffer.PContext = bufferPtr;
+            
+            var data = stackalloc byte[maxBufferSizeBytes];
+			buffer.Buffer.PAudioData = data;
 			return buffer;
```
ericwj, line 36: Excuse me if I speak prematurely, but I stumble upon this and it hurts my eyes. It is never valid to `stackalloc` and let the pointer escape the current method. `data` is implicitly discarded and will be reused for method calls immediately upon return from this method. This buffer should probably be a rented buffer and perhaps also allocated from the pinned object heap, or allocated using `NativeMemory`, since I assume it will be used for native interop and/or hardware.

sources/engine/Stride.Audio/Layers/XAudio/AudioProvider.cs
```diff
@@ -161,6 +167,11 @@ public unsafe Source SourceCreate(Listener listener, int sampleRate, int maxNumb
                 var delay = stackalloc float[AUDIO_CHANNELS];
                 source.dsp_settings.pDelayTimes = delay;
```
ericwj, line 168: Same.

sources/engine/Stride.Audio/Layers/OpenAL/AudioProvider.cs
```diff
@@ -255,7 +265,11 @@ public void SourceQueueBuffer(Source source, AudioBuffer buffer, nint pcm, int b
         buffer.Type = streamType;
         buffer.Size = bufferSize;
         al.BufferData(buffer.Value, source.Mono ? BufferFormat.Mono16 : BufferFormat.Stereo16, pcm.ToPointer(), bufferSize, source.SampleRate);
-        al.SourceQueueBuffers(source.Value, 1, &buffer.Value);
+        fixed (uint* bufferPtr = &buffer.Value)
+        {
+            al.SourceQueueBuffers(source.Value, 1, bufferPtr);
```
ericwj, line 270: The buffer will be used asynchronously, but it won't be pinned when it is used anymore, since the `fixed` statement is exited. To keep the buffer pinned, allocate it pinned, or from native memory, or use `GCHandle.Alloc(buffer.Value, GCHandleType.Pinned)` and then keep the `GCHandle` around and call `Free` on it once the buffer no longer needs to be pinned.

Eideren: Sorry for derailing the conversation slightly, but since we have you here, @ericwj can you quickly take a look at this comment we left a while ago on something you wrote https://github.com/stride3d/stride/pull/2393#discussion_r1682245735

Kryptos-FR:
I don't know if it is in the scope of this PR or should be its own issue, but can we lazy-load the audio layers and only have them if the loaded scene has an audio source and/or an audio listener?

I stumble upon an issue on WSL where I can't start a game because it crashes after failing to initialize the audio, while I don't use any audio in the game itself.

sources/engine/Stride.Audio/AudioEngine.cs
```diff
@@ -24,10 +24,7 @@ public class AudioEngine : ComponentBase
 
         static AudioEngine()
         {
-            if (!AudioLayer.Init())
-            {
-                throw new Exception("Failed to initialize the audio native layer.");
-            }
+            AudioLayer.Init();
```
Kryptos-FR, line 27: Maybe it's here that we could have a lazy loading. Just a thought, not to be done in that PR but in another one later for #2628.

Jklawreszuk: When I think about it it would be nice to lazy load other optional systems like VRDeviceSystem, DebugTextSystem, GameprofilingSystem etc. I'll try to concider such PR 🤔

Jklawreszuk:
I would like to start work from scratch so I am closing PR

PR #2457, opened 2024-09-19
fix: Cecil serialization inheritance cutoff with complex base types

Eideren:
# PR Details
The binary serializer can lose serialized data when round-tripped with a specific inheritance setup;
```cs
[DataContract]
public class D : C<AnyType>{ public bool bill; }

public class C<T> : B<T>{}
public class B<T> : A{}

[DataContract(Inherited = true)]
public class A : EntityComponent{}
```

The following logic fails to capture `B<T>`
https://github.com/stride3d/stride/blob/37bbf0c85262711f56786319effab2621479bab9/sources/core/Stride.Core.AssemblyProcessor/Serializers/CecilSerializerContext.cs#L198-L204
Which in turn means that all serialized data contained in its base types would be discarded, most obvious with the component Id inherited from `EntityComponent`, which right now shows up as a new Id on every save caused by the asset collision Id de-duplication using binary serialization through the `Clone` function:
https://github.com/stride3d/stride/blob/37bbf0c85262711f56786319effab2621479bab9/sources/assets/Stride.Core.Assets/Analysis/AssetCollision.cs#L33-L48

## Related Issue
Kind of hard to tell

## Types of changes

## Checklist

### DIFF
--- a/sources/core/Stride.Core.AssemblyProcessor/SerializationProcessor.cs
+++ b/sources/core/Stride.Core.AssemblyProcessor/SerializationProcessor.cs
@@ -82,9 +82,9 @@ namespace Stride.Core.AssemblyProcessor
 
                 TypeReference parentType = null;
                 FieldDefinition parentSerializerField = null;
-                if (complexType.Value.ComplexSerializerProcessParentType)
+                if (complexType.Value.ComplexSerializerProcessParentType != null)
                 {
-                    parentType = ResolveGenericsVisitor.Process(serializerType, type.BaseType);
+                    parentType = complexType.Value.ComplexSerializerProcessParentType;
                     serializerType.Fields.Add(parentSerializerField = new FieldDefinition("parentSerializer", Mono.Cecil.FieldAttributes.Private, dataSerializerTypeRef.MakeGenericType(parentType)));
 
                     hash.Write("parent");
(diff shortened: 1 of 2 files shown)


manio143 (approved):
Nice solution. Just a question - the check on generic type inside the for loop where we continue - what scenario is that skipping?

Kryptos-FR:
> Nice solution. Just a question - the check on generic type inside the for loop where we continue - what scenario is that skipping?

Would be nice to add a few comments on those lines that explain what the algorithm is doing.

Eideren:
Added a couple of comments to clarify.
@manio143 from what I can tell, this is a fallback for when `ResolveGenericsVisitor.Process` failed to resolve, which afaict happens when trying to serialize `B` as an open type with an inheritance like the following:
```cs
class A<T>{}
class B<T> : A<T>{}
```
Not 100% sure as I haven't tested it, here's the source for that method:
https://github.com/stride3d/stride/blob/beb2972d0ed333068f226db6773d225dcad83056/sources/core/Stride.Core.AssemblyProcessor/ResolveGenericsVisitor.cs#L27-L44

xen2:
Would it be easy to add a test (in https://github.com/stride3d/stride/blob/master/sources/core/Stride.Core.Tests/TestSerialization.cs) so that we make sure to not break it in the future? (i.e. if we switch to Roslyn)
If not trivial, no problem.

Eideren:
Looked into it - in that particular test context, cecil fails to generate a serializer for the class even with the exact same type definitions I used at edit and runtime to validate the changes in this PR. I don't have a lot of time to dig too deep into why the test fails while the real one passes, so I'll have to leave it as is for now

xen2:
Thanks for giving it a try, all good then!

PR #2460, opened 2024-09-22
feat: [Build] - Better Platform detection for Linux

Jklawreszuk:
# PR Details

My PR corrects graphics API detection for Linux. The main reason was to improve the graphics detection in code editors like VSCode or Rider : instead of OpenGL as the default api, Direct3D is always selected  which is not correct. 

## Types of changes

## Checklist

### DIFF
--- a/build/Stride.Build.props
+++ b/build/Stride.Build.props
@@ -2,9 +2,14 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
     <StrideCommonPreSettingsName>Stride</StrideCommonPreSettingsName>
-    <StridePlatforms Condition="'$(StridePlatforms)' == ''">Windows</StridePlatforms>
+    <StridePlatforms Condition="'$(StridePlatforms)' == '' And '$([MSBuild]::IsOSPlatform(Windows))'">Windows</StridePlatforms>
+    <StridePlatforms Condition="'$(StridePlatforms)' == '' And '$([MSBuild]::IsOSPlatform(Linux))'">Linux</StridePlatforms>
   </PropertyGroup>
   <PropertyGroup Condition="'$(StridePlatform)' == 'Windows'">
     <StrideGraphicsApis Condition="'$(StrideGraphicsApis)' == ''">Direct3D11</StrideGraphicsApis>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(StridePlatform)' == 'Linux'">
+    <StrideGraphicsApis Condition="'$(StrideGraphicsApis)' == ''">OpenGL</StrideGraphicsApis>
   </PropertyGroup>
 </Project>
--- a/sources/targets/Stride.Core.props
+++ b/sources/targets/Stride.Core.props
@@ -15,15 +15,13 @@
 
     <!-- Default values -->
     <StridePlatformOriginal>$(StridePlatform)</StridePlatformOriginal>
-    <StridePlatform Condition=" $(TargetFramework.StartsWith('net')) ">Windows</StridePlatform>
-    <StridePlatform Condition=" $(TargetFramework.StartsWith('net')) And $(RuntimeIdentifier.StartsWith('linux')) ">Linux</StridePlatform>
-    <StridePlatform Condition=" $(TargetFramework.StartsWith('net')) And $(RuntimeIdentifier.StartsWith('osx')) ">macOS</StridePlatform>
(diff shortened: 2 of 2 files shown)


sources/targets/Stride.Core.props
```diff
@@ -15,15 +15,13 @@
 
     <!-- Default values -->
     <StridePlatformOriginal>$(StridePlatform)</StridePlatformOriginal>
-    <StridePlatform Condition=" $(TargetFramework.StartsWith('net')) ">Windows</StridePlatform>
```
Jklawreszuk, line 18: This expression is always true and was only checked because of mono

sources/targets/Stride.Core.props
```diff
@@ -15,15 +15,13 @@
...
-    <StridePlatform Condition=" $(TargetFramework.StartsWith('net')) And $(RuntimeIdentifier.StartsWith('linux')) ">Linux</StridePlatform>
-    <StridePlatform Condition=" $(TargetFramework.StartsWith('net')) And $(RuntimeIdentifier.StartsWith('osx')) ">macOS</StridePlatform>
+    <StridePlatform>Windows</StridePlatform>
+    <StridePlatform Condition=" '$([MSBuild]::IsOSPlatform(Linux))' ">Linux</StridePlatform>
+    <StridePlatform Condition=  "$(RuntimeIdentifier.StartsWith('osx')) ">macOS</StridePlatform>
```
Kryptos-FR, line 20: Nit: space inconsistency.

```xml
<StridePlatform Condition=" $(RuntimeIdentifier.StartsWith('osx')) ">macOS</StridePlatform>
```

Jklawreszuk: Right, my mistake. Fixed!

sources/targets/Stride.Core.props
```diff
@@ -15,15 +15,13 @@
 
     <!-- Default values -->
     <StridePlatformOriginal>$(StridePlatform)</StridePlatformOriginal>
-    <StridePlatform Condition=" $(TargetFramework.StartsWith('net')) ">Windows</StridePlatform>
-    <StridePlatform Condition=" $(TargetFramework.StartsWith('net')) And $(RuntimeIdentifier.StartsWith('linux')) ">Linux</StridePlatform>
-    <StridePlatform Condition=" $(TargetFramework.StartsWith('net')) And $(RuntimeIdentifier.StartsWith('osx')) ">macOS</StridePlatform>
+    <StridePlatform>Windows</StridePlatform>
+    <StridePlatform Condition=" '$([MSBuild]::IsOSPlatform(Linux))' ">Linux</StridePlatform>
```
Kryptos-FR, line 19: Wouldn't that prevent building for Linux from a Windows machine?

build/Stride.Build.props
```diff
@@ -2,9 +2,14 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
     <StrideCommonPreSettingsName>Stride</StrideCommonPreSettingsName>
-    <StridePlatforms Condition="'$(StridePlatforms)' == ''">Windows</StridePlatforms>
+    <StridePlatforms Condition="'$(StridePlatforms)' == '' And '$([MSBuild]::IsOSPlatform(Windows))'">Windows</StridePlatforms>
+    <StridePlatforms Condition="'$(StridePlatforms)' == '' And '$([MSBuild]::IsOSPlatform(Linux))'">Linux</StridePlatforms>
```
Kryptos-FR, line 6: Same remark.

Jklawreszuk: I would prefer then without spacing 👀

Jklawreszuk: I dont think so, because is only chooses ONE platform. For Linux StridePlatforms should be 'Linux', for Windows 'Windows' etc. 

When release built is made StridePlatforms is collection of ALL platforms : 'Windows, Linux, MacOS, Android, iOS, UWP' 

So this is a default value for testing packages locally without worring about other platforms

sources/targets/Stride.Core.props
Jklawreszuk: @Kryptos-FR Huh...Yes, Good point. I realized that Linux detection gonna work only on Linux but on Windows not really

Jklawreszuk: I bringed back RuntimeIdentifier then

Kryptos-FR (approved):

Eideren:
Thanks !

Issue #2189, opened 2024-03-15
[TextBlock] - Runing game causes segmentation fault when Graphics Api is OpenGL 


Jklawreszuk:
**Release Type**: Official Release

**Version**: Latest

**Platform(s)**: Linux/Windows, OpenGL

**Describe the bug**

After launching, the game crashes, causing a segmentation error after exactly n seconds

**To Reproduce**
1. Create JumpyJet demo
2. Target Linux platform 
3. Build / Publish the game on Windows
4. Run executable on Linux after meeting [requirements](https://doc.stride3d.net/latest/en/manual/platforms/linux/setup-and-requirements.html?tabs=freetype-ubuntu%2Copenal-ubuntu%2Csdl2-ubuntu) or Windows
5. Application should crash after few seconds

**Expected behavior**
The application should work without any problems

**My observations**

Every application that had `TextBlock` objects caused this problem. When I removed all occurrences of it (for example, in JumpyJet) the problem did not appear

Jklawreszuk:
Update: There is something wrong with OpenGL api. Switched on Vulkan everything works fine

NicusorN5:
Do you have any stacktrace from the segfault?

Jklawreszuk:
@NicusorN5 No it just silently crash :(
I also updated title because it seems that OpenGL does not work properly on any platforn

Jklawreszuk:
I don't know if it be useful, but here is the frag of dotnet-dump tool analysis 
![test](https://github.com/user-attachments/assets/6e08671b-85ae-4960-9a8f-08408a5a8731)

Jklawreszuk:
I switched backed again to Windows and I've managed to receive stacktrace:
```
Fatal error. System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
   at Silk.NET.OpenGL.GL.BufferData[[System.IntPtr, System.Private.CoreLib, Version=8.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]](Silk.NET.OpenGL.BufferTargetARB, UIntPtr, IntPtr ByRef, Silk.NET.OpenGL.BufferUsageARB)
   at Stride.Graphics.CommandList.MapSubresource(Stride.Graphics.GraphicsResource, Int32, Stride.Graphics.MapMode, Boolean, Int32, Int32)
[...]

PR #2466, opened 2024-09-24
fix: (Partial) Resolve seg. fault when Graphics API is OpenGL #2189

Jklawreszuk:
# PR Details

PR is fixes crashing when UIPage component with Button is being added.
However it requires OpenGL 4.3+ 
Considering Steam Survey ~90% of Graphics Cards support DirectX 12 should also support OpenGL 4.6, so this is not a big problem for Desktop.
Note: This patch only fixes problems with OpenGL. When it comes OpenGL ES, it still may not work because of different instructions

## Related Issue

- #2189 

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Graphics/OpenGL/CommandList.OpenGL.cs
+++ b/sources/engine/Stride.Graphics/OpenGL/CommandList.OpenGL.cs
@@ -3,10 +3,7 @@
 #if STRIDE_GRAPHICS_API_OPENGL
 
 using System;
-using System.Threading;
-using Stride.Core;
 using Stride.Core.Mathematics;
-using Stride;
 using Stride.Shaders;
 using Color4 = Stride.Core.Mathematics.Color4;
 
@@ -53,6 +50,8 @@ namespace Stride.Graphics
         private IndexBufferView indexBuffer;
 
         private VertexBufferView[] vertexBuffers = new VertexBufferView[8];
+        private readonly int versionMajor;
+        private readonly int versionMinor;
 
 #if !STRIDE_GRAPHICS_API_OPENGLES
         private readonly float[] nativeViewports = new float[4 * MaxViewportAndScissorRectangleCount];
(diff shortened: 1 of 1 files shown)


sources/engine/Stride.Graphics/OpenGL/CommandList.OpenGL.cs
```diff
@@ -920,39 +921,31 @@ public unsafe MappedResource MapSubresource(GraphicsResource resource, int subRe
...
-                    // Orphan the buffer (let driver knows we don't need it anymore)
-                    if (mapMode == MapMode.WriteDiscard)
-                    {
-                        doNotWait = true;
-                        GL.BufferData(buffer.BufferTarget, (UIntPtr)buffer.Description.SizeInBytes, IntPtr.Zero, buffer.BufferUsageHint);
-                    }
+                    if(versionMajor >= 4 && versionMinor >= 3)
+                        GL.InvalidateBufferData(buffer.BufferId);
+                    else
+                        GL.BufferData(buffer.BufferTarget, (uint)buffer.Description.SizeInBytes, IntPtr.Zero, buffer.BufferUsageHint);
```
Jklawreszuk, line 937: My note here: Based on the various repositories I was inspired by, BufferData should work - Maybe the wrong size ? I don't know

Jklawreszuk:
Any thoughts ? Anyone?

sources/engine/Stride.Graphics/OpenGL/CommandList.OpenGL.cs
Jklawreszuk: I'll add a reference to the documentation just in case https://registry.khronos.org/OpenGL-Refpages/gl4/html/glInvalidateBufferData.xhtml

Eideren:
This is outside my area of expertise sadly, hopefully @xen2 or @tebjan can chime in

Jklawreszuk:
I am closing because I probably found proper the solution :)

PR #2468, opened 2024-09-24
fix: Reduce visual noise when loading in new properties in the property grid

Eideren:
# PR Details

The property grid is briefly emptied and categories are expanded after a short delay whenever the user swaps between two selected entities:

https://github.com/user-attachments/assets/b03fd1fa-7843-48b9-b83a-90d73704f1c2

This PR ensures the previous one is kept on screen while the other one loads in:

https://github.com/user-attachments/assets/e5b1050d-b9d9-4774-92b1-fd8ed57a3b0b

This improves readability when comparing two objects, or when quickly scanning for a given property or component.

Loading duration has been artificially lengthened in the two clips above for clarity.

This area probably has a bunch of conflicts with your branch @Kryptos-FR , do let me know if I should leave this PR for later! Also I have zero clue about WPF, probably did something wrong, let me know ;)

## Related Issue
None afaict

## Types of changes

## Checklist


Kryptos-FR:
I'm a bit sick at the moment so I can't build locally. Let me ask a few questions instead.

Did you check if it had any impacts when selecting entities either from the tree, or picking from the scene? Or with multiple selection as well as "navigating" back and forth with those selection?

Some of the previous code had checks to prevent reentries but I don't remember what scenarii were triggering it

Eideren:
There's a small issue right now where the selection's name label in the property grid is set to no selection when selecting an asset after selecting an entity, not yet sure how to fix that one without replacing the two observable collections with a cleaner pattern

xen2:

PR #2481, opened 2024-10-08
feat: Remove AssemblyProcessor unused option

Jklawreszuk:
# PR Details

This PR removes unused parameter for AssemblyProcessor. Because of recent changes  It is never used

## Checklist


Kryptos-FR:
How does it determine it needs to generate a module initializer then?

IXLLEGACYIXL:
> How does it determine it needs to generate a module initializer then? Another way to say it: how to we tell to not generate one?
> 
> Is it only based on the attribute? In that case, did we check that all previously generated initializer through the parameter do have that attribute?

if you dont want a module intializer generated then dont use stride.core.moduleinitializer
generated shaders use system moduleinitializer now
and all previous module initializers get generated through roslyn now in https://github.com/stride3d/stride/blob/master/sources/core/Stride.Core.CompilerServices/Generators/ModuleInitializerGenerator.cs

when is the scenario to not use an existing module initializer?

Eideren:
Tests failed in Teamcity: https://teamcity.stride3d.net/buildConfiguration/Engine_Tests_WindowsSimple/26996?expandBuildDeploymentsSection=false&hideTestsFromDependencies=false&expandPull+Request+Details=true&hideProblemsFromDependencies=false&expandBuildTestsSection=true&expandBuildProblemsSection=true

Issue #2496, opened 2024-10-21
[Linux] Issue tracker for Stride Engine Samples


Jklawreszuk:
# Description

Commit hash at the time of verification : 4c46fdcc4da0021514fbaaea3d432095d164b9b6 or 4.2.0.2232

## Context 

At the current state Stride 3D enables you to build you games on Linux...kind of. While is possible to build your game it may not be possible to run it because of many different runtime issues. The goal is to make clear which samples can be run and which do not work.

## List of Samples (OpenGL)

- Audio
  
  - [ ] Simple audio - App crashes on image click (OpenGL|Vulkan)
    
- Graphics
  
  - [x] AnimatedModel - `Error while compiling GLSL shader. []` #2487
  - [x] CustomEffect    
  - [x] MaterialShader    
  - [ ] SpriteFonts - Need to implement / port Font loading for Linux https://github.com/stride3d/stride/issues/3021
  - [x] SpriteStudioDemo
    
- Games
  
  - [x] JumpyJet
  - [x] SpaceEscape - `Error while compiling GLSL shader. []` #2487
    
- Input
  
  - [x] Gravity Sensor
  - [x] Touch Inputs - not sure about GamePad though
    
- Particles
  
  - [x] Physics Sample - `Error while compiling GLSL shader. []` #2487
    
- Templates
  
  - [x] First Person Shooter - `Error while compiling GLSL shader. []` #2487, game icon is missing
  - [x] Third Person Shooter - `Error while compiling GLSL shader. []` #2487, game icon is missing 
  - [x] Top Down RPG - `Error while compiling GLSL shader. []` #2487,, game icon is missing 
  - [ ] VR Sandbox - Need to implement / port Font loading for Linux, other issues (?) https://github.com/stride3d/stride/issues/3021
    
- Tutorials
  
  - [x] CsharpBegginer - #2487, game icon is missing
  - [x] CsharpIntermediate - #2487, game icon is missing
    
- UI
  
  - [x] GameMenu - `System.Exception: Depth buffer is not the same size as the render target` #2487
  - [x] UIElementLink - `Error while compiling GLSL shader. []` #2487    
  - [x] UIParticles

Jklawreszuk:
I may have found a potential cause for Stride.SimpleAudio crashing. I'm working now on finding a solution 😀
Edit: Well, maybe It'd better to continue my efforts on Stride.Audio port and fix it that way

Jklawreszuk:
[...]

PR #2504, opened 2024-10-25
feat: Enable to build SkyBoxAsset using OpenGL

Jklawreszuk:
# PR Details

This PR implements / fixes compiling Skybox Asset for OpenGL. I also made few related refactoring changes

_Example of a game using Skybox : FirstPersonShooter_
Screenshot from 2024-10-25 18-16-23

## Related Issue

https://github.com/stride3d/stride/issues/2496

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Graphics/OpenGL/GraphicsDevice.OpenGL.cs
+++ b/sources/engine/Stride.Graphics/OpenGL/GraphicsDevice.OpenGL.cs
@@ -658,23 +658,24 @@ namespace Stride.Graphics
 #endif
 
 #if STRIDE_UI_SDL
-            gameWindow = (Stride.Graphics.SDL.Window)windowHandle.NativeWindow;
+            // NOTE : This null handling is specific for Linux AssetCompiler #2504 (refactor required?)
+            gameWindow = windowHandle?.NativeWindow as SDL.Window ?? new SDL.Window("");
 
             var SDL = Graphics.SDL.Window.SDL;
 
 #if STRIDE_GRAPHICS_API_OPENGLES
-            SDL.GLSetAttribute(GLattr.GLContextProfileMask, (int)GLprofile.GLContextProfileES);
+            SDL.GLSetAttribute(GLattr.ContextProfileMask, (int)GLprofile.ES);
 #else
-            SDL.GLSetAttribute(GLattr.GLContextProfileMask, (int)GLprofile.GLContextProfileCore);
+            SDL.GLSetAttribute(GLattr.ContextProfileMask, (int)GLprofile.Core);
 #endif
 
 
             if (IsDebugMode)
-                SDL.GLSetAttribute(GLattr.GLContextFlags, (int)GLcontextFlag.GLContextDebugFlag);
+                SDL.GLSetAttribute(GLattr.ContextFlags, (int)GLcontextFlag.DebugFlag);
 
             // Setup version
-            SDL.GLSetAttribute(GLattr.GLContextMajorVersion, currentVersion / 100);
-            SDL.GLSetAttribute(GLattr.GLContextMinorVersion, (currentVersion / 10) % 10);
+            SDL.GLSetAttribute(GLattr.ContextMajorVersion, currentVersion / 100);
+            SDL.GLSetAttribute(GLattr.ContextMinorVersion, (currentVersion / 10) % 10);
 
             MainGraphicsContext = new SdlContext(SDL, (Silk.NET.SDL.Window*)gameWindow.SdlHandle);
             ((SdlContext)MainGraphicsContext).Create();
@@ -690,7 +691,7 @@ namespace Stride.Graphics
 #endif
 
             // Create shared context for creating graphics resources from other threads
-            SDL.GLSetAttribute(GLattr.GLShareWithCurrentContext, 1);
+            SDL.GLSetAttribute(GLattr.ShareWithCurrentContext, 1);
             deviceCreationContext = new SdlContext(SDL, (Silk.NET.SDL.Window*)gameWindow.SdlHandle);
             ((SdlContext)deviceCreationContext).Create();
 
--- a/sources/engine/Stride.Rendering/Rendering/ComputeEffect/GGXPrefiltering/Hammersley.sdsl
+++ b/sources/engine/Stride.Rendering/Rendering/ComputeEffect/GGXPrefiltering/Hammersley.sdsl
@@ -12,7 +12,7 @@ namespace Stride.Rendering.Images
         {
             var u = 0.0;
             var p = 0.5;
-            for (int kk=k; kk; p*=0.5, kk>>=1)
+            for (int kk=k; kk > 0; p*=0.5, kk>>=1)
             {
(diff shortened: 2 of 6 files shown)


sources/engine/Stride.Graphics/OpenGL/GraphicsDevice.OpenGL.cs
```diff
@@ -658,23 +658,23 @@ protected unsafe void InitializePlatformDevice(GraphicsProfile[] graphicsProfile
 #endif
 
 #if STRIDE_UI_SDL
-            gameWindow = (Stride.Graphics.SDL.Window)windowHandle.NativeWindow;
+            gameWindow = windowHandle?.NativeWindow as SDL.Window ?? new SDL.Window("");
```
Eideren, line 661: What's this new window for ?

sources/engine/Stride.Rendering/Rendering/ComputeEffect/GGXPrefiltering/Hammersley.sdsl
```diff
@@ -12,9 +12,9 @@ namespace Stride.Rendering.Images
         {
             var u = 0.0;
             var p = 0.5;
-            for (int kk=k; kk; p*=0.5, kk>>=1)
+            for (int kk=k; kk > 0; p *= 0.5, kk /= 2)
             {
-                if (kk & 1) // kk mod 2 == 1
+                if (kk % 2 == 1)
                     u += p;
             }
```
Eideren, line 19: Can you go over this change, the logic is equivalent of course, but is there a need to move from binary to math operator ?

Eideren (commented):
Thanks !

sources/engine/Stride.Graphics/OpenGL/GraphicsDevice.OpenGL.cs
Jklawreszuk: Dump window is created to initialize the SDLContext which is then used for the OpenGL api as a context (see 699 line).
The author assumed when casting that the `windowhandle` object was not `null` but it wasn't, so I handled it

sources/engine/Stride.Rendering/Rendering/ComputeEffect/GGXPrefiltering/Hammersley.sdsl
Jklawreszuk: @Eideren Thanks for pointing out. Looks like binary operators works well on GLSL so there is no need to change that since this is faster than division. Long story short there is `HLSLToGlslConvertor` class generated wrong Hammerslay glsl code that did not compile at all.
```glsl
{
    float u = 0.0;
    float p = 0.5;

    for (int kk = k; kk; p *= 0.5, kk >>= 1)
    {
        if (bool(bool(bool(kk & 1))))
            u += p;
    }
    float v = (float(k) + 0.5) / float(samplesCount);
    return vec2(u, v);
}
```
The main issue here is `kk` condition which is not understandable for OpenGL shader langauage

Jklawreszuk: So yeah, I'm leaving now only the explicit if more than zero condition

sources/engine/Stride.Graphics/OpenGL/GraphicsDevice.OpenGL.cs
Eideren: For opengl on windows this would load in sdl when using a winforms window then ?

Also, shouldn't it be initialized/set in a more appropriate spot then ? I.E.: ensure window is init first and has a native window handle, then this scope is called and retrieves the window without creating one ?

Jklawreszuk: When it comes to Windows, from what I've noticed, Stride Asset Compiler uses DirectX on Windows to compile assets and I haven't changed that. My change only applies to Linux (or MacOs if someone has decided to support it)

Jklawreszuk: As for the second question, this window object is used **before** the game starts so it will always be null. Also, I am not sure it even creates "real" window because Silk has command for this : SDL.CreateWindow()

Jklawreszuk: If you observe the flow of the program you will come to the conclusion that this is the earliest place where you can handle the `null` value is exactly in `InitializePlatformDevice` because only the `GraphicsDevice.Recreate` method which is common code for all platforms

Eideren: Checked Q#1 and games running on windows can enter this scope, but setting OpenGL as the backend forces SDL, this should be fine.
As for Q#2, just to clarify, the new window workaround here is specifically for the asset compiler, a game under any platform would have a valid native window value there ?

Jklawreszuk: Correct! Handling null is only for the purpose of AssetCompiler, more specifically `GraphicsDevice.New()` command - for DX11 it passes windowHandle with null and does nothing, for OpenGL it passes windowHandle and may throw NullRefException. 
When It comes to Stride runtime It shouldnt affect anything because creating game goes by `GraphicsDeviceManager.CreateDevice(...WindowHandle window)`

Jklawreszuk: Apparently no one predicted  it would be possible to run on Linux in the future 😅

Jklawreszuk: Anyways, If you have any suggestions feel free to give any ideas

Eideren: Not sure why the OGL stuff requires a window when the D3d one doesn't, but that's best left to a proper refactor I guess.
Can you split the ternary into a specific line with a comment mentioning that this is a workaround specifically for the asset compiler ?

Jklawreszuk: @Eideren Sure, Done!

Eideren:
Thanks !

sources/engine/Stride.Graphics/OpenGL/GraphicsDevice.OpenGL.cs
Ethereal77: Correct me if I'm wrong, but I think OpenGL has always needed a window to establish its context (the surface where to draw). Direct3D also needed it up until D3D9, but if I remember it correctly, it was D3D10 or 11 that started to allow setting up a swapchain without an associated HWND.
As Xenko supported both D3D9 and OpenGL, that may be the reason for requiring a window. Now that D3D9 is deprecated and D3D11 no longer needs it, only OpenGL is left with that requirement (as Vulkan can also render without a window).

PR #2517, opened 2024-11-11
Fix "Add Asset Pop-Up" height

minktusk:
# PR Details

The "Add Asset Pop-Up" does not have a height limit

Before:

gOl1S5DjA4

After:

qXfpZkc6EN

## Types of changes

## Checklist


Doprez:
Will these changes be consistent across different resolutions? It seems like this change only has an effect since the height was hard coded before to account for the list size so the scroll bar would have been available if any items were added in the future.

If it is not consistent across different resolutions or scales then I think we should consider that as a fix instead.

minktusk:
Sorry for not being more specific earlier!

Yes, this is a fix!

Issue #2515, opened 2024-11-10
Custom Hot Keys for Edit Text


TranquilAbyss:
**Is your feature request related to a problem? Please describe.**
I was trying to add an event to happen when EditText was selected and pressing the enter key.  I would like it for more than just Edit Text since I may come up with other elements to create hotkeys for.

**Describe the solution you'd like**
It would help to expose the existing KeyPressEvents on UI Elements.

**Describe alternatives you've considered**
1. Since the Key Press Events are not available I would have to capture a global key press and check every EditText to see which had focus. 
2. Or I would have to create a custom EditText class that provides access to the events.

**Additional context**
I am willing to create the PR myself, If there is any documentation that requires update, for changing an internal to a public class and event, let me know since I am new to how things get updated in this project.

PR #2518, opened 2024-11-12
feat: Made UIElement Key Events public

TranquilAbyss:
# PR Details
Made KeyPress Events public on UI Elements to allow for custom UI Hotkeys

## Related Issue
https://github.com/stride3d/stride/issues/2515

## Types of changes

## Checklist

This is my first PR, let me know if I am missing something.
All the UI tests passed, I am still working on getting my environment setup for other tests.

### DIFF
--- a/sources/engine/Stride.UI/KeyEventArgs.cs
+++ b/sources/engine/Stride.UI/KeyEventArgs.cs
@@ -8,7 +8,7 @@ namespace Stride.UI
     /// <summary>
     /// The arguments associated to an key event.
     /// </summary>
-    internal class KeyEventArgs : RoutedEventArgs
+    public class KeyEventArgs : RoutedEventArgs
     {
         /// <summary>
         /// The key that triggered the event.
--- a/sources/engine/Stride.UI/UIElement.Events.cs
+++ b/sources/engine/Stride.UI/UIElement.Events.cs
@@ -320,7 +320,7 @@ namespace Stride.UI
         /// Occurs when the element has the focus and the user press a key on the keyboard.
         /// </summary>
         /// <remarks>A key pressed event is bubbling</remarks>
-        internal event EventHandler<KeyEventArgs> KeyPressed
+        public event EventHandler<KeyEventArgs> KeyPressed
         {
             add { AddHandler(KeyPressedEvent, value); }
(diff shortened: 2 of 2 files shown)


Eideren:
Sorry for the late reply, completely missed this one somehow, this change kind of makes sense but I think the idea was that those events would be explicitly implemented on components were it make sense, for example, if you need to listen for text input, you can use `EditText.TextChanged`, there's also the text filter if you are listening for input specifically to ensure no invalid characters are entered.
I have not checked that element which do not contain editable text receive `Key*` events, if they do not, making them public may be counter intuitive.
Do you have a case were you need those events specifically instead of the specialized ones under the different component types ?

TranquilAbyss:
I may not answer you with a specific use case directly, right away but let me explain my use case.

The on focus functionality is hidden to Stride game developers, and the events the Edit text use implement Key Down on element being focused. The feature this is directly blocking for me is Send Chat message on Enter pressed, currently using a button for now but that wont work in the game due to taking too long to press. 

Future features low priority feature include, tab order manager between UI controls. Other features for Edit Text on focus include, Select all (Ctrl + A), hold down backspace, shift to select and arrows to characters, and a few more of nice to have keys for moving around the text. It seem like a lot of effort to add all this through Stride directly rather than allowing the user to customize this as needed. I can understand protecting the intent of the project though.

Eventually, in another PR, I would want access to the currently focused UIElement, I have a crazy workaround I am using just to keep track of which EditText is selected. But that is not high priority for me since I have the workaround.

Eideren:
@TranquilAbyss That actually makes sense, and given that the other events are public I'll merge this in.

Just in case here's the workaround I'm using to detect when the ui is hovered/selected, we definitely could do with something more straightforward though, yeah !
```cs
var uiProcessor = Entity.EntityManager.GetProcessor<UIRenderProcessor>();
bool uiHasFocus = false;
foreach (var uiElement in uiProcessor.UIRoots)
{
    if (uiElement.Page == null)
        continue;
    
    uiHasFocus |= uiElement.LastMouseOverElement is not null;
    uiHasFocus |= uiElement.Page.RootElement.FindVisualChildrenOfType<EditText>().Any(txt => txt.IsSelectionActive);
}
```

Issue #30, opened 2018-08-04
Create a new asset type equivalent to Unity's ScriptableObject.


dfkeenan:
Hi,

Though I have never used Unity I feel that the ScriptableObject is quite valuable. Being able to design a data object and save it with various values so scripts can share instances can really improve design time support in editor.

cheers,
dfkeenan

SaSiNO97:
I support this idea

brmassa:
for me is a long no-go for xenko.
game data asset is crucial

scottyboy805:
Just wanted to show support for this feature. I think a scriptable data asset would be a great addition to the engine. At the moment it is sort of possible in a hacky kind of way by using a prefab with an empty entity and single script, but I think a dedicated asset would be excellent. Many games need to store read only data that doesn't need to be in an external file for things like item configurations, level data etc. It makes sense to use the powerful editor and serialization that is already available.

Here's how I imagine it could be added:

1: Create a new base class 'DataScript' or similar that users would inherit from along with adding all the asset serializers and processors etc.
2: Add a way to create an asset from these scripts. 'Add Asset/Scripts/Data Script'? or something like that or maybe there is a better way.
3: Allow the content system to return instances of these scripts from the load methods. For example:
'MyDataScript a = Content.Load<MyDataScript>("Example");'
4: Enjoy better data support for your games.

I'm no expert, but those are my thoughts on the idea and think it would be a useful addition for many users.

makotech222:
I'm of the opinion the c# environment is powerful enough that scriptableobject isn't really needed. Yes, what you end up doing in Xenko is creating an entity with priority -1, and add a script to it. The script will be a class with a static reference to itself (Singleton), which holds whatever game data you need. That is how I'm currently handling it. I don't think we need a special entity to represent these easy steps.

scottyboy805:
Yes, that is fair enough. But what if the data should be accessible in multiple scenes? Then you would either have to load the data scene first and keep it alive, or clone the data which comes with its own issues, namley keeping the data in sync between scenes. This is why I use a prefab to store data but it just seems clumbsy and hacky. I come from a Unity background and when they added a scriptable object asset i didn't really see the point either. However, once you use it and see the flexibility and endless number of uses, it really does become a great feature to have.

tebjan:
So just for my understanding, the main feature is that you can edit the values in the editor, right? Otherwise, there are many ways to achieve that with common programming patterns, like singleton, services, and so on...

[...]

PR #2527, opened 2024-11-21
feat: User-defined assets support (ScriptableObject like data types and more)

Eideren:
# PR Details
WIP right now - it works but barely. I feel like most of the logic shouldn't be where I ended up shoving it, would need someone aware of how quantum, the assembly reloader and the undo system work to check over this one.

To create an asset, see this page https://github.com/stride3d/stride-docs/blob/master/en/manual/scripts/custom-assets.md or @manio143 's https://github.com/manio143/StrideCustomAsset

## Related Issue
#30 

## Types of changes

## Checklist


Eideren:
Cleaned up as much as I could, still a fair amount of logic that might be fragile, but I'm at a point were things are just too obtuse to investigate further.

xen2:
There is already a mechanism to handle type reloading, but it works only if it's **inside** an asset (as part of a member or list), not for the top-level item.

I am checking which option would make more sense:
- Handle specifically the top-level asset reloading for this use case (similar to your current approach). However we will need to add support for undo/redo, fix references, etc.
- Allow the other mechanism to work on top-level item (downside: `AssetViewModel.Asset` might become `null`, I think it might break a lot of stuff)
- Use `IUnloadable` which is what we do when a top-level type can't be initially loaded.

xen2:
I had to dig a bit deeper than expected because I ran into a situation that was not working properly (unrelated to your changes for custom asset).

So I spent some time to understand the various quantum nodes/reloading/propagator system and could finally fix this bug.

I did some tests and now it seems I can get proper reload for assets, components and references (tested on https://github.com/manio143/StrideCustomAsset).

Also, it should be not so difficult to add support to "unloadable assets" (just like we do for unloadable components/part of assets, this was just not possible at the top-level) since it would work the same way as we do for reload.

You can give it a try there:
https://github.com/xen2/stride/tree/improved_asset_reload
(I might still clean/polish/rewrite commits, and add support for unloaded assets before making final PR)

sources/editor/Stride.Core.Assets.Editor/ViewModel/AssetViewModel.cs
```diff
@@ -715,5 +683,33 @@ public void PrepareSave(ILogger logger)
         {
             PropertyGraph?.PrepareForSave(logger, assetItem);
         }
+
+        public void UpdateAsset(Asset newAsset, ILogger loggerResult)
```
Kryptos-FR, line 687: Any particular reason to move the method here? Usually we have public method first, then protected, then private.

sources/editor/Stride.Assets.Presentation/AssemblyReloading/GameStudioAssemblyReloader.cs
```diff
@@ -438,17 +427,17 @@ protected override bool ProcessObject(object obj, Type expectedType)
...
+                    if (CurrentPath.Match(expectedPath))
                     {
                         var eventReader = new EventReader(new MemoryParser(unloadedObject.ParsingEvents));
                         var settings = Log != null ? new SerializerContextSettings { Logger = Log } : null;
                         PropertyContainer properties;
                         unloadedObject.UpdatedObject = AssetYamlSerializer.Default.Deserialize(eventReader, null, unloadedObject.ExpectedType, out properties, settings);
                         // We will have broken references here because we are deserializing objects individually, so we don't pass any logger to discard warnings
                         var metadata = YamlAssetSerializer.CreateAndProcessMetadata(properties, unloadedObject.UpdatedObject, false);
-
+
```
Kryptos-FR, line 440: nit: remove spaces

sources/editor/Stride.Core.Assets.Editor/ViewModel/AssetViewModel.cs
Eideren: Moved it back, I'll let @xen2 chime in if this was on purpose

sources/assets/Stride.Core.Assets.Quantum/AssetNodeContainer.cs
```diff
@@ -33,10 +34,35 @@ public AssetNodeContainer()
...
+
+        var registeredTypes = new HashSet<Type>();
         // Register content types as primitive so they are not processed by Quantum
         foreach (var contentType in AssetRegistry.GetContentTypes())
         {
             NodeBuilder.RegisterPrimitiveType(contentType);
+            registeredTypes.Add(contentType);
         }
+
+        AssemblyRegistry.AssemblyRegistered += (sender, args) =>
```
Kryptos-FR, line 46: We might need to make `AssetNodeContainer` be `IDisposable` and unregister to those events when disposing. I'm thinking this might otherwise cause some issues when re-opening a projhect or opening a new project from within the Game Studio (without closing it).

sources/editor/Stride.Core.Assets.Editor/ViewModel/AssetViewModel.cs
```diff
@@ -182,7 +182,7 @@ protected AssetViewModel([NotNull] AssetViewModelConstructionParameters paramete
         [NotNull]
         public Asset Asset => AssetItem.Asset;
 
-        public AssetPropertyGraph PropertyGraph { get; }
+        public AssetPropertyGraph PropertyGraph { get; private set; }
```
Kryptos-FR, line 185: Is it necessary? Modern versions of C# implicitly have a private setter.

sources/assets/Stride.Core.Assets.Quantum/AssetNodeContainer.cs
Eideren: I ended up fixing this one in a different way as reaching it to call disposable ended up creating a significant amount of garbage logic
let me know what you think

Kryptos-FR: Not sure how, but would be nice to benchmark (or at least estimate) the performance of this version compared to before.

Eideren: The only way it could be slower is if the type being tested has close to or more base types than the amount of content types registered;
`AssetRegistry.IsExactContentType` does one hashtable test per base types the type has, while the `RegisterPrimitiveType` version does one `Type.IsAssignableFrom` tests for each registered content types.

I kept the rest of the logic as is, but it definitely can be optimized further if you think this would be necessary ?

NicusorN5:
TYSM for merging this! I've waited for so long for this PR! ❤️

PR #2531, opened 2024-11-24
refactor: FreeImage dependency update

Jklawreszuk:
# PR Details

This PR mainly removes the duplicate FreeImageNET source code, which is not used anywhere.
In addition, I decided to refactor code and remove the useless compatibility with `System.Drawing.Bitmap`, which is also not used anywhere. 

## Types of changes

## Checklist


Eideren:
There's a bunch of tabs/space mismatch in this PR, can't merge it as is sadly

sources/tools/Stride.TextureConverter/Backend/Wrappers/FINetWrapper/Classes/MetadataModels.cs
```diff
@@ -2790,14 +2779,12 @@ public TimeSpan? TimeStamp
             get
             {
                 FIURational[] stamp = GetTagArray<FIURational>("GPSTimeStamp");
-                if ((stamp == null) || stamp.Length != 3)
+                if (stamp is not { Length: 3 })
```
Eideren, line 2782: I think this one is just more confusing than the previous one; the negative and positive is merged into the same comparison

Jklawreszuk: No problem , I think it is more readable but I understand that it can be confusing, so I reverted it 😀

Jklawreszuk:
Huh... Maybe Rider uses spaces instead of tabs when I fixed the warnings. Now I formatted with VSCode,  so I hope that will do

Eideren:
Thanks !

PR #2533, opened 2024-11-29
Make Forward Renderer more flexible

Doprez:
# PR Details

This change allows users to add a more generic scene renderer as an alternative to only the clear renderer.

## Related Issue

none. I just want to use the default forward renderer but needed to use an alternative to the default clear option.

## Types of changes

## Checklist


Eideren:
Not sure about this one, the logic there expects an actual clear pass, this change makes clearing a bit more obtuse of an operation at the very least. I'll let someone else chime in though given how shallow my knowledge on that part of the engine is.

Doprez:
Totally fair, this can be closed then if there is hesitation. Its not as useful as I originally thought based on some more research into how rendering works with Stride.

Using a `RootRenderFeature` I was able to achieve what I needed. I thought what I wanted to do was run logic at the clear stage but I learned that I could just create a new `RootRenderFeature` that would replace the skybox instead.

PR #2555, opened 2024-12-11
feat: Process required keyword like MemberRequiredAttribute

Eideren:
# PR Details
Now that `required` is an actual keyword, we can use it to notify users when some of their components have missing references, just like our `MemberRequiredAttribute`.

`required` is an essential part of projects using the nullable reference feature, using it in stride to benefit from null-state analysis from both the compiler and our asset builder requires writing a fairly redundant property definition:
```cs
[MemberRequired] public required MyRef myRef { get; set; }
```
When the intent of the attribute is already made clear through the use of the `required` keyword.
```cs
public required MyRef myRef { get; set; }
```

## Related Issue
None

## Types of changes

## Checklist

Hopefully you have some time to get back to me about this one @manio143 , would like to hear your thoughts given that this is your feature.


Eideren:
There we go, covered both the keyword and the keyword with the attribute set to error. Can you check again @manio143 ?

manio143 (approved):

PR #2564, opened 2024-12-17
feat: LockFileEx/UnLockFileEx methods shouldn't be called explicitly

Jklawreszuk:
# PR Details

Stride should avoid direct references to any OS if possible. This PR Hides Lock/UnLockEx methods under already made Lock/Unlock methods that wrap them.

## Related Issue

None, but this relates to Stride.Launcher which is using `FileLock.Wait` method, so this provides cross-platform experience

## Types of changes

- [x] Breaking change - kind of, but only changes return type from void --> bool

## Checklist

### DIFF
--- a/sources/core/Stride.Core.Design/Windows/FileLock.cs
+++ b/sources/core/Stride.Core.Design/Windows/FileLock.cs
@@ -33,7 +33,7 @@ namespace Stride.Core.Windows
         {
             if (lockFile != null)
             {
-                NativeLockFile.UnlockFile(lockFile, 0, uint.MaxValue);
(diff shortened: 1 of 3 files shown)


Eideren:
Should this be renamed to `TryLockFile` instead ?

Jklawreszuk:
@Eideren Huh...Your're right, It should 😅 I have now added your suggestion

Jklawreszuk:
oh,yea...I forgot the parentheses. This is the last fix I promise 😆

Jklawreszuk:
Sorry, I should have just built it myself straight away before sending the PR 🤐 . I've built the project so the test should pass

Eideren:
Thanks !

PR #2567, opened 2024-12-19
docs: Bepu, missing property summaries

Eideren:
# PR Details
Docs are kind of the only thing I can contribute atm, not sure what we should do with `SpeculativeMargin`, is that even useful to surface given how it is described in bepu ... left it as is for now

## Related Issue
None.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.BepuPhysics/Stride.BepuPhysics/BodyComponent.cs
+++ b/sources/engine/Stride.BepuPhysics/Stride.BepuPhysics/BodyComponent.cs
@@ -193,7 +193,7 @@ public class BodyComponent : CollidableComponent
     }
 
     /// <summary>
-    /// The rotation velocity in unit per second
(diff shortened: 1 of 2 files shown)


sources/engine/Stride.BepuPhysics/Stride.BepuPhysics/BodyComponent.cs
```diff
@@ -167,6 +192,14 @@ public Vector3 LinearVelocity
         }
     }
 
+    /// <summary>
+    /// The rotation velocity in unit per second, also known as torque
```
Ethereal77, line 196: Torque refers to an angular force, not an angular velocity, isn't it?

Eideren: Yep, the distinction isn't huge in this context but best avoid any confusion, removed it from the summary

PR #2568, opened 2024-12-19
[Graphics] Fixes memory leak when creating D3D11 texture with shared options

azeno:
# PR Details

`QueryInterface` increases the reference count and must be paired with a call to `Release`.

## Types of changes

## Checklist


Eideren:
Do you guys have time to look into this one @arturoc @tebjan ?

tebjan:
This looks correct! Each interface that one gets with `QueryInterface` needs to be disposed. Even if the underlying native object is the same or should still exist.

tebjan:
Merged, thanks!

PR #2570, opened 2024-12-20
Introduce SharedCollidableComponent

Nicogo1705:
# PR Details

Bepu Physics was missing a feature: the ability to reuse shapes to avoid unnecessary allocations.

## Related Issue

Potential performance impact resulting in lower FPS.

## Types of Changes

## Checklist

- [X] My change requires documentation updates.
- [ ] All new and existing tests have passed.
- [X] I have built and run the editor to verify this change.


Nicogo1705:
@Eideren I let you review my nonsense and maybe test it a little more in depth, but it was working on my side!

Eideren:
Hey @Nicogo1705 thanks a bunch for looking into this one, do you think it would be possible to automate this process instead ? This sounds like it would be kind of a pain when you want to optimize for cases where you have arbitrary prefabs being spawned in.
Maybe checking the shape parameters for a shape we're about to create against a dictionary of those that already exists might work, similar idea to what we already have for the hulls actually: 
https://github.com/stride3d/stride/blob/b59fbc5dfa875895600e58bbddfa7da0517b00c8/sources/engine/Stride.BepuPhysics/Stride.BepuPhysics/Definitions/Colliders/ConvexHullCollider.cs#L43-L71

Nicogo1705:
@Eideren I can definitely add that as an additional feature. The approach I’m using offers some advantages over yours. For example, if you edit just the SharedCollider, all colliders on all associated bodies or static objects are automatically updated. By contrast, with your approach, editing one of the shared colliders creates a new shape in the pool, leaving the others to continue using the original shape.

Eideren:
@Nicogo1705 the bullet implementation had a similar idea were users could refer to a specific collider shape asset instead of an inline collider definition, issue is that that feature is just a less useful version of our prefab inheritance system - the only cases it doesn't cover would be for two distinct object sharing the same shape, which in itself is not recommended as it is fragile; since they are distinct objects, you have to double check that they still fit the shape every time you edit it.
I can't think of any case at runtime where it would be useful to edit multiple shapes at once ?

Nicogo1705:
@Eideren,
I see your point. Avoiding the “ref” might indeed simplify things in some scenarios. 

But it some case, e.g a door, you could open/close a door without rebuilding/checking a collider (Body.Collider = ClosedDoorSharedCollider or Body.Collider = OpenedDoorSharedCollider).

I’ve added automatic sharing —please take a look and let me know what you think. I haven’t tested it yet, but I’ll try tomorrow.

Nicogo1705:
I still haven't been able to test this PR, and it may take some time before I can. Could someone try it out for me?

xen2:

Issue #2576, opened 2024-12-22
Major breaking changes tracking


Doprez:
## Purpose
This issue will be an overview of planned changes that will be breaking to Strides core. This will allow us to consolidate discussions, PRs and issues into a single place for future planning.

## Current PRs
#2404 likely going to be closed in favour of #2574 
#2574 The main PR related to changing the core startup and surface rendering of Stride

## Current Discussions
#2524 The discussion related to strides startup changes
#2210 Tracking for talks about changing to `System.Numerics`

## Tasks
TBD I will fill this out better with breakdowns for what is needed to get going on the above discussions

#### Math
- [ ] Deprecate Stride math extensions in favour of `System.Numerics`
- [ ] Change core math types to the `System.Numerics` equivelent

#### Game startup
- [ ] Change to a more robust `Game.Window` 
- [ ] decouple the `GameWindow` from `GamePlatform` maybe in favour of something that defines just the update/render loop?
- [ ] remove `GameContext` it serves no real purpose and just adds an unnecessary layer. Should use `GameSettings` instead
- [ ] remove cyclical dependencies between systems. Discussed [here](https://github.com/stride3d/stride/discussions/2524#discussioncomment-11286302)

Jklawreszuk:
You can also add my two PRs which can be considered as breaking change:
- https://github.com/stride3d/stride/pull/2186
- https://github.com/stride3d/stride/pull/2261

Doprez:
I added the directx 9 and opengl 2 support to the main checklist but I think @xen2 wasn't convinced to remove metrics last I remember?

I also dont think its a breaking change so Im not sure that we would need to wait for 5.0 if we wanted to merge that in.

PR #2579, opened 2024-12-25
Deprecate stride math for numerics alternatives

Doprez:
# PR Details

Doing some benchmarks and discussing with @Eideren and @ericwj there are some major speed benefits using System.Numerics instead of Stride math. 

This PR is meant to deprecate the math functions that are available with the System.Numerics equivalents.

## Related Issue

https://github.com/stride3d/stride/pull/2238
https://github.com/stride3d/stride/issues/2576

## Types of changes

## Checklist

## IMPORTANT
https://github.com/stride3d/stride/pull/2238 must be merged first for the implicit conversions


Doprez:
first hurdle:
`Invert` seems to give NaN values where Stride did not. It seems like the 
```
Assert.Equal() Failure: Values differ
Expected: [M11:0 M12:0 M13:0 M14:0] [M21:0 M22:0 M23:0 M24:0] [M31:0 M32:0 M33:0 M34:0] [M41:0 M42:0 M43:0 M44:0]
Actual:   [M11:NaN M12:NaN M13:NaN M14:NaN] [M21:NaN M22:NaN M23:NaN M24:NaN] [M31:NaN M32:NaN M33:NaN M34:NaN] [M41:NaN M42:NaN M43:NaN M44:NaN]
```
same with `Orthonomolize`
```
Assert.Equal() Failure: Values differ
Expected: [M11:0.18257418 M12:0.36514837 M13:0.5477226 M14:0.73029673] [M21:0.8164966 M22:0.40824836 M23:-1.4600097E-07 M24:-0.40824822] [M31:0 M32:0 M33:9.536743E-07 M34:0] [M41:-0.4714045 M42:-0.4714045 M43:0.70710677 M44:-0.23570225]
Actual:   [M11:0.18257418 M12:0.36514837 M13:0.5477225 M14:0.73029673] [M21:0.8164966 M22:0.40824836 M23:1.4600097E-07 M24:-0.40824822] [M31:NaN M32:NaN M33:NaN M34:NaN] [M41:NaN M42:NaN M43:NaN M44:NaN]
```

So it seems like 0 values in System.Numerics arent handled in these methods?

Doprez:
Beginning benchmarks mostly for fun for now since `Invert` is broken:
image

sources/core/Stride.Core.Mathematics/Double2.cs
```diff
@@ -141,6 +140,18 @@ public double this[int index]
             }
         }
 
+        /// <summary>
+        /// Casts from System.Numerics to Stride.Maths vectors
+        /// </summary>
+        /// <param name="v">Value to cast</param>
+        public static implicit operator Double2(System.Numerics.Vector2 v) => new(v.X,v.Y);
```
ericwj, line 147: Missing a space here?

sources/core/Stride.Core.Mathematics/Int2.cs
```diff
@@ -157,6 +157,18 @@ public int this[int index]
             }
         }
 
+        /// <summary>
+        /// Casts from System.Numerics to Stride.Maths vectors
+        /// </summary>
+        /// <param name="v">Value to cast</param>
+        public static explicit operator Int2(System.Numerics.Vector2 v) => new((int)v.X,(int)v.Y);
```
ericwj, line 164: And here.

sources/core/Stride.Core.Mathematics/MathUtil.Matrix.cs
```diff
@@ -0,0 +1,97 @@
...
+
+namespace Stride.Core.Mathematics;
+public static partial class MathUtil
+{
+
+    public static Matrix4x4 Orthonormalize(Matrix4x4 matrix)
+    {
+        var result = matrix;
+
+        var row1 = new System.Numerics.Vector4(matrix.M11, matrix.M12, matrix.M13, matrix.M14);
```
ericwj, line 12: It is a bit unfortunate that `Matrix4x4` can't be read as 4 vectors, but it should be and you can do it from `M11`, `M21`, etc. There are several ways of doing that, e.g.:
```cs
var rows = MemoryMarshal.Cast<Matrix4x4, Vector4>(MemoryMarshal.CreateSpan(ref m, 1));
var row4 = rows[3]; // get the bounds check out of the way asap
var row3 = rows[2];
...
```
Or, even faster:
```
var row1 = Unsafe.As<float, Vector4>(ref m.M11);
var row2 = Unsafe.As<float, Vector4>(ref m.M21);
var row3 = Unsafe.As<float, Vector4>(ref m.M31);
var row4 = Unsafe.As<float, Vector4>(ref m.M41);
```
Could also use Bitcast, but since it is by value and the sizes don't match, I'd have to try. And either way not entirely sure the above code will be optimal.

Below you should do the reverse, since certainly 16 float assignments is a bad idea that you'd want to prevent everywhere you see them.

sources/core/Stride.Core.Mathematics/MathUtil.Matrix.cs
```diff
@@ -0,0 +1,97 @@
...
+        result.M34 = row3.W;
+        result.M41 = row4.X;
+        result.M42 = row4.Y;
+        result.M43 = row4.Z;
+        result.M44 = row4.W;
+
+        return result;
+    }
+
+    public static Matrix4x4 Invert(Matrix4x4 matrix)
```
ericwj, line 86: Generally speaking, not sure whether it is a good idea to add public API for `System.Numerics` to a type that may get scrapped when all deprecations and substitutions are done. Basically what you're doing by having these public is hinting at future work to get Stride.Core.Mathematics surface area on par between `System.Numerics` types and Stride types. Those two things are sort of in competition with each other.

Before that decision is made, I would keep these internal.

sources/core/Stride.Core.Mathematics/Matrix.cs
```diff
@@ -506,6 +528,7 @@ public float Determinant()
         /// Inverts the matrix.
         /// If the matrix cannot be inverted (eg. Determinant was zero), then the matrix will be set equivalent to <see cref="Zero"/>.
         /// </summary>
+        [Obsolete("Use MathUtil.Invert instead. This method will be removed in future versions.")]
```
ericwj, line 531: Use an attribute constructor that takes `DiagnosticId`.

```cs
[Obsolete("Use MathUtil.Invert instead.", DiagnosticId = "STRIDE2100")]
public void Invert();
```
or
```cs
[Obsolete("Use MathUtil.Invert instead.", DiagnosticId = "STRIDE2100", UrlFormat = "https://doc.stride3d.net/latest/en/whatever/{0}")]
public void Invert();
```
The diagnostic `STRIDE2000` was used previously, so just reserve the range 2100-2199. Use as few diagnostics codes as possible, but do use a different one if the situation is different - if people would need to supress them in different situations, for example. For example use `STRIDE2101` for deprecations of overloads that take `in`, `ref` or `ref ro` but which are otherwise operations that shouldn't be deprecated such as `Add`. `STRDIAG` is a bad prefix imho. `STR` is string, not Stride, at least that is the automatic context free association. And yeah its a diagnostic, why even bother saying that. Talk to the docs team on Discord (@VaclavElias) to figure out how much effort it is and what the URL format should be. There should be basically only one format for all of Stride.

ericwj:
My misplaced comment from the now merged PR #2238, slightly clarified.

I wouldn't add any surface area to Stride's math in terms of parameters it accepts. It accepts Stride types and returns Stride types now. Whatever how it is implemented is a private detail and that can use `System.Numerics` types. In addition these conversion methods what #2238 is about make it a lot easier to mix-and-match, so let people use them. The exception is `Matrix`, since the conversion involves a transpose and hence is not free.

Don't deprecate math that is perfectly fine and which can be optimized by changing private implementation. Deprecate performance smells such as by-ref and in-place modification (i.e. `void Transpose()` and `void Add(ref ro, ref ro, out)`, `Subtract(in, in)` etc).
* This will eventually reduce the surface area and align what remains with `System.Numerics` for as far as the features exist. Whether you use the one or the other, hence when/if they are deprecated, is moot.
* The discussion about removing Stride's math in favor of some third party library should be about what has no equivalent in `System.Numerics` and the occasional feature that with differences in behavior, such as your first [comment](https://github.com/stride3d/stride/pull/2579#issuecomment-2561979887) above.

One goal is to get Stride's math DRY. It is so freaking silly to have a handful of implementations for one operation! So if any Stride math is touched, for the purpose of deprecations or otherwise, have them all delegate to the most performant implementation. This will be the by-value one, which would work for `readonly struct`. And let the JIT handle the `ref` `in` `ref ro` dilemmas - it should often be able to inline and hide the smell when these methods are used in larger blocks in user code.

Kryptos-FR:
The tests `TestMathUtil_Invert` et `TestMathUtil_Orthonomolize` are failing.

Kryptos-FR:
The issue (identified by the failing tests) is that our version of those types added checks that are relevant to a game engine but not in the general case.

Namely, our implementation of `Stride.Core.MathematicsVector4.Normalize()` checks whether the vector is almost zero, and does nothing in that case:
```csharp
public void Normalize()
{
    float length = Length();
    if (length > MathUtil.ZeroTolerance)
    {
        float inverse = 1.0f / length;
        X *= inverse;
        Y *= inverse;
        Z *= inverse;
        W *= inverse;
    }
}
```
While  `System.Numerics.Vector4.Normalize()` just applies the numeric formula:
```csharp
public static Vector4 Normalize(Vector4 vector) => vector / vector.Length();
```
which makes sense since this can be SIMD'd.

We might need  to fix all usage of `Stride.Core.MathematicsVector4.Normalize()` to first check for zero vectors, or add a helper class and redirect code to that class even when Numerics will be adopted.

Note: there might be other APIs with the same issue.

xen2:

Doprez:
Im not sure if this should be continued. It was mostly functional I think I remember the last remaining weirdness was between how Stride and .NET numerics handle null float values in their respective numeric types. 

Might be better to leave the work to the users and keep the numeric conversions that were already merged previously so that users can see the breaking change themselves rather than relying on Stride.

PR #2598, opened 2025-01-19
fix: TextBlock enters an infinite loop if the given available space is too small to fit even one character.

### LINKED ISSUE #2514: Stride TextBlock newline infinite loop and memory leak
There's an infinite loop at TextBlock::UpdateWrappedText ( https://github.com/stride3d/stride/blame/master/sources/engine/Stride.UI/Controls/TextBlock.cs ) when using The text Combat that was most likely placed vertically. These are screenshots from debugging:

As far as I noticed, I think there's not enough space calculated for the first row, plus and the m character keeps being added and deleted because somehow indices are being set to 0 ad infinitum (or well, until the game crashes because of an out of memory error).

Image

net2cn:
# PR Details

If TextBlock is given an available space that the width is so small to fit in even one character, TextBlock will enter an infinite loop of repeatedly adding and removing the first character, and creating a new line in the wrapped text. This unexpected infinite loop will crash the editor due to a memory leak.

With this PR, the wrapped text will have at least one character (even if the character is zero width) from the original text in each new line.

Not wrapped

(Wrap Text option disabled, horizontal overflow)

Wrapped

(Wrap Text option enabled with the fix, horizontal wrapped with vertical overflow. Without this fix the engine will crash eventually given enough time)

## Related Issue

Fixes #2514

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.UI/Controls/TextBlock.cs
+++ b/sources/engine/Stride.UI/Controls/TextBlock.cs
@@ -330,7 +330,7 @@ namespace Stride.UI.Controls
                 }
 
                 // we reached the end of the line.
-                if (currentLine.Length <= 1 || CalculateTextSize(currentLine).X <= 0) // just one or all empty characters... just go one by one.
+                if (currentLine.Length <= 1 || lineCurrentSize <= 0) // just one or all empty characters... just go one by one.
                 {
                     currentLine.Clear();
                     currentLine.Append(currentCharacter);
--- a/sources/engine/Stride.UI/Controls/TextBlock.cs
+++ b/sources/engine/Stride.UI/Controls/TextBlock.cs
@@ -296,6 +296,7 @@ namespace Stride.UI.Controls
                 float lineCurrentSize;
                 var indexNextCharacter = 0;
                 var indexOfLastSpace = -1;
+                char currentCharacter = text[0];
 
                 while (true)
                 {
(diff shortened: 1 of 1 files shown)


sources/engine/Stride.UI/Controls/TextBlock.cs
```diff
@@ -329,7 +330,13 @@ private void UpdateWrappedText(Vector3 availableSpace)
                 }
 
                 // we reached the end of the line.
-                if (indexOfLastSpace < 0) // no space in the line
+                if (currentLine.Length <= 1 || CalculateTextSize(currentLine).X <= 0) // just one or all empty characters... just go one by one.
```
Eideren, line 333: Couldn't you replace `CalculateTextSize(currentLine).X` with `lineCurrentSize <= 0`

Eideren (commented):
Thanks for the contribution. One small note, looks good otherwise !

sources/engine/Stride.UI/Controls/TextBlock.cs
net2cn: Sorry for the late reply, had been on a vacation lately. You're right, they are equivalent. I'll clean it up a bit.

Kryptos-FR (commented):
From a quick reading it looks good to me. I'll let other people do another check, if possible.

Eideren:
Thanks !

Issue #2236, opened 2024-04-27
UI Page Will Crash if a Slider is added to a Canvas panel


MeharDT:
**Release Type**: Official Release

**Version**: 4.2.0.2122

**Platform(s)**: Windows

**Describe the bug**
The UI Page will crash within Stride Game Studio if the Page contains a Canvas panel with a Slider child.

**To Reproduce**
1. Start Stride 4.2.0.2122 and create a New Game project
2. Create a UI page in the Asset view
3. Open the above UI page and add a Canvas panel to the Grid
4. Add a Slider to the Canvas, the Slider should be a child of the Canvas element in the Visual Tree
5. The Page will crash with the below error

**Log and callstacks**
`System.ArithmeticException: Function does not accept floating point Not-a-Number values.
   at System.Math.Sign(Single value)
   at Stride.Core.Mathematics.CollisionHelper.RayIntersectsRectangle(Ray& ray, Matrix& rectangleWorldMatrix, Vector3& rectangleSize, Int32 normalAxis, Vector3& intersectionPoint) in C:\BuildAgent\work\b5f46e3c4829a09e\sources\core\Stride.Core.Mathematics\CollisionHelper.cs:line 753`

Full log:
[stride_uipage_crash.txt](https://github.com/stride3d/stride/files/15136887/stride_uipage_crash.txt)

**Additional context**
A work around is to first add the Slider to a Grid panel, resize the Slider, and then attach it to the Canvas. Once the Slider is added to the Canvas its prior dimensions can be restored without issue.

TranquilAbyss:
Can someone assign this to me?

PR #2604, opened 2025-01-26
fix: Slider put in Canvas NaN Exception

TranquilAbyss:
# PR Details
Handled `Slider` UIElement measurement when `availableSizeWithoutMargins` is infinity. This was done by setting it to max value. The UI math after this cannot handle infinity values. Also other UI Elements provide real numbers for desired size.

I have considered multiple values to put in as the default. But when the the UI Editor Slider Layout section is first opened the visual width is set to a high number, maybe max float. Due to canvas not having a limited size the Slider lacks the support to handle what to do in this situation. When the Layout is open the `availableSizeWithoutMargins.X` value becomes a high number, meaning that value is originally from the parent, this is is why I chose `float.Max`. 

It also worth noting that there is still a bug behind the original bug. The first slider put into a canvas wont resize on width change until it is re-selected. I think most UI elements measure override to smallest desired size, while the slider appears to use the `availableSizeWithoutMargins` which is infinity for this issue configuration. 

I plan on leaving the issue like this. Unless someone has a better default, please test this solution before suggesting.

## Related Issue
https://github.com/stride3d/stride/issues/2236

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.UI/Controls/Slider.cs
+++ b/sources/engine/Stride.UI/Controls/Slider.cs
@@ -347,7 +347,7 @@ namespace Stride.UI.Controls
             var idealSize = trackBackgroundSprite.SizeInPixels.Y;
             var desiredSize = new Vector3(idealSize, idealSize, 0)
             {
-                [(int)Orientation] = availableSizeWithoutMargins[(int)Orientation]
+                [(int)Orientation] = float.IsInfinity(availableSizeWithoutMargins[(int)Orientation]) ? float.MaxValue : availableSizeWithoutMargins[(int)Orientation]
             };
 
             return desiredSize;


sources/engine/Stride.UI/Controls/Slider.cs
```diff
@@ -347,7 +347,7 @@ protected override Vector3 MeasureOverride(Vector3 availableSizeWithoutMargins)
             var idealSize = trackBackgroundSprite.SizeInPixels.Y;
             var desiredSize = new Vector3(idealSize, idealSize, 0)
             {
-                [(int)Orientation] = availableSizeWithoutMargins[(int)Orientation]
+                [(int)Orientation] = float.IsInfinity(availableSizeWithoutMargins[(int)Orientation]) ? float.MaxValue : availableSizeWithoutMargins[(int)Orientation]
```
Eideren, line 350: I'm sorry I can't accept this change as is, this is a workaround, you're using `MaxValue` here because it doesn't cause an issue, not because it makes sense with the rest of the logic.
`availableSizeWithoutMargins` is positive infinity here I'm guessing, if that's the case you should fix this higher up the call stack. `Measure()` could be used to replace infinities with the actual maximum size the element can take inside the container it is in for example

TranquilAbyss: I put a Slider into a Canvas.
In `Measure` we use the `Canvas` `OverrideMeasure` to 
```
// compute the desired size for the children
var childrenDesiredSize = MeasureOverride(availableSizeWithoutMargins);
```
This is what happens in `Canvas` `OverrideMeasure`
image
Due to this this section of the code
image

It is the Canvas MeasureOverride that calls the Slider's Messure. With the following value
image

How can I use the Canvas's Measure size value when its still trying to calculate the measure of the children (Slider) to be use to get the parents initial measurement? Kind of a what comes first problem (Chicken or Egg)?

My guess was Canvas was suppose to provide infinity size so handle it within the Slider, since a Slider is the only UI Element with the issue.

Issue #2489, opened 2024-10-15
[Discussion] What to do with Depth in Stride UI?


MechWarrior99:
Currently the Stride runtime UI supports depth and thickness in the UI. This is a very cool feature when making world space UI. However part of this is the ability to make elements with depth/thickness. But there are some limitations and caveats that make it impracticle to use, and harder to add new features.

The main limitation that prevents it from being useful in its current form is that it only supports solid color cubes, meaning textures will just appear as flat planes. As such I don't see it ever actually being useful for a game's UI, and even if it was, it would be limited to a very small number of games with a _very_ specific UI style.

Examples:
![Image](https://github.com/user-attachments/assets/a43d5700-7125-4732-8c0d-eb8904c12cf2)
![Image](https://github.com/user-attachments/assets/77a99d38-c873-45e5-af4b-6591da69a87e)

Keeping it in as it is, also hase some issues. It adds complexity to the editor when creating the UI, and complexity to the engine codebase.
In the future, doing any form of vector graphic based UI also becomes many times more expensive to render, and shader based UI is more complex.

An example of this is if we wanted to have rounded corners. For 2D that is straightforward, but handling it for 3D cubes becomes much more complex because it introduces the concept of edges, along with 2x the number of corners (not impossible, just more complex).

So, with these in mind I want to start a discussion on what the ideal solution would be with how to handle elements being 3D, as currently I feel it is safe to say it is a 'useless' (though neat) feature for actual games.

Here are a few options I have thought of.

1. Remove it. 
    **Pro:** Easy and most straightforward option.
    **Con:** Could be removing a potentially cool and unique feature from the engine.

2. Change it so the texture gets 'stretched' along the side (like what dropped items in Minecraft look like)
    **Pro:** Keeps and supports a unique feature of the engine.
    **Pro:** Supports a wide range of styles.
    **Con:** Would require a custom shader or possibly mesh generation for the sides of the UIElement so that it would look correct for non-square textures. Meaning it would be more expensive to render, and develop and maintain.
    **Con:** Only works for UIs that have a flatter shaded style, as the outerpixels of a texture would be stretched, and look bad for more textured styles.

3. Don't support textures, but do support rounded corners. But only along the 'depth edge', meaning like you take a normal 2D rect with rounded corners and just extruded it. (This includes supporting borders as well)
    **Pro:** Keeps and supports a unique feature of the engine.
    **Pro:** Adds support for rounded corners to 2D as well without the use of hand made textures.
[...]

PR #2605, opened 2025-01-26
[Draft] Remove depth from Stride UI

MechWarrior99:
# PR Details

Removes most Depth related features of the runtime UI system. Including depth based layout, scrolling, and rendering thickness.
Elements can still be offset along the Z axis by setting the new `DepthOffset` property. Offset is relative to parent element's `DepthOffset.

## Related Issue

#2489

## Types of changes

## Checklist


MechWarrior99:
This is PR is not ready to merge yet.
[ ] Fix Test compile errors
[ ] Fix UI not rendering with an offset.

Could use a second set of eyes on the UI rendering, even test hard coding it I can't ge tthe boxes to render with a depth offset.

MechWarrior99:
So after a good bit of using the debugger to step through the code I found that the 'issue' with the UI depth rendering is with the [`ViewProjectionMatrix`](<https://github.com/stride3d/stride/blob/master/sources/engine/Stride.Graphics/UIBatch.cs#L242>). Specifically multiplying the matrix because `Row3` of it is all 0s.

The `ViewProjectionMatrix` is ultimately is set from the [`UIRenderFeature`](<https://github.com/stride3d/stride/blob/master/sources/engine/Stride.UI/Rendering/UI/UIRenderFeature.cs#L425>). However ths matrix math is really beyond my understanding of both matricies and rendering/perspective workings. So if somone feels like taking a poke at it that would be great. As I don't even know if `Row3` being all 0s is even the real issue.

TranquilAbyss:
I understand this PR is not ready, But glancing through there a lot of breaking changes in PR. Public variables being deleted, maybe some name changes, Public variable type changes.
I am not recommending leaving dead code, but maybe try
 1. Supporting the public old Vector3 and other 3D types by leaving them to call the new 2D versions
 3. Then deprecating the 3D public vars with `[Obsolete]` with a message of what replaced them

MechWarrior99:
Thanks for taking a look @TranquilAbyss! I thought about it, but there are a few reasons why I didn't do that. The UI system is pretty much un-used (as far as I can tell), and the when it is used, the positioning related APIs are even less likely to be touched (normally it is like images for items in a  inventories and such). So the number of people these breaking changes would effect are very small (again, as far as I can tell). So it didn't seem worth the overhead.
Some are unavoidedable, as otherwise you would be showing either duplicate or dead properties in the the Properties Grid window. Which would just add confusion and bloat.

Plus, by its nature it is a breaking change since it is removing a feature from the UI. My expectation was that the PR would only be merged in for the next major release because of this. So again, less worried about breaking changes.
But if others feel it would be worth it still, I am open to trying to make it less breaking. Not sure a good way to go about it though when taking in to account the Property Grid.

Eideren:
If we were to introduce multiple breaking changes I think it would make sense to bundle it with a couple features. And have at least one more version were those properties are marked as deprecated to notify users

MechWarrior99:
@Eideren Yeah totally makes sense to have some features as well. But wanted to avoid bloating a PR with multiple things. Though that is my tendency haha.

For deprecation, would it maybe make sense to have a version where the only change is marking things with `[Obsolete]` without any change in behavior? Though not sure how to handle properties like `Size` that has its type changed.

Eideren:
> But wanted to avoid bloating a PR with multiple things
I agree, you can create another PR forked from this branch were you introduce new features, we would merge this one first and the second one with new features soon after

> without any change in behavior
Yes, just mention that those still work as is but will be removed in a future version, and point to an existing stable alternative between the two versions if there is one.

> Though not sure how to handle properties like Size that has its type changed.
We can't do much in that regard

sources/core/Stride.Core.Mathematics/Size2.cs
```diff
@@ -66,6 +66,35 @@ public Size2(int width, int height)
...
+        /// </summary>
+        /// <value>The value of the Width or Height component, depending on the index.</value>
+        /// <param name="index">The index of the component to access. Use 0 for the Width component and 1 for the Height component.</param>
+        /// <returns>The value of the component at the specified index.</returns>
+        /// <exception cref="System.ArgumentOutOfRangeException">Thrown when the <paramref name="index"/> is out of the range [0, 1].</exception>
+        public int this[int index]
+        {
+            get
+            {
+                switch (index)
```
Kryptos-FR, line 81: You can use the switch pattern here. It has a nicer syntax.

sources/core/Stride.Core.Mathematics/Size2F.cs
```diff
@@ -122,6 +162,124 @@ public override int GetHashCode()
             return !left.Equals(right);
         }
 
+        /// <summary>
+        /// Implements the operator <c>/</c>, component wise.
```
Kryptos-FR, line 166: Description of a method shouldn't explain what they implement but what they do.

For instance: "Divides the first size by the second, component-wise".

sources/editor/Stride.Assets.Presentation/AssetEditors/UIEditor/Game/UIEditorGameAdornerService.Events.cs
```diff
@@ -276,15 +276,15 @@ private void DoHighlightingAtPosition(ref Vector3 worldPosition)
...
+            if (Math.Abs(worldPosition.X) > uiComponent.Resolution.Width * 0.5f ||
+                Math.Abs(worldPosition.Y) > uiComponent.Resolution.Height * 0.5f)
                 return null;
 
             var rootElement = uiComponent.Page?.RootElement;
             if (rootElement == null)
                 return null;
 
-            var ray = new Ray(new Vector3(worldPosition.X, worldPosition.Y, uiComponent.Resolution.Z + 1), -Vector3.UnitZ);
+            var ray = new Ray(new Vector3(worldPosition.X, worldPosition.Y, 1000 + 1), -Vector3.UnitZ);
```
Kryptos-FR, line 287: 1000 seems arbitrary. Consider naming a constant in the class and use it.

sources/core/Stride.Core.Mathematics/Size2.cs
MechWarrior99: I totally agree, I was opting to keep the changes to the type as similar to the other Stride Mathmatics types as possible to keep the source code and usage consistent. I figured it was better to just update the mathmatics types all at once in another PR, then to have some using one convention and others using another.

xen2:

PR #2629, opened 2025-02-05
fix: Add assets cleanup target when calling dotnet clean, clarify ref out of sync issues

Eideren:
# PR Details
Content out of sync exception triggers when removing/adding a property to a [custom asset](https://doc.stride3d.net/latest/en/manual/scripts/custom-assets.html), this change ensures users are aware of how to proceed when encountering this issue.

## Types of changes

## Checklist

Kindly pinging @Jklawreszuk to check this one as he is more experienced with the build system than I am

### DIFF
--- a/sources/assets/Stride.Core.Assets.CompilerApp/build/Stride.Core.Assets.CompilerApp.targets
+++ b/sources/assets/Stride.Core.Assets.CompilerApp/build/Stride.Core.Assets.CompilerApp.targets
@@ -160,6 +160,10 @@
   </Target>
 
   <!-- Clean assets -->
+  <Target Name="StrideCleanAssetsOnClean" AfterTargets="Clean">
+    <RemoveDir Condition="Exists('$(StrideCompileAssetBuildPath)')" ContinueOnError="true"  Directories="$(StrideCompileAssetBuildPath)"/>
+  </Target>
+
   <Target Name="StrideCleanAsset" Condition="'$(StrideIsExecutable)' == 'true' And '$(StrideCompilerSkipBuild)' != 'true'">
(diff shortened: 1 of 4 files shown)


sources/assets/Stride.Core.Assets.CompilerApp/build/Stride.Core.Assets.CompilerApp.targets
```diff
@@ -160,6 +160,11 @@
   </Target>
 
   <!-- Clean assets -->
```
Kryptos-FR, line 162: I think always cleaning all assets on clean is too heavy. Sometimes you need to call `dotnet clean` because of a code issue (for instance because of package dependency updates).

I'm ok with a full clean being the default, by I feel like we should have a `SkipAssetsClean` option.

sources/assets/Stride.Core.Assets.CompilerApp/build/Stride.Core.Assets.CompilerApp.targets
```diff
@@ -160,6 +160,11 @@
...
+  <Target Name="StrideCleanAssetsOnClean" AfterTargets="Clean">
```
Jklawreszuk, line 163: nit: Prehabs simply "StrideCleanAssets" ?

Eideren: That was what I went with initially but there's already a target named `StrideCleanAsset`

sources/assets/Stride.Core.Assets.CompilerApp/build/Stride.Core.Assets.CompilerApp.targets
Eideren: Added a condition, let me know

PR #2631, opened 2025-02-06
fix: Legacy DynamicNavigationMeshSystem bugs and memory leak

### LINKED ISSUE #2630: Legacy dynamic navigation mesh system bugs

**Version**: 4.2.0.2232

**Platform(s)**: Windows

**Describe the bug**
1. Disabling DynamicNavigationMeshSystem causes the error: `System.NullReferenceException: 'Object reference not set to an instance of an object.'`
2. Dynamic navigation doesn't work if DynamicNavigationMeshSystem is registered before NavigationProcessor

**To Reproduce**
For bug 1:
1. Get DynamicNavigationMeshSystem and Enable = true
2. Get DynamicNavigationMeshSystem and Enable = false
3. Game throws the error

For bug 2:
1. Load the game without any navigation components and manually add DynamicNavigationMeshSystem to IGameSystemCollection
2. Load/add scene with navigation components.
3. `NavigationComponent.TryFindPath` always fails because the nav mesh is never set.

Basewq:
# PR Details

1. Ensure NavigationProcessor looks for DynamicNavigationMeshSystem when added, in case DynamicNavigationMeshSystem exists before itself.
2. Guard against null `eventArgs` which occurs when DynamicNavigationMeshSystem.Enabled = false
3. Ensure `Navigation.DestroyBuilder(builder)` is always called otherwise memory is never released on failed cases.
4. Call dispose on BulletMesh & RecastNavigationMesh to ensure memory released.

## Related Issue

Fixes #2630

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Navigation/Processors/NavigationProcessor.cs
+++ b/sources/engine/Stride.Navigation/Processors/NavigationProcessor.cs
@@ -41,6 +41,7 @@ namespace Stride.Navigation.Processors
                 throw new Exception("NavigationProcessor can not access the game systems collection");
 
             gameSystemCollection.CollectionChanged += GameSystemsOnCollectionChanged;
+            TryRegisterDynamicNavigationMeshSystem();
         }
 
         /// <inheritdoc />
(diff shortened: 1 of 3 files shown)


sources/engine/Stride.Navigation/Processors/NavigationProcessor.cs
```diff
@@ -89,10 +90,10 @@ protected override void OnEntityComponentRemoved(Entity entity, NavigationCompon
 
         private void DynamicNavigationMeshSystemOnNavigationMeshUpdatedUpdated(object sender, NavigationMeshUpdatedEventArgs eventArgs)
         {
-            var newNavigationMesh = eventArgs.BuildResult.NavigationMesh;
             NavigationMeshData data;
-            if (eventArgs.OldNavigationMesh != null && loadedNavigationMeshes.TryGetValue(eventArgs.OldNavigationMesh, out data))
+            if (eventArgs != null && eventArgs.OldNavigationMesh != null && loadedNavigationMeshes.TryGetValue(eventArgs.OldNavigationMesh, out data))
```
Kryptos-FR, line 94: Can be merged into a single check: `eventArgs?.OldNavigationMesh is not null`

Basewq: Thanks for point this out. Thinking on this again, I'm realizing `eventArgs` shouldn't be null anyway, and we need the old nav mesh to remove it from the processor when DynamicNavigationMeshSystem gets disabled. Have converted this pull to draft to fix this.

Basewq: Ok, so I decided to just null check `eventArgs` instead of ensuring it is never null unless advised otherwise.
`UpdateNavigationMesh` already accounts for unloading the stale mesh in the `UpdateNavigationMesh` method.
Also changing this behavior would technically be breaking change, though this would only occur when DynamicNavigationMeshSystem gets disabled (which was broken anyway).

Kryptos-FR (approved):

PR #2636, opened 2025-02-12
fix: Correct Buffer size for #2530

Jklawreszuk:
# PR Details

Fixes an exception for some sizes

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Assets/SpriteFont/Compiler/TrueTypeImporter.cs
+++ b/sources/engine/Stride.Assets/SpriteFont/Compiler/TrueTypeImporter.cs
@@ -23,22 +23,15 @@
 
 using System;
 using System.Collections.Generic;
-using System.IO;
-
-using SharpDX.Direct2D1;
 using Stride.Graphics.Font;
 
 namespace Stride.Assets.SpriteFont.Compiler
 {
-    using System.Diagnostics;
     using System.Drawing;
-    using System.Drawing.Imaging;
-    using SharpDX.DirectWrite;
     using FreeImageAPI;
     using SharpDX.Mathematics.Interop;
     using SharpFont;
     using Stride.Core;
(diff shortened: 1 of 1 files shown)


sources/engine/Stride.Assets/SpriteFont/Compiler/TrueTypeImporter.cs
```diff
@@ -128,7 +121,7 @@ private unsafe Glyph ImportGlyph(Face face, char character, float fontSize, Font
                 face.LoadGlyph(index, LoadFlags.Render, LoadTarget.Normal);
 
                 var glyphBitmap = face.Glyph.Bitmap;
-                var bufferData = new Span<byte>((byte*)glyphBitmap.Buffer, glyphBitmap.Rows * glyphBitmap.Pitch);
+                var bufferData = new Span<byte>((byte*)glyphBitmap.Buffer, glyphBitmap.Rows * glyphBitmap.Width);
```
Ethereal77, line 124: It makes more sense that the size is `Rows` times `Pitch`. Why that results in an incorrect buffer size? Worst case should be that `Width == Pitch` if byte-sized and no paddings, no?

As `GetFont()` is a new method, if `Pitch` is not correctly returning the pitch, then maybe we should fix it there.

Also related: Why create a new `GetFont()` method returning a `Face` instead of modifying the already existing `GetFontFace()` that returns a `FontFace`?
I ask because maybe we should encapsulate `Face` inside `FontFace` (as it already has a correct `Pitch` property), and not change this line.

Ethereal77 (commented):
A few comments about the changes.

I think currently we are losing some features in the change to SharpFont, isn't it?
Specifically I see much more logic in the previous `GetFontFace()` methods that took the font styles into account.

Sorry if I'm misunderstanding the change.

PR #2642, opened 2025-02-16
feat: Remove unused PackageInstall Project

Jklawreszuk:
# PR Details

My PR removes the Stride.PackageInstall project. It looks like the main purpose was to install the prerequisites, but the prerequisites are automatically installed during the main installation, so It's no longer needed.

I ping @Kryptos-FR  just in case.

## Types of changes

## Checklist


xen2:
We still need this.
After installing launcher, future versions of Stride might need new/updated prerequisites.

Kryptos-FR:
I also think the same as @xen2. However we will need to revisit this when we want to have an xplat Launcher and to install/run the editor on Linux.

PR #2639, opened 2025-02-14
refactor: Nullable and Modernization for Stride.Core. & Stride.Core.AssemblyProcessor

Kryptos-FR:
# PR Details

In the same spirit of #2266, modernize `Stride.Core` and `Stride.Core.AssemblyProcessor`.

## Description
1. Enable `nullable` on both projects.
2. Use new language features such as collection initialization, pattern matching, range and indexers, etc.
3. Enable implicit usings, and use file-scoped namespaces.

## Related Issue

#2155
#2156

## Types of changes

- [x] Docs change / **refactoring** / dependency upgrade
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [ ] My change requires a change to the documentation.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.
- [x] **I have built and run the editor to try this change out.**

PR #2643, opened 2025-02-16
refactor: Nullable and Modernization for Core projects

Kryptos-FR:
# PR Details

Similar to #2639.

## Description

As in the previous PR, enable C# built-in nullable annotations on the project and remove references to our old custom ones. In addition, modernization of the code (pattern matching, file-scoped namespaces, collection initialization, etc.

### Equals
Worth of note is a fix of the `Equals(object?)` implementation for some structs in `Core.Mathematics`. While it was unlikely to be used (as the same structs also define a type version of `Equals`), the implementation was doing unnecessary checks and struct boxing (which also had Resharper detect the implementation as mutating).

Consider the previous implementation for `Vector3`:
```csharp
public override bool Equals(object value)
{
    if (value == null)
        return false;

    if (value.GetType() != GetType())
        return false;

     return Equals((Vector3)value);
}
```

The corresponding IL code is:
```asm
IL_0000: ldarg.1
IL_0001: brtrue.s IL_0005

IL_0003: ldc.i4.0
IL_0004: ret

IL_0005: ldarg.1
IL_0006: callvirt instance class [System.Runtime]System.Type [System.Runtime]System.Object::GetType()
IL_000b: ldarg.0
IL_000c: ldobj Vector3
IL_0011: box Vector3
IL_0016: call instance class [System.Runtime]System.Type [System.Runtime]System.Object::GetType()
IL_001b: call bool [System.Runtime]System.Type::op_Inequality(class [System.Runtime]System.Type, class [System.Runtime]System.Type)
IL_0020: brfalse.s IL_0024

IL_0022: ldc.i4.0
IL_0023: ret

IL_0024: ldarg.0
IL_0025: ldarg.1
IL_0026: unbox.any Vector3
IL_002b: call instance bool Vector3::Equals(valuetype Vector3)
IL_0030: ret
```
We can clearly see the virtual calls necessary for comparing the types as well as the boxing required for such call.

The new implementation is simpler:
```csharp
public override readonly bool Equals(object? value)
{
    return value is Vector3 vector && Equals(vector);
}
```

And the corresponding IL also:
```asm
IL_0000: ldarg.1
IL_0001: isinst Vector3
IL_0006: brfalse.s IL_0017

IL_0008: ldarg.1
IL_0009: unbox.any Vector3
IL_000e: stloc.0
IL_000f: ldarg.0
IL_0010: ldloc.0
IL_0011: call instance bool Vector3::Equals(valuetype Vector3)
IL_0016: ret

IL_0017: ldc.i4.0
IL_0018: ret
```

### Hashcode

Another point of improvement is the computation of hashcodes. In lots of places we had a custom implementation that wasn't necessary efficient or even correct (sometimes it was missing the `unchecked` context). .NET provides a battle-tested implementation with the right properties and efficiency.

Our previous implementation for `Matrix`:
```csharp
public override int GetHashCode()
{
    return M11.GetHashCode() + M12.GetHashCode() + M13.GetHashCode() + M14.GetHashCode() +
        M21.GetHashCode() + M22.GetHashCode() + M23.GetHashCode() + M24.GetHashCode() +
        M31.GetHashCode() + M32.GetHashCode() + M33.GetHashCode() + M34.GetHashCode() +
        M41.GetHashCode() + M42.GetHashCode() + M43.GetHashCode() + M44.GetHashCode();
}
```

Now, using `Hashcode.Combine`:
```csharp
public override readonly int GetHashCode()
{
    return HashCode.Combine(
        HashCode.Combine(M11, M12, M13, M14),
        HashCode.Combine(M21, M22, M23, M24),
        HashCode.Combine(M31, M32, M33, M34),
        HashCode.Combine(M41, M42, M43, M44)
    );
}
```

### Swapping

Another fun improvement is how to swap two values. The usual implementation was to use a temporary variable:
```csharp
var tmp = a;
a = b;
b = tmp;
```

We can now use a value tuple:
```csharp
(a, b) = (b, a);
```

In my opinion it does improve readability when applied to `Matrix.Transpose`.
Before:
```csharp
public void Transpose()
{
    float temp;
    temp = M21; M21 = M12; M12 = temp;
    temp = M31; M31 = M13; M13 = temp;
    temp = M41; M41 = M14; M14 = temp;
    
    temp = M32; M32 = M23; M23 = temp;
    temp = M42; M42 = M24; M24 = temp;
    
    temp = M43; M43 = M34; M34 = temp;
}
```
After:
```csharp
public void Transpose()
{
    (M21, M12) = (M12, M21);
    (M31, M13) = (M13, M31);
    (M41, M14) = (M14, M41);
    
    (M32, M23) = (M23, M32);
    (M42, M24) = (M24, M42);
    
    (M43, M34) = (M34, M43);
}
```

## Related Issue

#2155
#2156

## Types of changes

## Checklist


sources/core/Stride.Core.CompilerServices/Generators/ModuleInitializerGenerator.cs
```diff
@@ -50,7 +50,7 @@ private void Generate(SourceProductionContext context, ImmutableArray<OrderedMod
         var code = new StringBuilder();
         if (array.Length == 0)
             return;
-        foreach (var module in array.OrderBy(x => x.Order))
+        foreach (var module in array.OrderBy(x => x?.Order))
```
IXLLEGACYIXL, line 53: `provider2.Where(not null).Collect()` can make the immutable array non nullable , is most likely better overall so no module can be nullable

Kryptos-FR: I tried:
```csharp
var provider2 = provider.Where(x => x is not null).Collect();
```
It doesn't seem to satisfy the nullable analyzer: it still sees an `ImmutableArray<OrderedModuleInitializerMethod?>`.

I have to write it like that, which I find clumsy:
```csharp
IncrementalValueProvider<ImmutableArray<OrderedModuleInitializerMethod>> provider2 = provider.Where(x => x is not null).Collect()!;
```

IXLLEGACYIXL: allright, thank you

then i will dig deeper into it, you can leave it as is in your current PR

Kryptos-FR:
Note: I will do a few more projects in that PR (`Core.Yaml`, `Core.Design`, `Core.Translation`) as well as their corresponding test projects.

Kryptos-FR:
I finished all projects under `sources/core`.

One notable exception is `Stride.Core.Yaml`. I think we should review replacing it with https://github.com/aaubry/YamlDotNet/ (as it was based on an old version of it), and thus I thought it was not worth the effort modernizing/annotating it.

Eideren:
Welp, this one is going to create a whole bunch of merge conflicts, are you okay with taking care of the PRs that might be conflicting with this one if I were to merge this PR now ?

Kryptos-FR:
@Eideren I can take care of the conflicts assuming all those PRs allow maintainers to do changes.

Kryptos-FR:
PR with conflicts following the merge:
* https://github.com/stride3d/stride/pull/2605

PR #2397, opened 2024-07-21
WIP: Windows on Arm support

archanox:
# PR Details

This is a highly experimental WIP draft of getting the entire Stride suite to compile and run on Windows on Arm (WoA).
There has been a lot of broad stroke changes that change projects from win-x64 to win-arm64, just to test this proof of concept.
I have gone through every mention of x86/x64 through the repo to add in arm64 where appropriate, but there are some issues surrounding the native libraries. I feel that there's perhaps documentation somewhere on how to build these binaries that exist in the repo, maybe?

I am seeking any and all feedback and assistance with this PR where possible. I am on Discord in the Stride server as pierce @archanox

Milestone:
1. Getting the solution to build for arm64 with everything working (tests, packaged, installer built, etc)
2. Configure multiple targets for a win-x64 build and win-arm64

Current Issues checked in:
- ClearScript was only bundled with x86/x64 binaries. Dependencies for the Metrics were erroneously upgraded for the entire project. Untested.
- There are missing binaries for arm64 that do not exist in the repo
- 32bit/64bit checks assume x86 platform
- Native library build scripts appear to be pointing at an old version of visual studio
- MSBuild paths for x86/x64 have been temporarily removed until successful build & test.
- Preliminary ARM64 UWP changes were checked in, but will likely be reverted.

Goal of this PR:
- To develop on a WoA device, with Visual Studio integration and to build for Windows and Linux platforms.

Outside the scope of this PR:
- CI for automated builds of the arm64 version
- NuGet publishing

## Related Issue

#2018

## Types of changes

- [x] Docs change / refactoring / dependency upgrade
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] My change requires a change to the documentation.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.
- [ ] **I have built and run the editor to try this change out.**

PR #2644, opened 2025-02-16
feat: Adds native dependencies for win-arm64

azeno:
# PR Details

Allows to develop and also run games on `win-arm64`.

Needs MSVC v143 - VS2022 C++ ARM64 build tools (Latest) 

- The used versions of those dependencies are often rather dated and didn't build out of the box on win-arm64 / Visual Studio 2022, minimal patches had to be applied. Need to push/document those separately.
- As far as I understood PVRTexLib is delivered as binary only and there's no win-arm64 - therefor added a check to not load it at all. In the end it's only needed by the asset compiler, which in turn is platform neutral. However when invoking it on a arm64 machine, the `dotnet path/to/asset/compiler` will launch it in arm64. If there would be some option to launch it in x64 we wouldn't have any issues with asset compiling. Still the editor might have issues, since its running in process?
- Removed the FreeImageNET load - it's not used anywhere in code
- Fixes some PInvoke signatures in DxtWrapper

The game studio also runs on arm64, but one must set the runtime identifiers in Stride.GameStudio/Editor/Assets.Presentation accordingly. Would be nice to add an arm64 version to the installer, but different topic I'd say.

## FFmepg
checkout.bat says to use n3.3.3 tag however I didn't manage to get it to build. Was therefor using release/3.4 as a base.
Basically had to cherry-pick two commits from main branch in regards to arm64.
A quick test with importing a video asset and playing it later in the game worked. 
https://github.com/azeno/FFmpeg/tree/win-arm64

## FreeImage
- Added a checkout.bat file which will also apply the needed fixes

## Recast
Guessed the commit by comparing the checked in header files to those in the original repo.  Added the used commit-id to checkout.bat file.

## Needed PRs in other repos
- https://github.com/stride3d/BulletSharpPInvoke/pull/5
- https://github.com/stride3d/NativePath/pull/2
- https://github.com/stride3d/freetype/pull/1

## Related Issue
#2397 

## Types of changes

## Checklist

### DIFF
--- a/deps/FFmpeg/dotnet/build.sh
+++ b/deps/FFmpeg/dotnet/build.sh
@@ -44,6 +44,24 @@ configure_windows_x64_shared() {
     --prefix="${BASEDIR}/build/windows/x64"
 }
 
+configure_windows_arm64_shared() {
+  "${BASEDIR}/configure" \
+    --toolchain=msvc \
+    --arch=arm64 \
+    --enable-asm \
+    --enable-shared \
+    --enable-version3 \
+    --enable-w32threads \
+    --enable-yasm \
+    --disable-debug \
+    --disable-doc \
+    --disable-programs \
+    --disable-static \
+    --disable-x86asm \
+    --target-os=win64 \
+    --prefix="${BASEDIR}/build/windows/arm64"
+}
+
 configure_windows_x86_shared() {
   "${BASEDIR}/configure" \
     --toolchain=msvc \
@@ -99,6 +117,26 @@ configure_windows_x64_exe() {
     --prefix="${BASEDIR}/build/windows/x64"
 }
 
+configure_windows_arm64_exe() {
+  "${BASEDIR}/configure" \
+    --toolchain=msvc \
+    --arch=arm64 \
+    --enable-asm \
+    --enable-static \
+    --enable-version3 \
+    --enable-w32threads \
+    --enable-yasm \
+    --disable-debug \
+    --disable-doc \
+    --disable-ffplay \
+    --disable-ffprobe \
+    --disable-ffserver \
+    --disable-shared \
+    --disable-x86asm \
+    --target-os=win64 \
+    --prefix="${BASEDIR}/build/windows/arm64"
+}
+
 install() {
   make -j${NUMBER_OF_CORES} && make install
 }
@@ -110,6 +148,21 @@ build_windows_x64_shared() {
   
   # Configure ffmpeg (can be skipped if already done)
   configure_windows_x64_shared
+  # Make sure any previous build doesn't pollute the current one (can be skipped if configure didn't change since last time)
+  clean
+  # Build FFmpeg and copy binaries+include files into the installation folder (defined by --prefix in the configure command)
+  install
+  
+  popd
+}
+
+build_windows_arm64_shared() {
+  oot="${BASEDIR}/build/windows/arm64/oot/ffmpeg"
+  mkdir -p "$oot"
+  pushd "$oot"
+  
+  # Configure ffmpeg (can be skipped if already done)
+  configure_windows_arm64_shared
   # Make sure any previous build doesn't pollute the current one (can be skipped if configure didn't change since last time)
   clean
   # Build FFmpeg and copy binaries+include files into the installation folder (defined by --prefix in the configure command)
@@ -148,6 +201,21 @@ build_windows_x64_exe() {
   popd
 }
 
+build_windows_arm64_exe() {
+  oot="${BASEDIR}/build/windows/arm64/oot/ffmpeg"
+  mkdir -p "$oot"
+  pushd "$oot"
+  
+  # Configure ffmpeg (can be skipped if already done)
+  configure_windows_arm64_exe
+  # Make sure any previous build doesn't pollute the current one (can be skipped if configure didn't change since last time)
+  clean
+  # Build FFmpeg and copy binaries+include files into the installation folder (defined by --prefix in the configure command)
+  install
+  
+  popd
+}
+
 build_windows_x86_exe() {
   oot="${BASEDIR}/build/windows/x86/oot/ffmpeg"
   mkdir -p "$oot"
@@ -171,3 +239,4 @@ NUMBER_OF_CORES=$(nproc)
 build_windows_x86_shared
 #build_windows_x64_exe
 #build_windows_x86_exe
+#build_windows_arm64_exe
--- a/deps/NativePath/checkout.bat
+++ b/deps/NativePath/checkout.bat
@@ -5,4 +5,8 @@ IF NOT ERRORLEVEL 0 (
 ) 
 %GIT_CMD% clone https://github.com/stride3d/NativePath ../../externals/NativePath
 IF NOT ERRORLEVEL 0 PAUSE
+pushd ..\..\externals\NativePath
+%GIT_CMD% checkout e615840e57ea80ce30b9e9f29f5fd84ca9b4b253
+popd
+IF NOT ERRORLEVEL 0 PAUSE
 
--- a/sources/core/Stride.Core/Native/NativeLibraryHelper.cs
+++ b/sources/core/Stride.Core/Native/NativeLibraryHelper.cs
@@ -53,6 +53,7 @@ public static class NativeLibraryHelper
                 Architecture.X86 => "x86",
                 Architecture.X64 => "x64",
                 Architecture.Arm => "ARM",
+                Architecture.Arm64 => "arm64",
                 _ => throw new PlatformNotSupportedException(),
             };
 
--- a/sources/engine/Stride.Audio/Stride.Native.Libs.targets
+++ b/sources/engine/Stride.Audio/Stride.Native.Libs.targets
@@ -1,13 +1,13 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
-    <StrideNativePathLibsWindows>libCompilerRt.lib libCelt.lib</StrideNativePathLibsWindows>
-    <StrideNativePathLibsUWP>libCompilerRt.lib libCelt.lib Xaudio2.lib</StrideNativePathLibsUWP>
+    <StrideNativePathLibsWindows>libCelt.lib</StrideNativePathLibsWindows>
+    <StrideNativePathLibsUWP>libCelt.lib Xaudio2.lib</StrideNativePathLibsUWP>
   </PropertyGroup>
 
   <ItemGroup>
     <StrideNativePathLibsiOS Include="$(MSBuildThisFileDirectory)..\..\..\deps\NativePath\iOS\libCelt.a" />
     <StrideNativePathLibsmacOS Include="$(MSBuildThisFileDirectory)..\..\..\deps\NativePath\iOS\libCelt.a" />
-    <StrideNativePathLibsLinux Include="libCelt.a;libCompilerRt.a" />
-    <StrideNativePathLibsAndroid Include="libCelt.a;libCompilerRt.a" />
+    <StrideNativePathLibsLinux Include="libCelt.a" />
+    <StrideNativePathLibsAndroid Include="libCelt.a" />
   </ItemGroup>
 </Project>
--- a/sources/native/Stride.Native.targets
+++ b/sources/native/Stride.Native.targets
@@ -26,6 +26,8 @@
   
     <!--<StrideNativeOutputPath>$([MSBuild]::MakeRelative('$(OutputPath)', '$(StridePackageStridePlatformBin)\'))</StrideNativeOutputPath>-->
     <AllowedOutputExtensionsInPackageBuildOutputFolder>.so; .a; $(AllowedOutputExtensionsInPackageBuildOutputFolder)</AllowedOutputExtensionsInPackageBuildOutputFolder>
+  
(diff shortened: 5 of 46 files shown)


deps/FFmpeg/dotnet/build.sh
```diff
@@ -168,6 +236,8 @@ NUMBER_OF_CORES=$(nproc)
 
 # Uncomment a line to run that configuration
 #build_windows_x64_shared
-build_windows_x86_shared
+#build_windows_x86_shared
```
Kryptos-FR, line 239: Is x86 disabled on purpose?

sources/core/Stride.Core/Native/NativeLibraryHelper.cs
```diff
@@ -56,6 +56,7 @@ public static void PreloadLibrary(string libraryName, Type owner)
                     Architecture.X86 => "x86",
                     Architecture.X64 => "x64",
                     Architecture.Arm => "ARM",
+                    Architecture.Arm64 => "arm64",
```
Kryptos-FR, line 59: Is it case-sensitive? I see `ARM64` in some other places.

sources/engine/Stride.Audio/Stride.Native.Libs.targets
```diff
@@ -1,13 +1,13 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
-    <StrideNativePathLibsWindows>libCompilerRt.lib libCelt.lib</StrideNativePathLibsWindows>
-    <StrideNativePathLibsUWP>libCompilerRt.lib libCelt.lib Xaudio2.lib</StrideNativePathLibsUWP>
+    <StrideNativePathLibsWindows>libCelt.lib</StrideNativePathLibsWindows>
+    <StrideNativePathLibsUWP>libCelt.lib Xaudio2.lib</StrideNativePathLibsUWP>
```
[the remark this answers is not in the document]
azeno: The linker was happy. It would've complained I assume.

sources/core/Stride.Core/Native/NativeLibraryHelper.cs
azeno: This ends up as the runtime identifier (RID), for example `win-arm64` - to my knowledge that's lower case. Why the upper entry isn't done that way I don't know.

deps/FFmpeg/dotnet/build.sh
azeno: Will revert that one - in general all those custom build scripts need to be taken with a grain of salt. In their current state you can use them as some sort of reference point, but none of them were working out of the box. Either wrong paths or out-dated ones, wrong output paths, out-commented code, etc.

sources/tools/Stride.TextureConverter/Backend/Wrappers/DxtNetWrapper.cs
```diff
@@ -515,10 +515,10 @@ internal class Utilities
         [DllImport("DxtWrapper", CallingConvention = CallingConvention.Cdecl, CharSet = CharSet.Unicode), SuppressUnmanagedCodeSecurity]
         private extern static void dxtComputePitch(DXGI_FORMAT fmt, int width, int height, out int rowPitch, out int slicePitch, CP_FLAGS flags);
 
-        [DllImport("DxtWrapper", CallingConvention = CallingConvention.Cdecl, CharSet = CharSet.Auto), SuppressUnmanagedCodeSecurity]
+        [DllImport("DxtWrapper", CallingConvention = CallingConvention.Cdecl, CharSet = CharSet.Ansi), SuppressUnmanagedCodeSecurity]
```
Jklawreszuk, line 518: Ansi ? Why?

azeno: If we look in the C header, the method is defined with `char` (1 byte). With Auto we don't know what it will take, clearly on arm64 it took the wrong one. I debugged that part, the data I got on unmanaged was garbage with Auto. With Ansi all good.

Jklawreszuk: Seems legit. I quickly tested on Linux DDS and TGA files, no regression here 👍

azeno:
@Kryptos-FR @Jklawreszuk I just force pushed - added more info in the PR summary. For those deps where Stride maintains a fork I created PRs, and for those where it doesn't I added patch files or forked myself if the original repo was hosted on github.

sources/native/Stride.Native.targets
```diff
@@ -58,7 +60,7 @@
   </ItemGroup>
   <!-- Can be deleted -->
   <ItemGroup Condition="'$(TargetFramework)' == '$(StrideFrameworkUWP)'">
-    <StrideNativeCPU Include="x86;x64;ARM" LibraryExtension=".dll" />
+    <StrideNativeCPU Include="x86;x64;ARM;ARM64" LibraryExtension=".dll" />
```
anthony-linaro, line 63: Is it worth removing the old ARM32 (ie, plain "ARM") from here, so users don't accidentally use the wrong arch? It hasn't been relevant since WinRT

sources/native/Stride.Native.targets
```diff
@@ -138,6 +144,11 @@
...
     <MSBuild Projects="$(MSBuildThisFileDirectory)WindowsProjects\WindowsDesktop\WindowsDesktop.vcxproj" Targets="Build" Properties="StrideNativeOutputName=$(StrideNativeOutputName);StrideNativeOutputDir=$(StrideNativeOutputPath)\runtimes\win-x64\native;StrideDependenciesDir=$(MSBuildThisFileDirectory)..\..\deps\;StrideNativePathLibs=libNativePath.lib $(StrideNativePathLibsWindows);StrideNativeProjectFolder=$(ProjectDir);StrideNativeProjectObjFolder=$(OutputObjectPath)\win-x64;Configuration=$(Configuration);Platform=x64" StopOnFirstFailure="true" />
 
+    <MakeDir Directories="$(OutputObjectPath)\win-arm64" Condition="'$(StrideNativeWindowsArm64Enabled)' == 'true'"/>
+    <Exec Condition="'%(StrideNativeCFile.Extension)' != '.cpp' AND '$(StrideNativeWindowsArm64Enabled)' == 'true'" Command="&quot;$(StrideNativeClangCommand)&quot; -gcodeview -fno-ms-extensions -nobuiltininc -nostdinc++ $(StrideNativeClang) -DNEED_DLL_EXPORT -o &quot;$(OutputObjectPath)\win-arm64\%(StrideNativeCFile.Filename).obj&quot; -c &quot;%(StrideNativeCFile.FullPath)&quot; -fms-extensions -DWINDOWS_DESKTOP -target aarch64-pc-windows-msvc" />
```
anthony-linaro, line 148: Excuse my unfamiliarity with the Stride3D codebase here, but have you tested this with clang?

If it works, it might even be beneficial to compile everything with clang, instead of MSVC, I found when I moved Blender builds to clang-cl instead of MSVC, I got a 46% perf improvement in rendering workloads.

anthony-linaro (commented):
A quick initial pass, just based on a read through

sources/native/Stride.Native.targets
azeno: I wanted to keep the changes arm64 related only. Modernising the native build chain would certainly be great but I believe that should be done separately. There are already attempts to so, for example one using zig which in turn uses clang.

sources/native/Stride.Native.targets
azeno: Are we even supporting UWP still? There were discussions to remove it entirely, no?

azeno: But then again - removing UWP should be done separately I'd say.

azeno: Good to know. I was referring to the line of code you put the comment on which is only relevant if compiling for UWP. And if UWP is out of the picture, that whole block of code can go out, not just the ARM32 part.

azeno: Haha it says even in the comment one line above, "can be deleted" :)

archanox:
Awesome job on picking this up again after I unceremoniously gave up.
Looks like you actually did the bits with the precompiled build artifacts successfully, which were too daunting for me due to how inconsistent it was.
Ideally it would be all built as part of the entire build process as to avoid future issues with additional platform bring-up.

I know my attempt was pretty messy, but hopefully there was something worthwhile to be taken from it, in particular the strange arm64 msbuild registry location.

azeno:
While trying to get a grip on the build error of the Visual Studio package on teamcity:
- Did a console build locally and saw that one invocation of Stride.Core.Tasks was using a wrong path. See my commit above for details.
- After that it was building fine locally. However on teamcity it's failing with a different error at `MSBuildLocator.RegisterDefaults()` in Stride.Core.Tasks - anyone any idea what's going on there?

sources/native/Stride.Native.targets
Kryptos-FR: Leave that line as-is for now, since dealing with UWP is another issue.

These lines won't be it by other configurations.

You could remove your addition of ARM64 though since it might not be relevant to UWP even if it makes a comeback.

sources/native/Stride.Native.targets
Kryptos-FR: Yes I think we need another issue for it.

Kryptos-FR:
Did you try to build on Linux or Windows? It's ok to fail on Linux because those package are for Visual Studio which is Windows-only.

sources/tools/Stride.VisualStudio.Package.Tests/Stride.VisualStudio.Package.Tests.csproj
```diff
@@ -76,7 +76,7 @@
   <Import Project="$(StrideSdkTargets)" />
   <Target Name="LocateDevenv" AfterTargets="PrepareForBuild">
     <!-- Compute and save VisualStudio path to a file so that it can be used when running the test. Note: ideally we should use a Task, but Visual Studio lock the files -->
-    <Exec Command="&quot;..\..\core\Stride.Core.Tasks\bin\$(Configuration)\$(StrideEditorTargetFramework)\Stride.Core.Tasks.exe&quot; locate-devenv &quot;$(MSBuildBinPath)&quot;" ConsoleToMsBuild="true">
+    <Exec Command="&quot;..\..\core\Stride.Core.Tasks\bin\$(Configuration)\$(StrideXplatEditorTargetFramework)\Stride.Core.Tasks.exe&quot; locate-devenv &quot;$(MSBuildBinPath)&quot;" ConsoleToMsBuild="true">
```
Kryptos-FR, line 79: Good catch.

azeno:
Turns out the Visual Studio update removed .NET 8 on the build machine. @xen2 just re-installed it, should be ok now.

sources/tools/Stride.VisualStudio.Package.Tests/Stride.VisualStudio.Package.Tests.csproj
Kryptos-FR: The funny part is that this test project does nothing. It hasn't been updated for ages and doesn't work with current Stride (we could make it work but it's not trivial).

archanox:
@azeno is there anything that I can help with?

Are we in a position to update vstools? If so, I believe we can consolidate the references to the one version iirc.

How far did you get with getting the launcher and editor to build and run?

azeno:
Basically waiting that the PRs in the other repos get accepted. Once that happens I can update this one again with the correct URLs/commit ids.

Editor is working. Asset compiler also. Like I wrote in the description, it's only one library for which there's no arm64 version.

azeno:
@xen2 Thanks for merging the external repos. Once NativePath is merged (I've modified it based on our comment) I'll put this PR out of draft state.

azeno:
Updated the checkout scripts once more, I think all external changes are now documented. Moving out of draft state.

Kryptos-FR:
@anthony-linaro as far as I understand it, runtime identifiers have no impact on non-executable project (i.e. libraries). It only matters for executables (such as the tests).

I'm actually adding `linux-x64` to some of those tests. This PR should add `win-arm64` as well but not replace or remove `win-x64`.

So basically, it should go from:
```xml
<Project>
  <PropertyGroup>
    <RuntimeIdentifier>win-x64</RuntimeIdentifier>
  </PropertyGroup>
</Project>
```
to:
```xml
<Project>
  <PropertyGroup>
    <RuntimeIdentifiers>win-arm64;win-x64</RuntimeIdentifiers>
  </PropertyGroup>
</Project>
```
Note that `RuntimeIdentifier` (singular) becomes `RuntimeIdentifiers` (plural).

deps/NativePath/checkout.bat
```diff
@@ -5,4 +5,8 @@ IF NOT ERRORLEVEL 0 (
 ) 
 %GIT_CMD% clone https://github.com/stride3d/NativePath ../../externals/NativePath
 IF NOT ERRORLEVEL 0 PAUSE
+pushd ..\..\externals\NativePath
+%GIT_CMD% checkout e615840e57ea80ce30b9e9f29f5fd84ca9b4b253
```
xen2, line 9: Good idea!

azeno:
@Kryptos-FR @Jklawreszuk Please tell me if there's anything missing or you need more info to be able to merge this PR.

Jklawreszuk (approved):

Eideren:
Thanks !

PR #2650, opened 2025-02-22
fix: Ensure teleporting bodies notify bepu's data structure

### LINKED ISSUE #2649: Bepu Physics - Raycast does not intercept Entity with BodyComponent
**Release Type**: Official Release 4.2.0.2293

**Platform(s)**: Windows

**Describe the bug**
I have a sample project where I have 2 spheres, using Bepu Physics.
The first one, on the right with gray metal material, has a `StaticComponent` and a Sphere Collider.
The second one, on the left with golden metal material, has a `BodyComponent` flagged as `Kinematic` and a Sphere Collider as well.

I tweaked the default `BasicCameraController.cs` created by default Game Studio to add a raycast in order to display the hovering entity by using `DebugText.Print()` method.
✔When hovering on the "Static" sphere, even when it is moved away from its initial position, is working well.
❌When hovering on the "Body" sphere, raycasting returns that entity only when the sphere is in its initial position. If the user left clicks on the screen, the sphere moves to that point and raycasting stops working.

Default initial position:

Image

After moving "Body" sphere away from initial position raycast does not intercept the sphere anymore ("hovering on Ground" tells me that):

Image

**If I return the sphere back to origin, raycast starts working ok again.**

Raycasting on "Static" sphere always works ok.

**To Reproduce**
Clone repro repo here https://github.com/luca-domenichini/StrideRaycastIssueRepro

My doubt is that, maybe, I am not moving the "body sphere" the right way.
To move that sphere I am using a `MoverScript` attached to `Ground` entity.
In that script I set `BodyComponent::Position` property instead of the usual `Entity.Transform.Position`.

This is the line of code where "lerping" the position:
https://github.com/luca-domenichini/StrideRaycastIssueRepro/blob/bcf1c70ceb8cf1ace3fc30bc0fbd856b9765bc38/StrideRaycastIssue/MoverScript.cs#L73

**Expected behavior**
Raycast should work fine on moved BodyComponent entities.

**Additional context**
I also tried to move the sphere by setting `Entity.Transform.Position`, but that seems to be worse since the Entity is moving but position of the BodyComponent stands still at (0, 0.5, 1) insted of following the Entity.. but maybe I am missing something..

Image

Eideren:
# PR Details
We have to explicitly notify bepu when moving a body, even more so when sleeping, otherwise its internal representation inside the physics scene is not updated, leading to misses when doing physics tests

## Related Issue
fix #2649

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.BepuPhysics/Stride.BepuPhysics/BodyComponent.cs
+++ b/sources/engine/Stride.BepuPhysics/Stride.BepuPhysics/BodyComponent.cs
@@ -212,7 +212,7 @@ public class BodyComponent : CollidableComponent
     }
 
     /// <summary>
-    /// The position of this body in the physics scene, setting it will teleport this object to the positoin provided.
+    /// The position of this body in the physics scene, setting it will teleport this object to the position provided.
     /// </summary>
     /// <remarks>
     /// Using this property to move objects around is not recommended,
@@ -226,12 +226,14 @@ public class BodyComponent : CollidableComponent
         set
         {
             if (BodyReference is { } bodyRef)
+            {
                 bodyRef.Pose.Position = PreviousPose.Position = value.ToNumeric();
+                bodyRef.UpdateBounds();
+            }
 
             Quaternion dummy = default;
             WorldToLocal(ref value, ref dummy);
             Entity.Transform.Position = value;
-
         }
     }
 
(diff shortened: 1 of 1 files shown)


luca-domenichini:
@Eideren maybe the same fix should be applied to `Orientation` property? https://github.com/stride3d/stride/blob/257d6afebc9f14835db7b7845fcfb67926864870/sources/engine/Stride.BepuPhysics/Stride.BepuPhysics/BodyComponent.cs#L249

And maybe, it would be even better to have a method to set `Position` and `Orientation` in just one call, to minimze calls to `bodyRef.UpdateBounds();` internally?

Eideren:
Kind of embarrassing to have missed this discrepancy, I'll chalk it up to being too short on time. Thanks for that !
Added the suggested method as `SetPose`

PR #2651, opened 2025-02-23
fix: Temporarily disable the resolver UI as there is a dependency issue with Avalonia.

Kryptos-FR:
Hotfix : UI nuget resolver doesn't work because it needs Avalonia which requires... running the nuget resolver 🙃

## Types of changes

## Checklist


xen2:
For reference, here is where `.dll` can be embedded in the top level package that will do nuget resolving:
https://github.com/stride3d/stride/blob/master/sources/shared/Stride.NuGetResolver.Targets/Stride.NuGetResolver.Targets.projitems#L36

Kryptos-FR:
@xen2 I did't want to embed Avalonia because it has a lot of dependencies, and some are platform-dependent. I found it easier to resolve it.

xen2:
Yes I just saw #2653
Agree with you, I wanted to mention about embedding dll not being ideal anyway.
It's an even better way to fix it, thanks!

PR #2653, opened 2025-02-23
fix: Restore Nuget Resolver UI

Kryptos-FR:
# PR Details

Following #2651, properly solves the issue with displaying the new Nuget Resolver UI that broke `releases/4.2.0.2371`.

## Description

The new Resolver UI uses Avalonia instead of WPF.

By design (and because of its function) the resolver runs very early in the startup of the GameStudio (or of the ConnectionRouter). Stride packages(and their dependencies) might not have ever been resolved on the user machine and the purpose of the resolver is also to locate such packages so that their paths can be known and their respective assembly resolved. It is also required because the GameStudio is started from its nuget package which doesn't bring all of its dependencies with it (but rather links them as is the usual way).

However, Avalonia itself is also not resolved and the UI cannot be loaded because of that. On top of that, Avalonia has its own dependencies, some of which are native.

To add to the fun, when building and running the GameStudio locally, the issue would not be apparent because the Stride.NugetResolver.UI project was referencing Avalonia as a direct dependency and its dlls would end up in the bin folder of the local build, which .NET is perfectly able to resolve by itself.

We have a mechanism to load native libraries, but this require the `NativeLibraryHelper` which is located in `Stride.Core` which itself is also not resolved/loaded.

In short, the following was needed to make it work :
* First resolve `Stride.Core`
* Then resolve Avalonia packages (`Avalonia.Themes.Fluent` and `Avalonia.Desktop`)
* Then finally resolve `Stride.GameStudio`.
* For all of the above resolve their native dependencies.

## How to test this PR

0. Clean the bin folder in your local Stride clone as well as `%LOCALAPPDATA%\Stride\NugetDev`
1. Build the GameStudio
2. Run it once to make sure it works locally.
3. Close it.
4. Open the launcher
5. Select the dev version (`4.2.0.1`)
6. If it's already "downloaded", remove it.
7. Go to your local nuget packages cache (should be `%USERPROFILE%\.nuget\packages`)
8. Locate and remove all Avalonia packages from the cache
9. Back to the launcher, install the dev version (`4.2.0.1`)
10. Make sure no Avalonia packages appeared in the nuget cache
11. Run the dev version from the launcher
12. It should resolve Avalonia, display the resolver UI and then open the GameStudio
13. Avalonia packages should be visible in the nuget cache

## Types of changes

## Checklist

### DIFF
--- a/sources/shared/Stride.NuGetResolver/NuGetAssemblyResolver.cs
+++ b/sources/shared/Stride.NuGetResolver/NuGetAssemblyResolver.cs
@@ -78,6 +78,7 @@ public static partial class NuGetAssemblyResolver
                     const string AvaloniaPackageName = "Avalonia.Desktop";
                     const string AvaloniaVersion = "11.0.6"; // FIXME find a way to inject that version number
                     var i = 0;
+                    packagesConfigs.Insert(i++, (packagesConfigs[0].targetFramework, "Stride.Core", packagesConfigs[0].packageVersion));
                     packagesConfigs.Insert(i++, (packagesConfigs[0].targetFramework, "Avalonia.Themes.Fluent", AvaloniaVersion));
                     packagesConfigs.Insert(i++, (packagesConfigs[0].targetFramework, AvaloniaPackageName, AvaloniaVersion));
 
@@ -142,15 +143,13 @@ public static partial class NuGetAssemblyResolver
                             }
 
                             // Register the native libraries
-                            if (packageName.StartsWith("Stride."))
-                            {
-                                var nativeLibs = RestoreHelper.ListNativeLibs(result.LockFile);
-                                RegisterNativeDependencies(assemblyNameToPath, nativeLibs);
-                            }
+                            var nativeLibs = RestoreHelper.ListNativeLibs(result.LockFile);
+                            RegisterNativeDependencies(assemblyNameToPath, nativeLibs);
 
 #if STRIDE_NUGET_RESOLVER_UI
                             if (packageName == AvaloniaPackageName)
                             {
+                                LoadNativeDependencies(assemblyNameToPath, nativeLibs);
                                 avaloniaLoaded.TrySetResult();
                             }
 #endif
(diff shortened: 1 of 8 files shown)


sources/shared/Stride.NuGetResolver/NuGetAssemblyResolver.cs
```diff
@@ -1,116 +1,114 @@
...
-                        var dialogNotNeeded = new TaskCompletionSource<bool>();
-                        var dialogClosed = new TaskCompletionSource<bool>();
-
+                    var avaloniaLoaded = new TaskCompletionSource();
+                    var dialogNotNeeded = new TaskCompletionSource();
+                    var dialogClosed = new TaskCompletionSource();
+
+                    // We need to make sure Avalonia has been restored before we can display the UI.
+                    const string AvaloniaPackageName = "Avalonia.Desktop";
+                    const string AvaloniaVersion = "11.0.6"; // FIXME find a way to inject that version number
```
Kryptos-FR, line 79: The version of Avalonia is hardcoded. This needs to be fixed.

Kryptos-FR: Found a way.

sources/shared/Stride.NuGetResolver/NuGetAssemblyResolver.cs
```diff
@@ -165,180 +174,189 @@ public static void SetupNuGet(string targetFramework, string packageName, string
...
             {
-                lock (logLock)
-                {
-                    this.action = action;
-                    foreach (var log in Logs)
-                        action.Invoke(log.Level, log.Message);
-                }
+                this.action = action;
+                foreach (var (level, Message) in Logs)
+                    action.Invoke(level, Message);
```
Eideren, line 291: `level` is lower case but `Message` is upper case ;)

Eideren: Nevermind, took care of it instead

PR #2654, opened 2025-02-25
refactor: Nullable and Modernization for Core Presentation projects

Kryptos-FR:
# PR Details

Similar to #2639 and #2643.

## Description

Modernize the code, use nullable annotations, try to fix null reference warnings when possible.

Affected projects:
* `Stride.Core.Presentation`
* `Stride.Core.Quantum`
* `Stride.Core.Quantum.Tests`
* `Stride.Core.Presentation.Quantum`
* `Stride.Core.Presentation.Quantum.Tests`

I purposely left out `Stride.Core.Presentation.Wpf` as well as `Stride.Core.Presentation.Tests` as I consider those legacy since we are moving towards using Avalonia in the near future.

### Dispose pattern

I also fix the `Dispose` pattern in some places. It was previously implemented as follow:
```csharp
class BaseClass : IDisposable
{
    public virtual void Dispose()
    {
        // dispose stuff
    }
}

class DerivedClass : BaseClass
{
    public override void Dispose()
    {
        // dispose other stuff
        base.Dispose();
    }
}
```

The main issue is that if any derived class does have a finalizer (because it has to release unmanaged resources), it will be broken : finalizer are not supposed to call managed code (included from ancestor classes), as it can lead to unpredictable/undefined behavior. More details https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose and https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/dispose-pattern.

The correct implementation (according to the guidelines linked above) is:
```csharp
class BaseClass : IDisposable
{
    public void Dispose()
    {
        Dispose(true);
        GC.SuppressFinalize(this);
    }

    protected virtual void Dispose(bool disposing)
    {
        if (disposing)
        {
            // dispose stuff
        }
    }
}

class DerivedClass : BaseClass
{
    protected override void Dispose(bool disposing)
    {
        if (disposing)
        {
            // dispose other stuff
        }
        base.Dispose(disposing);
    }
}
```

## Related issues

#2155
#2156

## Types of changes

## Checklist

### DIFF
--- a/sources/core/Stride.Core.Design/Extensions/ObjectExtensions.cs
+++ b/sources/core/Stride.Core.Design/Extensions/ObjectExtensions.cs
@@ -3,7 +3,6 @@
 
 using System.Collections;
 using System.Runtime.CompilerServices;
-using Stride.Core.Annotations;
 
 namespace Stride.Core.Extensions;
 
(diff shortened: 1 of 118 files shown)


Eideren:
Thanks !

PR #2661, opened 2025-03-04
feat: Enable to build WPF projects outside windows

Jklawreszuk:
# PR Details

Small and temp feature that enables to build WPF projects that are the part of Stride solution. 
Once we migrate to Avalonia this feature will no longer be needed, just for sake of convenience.

## Types of changes

## Checklist


Kryptos-FR:
It's probably not enough because some distributions don't include the Windows workloads by default in their .NET packages. But it shouldn't hurt. I'll test once to build a few projects on Linux and approve if it's fine.

Kryptos-FR (approved):

Jklawreszuk:
Really ? I'm not sure about that since I was installing dotnet from shell script and all I had to do was setting this property

Jklawreszuk:
But yeah that's better than nothing

PR #2671, opened 2025-03-10
fix: CollisionHelper.RayIntersectsPlane to return the correct hit position

### LINKED ISSUE #2670: Ray-Plane intersection test returns wrong hit position
EDIT: This may not be an issue. `CollisionHelper.RayIntersectsPlane` appears to be usable for Planes constructed in the form `Ax+By+Cz+D=0`
Refer to #2672 to determine the final answer for this issue.

**Version**: All versions up to the latest (4.2.0.2381 as of this issue)

**Platform(s)**: All (tested on Windows)

**Describe the bug**
When the plane's `D` value is non-zero, `CollisionHelper.RayIntersectsPlane` returns the wrong hit position (if it hits).

Consider the plane, flat on XZ with Y = 1.
Have a ray sitting at Y = 2, pointing directly down.
The ray should hit at Y = 1.
However `CollisionHelper.RayIntersectsPlane` says the hit is at Y = -1.

**To Reproduce**
1. Run the following code:
```cs
var rayPosition = new Vector3(0, 2, 0);
var rayDirection = new Vector3(0, -1, 0);
var ray = new Ray(position: rayPosition, direction: rayDirection);
var plane = new Plane(point: Vector3.UnitY, normal: Vector3.UnitY);     // Alternatively new Plane(Vector3.UnitY, d: 1)
CollisionHelper.RayIntersectsPlane(in ray, in plane, out Vector3 hitPoint);
System.Diagnostics.Debug.WriteLine($"Hit: {hitPoint}");
```
Output will say:
```
// Hit: X:0 Y:-1 Z:0
```

**Expected behavior**
Expect hit to be at X: 0, Y: 1, Z: 0

**Additional context**
The problem is the following line:
https://github.com/stride3d/stride/blob/975c8a8ea55cf2ab20a64a32a261eb62d6922bce/sources/core/Stride.Core.Mathematics/CollisionHelper.cs#L502

According to the comments above this code,
https://github.com/stride3d/stride/blob/975c8a8ea55cf2ab20a64a32a261eb62d6922bce/sources/core/Stride.Core.Mathematics/CollisionHelper.cs#L490-L491
however the equation in the book uses positive D not negative.

If you don't have the book, you can also confirm with the following (random googling):
https://www.cs.princeton.edu/courses/archive/fall00/cs426/lectures/raycast/sld017.htm
(image of page copied below in case it disappears)

Image

This shows the code matches this solution (ie. negative D), where the plane is assumed `P.N + d = 0`
However, the plane equation Stride uses is `P.N = d` (and as per the Plane's documentation, `The distance of the plane along its normal from the origin.`)
Substituting the equation with `P.N = d`, it should become clear we need D to be positive.

**Workaround**
Copy the entire static method and change
`distance = (-plane.D - position) / direction;`
to
`distance = (plane.D - position) / direction;`

Basewq:
# PR Details

Fix plane-ray equation which was assuming the plane was `P.N + d = 0` whereas the plane equation in Stride is `P.N = d`
See #2670 for the full explanation of the equation.

## Related Issue

Fixes #2670

## Types of changes

## Checklist

### DIFF
--- a/sources/core/Stride.Core.Mathematics.Tests/TestCollisionHelper.cs
+++ b/sources/core/Stride.Core.Mathematics.Tests/TestCollisionHelper.cs
@@ -1,0 +1,126 @@
+// Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net)
+// Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
+using System.Collections;
+using System.Collections.Generic;
+using Xunit;
+
+namespace Stride.Core.Mathematics.Tests;
+
+public class TestCollisionHelper
+{
+    [Theory, ClassData(typeof(RayPlaneIntersectionTestData))]
+    public void TestRayPlaneIntersections(Ray ray, Plane plane, bool expectedIsHit, Vector3? expectedHitPoint)
+    {
+        bool isHit = CollisionHelper.RayIntersectsPlane(in ray, in plane, out Vector3 hitPoint);
+        Assert.Equal(expectedIsHit, isHit);
+        if (isHit)
+        {
+            Assert.Equal(expectedHitPoint, hitPoint);
+        }
+    }
+
+    private class RayPlaneIntersectionTestData : IEnumerable<object[]>
+    {
+        public IEnumerator<object[]> GetEnumerator()
+        {
+            var result = new List<object[]>();
+
+            /* Ray hits */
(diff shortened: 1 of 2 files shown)


Basewq:
Back to draft, for the time being.
Need confirmation from #2672 on which way `D` works. If positive `D` means the plane moves in the same direction as the its normal vector, this pull is fine as-is.

PR #2684, opened 2025-03-22
Vulkan concurrency and validation error fixes

johang88:
# PR Details

* Fixed concurrency issues that caused a full driver crash when texture streaming is enabled.
* Fixed a validation error casued by the default `NoSampler` that is used when converting `Texture.Load` to `texelFetch`.

## Related Issue

Found some validation errors when running in debug mode and also full driver crash when texture streaming was enabled.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Graphics/Vulkan/GraphicsDevice.Vulkan.cs
+++ b/sources/engine/Stride.Graphics/Vulkan/GraphicsDevice.Vulkan.cs
@@ -40,6 +40,7 @@ namespace Stride.Graphics
         private IntPtr nativeUploadBufferStart;
         private int nativeUploadBufferSize;
         private int nativeUploadBufferOffset;
+        private object nativeUploadBufferLock = new();
 
         private Queue<KeyValuePair<long, VkFence>> nativeFences = new Queue<KeyValuePair<long, VkFence>>();
         private long lastCompletedFence;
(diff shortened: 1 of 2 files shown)


johang88:
These changes allowed me to run the default hdr new game sample in vulkan with no crashes, which is an improvement! There are probably more issues that need fixing but these should be mergable.

sources/engine/Stride.Graphics/Vulkan/GraphicsDevice.Vulkan.cs
```diff
@@ -366,40 +371,43 @@ private unsafe void InitializePlatformDevice(GraphicsProfile[] graphicsProfiles,
...
+                resource = nativeUploadBuffer;
+                offset = nativeUploadBufferOffset;
+                nativeUploadBufferOffset += size;
             }
 
-            // Bump allocate
-            resource = nativeUploadBuffer;
-            offset = nativeUploadBufferOffset;
-            nativeUploadBufferOffset += size;
             return nativeUploadBufferStart + offset;
```
xen2, line 411: Shouldn't a copy of `nativeUploadBufferStart` be made inside the lock?

johang88: makes sense, I'll move the return inside the lock and that should take care of it.

Eideren:
Thanks !

PR #2686, opened 2025-03-23
refactor: Remove "Refactor" project

Jklawreszuk:
# PR Details

PR removes unnecessary project containing only redundant wrapper around NotImplementedException class.
Description states that it " can be easily tracked by finding usages", but the same goal can be easily achieved by using a global search in any IDE ... (CTRL + SHIFT + F)

## Types of changes

## Checklist


Kryptos-FR:
I think the original idea was to be able to find these specific not implemented features, as opposed to getting every other not implemented cases.

However, since it's an old TODO, I think it's fair to remove them. I didn't check by maybe the error message can be put into a constant (it looks like it's the same in most places) so that we can easily find again those same places.

sources/engine/Stride.Graphics/OpenGL/Buffer.OpenGL.cs
```diff
@@ -70,7 +70,7 @@ public void Recreate(IntPtr dataPointer)
             else if ((ViewFlags & BufferFlags.ShaderResource) == BufferFlags.ShaderResource && GraphicsDevice.HasTextureBuffers)
             {
 #if STRIDE_GRAPHICS_API_OPENGLES
-                Internal.Refactor.ThrowNotImplementedException();
+                throw new NotImplementedException();
```
Kryptos-FR, line 73: In such case I would use a NotSupportedException instead.

sources/engine/Stride.Graphics/OpenGL/CommandList.OpenGL.cs
```diff
@@ -332,7 +332,7 @@ public unsafe void CopyRegion(GraphicsResource source, int sourceSubresource, Re
 
             if (sourceTexture == null || destTexture == null)
             {
-                throw Internal.Refactor.NewNotImplementedException("Copy is only implemented for Texture objects.");
+                throw new NotImplementedException("Copy is only implemented for Texture objects.");
```
Kryptos-FR, line 335: Again: NotSupportedException

Jklawreszuk:
Thanks for explanation, it might be worthwhile to make a compilation of features to be implemented as a github issue 🤔

sources/engine/Stride.Graphics/OpenGL/Buffer.OpenGL.cs
```diff
@@ -168,7 +168,7 @@ protected void Init(IntPtr dataPointer)
                     if ((Flags & BufferFlags.ShaderResource) != 0)
                     {
 #if STRIDE_GRAPHICS_API_OPENGLES
-                        Internal.Refactor.ThrowNotImplementedException();
+                        throw new NotImplementedException();
```
Kryptos-FR, line 171: I meant to use NotSupportedException in all similar cases. I didn't comment every single file.

Jklawreszuk: Right, I updated branch now. Let me know if that's okay 😅

Kryptos-FR (approved):

Eideren:
Thanks !

PR #2320, opened 2024-06-10
feat: Save Asset View Options Settings (#147)

dloe:
# PR Details

Saving out asset view options (Filter tags, zoom, etc) so that new instances of the editor dont repeatively use default values. Some of the view options actually are already able to be saved between instances, added the filter tags and tile and grid zoom to those saved out.

## Related Issue

Issue: https://github.com/stride3d/stride/issues/147#issuecomment-2157012923

## Types of changes

- [ ] Docs change / refactoring / dependency upgrade
- [ ] Bug fix (non-breaking change which fixes an issue)
- [X ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [ ] My change requires a change to the documentation.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.
- [X] **I have built and run the editor to try this change out.**

PR #2691, opened 2025-03-29
fix: Prevent editor-inserted filters from persisting across views

Eideren:
# PR Details
Followup to #2320 - filters that are added by the engine, like `UIPage` when opening the picker for field of that type, persisted to new instance of the picker, meaning that new pickers for `Texture` would have a `UIPage` filter added in as well.
This change ensures that we only persist user-defined filters across instances.

## Types of changes

## Checklist


Doprez:
Did some testing here with adding and removing filters as well as opening predefined filtered windows, I dont see any issues here.

LGTM

PR #2693, opened 2025-03-29
make default x64 and fix .dd imported files

Doprez:
# PR Details

Changes the default build setting changed by https://github.com/stride3d/stride/pull/2644
Also removes a change in the DXTWrapper 

## Related Issue

None, found in master.

## Types of changes

## Checklist

### DIFF
--- a/sources/native/Stride.Native.targets
+++ b/sources/native/Stride.Native.targets
@@ -27,7 +27,7 @@
     <!--<StrideNativeOutputPath>$([MSBuild]::MakeRelative('$(OutputPath)', '$(StridePackageStridePlatformBin)\'))</StrideNativeOutputPath>-->
     <AllowedOutputExtensionsInPackageBuildOutputFolder>.so; .a; $(AllowedOutputExtensionsInPackageBuildOutputFolder)</AllowedOutputExtensionsInPackageBuildOutputFolder>
   
-    <StrideNativeWindowsArm64Enabled Condition="'$(StrideNativeWindowsArm64Enabled)' == ''">true</StrideNativeWindowsArm64Enabled>
+    <StrideNativeWindowsArm64Enabled Condition="'$(StrideNativeWindowsArm64Enabled)' == ''">false</StrideNativeWindowsArm64Enabled>
   </PropertyGroup>
 
   <Choose>
--- a/README.md
+++ b/README.md
@@ -50,10 +50,11 @@ Our [Roadmap](https://doc.stride3d.net/latest/en/contributors/roadmap.html) comm
    - **Desktop development with C++** with:
      - **Windows 11 SDK (10.0.22621.0)** or a later version (should be enabled by default)
      - **MSVC v143 - VS2022 C++ x64/x86 build tools (Latest)** (should be enabled by default)
-     - **MSVC v143 - VS2022 C++ ARM64 build tools (Latest)**
+     - **MSVC v143 - VS2022 C++ ARM64 build tools (Latest)** *(not enabled by default)*
      - **C++/CLI support for v143 build tools (Latest)** *(not enabled by default)*
    - *Optional* (to target iOS/Android): **.NET Multi-platform App UI development** and the **Android SDK setup** individual component (enabled by default). Then, in Visual Studio, go to `Tools > Android > Android SDK Manager` and install **NDK** (version 20.1+) from the `Tools` tab.
    - *Optional* (to build the VSIX package): **Visual Studio extension development**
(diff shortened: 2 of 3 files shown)


sources/native/Stride.Native.targets
```diff
@@ -27,7 +27,7 @@
     <!--<StrideNativeOutputPath>$([MSBuild]::MakeRelative('$(OutputPath)', '$(StridePackageStridePlatformBin)\'))</StrideNativeOutputPath>-->
     <AllowedOutputExtensionsInPackageBuildOutputFolder>.so; .a; $(AllowedOutputExtensionsInPackageBuildOutputFolder)</AllowedOutputExtensionsInPackageBuildOutputFolder>
   
-    <StrideNativeWindowsArm64Enabled Condition="'$(StrideNativeWindowsArm64Enabled)' == ''">true</StrideNativeWindowsArm64Enabled>
+    <StrideNativeWindowsArm64Enabled Condition="'$(StrideNativeWindowsArm64Enabled)' == ''">false</StrideNativeWindowsArm64Enabled>
```
Jklawreszuk, line 30: Isn't that change forces to set to True when releasing nuget packages?

Doprez: What do you mean? Do you mean, does this get affected by the team city builds?

Jklawreszuk: I .. think so, since by default is true then it's generating arm64 dll files without any additional paramateres. I'm just afraid that's gonna require some additional modification.

Jklawreszuk: My proposal would be something like this
```suggestion
<StrideNativeWindowsArm64Enabled Condition="'$(StrideNativeWindowsArm64Enabled)' == '' AND '$(Configuration)' == 'Release'">true</StrideNativeWindowsArm64Enabled>
<StrideNativeWindowsArm64Enabled Condition="'$(StrideNativeWindowsArm64Enabled)' == '' AND '$(Configuration)' != 'Release'">false</StrideNativeWindowsArm64Enabled>
```

Jklawreszuk: Would be great if someone else would comment If I'm correct

azeno:
The changes in DXTWrapper were made on purpose. Without them the calls would not work on arm64. Are you saying they don't work now on x64?

sources/native/Stride.Native.targets
azeno: I share your concerns :) But I also understand the intent to make them an opt in feature. Important is that they geht included on TC builds. What about we enable that flag in the build invocation?

Jklawreszuk: If it's easy to configure in TC then why, not otherwise I'd go with my change (i.e if value is NaN then set true, then if is not relase, dont geneerate arm files)

azeno:
Interesting. Then I'll have to read up on it again and find a different solution.

sources/native/Stride.Native.targets
azeno: It could be irritating that the debug build works and once you switch to release it breaks. I'd prefer passing the parameter explicitly to TC build.

Jklawreszuk: makes sense

Doprez: Im not very familiar with conditionals in targets files. If we make the change through TC then would my current change be ok currently or would it cause a problem?

Doprez:
I might be able to use an if def in the class with a const to change the value depending on architecture.

Doprez:
@azeno would you be able to test out this branch when you have some time to make sure it still works for windows arm?

I just added a check for the value you added in the targets file `StrideNativeWindowsArm64Enabled` that if I understand correctly should hopefully make sense.

sources/native/Stride.Native.targets
azeno: It should be ok. I think the file we should modify is `build/Stride.build` which contains a property `BuildProperties`, adding `StrideNativeWindowsArm64Enabled=true` should enable it for command line / TC builds.

azeno:
@Doprez Sure can do, but I was just wondering if we should change the native code so that it uses Unicode like other methods in that file do. Those ifdefs are such a pain.

Doprez:
100% hate what I did here lol I am not a fan of them either. I just wasn't sure on the fix complexity with the native code so this was a quick work around fix.

azeno:
Let me check the native part ..

azeno:
@Doprez Looks like simply re-compiling `sources\tools\Stride.TextureConverter.Wrappers` for win-x64 did the trick. I'll make a PR.

azeno:
@Doprez See #2694 - undo your changes, and make sure to do a clean build so you don't end up with the old dlls.

Doprez:
Found an alternative to detect CPU architecture instead that I think should work better assuming I did it right.

Will make a new PR that should be simpler.

PR #2694, opened 2025-03-30
Fixes output directories of DxtWrapper and re-builds with MSVC v143

azeno:
The dlls were built using `sources\tools\Stride.TextureConverter.Wrappers\build-deps.bat` inside of developer command prompt of VS2022

# PR Details

After #2644 dds files did no longer load on `win-x64`. That PR changed the string marshalling from `Auto` to `Ansi`, which is correct if we look at the native call site (https://github.com/stride3d/stride/blob/1925a7bdb9e5627428fab43de36a911f69fcdf7c/sources/tools/Stride.TextureConverter.Wrappers/DxtWrapper/dxt_wrapper.h#L40) and added DxtWrapper.dll for win-arm64, but it didn't touch the already checked-in versions for win-x64 and win-x86. 

This PR fixes that, by re-building those native assets with current compiler (MSVC v143) as well.

## Related Issue

#2693

## Types of changes

## Checklist


Doprez:
Thank you Azeno! I can confirm with your branch I am able to now build on win-64 without the texture issues:
{3949D98A-AE95-4681-A608-EEE7A4F47EF3}

I did have to change the Stride.Native.Targets to make the arm setting false but I will redo my PR and try to follow your suggestion of the Stride.Build config changes.

Eideren:
Thanks !

Issue #2487, opened 2024-10-14
[Linux] SDL: MakeCurrent returns sometimes EGL_BAD_ACCESS error on Wayland


Jklawreszuk:
**Release Type**: Official Release

**Version**: 4.1.0.1805+

**Platform(s)**: Linux, Fedora (OpenGL/Vulkan)

**Describe the bug**

It is not possible to play any 3D game during runtime because of error : `Error while compiling GLSL shader. []` using Wayland driver

**To Reproduce**
1. Run GameStudio
2. Create new project based on SpaceEscape template
3. Add Linux platform
4. Go to Linux platform project and build It to test executable on Linux
5. `dotnet run`

**Expected behavior**
The application should work flawlessly

**Log and callstacks**
[stacktrace.txt](https://github.com/user-attachments/files/17360344/stacktrace.txt)

**Additional context**
Described bug exists only for Wayland. If you switch to X11 by setting following variable
```
export SDL_VIDEODRIVER=x11
```
then problem does not occur.

Jklawreszuk:
I found another way to reproduce the error:
1. create a project from an empty template
2. remove skybox and leave only two elements for Main Scene - camera and light.
3. import any custom model (for testing purposes I used royality free Train kit from [Kenney](https://www.kenney.nl/assets/train-kit)).

Same result

Jklawreszuk:
I got the same result when I added only a procedural cube without material to the scene

Jklawreszuk:
Another use case:

> 1 . create a project from an empty template
> 2. remove skybox and leave only two elements for Main Scene - camera and light.
3.  this time import only spritesheet (download any image from the internet or create by yourself)

Works

Jklawreszuk:
Conclusion : Both Procedural and Imported models may share a common code that is flawed

Jklawreszuk:
Okay...That's new.

Another use case
1 . create a project from an empty template, but this time **don't check the HDR option** and use **LDR* instead - also pick lowest API level - Level 9
2. remove skybox and leave only two elements for Main Scene - camera and light.
3. import any model - procedural cube for example

Works!

Doprez:
Those steps sound like a similar issue to the mobile issue steps https://github.com/stride3d/stride/issues/2485

You might be able to just uncheck `Generate mipmaps` if Im right? I dont know how much that info actually helps you but if its true it may be useful.

Jklawreszuk:
[...]

PR #2698, opened 2025-04-04
fix: Wayland crash workaround

Jklawreszuk:
# PR Details

This PR fixes an issue for Wayland based DE like Gnome that causes a crash. Issue was fixed by setting driver to X11. 

## Related Issue

https://github.com/stride3d/stride/issues/2487

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Graphics/SDL/Window.cs
+++ b/sources/engine/Stride.Graphics/SDL/Window.cs
@@ -24,6 +24,11 @@ namespace Stride.Graphics.SDL
         {
             SDL = Silk.NET.SDL.Sdl.GetApi();
 
+            // jklawreszuk: Workaround for wayland (see #2487 for more details)  
+            // Perhabs there is more elegant solution?
+            if (OperatingSystem.IsLinux())
+                SDL.SetHint("SDL_VIDEODRIVER", "x11");
+
             SDL.Init(Sdl.InitEverything);
 
             // Pass first mouse event when user clicked on window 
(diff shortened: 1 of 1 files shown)


Jklawreszuk:
I've read on many forums that setting SDL_VIDEODRIVER=wayland is causing a lot of problems but yet I consider it a workaround anyway so I added FIXME tag

Jklawreszuk:
SDL3 uses wayland as a driver by default we can go back to that later when migrating

Eideren:
Is this an actual fix or is it a workaround, specifically in the context of #2487 ? Meaning, should we added a note to that issue that there is a workaround in place for it which would need to be replaced, or should we instead create a new issue specifically to replace this workaround and linking to #2487 as added context ?

Jklawreszuk:
Huh...good question..After some more research I found that SDL_EGL_MakeCurrent when using wayland  does not "cover multicontext scenario" unlike x11. I have now added a link to the issue to the SDL repository

Jklawreszuk:
So guess we have to wait until someone fixes it or revisit our rendering system to keep eye on making only one context specifically for wayland... I think?

Jklawreszuk:
According to one of the SDL contributors, it seems that SDL_EG_MakeCurrent is limited to a single context for wayland. See https://github.com/libsdl-org/SDL/issues/9072 for more details.
I propose as temp workaroud to set X11 as a drvier until someone fixes it in SDL repo or we should adapt our rendering system. 
Perhabs someone smarter than me could figure out

Jklawreszuk:
Anyway I added that note to the issue and I think I could close as "won't fix"

Eideren:
Thanks !

PR #2702, opened 2025-04-06
fix: Handle initial collection items appropriately, YAML

Eideren:
# PR Details
The yaml serializer adds its serialized items on top of those that may already be contained in the list. The following snippet would continuously add one more `0.5` to the collection's serialization data on every save.
```cs
public List<float> Coll { get; } = new List<float>(){ 0.5f };
```
This PR ensures that, if yaml has data about items for a collection, the pre-existing ones are cleared before adding those that are serialized.

Do note that this was already handled on the `AssemblyProcessor` side, see `ListSerializer.PreSerialize()` and other collection serializers.

Pinging @IXLLEGACYIXL to let him know of this rule given the work he is doing for the next yaml serializer.

## Related Issue
None logged afaict

## Types of changes

## Checklist


IXLLEGACYIXL:
this would only work for equatables and structs?

Eideren:
I don't know what you mean, this just clears the collection before adding to it ?

IXLLEGACYIXL:
ohh now i get it, yep will implement it

Issue #2703, opened 2025-04-06
Bepu physics out of bounds exception


Doprez:
**Release Type**: GitHub

**Version**: master

**Platform(s)**: Windows

**Describe the bug**
I'm currently re-adding the demo scenes as Stride project templates and the Car scene causes a crash.

**To Reproduce**
1. load the new template
2. run the game
3. select the Car demo
4. crash 

**Log and callstacks**
```
[Game]: Error: Unexpected exception. System.IndexOutOfRangeException: Index was outside the bounds of the array.
   at System.Collections.Generic.HashSet`1.AddIfNotPresent(T value, Int32& location)
   at System.Collections.Generic.HashSet`1.Add(T item)
   at Stride.BepuPhysics.Definitions.Contacts.ContactEventsManager.TrackActivePairs(Single dt, IThreadDispatcher threadDispatcher) in D:\dev\GitControlledProjects\stride\sources\engine\Stride.BepuPhysics\Stride.BepuPhysics\Definitions\Contacts\ContactEventsManager.cs:line 386
   at BepuPhysics.DefaultTimestepper.Timestep(Simulation simulation, Single dt, IThreadDispatcher threadDispatcher)
   at BepuPhysics.Simulation.Timestep(Single dt, IThreadDispatcher threadDispatcher)
   at Stride.BepuPhysics.BepuSimulation.Update(TimeSpan elapsed) in D:\dev\GitControlledProjects\stride\sources\engine\Stride.BepuPhysics\Stride.BepuPhysics\BepuSimulation.cs:line 729
   at Stride.BepuPhysics.Systems.PhysicsGameSystem.Update(GameTime time) in D:\dev\GitControlledProjects\stride\sources\engine\Stride.BepuPhysics\Stride.BepuPhysics\Systems\PhysicsGameSystem.cs:line 35
   at Stride.Games.GameSystemCollection.Update(GameTime gameTime)
   at Stride.Games.GameBase.Update(GameTime gameTime)
   at Stride.Games.GameBase.RawTick(TimeSpan elapsedTimePerUpdate, Int32 updateCount, Single drawInterpolationFactor, Boolean drawFrame)
   at Stride.Games.GameBase.RawTickProducer()
System.IndexOutOfRangeException: Index was outside the bounds of the array.
   at System.Collections.Generic.HashSet`1.AddIfNotPresent(T value, Int32& location)
   at System.Collections.Generic.HashSet`1.Add(T item)
[...]

PR #2704, opened 2025-04-06
feat: Create Bepu demo sample templates.

Doprez:
# PR Details

This PR re-configures the old demos into a project template for users to look at how some of the features are meant to work in a playable environment.

I have also updated Bepu to the latest version which seems to fix the convex creation issue.

## Related Issue

There have been a few mentions in discord about the lack of docs/examples for the Bepu library.
https://github.com/stride3d/stride/issues/2703

## Types of changes

## Checklist


Doprez:
Couple more things to fix:
- Controls per scene dont work (Should be simple to fix I just made the main camera centralized which is the issue)
- Car scene crashes on load due to https://github.com/stride3d/stride/issues/2703

Doprez:
This is pretty much done but the Car scene still has the crash issue. It was happening before the changes here though so it is not an issue created in this project. (Unless there was something done incorrectly in the Car scene specifially of course)

Eideren:
Doesn't look like this is part of the samples that the project window can create, I might be missing something though ... ?

Doprez:
Sorry, you were not missing anything. I forgot to add the reference to the template file for the launcher.

Although I noticed with the template its improperly creating a separate Windows project so Im just looking into that now.

Doprez:
Ok so it turns out you **must** add the references to the non main scenes as root assets in the sdpkg for the template to work properly. 

One final issue is the resource files not being added for some reason. Almost there though.

Doprez:
ok, **now** this should be mostly working again apart form the car demo due to the previous issue. Sorry again Eideren, I should have made sure to properly test the template in the launcher not just the project.

Eideren:
All good, getting a 
```
[EXEC(0,0)]: Error: [ModelAssetCompiler] Unable to find the source file 'D:/Documents/Stride Projects/BepuSample/Resources/Models/Hex.obj' for Asset [[ModelAsset] Models/Hex]
```
now when building the project, same with the dungeon and hex models

Doprez:
hmmm 
image

Also tested in my Document folder just in case:
image

Opened the scene with the models and it looks fine:
image

It also seems to load at runtime as well. Can you confirm if you see the assets in the projects resource folder? `{ProjectName}\Resources\Models`

Doprez:
good catch, for some reason the gitignores had .obj files? Not usre if there is some greater risk with removing those lines but it only added the ones I wanted in this case.

Eideren:
The samples are a bit too rough right now I think. The car scene, for example, captures the mouse on start, if the user presses tab the mouse is unlocked but the scene selection menu disappears at the same time so we can't easily go back to another scene.

I think we should remove the weird platform from the car scene as well given how much of a performance issue it is at the start but even afterward. Doesn't make for a good best practice example.

And we'll also have to wait for #2722 to be merged before merging this one.

Doprez:
Sounds good, I will clean it up with a proper state machine and maybe just merge the FPS and Car scene into one in that case, it may be more interesting for users anyway.

And thank you for figuring out that crash!

Doprez:
I cleaned up the demo and removed the character and car examples as I have a plan to make a separate Bepu template dedicated to being more of a simple game example.

I will make a new PR for a template that will have an first person controller that can enter and exit a car and depepdning on my time I will try to add the same shooter functionality from the FPS demo that exists with bullet.

Doprez:
Oh I also added a couple more constraint examples (ball socket and distance).

.gitignore
```diff
@@ -11,7 +11,6 @@ _ReSharper.*
 *.Cache
 *.tlog
 *.log
```
Kryptos-FR, line 13: This shouldn't be removed. the `.obj` extension is found in C/C++ projects for instance.

You don't need to touch the .gitignore file even if you want to add a file which extension is ignored. It can be done from the command line (using the `-f` switch).

samples/.gitignore
```diff
@@ -11,7 +11,6 @@ _ReSharper.*
 *.cache
 *.tlog
 *.log
-*.obj
```
Kryptos-FR, line 14: Same remark. As we want this `.gitignore` file to be a sample by itself, we shouldn't touch it.

Kryptos-FR (changes requested):

.gitignore
Doprez: Should be back. Its good to know about the C++ stuff I figured it was just for projects without LFS set up.

Eideren:
Thanks !

PR #2712, opened 2025-04-13
feat: Implement Help menu

Jklawreszuk:
# PR Details

PR implements Help menu from WPF Gamestudio
I also used RelayCommandAttribute from CommunityToolkit.MVVM to reduce Boilerplate code

## Types of changes

## Checklist

### DIFF
--- a/sources/Directory.Packages.props
+++ b/sources/Directory.Packages.props
@@ -42,7 +42,6 @@
   </ItemGroup>
   <!-- Other dependencies -->
   <ItemGroup>
-    <PackageVersion Include="CommunityToolkit.Mvvm" Version="8.4.0" />
     <PackageVersion Include="JetBrains.Rider.PathLocator" Version="1.0.12" />
     <PackageVersion Include="Microsoft.AspNet.WebApi.Client" Version="5.2.7" />
     <PackageVersion Include="Microsoft.Build" Version="17.0.0" />
--- a/sources/Directory.Packages.props
+++ b/sources/Directory.Packages.props
@@ -42,6 +42,7 @@
   </ItemGroup>
   <!-- Other dependencies -->
   <ItemGroup>
+    <PackageVersion Include="CommunityToolkit.Mvvm" Version="8.4.0" />
     <PackageVersion Include="JetBrains.Rider.PathLocator" Version="1.0.12" />
     <PackageVersion Include="Microsoft.AspNet.WebApi.Client" Version="5.2.7" />
     <PackageVersion Include="Microsoft.Build" Version="17.0.0" />
--- a/sources/editor/Stride.GameStudio.Avalonia/Helpers/StrideGameStudio.cs
+++ b/sources/editor/Stride.GameStudio.Avalonia/Helpers/StrideGameStudio.cs
@@ -3,12 +3,12 @@
 
 namespace Stride.GameStudio.Avalonia.Helpers;
 
-public class StrideGameStudio
+public static class StrideGameStudio
 {
     public static string EditorVersion => StrideVersion.NuGetVersion;
     public static string EditorVersionMajor => new Version(StrideVersion.PublicVersion).ToString(2);
-    public static string AnswersUrl => "https://discord.gg/f6aerfE";
+    public static string AnswersUrl => "https://github.com/stride3d/stride/discussions";
     public static string DocumentationUrl => $"https://doc.stride3d.net/{EditorVersionMajor}/";
-    public static string ForumsUrl => "https://github.com/stride3d/stride/discussions";
+    public static string ForumsUrl => "https://discord.gg/f6aerfE";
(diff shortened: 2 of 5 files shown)


Kryptos-FR:
We don't use the community toolkit. Look at the existing code we already have all that's needed for commands.

sources/Directory.Packages.props
```diff
@@ -42,6 +42,7 @@
   </ItemGroup>
   <!-- Other dependencies -->
   <ItemGroup>
+    <PackageVersion Include="CommunityToolkit.Mvvm" Version="8.4.0" />
```
Kryptos-FR, line 45: Not needed.

sources/editor/Stride.GameStudio.Avalonia/Helpers/StrideGameStudio.cs
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net)
+// Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
+
+namespace Stride.GameStudio.Avalonia.Helpers;
+
+public class StrideGameStudio
+{
+    public static string EditorVersion => StrideVersion.NuGetVersion;
+    public static string EditorVersionMajor => new Version(StrideVersion.PublicVersion).ToString(2);
+    public static string AnswersUrl => "https://gamedev.stackexchange.com/tags/stride"; // #706
```
Kryptos-FR, line 10: Is it active? Do people from our community moderate it?

If not, I'd rather we link to Discord.

sources/editor/Stride.GameStudio.Avalonia/ViewModels/MainViewModel.cs
```diff
@@ -136,6 +141,21 @@ private Task OnOpen(UFile? initialPath)
         return OpenSession(initialPath);
     }
 
+    [RelayCommand]
```
Kryptos-FR, line 144: Use ICommandBase which already implements the command pattern with additional features.

Kryptos-FR (changes requested):

sources/editor/Stride.GameStudio.Avalonia/Helpers/StrideGameStudio.cs
Jklawreszuk: I'm not sure, but at first look there is not much traffic, I've added discord link 👍

sources/editor/Stride.GameStudio.Avalonia/ViewModels/MainViewModel.cs
Jklawreszuk: Yes, but I though that It might be better to reduce redundant code by using code generators like community toolkit have

Jklawreszuk: Anyway I've applied the changes

sources/editor/Stride.GameStudio.Avalonia/Views/MainView.axaml
```diff
@@ -6,6 +6,7 @@
              xmlns:ctrl="using:Stride.Core.Assets.Editor.Avalonia.Controls"
              xmlns:vm="using:Stride.GameStudio.Avalonia.ViewModels"
              xmlns:vw="using:Stride.GameStudio.Avalonia.Views"
+             xmlns:gh="clr-namespace:Stride.GameStudio.Avalonia.Helpers"
```
Kryptos-FR, line 9: nit: prefer the `using` syntax. `clr-namespace` is the old WPF way.

Kryptos-FR (commented):
Just one final remark. Otherwise LGTM.

sources/editor/Stride.GameStudio.Avalonia/ViewModels/MainViewModel.cs
Kryptos-FR: Once we have finished the rewrite, maybe we can look at it. But for now we have code that is shared or duplicated between WPF and Avalonia, and I prefer to not have two diferent implementations at the same time.

sources/editor/Stride.GameStudio.Avalonia/Views/MainView.axaml
Jklawreszuk: oh, Didn't know that ! I made fix now

sources/editor/Stride.GameStudio.Avalonia/ViewModels/MainViewModel.cs
Jklawreszuk: Right that's not urgent, so we can discuss this later

Kryptos-FR (approved):

Issue #2716, opened 2025-04-18
Overriden collection values in prefab components are lost upon reloading an assembly defining that component


Eideren:
**Release Type**: Git

**Version**: Latest

**Platform(s)**: Windows editor

**Describe the bug**
Let's say you have a collection defined on one of your component:
```cs
public class MyComponent : StartupScript
{
    public List<string> MyStrings { get; set; } = new();
}
```
You then 
1. Put that component in a prefab
2. Add that prefab to a scene
3. And add in a string to the `MyStrings` collection on that prefab in the scene. 

The yaml for the component in that scene would look something like so:
```yaml
MyStrings:
    0a0000000a0000000a0000000a000000*: MyDerivedString
```
Notice the override marker (`*`) at the end of the Id for that item.
Now, make any change to your source and reload the editor, the next time you save, the collection would be empty in the scene;
```yaml
MyStrings: { }
```

**To Reproduce**
See above.

**Expected behavior**
Reloads without loss of data.

**Additional context**
I'm currently working on this one.

PR #2717, opened 2025-04-18
fix: Optimize assembly reloading

Eideren:
# PR Details
Replaced the `ReloadingVisitor` - we already have the paths to the items we need to reload, no need to go through every leaf of the tree of the asset.

https://github.com/stride3d/stride/blob/8d3547a7de1d3f3fdbd796a998915c461bbaa1e9/sources/editor/Stride.Assets.Presentation/AssemblyReloading/GameStudioAssemblyReloader.cs#L300-L305
Calls `ProcessObject` twice for the same object/path combination, once right in this scope, another one through `VisitObject` down the base call.

Here I'm not 100% certain:
https://github.com/stride3d/stride/blob/8d3547a7de1d3f3fdbd796a998915c461bbaa1e9/sources/editor/Stride.Assets.Presentation/AssetEditors/GameEditor/ViewModels/GameEditorChangePropagator.cs#L170-L178
The use of `CanBeAssignedToContentTypes()` with `GetInnerCollectionType()` is to ignore `List<MyCustomAsset>` and `MyCustomAsset`, then it would make sense that both `int` and `List<int>` not just `int` should pass the `IsPrimitiveType` branch.

## Related Issue
Kind of a fix for #2716 - the previous logic processed the objects twice, see the blurb about `ProcessObject` above. The first pass consumes the override tag when reading the yaml events because of the `key.Value = keyName;` there
https://github.com/stride3d/stride/blob/8d3547a7de1d3f3fdbd796a998915c461bbaa1e9/sources/assets/Stride.Core.Assets/Yaml/AssetObjectSerializerBackend.cs#L153-L159
It would remove the override tag, so the next read would be without the override metadata, which in term gets the collection item removed as it is not part of its base asset.

## Types of changes

## Checklist

### DIFF
--- a/sources/editor/Stride.Assets.Presentation/AssemblyReloading/GameStudioAssemblyReloader.cs
+++ b/sources/editor/Stride.Assets.Presentation/AssemblyReloading/GameStudioAssemblyReloader.cs
@@ -58,7 +58,7 @@ namespace Stride.Assets.Presentation.AssemblyReloading
                     throw;
                 }
 
-                var reloadOperation = new ReloadAssembliesOperation(assemblyContainer, modifiedAssemblies, Enumerable.Empty<IDirtiable>());
+                var reloadOperation = new ReloadAssembliesOperation(assemblyContainer, modifiedAssemblies, []);
                 session.UndoRedoService.SetName(reloadOperation, "Reload assemblies");
 
                 // Reload assemblies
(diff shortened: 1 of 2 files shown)


xen2:
I looked at it and it seems OK from what I remember.
I like the simplifications, thanks for investigating!

However I am not sure about some of the more advanced situations (I am not that familiar with the code, and it's been a while).

Maybe we can merge this in if it works well enough for you (which seem to be a complex enough use case), and see if anybody else run into issues and/or some other yet-unknown issue pop up later.

xen2 (approved):

Kryptos-FR:
Do we cover these scenarios (or similar) in test already? Might be worth to add a few.

Otherwise, like @xen2 it looks ok to me.

Eideren:
We don't have any tests covering gamestudio reloads, which is likely why it's so broken, but it's a huge amount of work, so not really feasible at the moment. Thanks for the review.

PR #2718, opened 2025-04-18
fix: yaml, rollback trimming of the override tag when parsing collection index to ensure inputs can be re-used

Eideren:
# PR Details
The parsing events, which are effectively tokens that the deserializer reads to create the final object, are mutated when deserializing. If you were to re-use those parsing events a second time, it would generate the same object, but the metadata would be partially lost.

A reader-like object should not be writing to the stream it is reading from.

The spot where it is called twice with the same parsing events is located here:
https://github.com/stride3d/stride/blob/8d3547a7de1d3f3fdbd796a998915c461bbaa1e9/sources/editor/Stride.Assets.Presentation/AssemblyReloading/GameStudioAssemblyReloader.cs#L423-L466

And it is because
https://github.com/stride3d/stride/blob/8d3547a7de1d3f3fdbd796a998915c461bbaa1e9/sources/editor/Stride.Assets.Presentation/AssemblyReloading/GameStudioAssemblyReloader.cs#L300-L305
Calls `ProcessObject` twice for the same object/path combination, once right in this scope, another one through `VisitObject` down the base call.
I have another PR covering that scope.

## Related Issue
Fix: #2716 
The previous logic processed the objects twice, see the blurb about `ProcessObject` above. The first pass consumes the override tag when reading the yaml events because of the `key.Value = keyName;` there
https://github.com/stride3d/stride/blob/8d3547a7de1d3f3fdbd796a998915c461bbaa1e9/sources/assets/Stride.Core.Assets/Yaml/AssetObjectSerializerBackend.cs#L153-L159
It would remove the override tag, so the next read would be without the override metadata, which in term gets the collection item removed as it is not part of its base asset.

## Types of changes

## Checklist

### DIFF
--- a/sources/assets/Stride.Core.Assets.Tests/TestAssetInheritance.cs
+++ b/sources/assets/Stride.Core.Assets.Tests/TestAssetInheritance.cs
@@ -1,10 +1,28 @@
 // Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net) and Silicon Studio Corp. (https://www.siliconstudio.co.jp)
 // Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
 
+using Stride.Core.Assets.Serializers;
+using Stride.Core.Reflection;
+using Stride.Core.Yaml;
+using Stride.Core.Yaml.Events;
+
 namespace Stride.Core.Assets.Tests;
 
 public class TestAssetInheritance
 {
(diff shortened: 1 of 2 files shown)


Kryptos-FR:
@Eideren Can you provide one or two YAML samples (e.g. a few stride assets) with the issue highlighted, and/or a screenshot if the issue is only "visible" in the editor?

Kryptos-FR (approved):
LGTM. Thanks for the test case.

PR #2719, opened 2025-04-18
fix: internal exception being thrown when selecting a EntityComponent type in interface properties + drag and drop support

Eideren:
# PR Details
The exception thrown is because the handler always tries to create a new instance of whichever type is selected while the editor expects `EntityComponent` to live on an entity in a scene.

Drag and drop support for interface properties specifically, other property types already works.
Here's how it looks like *without* this PR:

https://github.com/user-attachments/assets/e04b40b8-4c6d-48fc-b33d-3b6db4b6dc1a

## Related Issue
None reported afaict.

## Types of changes

## Checklist

### DIFF
--- a/sources/editor/Stride.Core.Assets.Editor/Quantum/NodePresenters/Updaters/AbstractNodeEntryNodeUpdater.cs
+++ b/sources/editor/Stride.Core.Assets.Editor/Quantum/NodePresenters/Updaters/AbstractNodeEntryNodeUpdater.cs
@@ -19,15 +19,34 @@ namespace Stride.Core.Assets.Editor.Quantum.NodePresenters.Updaters
         {
             var type = node.Descriptor.GetInnerCollectionType();
 
-            IEnumerable<AbstractNodeEntry> abstractNodeMatchingEntries = AbstractNodeType.GetInheritedInstantiableTypes(type);
+            var abstractNodeMatchingEntries = AbstractNodeType.GetInheritedInstantiableTypes(type);
+            IEnumerable<AbstractNodeEntry> abstractNodeMatchingEntries2 = [];
+            foreach (var nodeType in abstractNodeMatchingEntries)
+            {
+                AbstractNodeEntry nodeEntry = IsEntityComponent(nodeType.Type) ? new AbstractNodeValue(null, nodeType.Type.Name, 0) : nodeType;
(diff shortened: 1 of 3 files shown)


sources/editor/Stride.Core.Assets.Editor/Quantum/NodePresenters/Updaters/AbstractNodeEntryNodeUpdater.cs
```diff
@@ -19,15 +19,34 @@ public static IEnumerable<AbstractNodeEntry> FillDefaultAbstractNodeEntry(IAsset
...
-                    abstractNodeMatchingEntries = AbstractNodeValue.Null.Yield().Concat(abstractNodeMatchingEntries);
+                    abstractNodeMatchingEntries2 = AbstractNodeValue.Null.Yield().Concat(abstractNodeMatchingEntries2);
+            }
+            return abstractNodeMatchingEntries2;
+
+            static bool IsEntityComponent(Type type)
+            {
+                for (var t = type; t != null; t = t.BaseType)
+                {
+                    if (t.Name == "EntityComponent" && t.FullName == "Stride.Engine.EntityComponent")
```
Kryptos-FR, line 42: Why comparing the name and not `typeof(EntityComponent)`?

Eideren: Right, forgot to mention that, I would have to add a reference to the engine project, not let me know if that's alright

Kryptos-FR:
Could you add a short GIF illustrating the issue with the drag and drop operation? It will help later if we encounter a similar issue.

Eideren:
Added a video to the main post

sources/editor/Stride.Core.Assets.Editor/Quantum/NodePresenters/Updaters/AbstractNodeEntryNodeUpdater.cs
Eideren: Yep, sounds good, I defaulted to that implementation for this very reason.

Kryptos-FR (approved):

Eideren:
Thanks for the review @Kryptos-FR

PR #2720, opened 2025-04-18
WIP: feat: Implement Project selection/creation

Jklawreszuk:
# PR Details

WIP, This PR represents task https://github.com/orgs/stride3d/projects/6/views/1?pane=issue&itemId=106083217 which I would like to implement 

## Types of changes

## Checklist

### DIFF
--- a/build/Stride.sln
+++ b/build/Stride.sln
@@ -374,6 +374,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Avalonia", "Avalonia", "{C4
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Plugins", "Plugins", "{BFF9AA54-F9DD-42BB-8163-6C03E7F51623}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Stride.Core.Assets.Editor.Avalonia", "..\sources\editor\Stride.Core.Assets.Editor.Avalonia\Stride.Core.Assets.Editor.Avalonia.csproj", "{67142D11-FA16-47E3-9894-6EE071F4F3B2}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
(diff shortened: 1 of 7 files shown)


sources/editor/Stride.Core.Assets.Editor.Avalonia/NewOrOpenSessionTemplateCollectionViewModel.cs
```diff
@@ -0,0 +1,21 @@
+using Stride.Core.Presentation.Commands;
```
Kryptos-FR, line 1: Note: view models should be as much as possible outside Avalonia projects. The idea being that they could be reused by another UI library if it were to change in the future (or if we decide to keep WPF around and share mist of the view models between the two).

It implies that view models cannot use or reference any UI-specific types. Even without reuse, it helps keeping the view and model logic separated.

sources/editor/Stride.GameStudio.Avalonia/App.axaml.cs
```diff
@@ -31,7 +32,7 @@ public override void OnFrameworkInitializationCompleted()
 
         if (ApplicationLifetime is IClassicDesktopStyleApplicationLifetime desktop)
         {
-            desktop.MainWindow = new MainWindow();
+            desktop.MainWindow = new ProjectSelectionWindow();
```
Kryptos-FR, line 35: I'm not quite decided about that one. I know the WPF version does start with this window, but I don't like it. I'd prefer we change the logic and start with the main window. We could force open the new/select window after the main one is created if needed.

I made sure that a new session can be opened after an already loaded one. So the trick to basically restart everything for a new project isn't needed anymore. At least for now.

build/Stride.sln
```diff
@@ -374,6 +374,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Avalonia", "Avalonia", "{C4
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Plugins", "Plugins", "{BFF9AA54-F9DD-42BB-8163-6C03E7F51623}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Stride.Core.Assets.Editor.Avalonia", "..\sources\editor\Stride.Core.Assets.Editor.Avalonia\Stride.Core.Assets.Editor.Avalonia.csproj", "{67142D11-FA16-47E3-9894-6EE071F4F3B2}"
```
Kryptos-FR, line 377: I also added this project in my `xplat/property-grid` branch. If possible can you cherry-pick or copy this project creation so we avoid conflicts (esp. on the .sln).

I'll open a draft PR also for that branch so it's easier to find.

sources/editor/Stride.Core.Assets.Editor.Avalonia/NewOrOpenSessionTemplateCollectionViewModel.cs
Jklawreszuk: I agree, I moved VMs here temperarily for just convenience 😅

Issue #2730, opened 2025-04-24
UIPage in scene and prefab throws when instantiated


Eideren:
**Release Type**: Github

**Version**: Latest as of writing

**Platform(s)**: Windows running game

**Describe the bug**
References to UIPages in Prefabs and in scene components point to the same object instance by default.

So, when a UIPage is in the scene, its `UIElement.DependencyProperties` will be filled with `ContentArrangeMatrixPropertyKey` and `PanelArrangeMatrixPropertyKey` properties, meaning the prefab's page is also filled in since it's the same object.

Those two properties do not have any serializer bound to their type (`PropertyKey<Matrix>`), so when instantiating a prefab containing that UIPage, the engine will throw when it goes through those types and fails to retrieve the serializer.

**To Reproduce**
See explanation above or download this project with everything already setup, do note that it is running on the oldest available version of stride.
[Repro for #2730.zip](https://github.com/user-attachments/files/19922389/Repro.for.2730.zip)

**Additional context**
Taking care of this one right now.

Kryptos-FR:
I feel like the issue is elsewhere (see my comment on the PR) and 1 likely regression. Can we try to figure out since when this issue occurred?

Eideren:
I remember having this issue years ago when I was cloning UIElements for a UI that was already in a scene instead of using a UILibrary. I'll check earlier versions of stride, but UI logic hasn't changed at all in all those years, and serialization changes in the assembly processor has only changed in how we treat access modifiers, not the types themselves, so it's very unlikely.

Eideren:
@Kryptos-FR Can be reproduced under `4.1.0.1728` which dates back to july of 2022, all anterior versions to this one fail as the `SharpDX` dependencies are no longer available. I added a very straightforward repro to the main post as well, the repro just goes through the steps I've explained in a minimal fashion.

PR #2731, opened 2025-04-24
fix: Generate serializers for Non-Private PropertyKey<T>

Eideren:
# PR Details
`PropertyKey` is the abstract class for a bunch of generic types (e.g.: `PropertyKey<T>`) and as such, has a unique serialization method; the assembly processor generates serializers only for concrete implementation of `PropertyKey`, e.g.: `PropertyKey<float>` and this only if there is a static public field of that type that exists in the source traversed.

https://github.com/stride3d/stride/blob/dd6adf2b07d3df0ab6106cf34f15b94c1fe99a01/sources/engine/Stride.UI/UIElement.cs#L97-L104

Now, `UIElement.DependencyProperties` may hold `ContentArrangeMatrixPropertyKey` and `PanelArrangeMatrixPropertyKey` keys, which both are *not* defined as public keys, they are protected

https://github.com/stride3d/stride/blob/dd6adf2b07d3df0ab6106cf34f15b94c1fe99a01/sources/engine/Stride.UI/Controls/ContentControl.cs#L27
https://github.com/stride3d/stride/blob/dd6adf2b07d3df0ab6106cf34f15b94c1fe99a01/sources/engine/Stride.UI/Panels/Panel.cs#L33

This PR makes sure we generate a serializer for protected as well to ensure those key types can be serialized.

There is no reason not to generate serializers for all PropertyKey<T> types, not just public/protected; if we need one of those properties to not be serialized, not generating a serializer is not the right way to go about it since another public property of the same type could be created which would in turn enable serialization of that first one, still I kept it for non-public instead of any for performance purposes. Something we can investigate again once Joreyk's new source generated serializer is in.

## Related Issue
Fix: #2730

## Types of changes

## Checklist


IXLLEGACYIXL:
we cannot unsafe access protected properties fields, and privates are not at all serialized.. internal wopuldnt be aproblem

Kryptos-FR:
This seems weird to me since we have had UI samples for years (including using prefabs). Did something else create a regression?

I could be wrong but aren't those properties supposed to be calculated during layouting and hence why there aren't serialized?

Otherwise why did we make them protected and not public if they were meant to be serialized?

Eideren:
@Kryptos-FR the issue I laid out in #2730 is fairly specific, there isn't many cases were one would need multiple instances of the same UI page on the same scene, with one of those instance already being inside the scene from the very start.

They are static fields, statics fields aren't serialized, it's the objects those two field contains which are being serialized because they are reference in other fields, and at that point, access modifiers of the field it originated from cannot be taken into account.

All `PropertyKey<T>` should have a serializer generated, it is a serializable type, the reason we're not doing that is because we have performance considerations since the assembly processor has to go through every field declared in the solution for that.

The reason it's protected is because that property only makes sense when you derive from UIElements to make your own layout type. If the writer of that logic actually intended this specific field's object to not be serialized, they should have architected their logic differently, they shouldn't have stored a type which is supposed to be serialized inside of a collection which is definitely serialized.

I can just declare a `public static PropertyKey<Matrix>` in my game solution and this PR wouldn't even be required as the assembly processor would notice it and generate a serializer for the type, which would then be used for those two fields in UI. So the PR is definitely fine by itself, if we actually want to avoid serializing those two fields' values, it's the UI logic that should be changed then.

Kryptos-FR:
@Eideren I can't remember why those fields weren't public in the first place 😅.

I'm ok with this PR, but for our sake can we list all `PropertyKey`, `ParameterKey` (etc.) that are affected by this (beside the ones mentioned in the UI)? I want to make sure it will not have another side effect, and in such case decorate properties we don't want to serialize with a `DataMemberIgnore`.

Eideren:
Also, a thing of note is that there are a bunch of other property keys in the ui solution related to layout and such, and those are all public.

If something else would be affected by this change we would have noticed since it would have crashed the engine like this one did ;)

Still, here's all of the non-public and non-private which would now have a serializer generated for them:

## As PropertyKey<T>
`GraphicsPresenter.ForcedPresentInterval` which is set on the graphics device - not serialized
`CollectionWithIdsSerializerBase.InstanceInfoKey` which is used through YAML serialization but is not actually part of the serialized output
`CollectionWithIdsSerializerBase.DeletedItemsKey` same
`EditorGameComponentGizmoService.ReferencedEntityKey` unused
`AssetFromFileTemplateGenerator.SourceFilesPathKey` stored inside of a serialized collection, but the object containing that collection is not serialized

Those are `PropertyKey<bool>`; even if they aren't public, a serializer to handle them was generated through other public `PropertyKey<bool>` like `Canvas.UseAbsolutePositionPropertyKey` for example.
`ModelFromFileTemplateGenerator.ImportMaterialsKey`
`ModelFromFileTemplateGenerator.DeduplicateMaterialsKey`
`ModelFromFileTemplateGenerator.ImportTexturesKey`
`ModelFromFileTemplateGenerator.ImportAnimationsKey`
`ModelFromFileTemplateGenerator.ImportSkeletonKey`

`ModelFromFileTemplateGenerator.SkeletonToUseKey`
`ModelFromFileTemplateGenerator` keys are used to bind the import configuration for models to be read by the importer, set while processing the model for import based on the file/ui content but not actually serialized to the file

## As ValueParameterKey<T>
`GridGizmoBase.GridColorKey` editor specific, used as a key to write to the grid material, not serialized, but no changes anyway since a serializer would have been serialized from other public `ValueParameterKey<Color4>` like `SpriteEffectKeys.Color`

## As PermutationParameterKey<T>
`DepthMinMax.IsFirstPassKey` used to write to a post effect, this effect is private, it won't be serialized, but no change either since other public `PermutationParameterKey<bool>` like `MaterialKeys.HasSkinningPosition` would have generated a serializer for it.

Kryptos-FR:
Ok with the change then.

Kryptos-FR (approved):

Issue #2733, opened 2025-04-27
compile.bat outputs error message during duration computation


hoelzl:
**Release Type**: GitHub

**Version**: git master (6938820d)

**Platform(s)**: Windows 11

**Describe the bug**
After successfully compiling the project and displaying the time, the `compile.bat` script displays an error message instead of a duration:

```
Time Elapsed 00:01:19.61
Done compiling project: Stride.sln


Starting time was:  1:31:17,08
Ending time is   :  1:32:37,81
Unbalanced parenthesis.
Unbalanced parenthesis.
Missing operator.
81 was unexpected at this time.
```

**To Reproduce**
1. Run `compile.bat` between midnight and 10 o'clock

This probably only happens on systems where one of `ShortTimePattern` or `LongTimePattern` has the form `h:mm...` (with a single `h`, not sure which one of these patterns is relevant or who is responsible for setting them).

**Expected behavior**
Output of the build duration.

**Additional context**
These are the culture settings from my machine. 
```
PS F:\Stride\Git\stride\build> Get-Culture

LCID             Name             DisplayName
----             ----             -----------
1033             en-US            English (United States)


PS F:\Stride\Git\stride\build> (Get-Culture).DateTimeFormat

AMDesignator                     : AM
Calendar                         : System.Globalization.GregorianCalendar
DateSeparator                    : /
FirstDayOfWeek                   : Sunday
CalendarWeekRule                 : FirstDay
FullDateTimePattern              : dddd, MMMM d, yyyy h:mm:ss tt
LongDatePattern                  : dddd, MMMM d, yyyy
LongTimePattern                  : h:mm:ss tt
MonthDayPattern                  : MMMM d
PMDesignator                     : PM
RFC1123Pattern                   : ddd, dd MMM yyyy HH':'mm':'ss 'GMT'
ShortDatePattern                 : M/d/yyyy
ShortTimePattern                 : h:mm tt
SortableDateTimePattern          : yyyy'-'MM'-'dd'T'HH':'mm':'ss
TimeSeparator                    : :
UniversalSortableDateTimePattern : yyyy'-'MM'-'dd HH':'mm':'ss'Z'
YearMonthPattern                 : MMMM yyyy
AbbreviatedDayNames              : {Sun, Mon, Tue, Wed...}
ShortestDayNames                 : {Su, Mo, Tu, We...}
DayNames                         : {Sunday, Monday, Tuesday, Wednesday...}
[...]

PR #2755, opened 2025-04-30
fix: compile.bat error when computing duration (#2733)

hoelzl:
# PR Details

Fixes the error reported in #2733 in a (hopefully) minimally invasive way. 

I also tried another approach to fix this [here](https://github.com/hoelzl/stride/tree/fix-build-de). But that is probably too much code for such a simple task, although it seems slightly more robust. 

I know very little about writing CMD scripts, however, so both approaches might be flawed.

## Related Issue

#2733 

## Types of changes

## Checklist

### DIFF
--- a/build/compile.bat
+++ b/build/compile.bat
@@ -2,7 +2,7 @@
 
 setlocal
 
-set STARTTIME=%TIME%
+set STARTTIME=%TIME: =0%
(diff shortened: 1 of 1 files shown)


hoelzl:
@dotnet-policy-service agree

Eideren:
Thanks !

Issue #2678, opened 2025-03-12
Add interfaces to unify behaviors across constraint components


VaclavElias:
**Is your feature request related to a problem? Please describe.**

This is a follow up on https://github.com/stride3d/stride/pull/2633.

This issue proposes adding:

1.  New `ISpringConstraint` interface to establish a common API for all constraint components that include spring behaviors. Currently, multiple constraint components duplicate the same spring-related properties.
    -  `CollidableComponent`, `AngularHingeConstraintComponent`, `AngularServoConstraintComponent`,  `AngularSwivelHingeConstraintComponent`, `AreaConstraintComponent`, `BallSocketConstraintComponent` and so on. 
1. New `IMotorConstraint` interface to establish a common API for all constraint components that include motor behaviors. Currently, multiple constraint components duplicate the same motor-related properties.
    -  `AngularAxisGearMotorConstraintComponent`, `AngularAxisMotorConstraintComponent`, `AngularMotorConstraintComponent`, `BallSocketMotorConstraintComponent` and so on.
1.  New `IServoConstraint` interface to establish a common API for all constraint components that include servo behaviors. Currently, multiple constraint components duplicate the same motor-related properties.  
    -  `AngularServoConstraintComponent`, `BallSocketServoConstraintComponent`, `DistanceServoConstraintComponent`, `LinearAxisServoConstraintComponent` and so on.

**Describe the solution you'd like**

Create an `ISpringConstraint` interface:

```csharp
/// <summary>
/// Interface for constraints that have spring behavior.
/// </summary>
/// <remarks>
/// Spring constraints allow for soft connections between bodies with configurable
/// <see cref="SpringFrequency"/> and <see cref="SpringDampingRatio"/>.
/// </remarks>
public interface ISpringConstraint
{
    /// <summary>
    /// Gets or sets the target number of undamped oscillations per unit of time.
    /// Higher frequency values create stiffer connections, while lower values allow more elasticity in the joint.
    /// </summary>
    /// <remarks>
    /// Controls how quickly the spring tries to return to its target position.
    /// </remarks>
    [DataMember]
    [Display("Frequency", "Spring")]
    float SpringFrequency { get; set; }

    /// <summary>
    /// Gets or sets the ratio of the spring's actual damping to its critical damping. 0 is undamped, 1 is critically damped, and higher values are overdamped.
    /// Higher damping ratios reduce oscillations and make the connection less elastic.
    /// </summary>
    /// <remarks>
[...]

PR #2763, opened 2025-05-01
docs: Add documentation to bepu constraints

Eideren:
# PR Details
See title, also added interfaces to those constraints to help users create more generalized logic if they need to.

Note: Having all of those duplicate properties everywhere is a pain, but we can't generalize this logic because it reads and writes into bepu's structs under the hood, which do not share implementation between motors, servos, and springs for example.

## Related Issue
Fix: #2678

## Types of changes

## Checklist


VaclavElias:
Thanks taking over this. I will review it!

Eideren:
Added a couple more summaries but I'm done for this PR, the rest will have to be left as is until I have more time/someone else finishes the work.

VaclavElias:
I will be reviewing it today :)

VaclavElias (approved):
Thanks. This is good. If needed we can now improve all related XML comments from interfaces.

PR #2764, opened 2025-05-01
feat: property grid improvements

Kryptos-FR:
# PR Details

Some improvements of the property grid to bring it closer to feature parity with the old editor.

* highlighting
* rule-based auto-expanding

As well as some Avalonia-specific improvements such as using compiled bindings by default, which should bring better performance (https://docs.avaloniaui.net/docs/basics/data/data-binding/compiled-bindings)

## Related Issue

#2739

## Types of changes

## Checklist

### DIFF
--- a/sources/editor/Stride.Assets.Editor.Avalonia/EditorViewSelector.cs
+++ b/sources/editor/Stride.Assets.Editor.Avalonia/EditorViewSelector.cs
@@ -9,7 +9,6 @@ using Stride.Core.Assets.Editor.ViewModels;
 
 namespace Stride.Assets.Editor.Avalonia;
 
-// TODO xplat-editor consider moving this class to Stride.Assets.Editor.Avalonia
 public sealed class EditorViewSelector : AvaloniaObject, IDataTemplate
 {
     private SessionViewModel? session;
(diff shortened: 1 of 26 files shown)


Kryptos-FR:
@Jklawreszuk it shouldn't create conflict with your branch. However, you might have some bindings to fix since I have enabled compiled bindings by default.

Eideren:
Tried it out, looks good. I had a decimal to float conversion exception when selecting a procedural sphere which you likely are already aware of.
Would now be the right time to fix the delayed property expansion ? It is very distracting when trying to compare the properties of two similar objects visually

Kryptos-FR:
The issue with decimal/double/float is that the NumericUpDown in Avalonia only deals with decimal and does conversion on the fly but double.MaxValue is bigger than decimal.MaxValue. I'll probably have to rewrite a custom control that works exclusively with double to prevent unnecessary conversions and avoid exceptions. I'll try to do it in this PR.

The delayed expansion might be trickier. Avalonia doesn't have async properties and on top of that the attached behavior library is a 3rd party. It looks like it needs to be first rendered once before the behavior is executed. Since the property grid is "recursive", each level seems to be rendered one frame after the other, causing this visual glitch.

I won't fix that for now (phase 1), but it's good to list those issues so that we can tackle them later.

Eideren:
Looked at the new changes you introduced, LGTM, feel free to merge when you are ready

PR #2765, opened 2025-05-01
fix: Slider put in Canvas NaN Exception

TranquilAbyss:
Replacing PR #2604 for a more generalize approach as asked.

# PR Details
Canvas can provides children with Infinity values for `availableSizeWithMargins` when measuring. Most `UIElement`s handle Infinity in `MeasureOverride`. For UIElments that do not handle Infinity, an Infinity check was added to UIElement.Measure as a default behavior to check for Parent.RenderSize recursively in `InfinityToParentRenderSize`. 

This was done recursively in since the `Arrange` happens after `Measure`, and  `Arrange` is when RenderSize is normally set except for the root element RenderSize is set earlier. 

Note on UIElments Stretch:
I have tested UIElements like Canvas inside of another Canvas and currently the second Canvas size is set to 0. This appears to be caused by `childrenDesiredSize` being set to `NaN` then being replace in `Measure` with a size of 0. The recursion back to the root element in `InfinityToParentRenderSize` is never reached due to the second `Canvas` `childrenDesiredSize` is not set to infinity, so the parent s. Which means that the Stretch behavior is ignored due how the UI currently works. I choose to leave the recursion since it would allow for Stretch to behave as expected, rather than setting values to NaN, then to 0.

Fixing how the code is ignoring UIElement Stretch seemed to me to be scopecreep. I did not want to ignore the structure of the code by having the `InfinityToParentRenderSize` only look back one parent.

## Related Issue

https://github.com/stride3d/stride/issues/2236

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.UI/UIElement.cs
+++ b/sources/engine/Stride.UI/UIElement.cs
@@ -925,6 +925,10 @@ namespace Stride.UI
             if (float.IsNaN(desiredSize.Z))
                 desiredSize.Z = DefaultDepth;
 
+            // replace infinity with the parent's RenderSize
+            if (float.IsInfinity(desiredSize.X) || float.IsInfinity(desiredSize.Y) || float.IsInfinity(desiredSize.Z))
+                desiredSize = InfinityToParentRenderSize(desiredSize);
+
             // clamp the desired size between the maximum and minimum width/height of the UIElement
             desiredSize = new Vector3(
                 Math.Max(MinimumWidth, Math.Min(MaximumWidth, desiredSize.X)),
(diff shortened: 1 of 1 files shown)


Eideren:
Thanks !

Kryptos-FR:
I didn't ave time to look at it. Just to be sure I understand, the behavior for canvas was unchanged, right? 

If I remember correctly, canvas are by design not supposed to be affected by stretch or most layouting beside setting their position/anchor relative to their parent. That's their point: to give more flexibility to for example draw stuff on top of each other.

See for example the doc for the WPF version: https://learn.microsoft.com/en-us/dotnet/api/system.windows.controls.canvas?view=windowsdesktop-9.0

> Canvas is the only panel element that has no inherent layout characteristics. A Canvas has default Height and Width properties of zero, unless it is the child of an element that automatically sizes its child elements. Child elements of a Canvas are never resized, they are just positioned at their designated coordinates. This provides flexibility for situations in which inherent sizing constraints or alignment are not needed or wanted. For cases in which you want child content to be automatically resized and aligned, it is usually best to use a Grid element.

TranquilAbyss:
The only change made is that after the child determines it size, if the size has an infinity in it then set it to parent RenderSize.

For Canvas in Canvas the recursion currently does nothing and can be removed, since the infinity check logic is bypassed by NaN checks layer happens first and setting the second Canvas size to 0. 

I wish I had your comment earlier, since I did not know the goal was to have child objects default to 0 when undefined in a Canvas. The condition of a Slider in a Canvas was a crash, since Infinity was passed as the Size of the slider which turned into NaN a few frames later. So Sliders now use parent RenderSize instead of  0. I am oversimplifying the logic, but you can see the code to see what I mean exactly.

Kryptos-FR:
@TranquilAbyss to be fair, I'm not 100% sure of what I wrote. I just kind of remember that when the UI system was designed, we took inspiration from WPF.

No need to revert anything right now, we can try to make a few samples with complicated UI combining canvas and slider and see if the result look ok.

PR #2772, opened 2025-05-04
Feat: Add local nugget package sources

Arc-huangjingtong:
# PR Details
This commit is small ,but its will help developer use local code to test project in samples folder

## Types of changes

### DIFF
--- a/nuget.config
+++ b/nuget.config
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <packageSources>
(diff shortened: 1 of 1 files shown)


Kryptos-FR:
The root nuget.config shouldn't be changed for such case. It will interact with other parts which might be either unnecessary or error prone.

Samples can be built against the local version if running from the locally built GameStudio (which should look for the locally built packages and copy them to the StrideDev source location).

If that's really needed for a case not covered here, at the very least if should be a nuget.config file within the samples folder.

PR #2790, opened 2025-05-17
feat: Adds the 8BitDo layout to for proper GamePad support.

Doprez:
# PR Details

Currently 8BitDo controllers dont have a registered layout so they will not work as GamePads. This adds the vendor Id and layout so that it will work outside of an XInput context. 

## Related Issue

No related issues, I found this by trying to use SDL on linux and the GamePad always being null.

## Types of changes

## Checklist


Kryptos-FR:
Do you have a link to the vendor specs? If so can it be added as comment?

We might need to do fixes in the future so it's good to have a reference.

Doprez:
Unfortunately not a full spec sheet I only have the vendor Id and product Id from the mapping on my machine `\\?\hid#vid_2dc8&pid_310a ...` and using [Device Hunt](https://devicehunt.com/view/type/usb/vendor/2DC8) and there is the USB spec sheet for general USB debugging https://www.usb.org/sites/default/files/hut1_4.pdf

Although its important to note that this mapping is probably only useful for the SDL context since the mappings are coming from the [`JoyStick` controller mappings](https://github.com/stride3d/stride/blob/6183210a2ba67f82fe3c3321fcf8ba89009e31d1/sources/engine/Stride.Input/SDL/InputSourceSDL.cs#L113) so maybe I should check that this is also coming from the SDL input source?

sources/engine/Stride.Input/GamePadLayouts/GamePadLayout8BitDo.cs
```diff
@@ -0,0 +1,37 @@
...
+        AddAxisToAxis(5, GamePadAxis.RightTrigger, remap: true);
+    }
+
+    public override bool MatchDevice(IInputSource source, IGameControllerDevice device)
+    {
+        byte[] guidBytes = device.ProductId.ToByteArray();
+
+        ushort vendorId = BitConverter.ToUInt16(guidBytes, 4);
+
+        return vendorId == _vendorId;
```
Kryptos-FR, line 35: Is there a way to do a non-alloc comparison? Maybe with a `Span<byte>` or a similar API.

sources/engine/Stride.Input/GamePadLayouts/GamePadLayout8BitDo.cs
```diff
@@ -0,0 +1,37 @@
+using System;
+
+namespace Stride.Input;
+
+public class GamePadLayout8BitDo : GamePadLayout
```
Kryptos-FR, line 5: If it is not meant to be derived, better seal the class.

Eideren:
Moving this to draft, let us know when you're done working on this one :)

Doprez:
Makes sense, I think I may replace this with a more generic solution because of the "only useful for the SDL context since the mappings are coming from the [JoyStick controller mappings](https://github.com/stride3d/stride/blob/6183210a2ba67f82fe3c3321fcf8ba89009e31d1/sources/engine/Stride.Input/SDL/InputSourceSDL.cs#L113)" issue. 

I did start using the [HIDDevice](https://github.com/DevDecoder/HIDDevices) lib from DevDecoder so that the controllers would work consistently between different windowing backends.

https://github.com/Doprez/Doprez.Stride.MoreInput/blob/main/src/GameControllerHIDDevice.cs

xen2:

Issue #2628, opened 2025-02-03
Lazy-load audio engine


Kryptos-FR:
**Is your feature request related to a problem? Please describe.**
The audio providers/components/etc. are always initialized in a game, even if all scenes don't have any audio component/listener. 

**Describe the solution you'd like**
Have the audio system initialize on-demand (when it is first needed).

**Describe alternatives you've considered**
This behavior could be driven by a configuration option, where audio is
- *opt-out*: in which case any attempt to initialize it would throw
- *lazy-load*: initialized on first use
- *pre-load* (current behavior): initialized at the start

**Additional context**
While rare, some games might not need any audio. Not initializing that part of the engine might save some memory and improve performance.

Another use case is when trying to run a game from WSL where audio is not supported by default. Granted, it is niche but very useful for developers to quickly test whether a game works correctly on Linux without requiring dual-boot or another machine.

VaclavElias:
Also, just when learning game phycics and gaming logic and concepts, one doesn't need necessarily audio.

Doprez:
Could an EntityProcessor be the solution for this? 

Have a component for audio and when its added use the `OnSystemAdd` from the processor to initialize the audio provider.

PR #2808, opened 2025-06-01
feat: Load Services only on first call

Doprez:
# PR Details

This PR is to allow systems to be modified from the `Game` class and not set services until it is first called.

## Related Issue

https://github.com/stride3d/stride/issues/2628 Semi related 
https://github.com/stride3d/stride/pull/2806 This will be useful when avoiding injecting systems at startup.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Engine/Engine/ScriptComponent.cs
+++ b/sources/engine/Stride.Engine/Engine/ScriptComponent.cs
@@ -57,14 +57,6 @@ namespace Stride.Engine
 
             Game = Services.GetSafeServiceAs<IGame>();
             Content = (ContentManager)Services.GetSafeServiceAs<IContentManager>();
-            Input = Services.GetSafeServiceAs<InputManager>();
-            Script = Services.GetSafeServiceAs<ScriptSystem>();
-            SceneSystem = Services.GetSafeServiceAs<SceneSystem>();
-            EffectSystem = Services.GetSafeServiceAs<EffectSystem>();
-            Audio = Services.GetSafeServiceAs<AudioSystem>();
-            SpriteAnimation = Services.GetSafeServiceAs<SpriteAnimationSystem>();
-            GameProfiler = Services.GetSafeServiceAs<GameProfilingSystem>();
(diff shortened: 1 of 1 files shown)


Doprez:
Adding onto this I wonder if it would be useful to have a load on call extension for the IServiceregistry? This could be shortened to something like 
```
public AudioSystem Audio => Services.CachedService<AudioSystem>();
```

Not sure if that would be useful or just push poor design practices.

Kryptos-FR:
Are calls to Services.GetSafeServiceAs<T>()` thread-safe? If not, two threads initializing for the first time could get different instances.

Doprez:
~~None of the service registry calls are thread safe AFAICT~~. It is locked on call.
```csharp
    public T? GetService<T>()
        where T : class
    {
        var type = typeof(T);
        lock (registeredService)
        {
            if (registeredService.TryGetValue(type, out var service))
                return (T)service;
        }

        return null;
    }
```

Although that does bring up a great point and the threaded instance references issues could be something to consider here.

`GetSafeServiceAs` is just meant to throw when the service is null instead of just returning the null, non-existent reference.
```csharp
    public static T GetSafeServiceAs<T>(this IServiceRegistry registry)
        where T : class
    {
        return registry.GetService<T>() ?? throw new ServiceNotFoundException(typeof(T));
    }
```

Eideren:
`GetService` does not create the service, and since calls to that are thread safe as Doprez mentioned, the only way for a user to get two different instances of a given service type would be if they are removing the service from a different thread while accessing it from the main thread. I think it's fair to leave this edge case as is. The user is clearly playing with fire here.
Is this ready to be merged ?
What's up with the Streaming and Content service being omitted ?

Doprez:
>What's up with the Streaming and Content service being omitted ? 

No real reason other than they are very core concepts so I must have thought they werent needed at the time. I can add the same functionality for consistency though.

>Is this ready to be merged ?

Ill add those 2 today and it can be merged. Might as well make it a bit more modular.

Doprez:
All done

Eideren:
Thanks !

PR #2816, opened 2025-06-13
feat: Settings Window

Kryptos-FR:
# PR Details

Adds a Window for editing the settings.

## Types of changes

## Checklist

### DIFF
--- a/sources/editor/Stride.Core.Assets.Editor/Settings/EditorSettings.cs
+++ b/sources/editor/Stride.Core.Assets.Editor/Settings/EditorSettings.cs
@@ -1,0 +1,41 @@
+// Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net)
+// Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using Stride.Core.Settings;
+using Stride.Core.Translation;
+
+namespace Stride.Core.Assets.Editor.Settings;
+
+public static class EditorSettings
+{
+    private static SettingsProfile? profile;
+    public static SettingsContainer SettingsContainer { get; } = new();
+
+    // Categories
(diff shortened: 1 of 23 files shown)


sources/editor/Stride.Core.Assets.Editor/Settings/EditorSettings.cs
```diff
@@ -0,0 +1,41 @@
+// Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net)
+// Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using Stride.Core.Settings;
+using Stride.Core.Translation;
+
+namespace Stride.Core.Assets.Editor.Settings;
+
+public static class EditorSettings
```
IXLLEGACYIXL, line 9: why make it a singleton?
it forces to have only one setting, what about swapping between settings etc, no need for it to be singleton and static?

Kryptos-FR: What do you mean by swapping between settings? Can you explain the use cases?

IXLLEGACYIXL: you have a folder with X different settings bundles and you should be able to swap between them
like there is no reason for it to be singleton, graphicscontext is also not a singleton for example

Kryptos-FR: Why would you have different bundles of settings files for the editor? Those are basic settings like theme, language, list of recent projects, etc.

If you think about having several user profiles, that could be a new feature but it is out of scope for the rewrite effort. Here I'm just porting the existing settings system from the old editor, and such feature is not implemented there either. Feel free to open a feature request.

PR #2821, opened 2025-06-21
fix: DoubleExtension & IntExtension

Jklawreszuk:
# PR Details

PR fixes following bug: 

1. Run new Avalonia Editor on Windows (`dotnet run`)
2. Load any game solution file.
3. Select any asset or go to settings.
4. Crash :

```
Exception: FormatException: The input string '0.5' was not in a correct format.
```

I simply used passed additional CultureInfo argument to take into account the different formats (dots or commas) that occur between languages.

My Current LocaleInfo : pl-PL

## Types of changes

## Checklist

### DIFF
--- a/sources/presentation/Stride.Core.Presentation.Avalonia/MarkupExtensions/DoubleExtension.cs
+++ b/sources/presentation/Stride.Core.Presentation.Avalonia/MarkupExtensions/DoubleExtension.cs
@@ -18,7 +18,7 @@ public sealed class DoubleExtension : MarkupExtension
         Value = 0.0;
     }
 
-    public DoubleExtension(string value)
+    public DoubleExtension(object value)
     {
         Value = Convert.ToDouble(value, CultureInfo.InvariantCulture);
     }
--- a/sources/presentation/Stride.Core.Presentation.Avalonia/MarkupExtensions/DoubleExtension.cs
+++ b/sources/presentation/Stride.Core.Presentation.Avalonia/MarkupExtensions/DoubleExtension.cs
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net) and Silicon Studio Corp. (https://www.siliconstudio.co.jp)
 // Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
 
+using System;
 using System.Globalization;
 using Avalonia.Markup.Xaml;
 using Avalonia.Metadata;
@@ -19,7 +20,7 @@ public sealed class DoubleExtension : MarkupExtension
 
     public DoubleExtension(string value)
     {
-        Value = double.Parse(value, CultureInfo.InvariantCulture);
+        Value = Convert.ToDouble(value, CultureInfo.InvariantCulture);
     }
     
     /// <inheritdoc/>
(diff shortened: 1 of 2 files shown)


sources/presentation/Stride.Core.Presentation.Avalonia/MarkupExtensions/DoubleExtension.cs
```diff
@@ -16,9 +17,9 @@ public DoubleExtension()
         Value = 0.0;
     }
 
-    public DoubleExtension(object value)
+    public DoubleExtension(string value)
```
Jklawreszuk: We dont need to use object as type since its always a string

Jklawreszuk:
Huh.. I realized that Convert.ToDouble(string, IFormarProvider) uses double.Parse under the hood but is checking for nulls. Lets use that 😄

Kryptos-FR:
I don't have that crash. Double extension is a markup extension which is only used on the xaml side.

Can you give more details on your config. For instance, what's the current language and locale of your machine?

sources/presentation/Stride.Core.Presentation.Avalonia/MarkupExtensions/DoubleExtension.cs
```diff
@@ -16,9 +18,9 @@ public DoubleExtension()
...
```
Kryptos-FR: I'd prefer to keep an `object` argument so that it eventually works for any type that can be converted.

sources/presentation/Stride.Core.Presentation.Avalonia/MarkupExtensions/DoubleExtension.cs
```diff
@@ -16,9 +18,9 @@ public DoubleExtension()
...
     {
-        Value = Convert.ToDouble(value);
+        Value = Convert.ToDouble(value, CultureInfo.InvariantCulture);
```
Kryptos-FR, line 22: Only in the case of a `string` use the overload that takes a `CultureInfo`.

```
if (value is string s)...
```

Kryptos-FR (commented):
I haven't tested with other culture so I kind of understand why the fix could be needed.

With that said, `DoubleExtension` is not the only one doing a conversion. Others should be fixed as well.

Jklawreszuk:
> Double extension is a markup extension which is only used on the xaml side.

Yeah, because it is parsed as text, I concluded that it would be better to declare it as `string`, for now 😅 But yeah I guess consequently it would be useful to change the rest of the extensions in this way, so I'm going to undo that

> I don't have that crash. 

That's because my case is niche, I think? 😅 My current Locale is pl-PL, but OS language is en-US. I'm just used to various settings written in English etc. This is probably the cause of the issue

Kryptos-FR:
So the reasons it didn't crash in WPF is that bindings are more tolerant in WPF (exception is caught somewhere and doesn't crash the whole app), and secondly it doesn't look like we had cases where the value had a decimal part (and hence no formatting issue).

Kryptos-FR:
> With that said, `DoubleExtension` is not the only one doing a conversion. Others should be fixed as well.

The only other one actually is IntExtension which shouldn't have the same issue since there are no decimal parts. To be safer, let's also force the use of the InvariantCulture for IntExtension.

Jklawreszuk:
@Kryptos-FR Sounds good to me, let's handle it even though it may be a small chance for an exception.

Issue #1791, opened 2023-09-18
Setting the .net SDK of a project to "net6.0" instead of "net6.0-windows" or "netstandard2.0" results in inconsistent controller behavior


SeleDreams:
**Release Type**:  Official release

**Version**: 4.1.0.1898

**Platform(s)**: Windows 11

**Describe the bug**
When the target sdk of the csproj of a project is set to net6.0, its controller detection will be inconsistent with both net6.0-windows and netstandard2.0. This caused a lot of trouble when porting an old project.
Basically, if a game is compiled with net6.0, it will detect some controllers (XInput controllers) as GameControllers instead of gamepads and as such, those won't be detected by gamepad functions and will not work with the virtual key binding API.
However if a game is compiled with net6.0-windows specifically or netstandard2.0 (for the game's library), then the input system will behave as intended.

**To Reproduce**
1. Create a project with stride
2. Define the net version as "net6.0" for both the game library and game binary
3. Xbox One and Xbox Series X controllers are detected as GameController and will not work with the virtual input system

**Expected behavior**
net6.0 should behave the same as net6.0-windows and netstandard2.0

KanterVault:
It would be great if GameController containing the word Xbox in its name were simply sent to the Input.GamePads list. That would immediately solve all the problems.
![image](https://github.com/stride3d/stride/assets/56644961/38b0923a-7c80-432f-846e-1de55e8882eb)

PR #2826, opened 2025-06-25
fix: Gamepad not detected due to window handler definition error

w0wca7a:
# PR Details

When connecting a Logitech RumblePad II gamepad, `SharpDX.DirectInput` returns an error related to an invalid window identifier.

## Related Issue

[#1791](https://github.com/stride3d/stride/issues/1791)

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Input/Windows/RawInput/Win32.cs
+++ b/sources/engine/Stride.Input/Windows/RawInput/Win32.cs
@@ -16,6 +16,9 @@ namespace Stride.Input.RawInput
 
         [DllImport("user32.dll")]
         public static extern void ClipCursor(IntPtr rect);
+
+        [DllImport("user32.dll")] 
+        public static extern IntPtr GetForegroundWindow();
     }
 }
 #endif
--- a/sources/assets/Stride.Core.Assets.Quantum/AssetCompositeHierarchyPropertyGraph.cs
+++ b/sources/assets/Stride.Core.Assets.Quantum/AssetCompositeHierarchyPropertyGraph.cs
@@ -621,7 +621,7 @@ public abstract class AssetCompositeHierarchyPropertyGraph<TAssetPartDesign, TAs
         var currentParts = Asset.Hierarchy.RootParts.DepthFirst(x => Asset.EnumerateChildParts(x, false)).Select(x => Asset.Hierarchy.Parts[x.Id]);
         foreach (var part in currentParts)
         {
-            if (part.Base is null)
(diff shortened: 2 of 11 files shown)


w0wca7a:
@dotnet-policy-service agree

sources/engine/Stride.Input/Windows/RawInput/Win32.cs
```diff
@@ -16,6 +16,9 @@ internal static class Win32
 
         [DllImport("user32.dll")]
         public static extern void ClipCursor(IntPtr rect);
+
+        [DllImport("user32.dll")] 
+        public static extern IntPtr GetForegroundWindow();
```
Doprez, line 21: How reliable is this to *always* get the game window? If there is a possibility that the game window may not be in focus, there may be a better way to do this with a [`GameWindow`](https://github.com/stride3d/stride/blob/27f87bf78e20cad0c5671ed782e0b46b6fef35e8/sources/engine/Stride.Games/GameWindow.cs#L123) reference.

Eideren:
Converting this to draft while the contributor gets back to us

w0wca7a:
Sorry for the long delay. I can't get back now gamepads from true gamers. I hope to return to this PR in 2-3 months

sources/engine/Stride.Input/Windows/RawInput/Win32.cs
w0wca7a: Thanks for the review. It doesn't really matter how we get the Window handler, gamepad working with `SharpDX.DirectInput` depends from the `CooperativeLevel` mode. It seems that if I can make `CooperativeLevel` work in `Exclusive` mode, it will allow ForceFeedback to work. At the moment, `Exclusive` mode will crash `Stride.Input` when window not in a focus. I will try your recomendation.

w0wca7a:
I tested connection/disconnection a game controller in a foreground/background game window, and also running two games that support game controllers. The game controller works in both simultaneously.
There were no crashes when the game window lost focus.
In any case, this is a temporary solution that allows game controllers to work out of the box. [Silk.Net.Input](https://github.com/dotnet/Silk.NET/tree/main/src/Input) Silk.Net.Input looks more promising.
This [script ](https://gist.github.com/w0wca7a/2b1208cc014329b20011b7f8e1b0d885)is for testing. Any comments are welcome.

w0wca7a:
This PR should be closed. It's better to focus on Silk.NET.Input.

Issue #2412, opened 2024-08-14
Cant resize window in OpenGL


Doprez:
**Release Type**: Official Release

**Version**: 4.2.0.2188

**Platform(s)**: Windows

**Describe the bug**
The Depth Buffer seems to be stuck at a resolution of 1280 x 720. When I try to resize the window this will cause a crash saying it does not match the render target.

**To Reproduce**
1. use the OpenGL context
2. change the screen resolution to anything outside of 1280 x 720
3. witness crash

**Expected behavior**
It should resize

**Screenshots**
![image](https://github.com/user-attachments/assets/804fbe46-8a84-40e4-9806-5f84a083adcb)

**Additional context**
This is the code I used to get the error:
```cs
protected override void BeginRun()
{
    Window.PreferredWindowedSize = (new Int2(2880, 1920));
    Window.AllowUserResizing = true;
    Window.Position = new Int2(0, 0);
    Window.SetSize(Window.PreferredWindowedSize);
}
```

I also tried calling resize on the presenter but it did not seem to have any affect:
```cs
GraphicsDevice.Presenter.Resize(2880, 1920, Graphics.PixelFormat.R8G8B8A8_UNorm_SRgb);
```

Jklawreszuk:
@Doprez I commented out the exception condition and it seems to work somehow.

Jklawreszuk:
Now the mystery remains why and how to handle it 😅. Also resizing unlike for DX seems to be slow but is acceptable

Jklawreszuk:
I also noticed that depth stecil buffer dimension size never changes (Resize required?)

Doprez:
If Im not mistaken, the Directx implementation recreates the depth buffer on resize so I think it would make sense.

Eideren:
Fixed through #2845

PR #2834, opened 2025-07-07
[OpenGL] Removed an unnecessary check in CommandList that crashed the engine when resizing the window

DockFrankenstein:
# PR Details

When using OpenGL as the rendering API, resizing the window would cause an exception in `CommandList.OpenGL.cs` "Depth buffer is not the same size as the render target". This PR removes this check, which fixes the issue.

The reason why this check is unnecessary, is because the depth buffer eventually resizes itself (after a delay of 1 to a couple of frames) and during that time, nothing bad happens. The Direct X implementation - which is probably the best guide to how other API's should work - doesn't do any checks like this and also works just fine.

**As to why this check was there in the first place**: my idea is it's because this check is a relic of an older version of the engine or that it was just unfinished. Seeing how there are many methods and properties returning a default value, the second one is very likely. It's clear that the priority was put on Direct X, while Vulkan and OpenGL were only brought to a good enough state, to make the window appear and display the game.

Do note, I haven't tested if the depth buffer fails to resize itself on Direct X, but either way, this check shouldn't be there and it's already proven to work.

### What this PR doesn't fix
The depth buffer not being resized for a couple of frames seems to be an issue worth investigating, but because that isn't causing any issues on OpenGL and Direct X (if the problem is also present there), I am leaving it out of this PR.

### What this PR can break
Can't break something that's already broken. From my testing on my machine, this fixes the problem with resizing. In case that under some circumstances it doesn't work, then we won't really introduce a new bug, seeing as that would have caused a crash previously anyway.

## Related Issue

* https://github.com/stride3d/stride/issues/2412

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Graphics/OpenGL/CommandList.OpenGL.cs
+++ b/sources/engine/Stride.Graphics/OpenGL/CommandList.OpenGL.cs
@@ -1222,11 +1222,6 @@ namespace Stride.Graphics
                 // ensure size is coherent
                 var expectedWidth = renderTargets[0].Width;
                 var expectedHeight = renderTargets[0].Height;
-                if (depthStencilBuffer != null)
-                {
-                    if (expectedWidth != depthStencilBuffer.Width || expectedHeight != depthStencilBuffer.Height)
-                        throw new Exception("Depth buffer is not the same size as the render target");
-                }
                 for (int i = 1; i < renderTargetCount; ++i)
                 {
                     if (renderTargets[i] != null && (expectedWidth != renderTargets[i].Width || expectedHeight != renderTargets[i].Height))


ferafiks:
@dotnet-policy-service agree

Jklawreszuk:
>because the depth buffer eventually resizes itself

I am not really sure about this statement. That is absolutely true when it comes to render target texture, but when I tried to observe depth stecil buffer texture size I receive always "default" screen size:

image

ferafiks:
weird, it didn't do that when I tested. I guess I'm gonna have to investigate this further

Jklawreszuk:
@DockFrankenstein Sure, maybe we can somehow find a solution together 😅

ferafiks:
After some investigation, it seems that the check fixed window resizing only on windows 11. Everywhere else (linux, win10), the behaviour varies depending on how the system manages windows. I think removing the check is still worth it, even if it means it fixes the problem only on windows 11. However, I'm looking into fixing the problem entirely, so for now, I'm closing this PR

PR #2837, opened 2025-07-10
Allow editing the animations on AnimationComponent

Keepsie:
# PR Details

Enable programmatic animation assignment instead of requiring just editor setup.
Currently, users must configure animations in the editor, then reference them by name in code. This creates duplicate work and limits flexibility.

Use Case: We created an HSAnimator script that manages all animation clips and names in one place. By adding the AnimationComponent via code and populating its animations programmatically, we eliminate the need to configure animations twice and gain full control over the animation workflow.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Engine/Engine/AnimationComponent.cs
+++ b/sources/engine/Stride.Engine/Engine/AnimationComponent.cs
@@ -30,7 +30,7 @@ namespace Stride.Engine
         /// Gets the animations associated to the component.
         /// </summary>
         /// <userdoc>The list of the animation associated to the entity.</userdoc>
-        public Dictionary<string, AnimationClip> Animations { get; } = new();
+        public Dictionary<string, AnimationClip> Animations { get; set; } = new();
 
         /// <summary>
         /// Gets list of active animations. Use this to customize startup animations.


Keepsie:
@dotnet-policy-service agree company="Happenstance Games"

Kryptos-FR:
I believe the property only has a getter because it might be cached somewhere. With a setter we would have to check if the reference changed during execution, everytime we have to access it. And this could break assumptions especially in a multi-threaded environment.

Currently I don't see what this PR adds that can't already be done. Collections and dictionaries have syntax that allows them to be initialized inline without requiring a setter.

For example, this already works:

```csharp
new AnimationComponent
{
    Animations =
    {
        { "Key", animationClip }
    }
}
```

In case the dictionary needs to be filled up later (after initialization), it can be done with `.AddRange()` taking another dictionary as argument:

```csharp
var dic = new Dictionary<string, AnimationClip>
{
    // ...
};

animComponent.Animations.AddRange(dic);
```

Keepsie:
I was just trying to replace the whole list since it was already done. I didn't really think about AddRange. Looking at all the refs, I don't see anything that stands out that would be an issue either way.  

refs_for animation

Kryptos-FR:
It might not be an issue now. But if we allow having a setter and later we do need to do some caching, it will make it harder. We would basically have to maintain the expectation that the whole dictionary can be modified at any point during execution.

If it isn't required to solve your issue at the moment, then better not modify the current API.

Keepsie:
Actually after testing AddRange() doesn't work. So now I would need to loop and add which is more work for no reason. 

Addrange

Kryptos-FR:
We have an extension for it here that you can copy: https://github.com/stride3d/stride/blob/master/sources%2Fcore%2FStride.Core.Design%2FExtensions%2FDictionaryExtensions.cs

I don't know what other maintainers think but I'm not in favor of introducing maintenance burden to the API (with the potential future issues I mentioned) for the sake of convenience.

Keepsie:
Fair enough, it’s not that big of a deal to argue about further. Thanks!

PR #2842, opened 2025-07-18
fix: crash when resizing render window and at the same time creating textures in a background thread

azeno:
# PR Details

A relevant lock statement on the internal GraphicsDevice.Resources collection was missing.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Graphics/ResumeManager.cs
+++ b/sources/engine/Stride.Graphics/ResumeManager.cs
@@ -41,22 +41,28 @@ namespace Stride.Graphics
 
         public void Pause()
         {
-            foreach (var resource in graphicsDevice.Resources)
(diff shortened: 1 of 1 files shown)


Eideren:
Looks like there are other places in the source where `device.Resources` is accessed without a lock, it looks to me like those spots could also be problematic ?

azeno:
Ok can you point me to them? I remember I did a quick search and all the other places looked good to me.

Eideren:
Entirely within this file actually: https://github.com/stride3d/stride/blob/master/sources/engine/Stride.Graphics/ResumeManager.cs

azeno:
@Eideren Thanks, updated

Eideren:
Thanks !

Issue #2828, opened 2025-07-01
[OpenGL] Game.GraphicsDevice.Adapter.Outputs being incorrectly set on OpenGL


DockFrankenstein:
**Release Type**: Official Release

**Version**: 4.2.0.2381

**Platform(s)**: Windows and Linux (with OpenGL Graphics API)

**Describe the bug**
When using OpenGL as the Graphics API, `Game.GraphicsDevice.Adapter.Outputs` contains only 1 output no matter how many monitors are connected to the computer, with all the values set to 0,0. From what I've seen by skimming through Stride's source code at 3 am while being sleep deprived, the same thing probably happens with Vulkan as well.

**To Reproduce**
1. Create a blank Stride project.
2. Create a script that prints either to the command line or using `DebugText` all of the outputs in `Game.GraphicsDevice.Adapter.Outputs` with their position and size.
3. In `MyGame.Windows.csproj` add `<StrideGraphicsApi>OpenGL</StrideGraphicsApi>` to the same group as `<TargetFramework>`.
4. Run the game and observe 

**Expected behavior**

**Screenshots**

On Linux using OpenGL

![Image](https://github.com/user-attachments/assets/055996fe-fe5c-4b93-b2f5-cfe909ea9f52)

On Windows using OpenGL

![Image](https://github.com/user-attachments/assets/cca33e80-1a66-4100-9f95-453d44ceb901)

On Windows using the default API

![Image](https://github.com/user-attachments/assets/c3292916-860c-4958-bddd-2f55b1a4e268)

Eideren:
Fixed through #2845

Issue #2411, opened 2024-08-14
Monitor handle missing with OpenGL


Doprez:
**Release Type**: Official Release

**Version**: 4.2.0.2188

**Platform(s)**: Windows

**Describe the bug**
I was trying to get the screen bounds to be able to scale the window based on the Users screen resolution. However the monitor handle is a null pointer when the Graphics API is OpenGL. I have not tested this on Linux but this will be a useful to know since this would be a useful feature for devs.

**To Reproduce**
1. Inherit from the Game class
2. override `BeginRun()`
3. add the following code:
```cs
protected override void BeginRun()
{
  base.BeginRun();
  var desktopBounds = GraphicsDevice.Adapter.Outputs[0].DesktopBounds;
}
```
4. place breakpoint after desktop bounds and see broken resolution.

**Expected behavior**
This should function the same way that the DirectX graphics API does and have the monitor information.

**Screenshots**
![image](https://github.com/user-attachments/assets/e337620f-734b-44c2-a826-7ab9cccc0f27)

**Additional context**
Im not sure if its related but you also can not resize the window while using OpenGL but that might need a separate issue.

Eideren:
Fixed through #2845

PR #2845, opened 2025-07-20
fix: OpenGL window resizing and GraphicsAdapter.Outputs not being properly populated

ferafiks:
# PR Details

This PR fixes a crash that would occur when trying to resize a window on OpenGL, other issues with resizing the depth buffer, a fix to `GraphicsAdapter.Outputs` not being populated with the right values and a refactor to how SDL window resize events are handled

## Related PRs
* https://github.com/stride3d/stride/pull/2834

## Related Issue
* https://github.com/stride3d/stride/issues/2828
* https://github.com/stride3d/stride/issues/2412
* https://github.com/stride3d/stride/issues/2411

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Graphics/OpenGL/GraphicsOutput.OpenGL.cs
+++ b/sources/engine/Stride.Graphics/OpenGL/GraphicsOutput.OpenGL.cs
@@ -10,6 +10,15 @@ namespace Stride.Graphics
 {
     public partial class GraphicsOutput
     {
+        private readonly int displayIndex;
+
+        public string DisplayName { get; init; }
+
+        public IntPtr MonitorHandle
+        {
+            get { return IntPtr.Zero; }
+        }
+
         /// <summary>
         /// Initializes a new instance of <see cref="GraphicsOutput" />.
         /// </summary>
@@ -38,10 +47,6 @@ namespace Stride.Graphics
             DisplayName = SDL.GetDisplayNameS(displayIndex);
         }
 
-        private readonly int displayIndex;
-
-        public string DisplayName { get; init; }
-
         /// <summary>
         /// Find the display mode that most closely matches the requested display mode.
         /// </summary>
@@ -65,16 +70,11 @@ namespace Stride.Graphics
             return closest;
         }
 
-        public IntPtr MonitorHandle
-        {
-            get { return IntPtr.Zero; }
-        }
-
         private void InitializeSupportedDisplayModes()
         {
             var SDL = Stride.Graphics.SDL.Window.SDL;
 
-            var modesMap = new Dictionary<string, DisplayMode>();
+            var modesMap = new Dictionary<(int w, int h, int refreshRate), DisplayMode>();
             var modeCount = SDL.GetNumDisplayModes(displayIndex);
 
             unsafe
@@ -84,7 +84,7 @@ namespace Stride.Graphics
                     var sdlMode = new Silk.NET.SDL.DisplayMode();
                     SDL.GetDisplayMode(displayIndex, i, &sdlMode);
 
-                    var key = $"{sdlMode.W};{sdlMode.H};{sdlMode.RefreshRate}";
+                    var key = (sdlMode.W, sdlMode.H, sdlMode.RefreshRate);
 
                     if (!modesMap.ContainsKey(key))
                     {
@@ -95,9 +95,7 @@ namespace Stride.Graphics
                 }
             }
 
-            supportedDisplayModes = modesMap
-                .Select(x => x.Value)
-                .ToArray();
+            supportedDisplayModes = modesMap.Values.ToArray();
         }
 
         private void InitializeCurrentDisplayMode()
--- a/sources/engine/Stride.Graphics/OpenGL/SwapChainGraphicsPresenter.OpenGL.cs
+++ b/sources/engine/Stride.Graphics/OpenGL/SwapChainGraphicsPresenter.OpenGL.cs
@@ -1,7 +1,7 @@
 // Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net) and Silicon Studio Corp. (https://www.siliconstudio.co.jp)
 // Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
 #if STRIDE_GRAPHICS_API_OPENGL
-using Stride.Core.Collections;
+using System.Collections.Generic;
 using Rectangle = Stride.Core.Mathematics.Rectangle;
 using Window = Stride.Graphics.SDL.Window;
 using WindowState = Stride.Graphics.SDL.FormWindowState;
@@ -79,12 +79,12 @@ namespace Stride.Graphics
             // Manually update the texture
             backBuffer.OnDestroyed();
 
-            var fastList = DestroyChildrenTextures(backBuffer);
+            var list = DestroyChildrenTextures(backBuffer);
 
             // Put it in our back buffer texture
             backBuffer.InitializeFrom(newTextureDescrition);
 
-            foreach (var texture in fastList)
+            foreach (var texture in list)
             {
                 texture.InitializeFrom(backBuffer, texture.ViewDescription);
             }
@@ -99,12 +99,12 @@ namespace Stride.Graphics
             // Manually update the texture
             DepthStencilBuffer.OnDestroyed();
 
-            var fastList = DestroyChildrenTextures(DepthStencilBuffer);
+            var list = DestroyChildrenTextures(DepthStencilBuffer);
 
             // Put it in our back buffer texture
             DepthStencilBuffer.InitializeFrom(newTextureDescrition);
 
-            foreach (var texture in fastList)
+            foreach (var texture in list)
             {
                 texture.InitializeFrom(DepthStencilBuffer, texture.ViewDescription);
             }
@@ -115,20 +115,19 @@ namespace Stride.Graphics
         /// </summary>
         /// <param name="parentTexture">Specified parent texture</param>
         /// <returns>A list of the children textures which were destroyed</returns>
-        private FastList<Texture> DestroyChildrenTextures(Texture parentTexture)
+        private List<Texture> DestroyChildrenTextures(Texture parentTexture)
         {
-            var fastList = new FastList<Texture>();
+            var list = new List<Texture>();
             foreach (var resource in GraphicsDevice.Resources)
             {
-                var texture = resource as Texture;
-                if (texture != null && texture.ParentTexture == parentTexture)
+                if (resource is Texture texture && texture.ParentTexture == parentTexture)
                 {
                     texture.OnDestroyed();
-                    fastList.Add(texture);
+                    list.Add(texture);
                 }
             }
 
-            return fastList;
+            return list;
         }
     }
 }


sources/engine/Stride.Graphics/OpenGL/CommandList.OpenGL.cs
```diff
@@ -1222,11 +1222,6 @@ private void SetRenderTargetsImpl(Texture depthStencilBuffer, int renderTargetCo
                 // ensure size is coherent
                 var expectedWidth = renderTargets[0].Width;
                 var expectedHeight = renderTargets[0].Height;
-                if (depthStencilBuffer != null)
-                {
-                    if (expectedWidth != depthStencilBuffer.Width || expectedHeight != depthStencilBuffer.Height)
-                        throw new Exception("Depth buffer is not the same size as the render target");
-                }
```
Ethereal77: If the problem is a mix of SDL not doing the right thing and some faulty logic in the way fullscreen is being treated, why eliminate this check?

Is is not reasonable that the CommandList checks that the depth buffer is the same size as the render targets?

ferafiks: Main reason: direct x doesn't check that and previously even without those checks, it would make the resizing work in some cases

ferafiks: Basically, I don't see why would we want to crash people's games when they could work just fine and even for debugging purposes, if there was an issue with resizing the depth buffer again, it will let know of itself in better ways

Also, the depth buffer is resized by stride, so we can be safe trusting that SDL won't break anything related to that

Ethereal77: Crash aside, if the depth buffer isn't being resized correctly elsewhere in the engine, that’s its own issue. But keeping render targets the same size as the depth buffer is standard practice.

Also, we are talking here about a low-level graphics abstraction, the `CommandList`. Mantaining a validation check in this context can be vital. Some graphics APIs assume or require that depth buffers match the size of render targets for proper depth testing. Even when not required, failing fast with a clear error (like this check) makes bugs easier to diagnose, instead of silent misbehavior.

If some part of the engine fails to resize the depth buffer properly during window resize, then skipping the check only papers over the real problem.
So, I think we should strive to fix the unintended behavior when resizing a window, not eliminating validations in lower-level APIs.

manio143: Maybe `Debug.Assert` would be better here than `if() throw`?

ferafiks: I think if a check like this should exist, it should be done in SetRenderTargets and not SetRenderTargetsImpl for every Graphics API, but I don't want to accidentally break anything, so I'm gonna re-add it as Debug.Assert for OpenGL as suggested by @manio143

Ethereal77: `Debug.Assert()` would be a nice option for validation. The problem with that is that users that download Game Studio get the release builds of the assemblies. Even if they have the project configured for `DEBUG` build, the Stride NuGet packages only contain the release assemblies, with the `Debug.Assert()` methods stripped out (they have the `[Conditional("DEBUG")]` attribute if I remember correctly. If would be nice if NuGet allowed us an easy way to package both `DEBUG` and `RELEASE` builds and decide which to consume based on project configuration, but for now...

Maybe moving the check to the common `SetRenderTargets` would make sense, yes. Although the `Impl` methods are there to apply platform- and API-specific validation and logic. As I've written in my above comment, some APIs may expect or even require this match in dimensions, and others may not be so strict.
Also, the `Impl` methods are the common ground where the logic takes place, and it is called by many other methods. Having the validation in a single method is better than repeating the same validation over several other methods that all end up calling a common implementation.

I don't want to sound too negative. I'm not completely opposed to what you are saying. Just exposing facts, and explaining what I know about graphics APIs in general, and about Stride's abstraction in particular. For reference, I'm the one currently refactoring a big chunk of all that in the Silk.NET conversion PR.

Ethereal77: A bit off-topic: Some times I've been thinking if it would make sense to have something like `Guard.Assert()`-like methods that would validate according to a `static readonly` flag configurable by the application, that the JIT can remove easily in release builds. That way you can, for example, throw in `DEBUG` but just log an error or elide the check completely in `RELEASE`. I've not tried if this would work though.

Speaking about `SetRenderTargets`, I forgot to mention that as `CommandList` is a low-level abstraction, imagine it being used isolated for some graphics task. Not always is going to be used with a `GameWindow` or even a `Game` that knows the concept of window resizing.
I say this because if you remove the check because a higher-level concept like a window fails when resizing because of `CommandList`, then those other uses of `CommandList` that do not depend on a higher-level concept suffer from worse validation and possibly subtle and difficult to pinpoint incorrect behaviors.

ferafiks: @Ethereal77 I restored the original checks

manio143: Not to make this thread longer, but @Ethereal77 I think the audience of the exception is important to take into consideration. If this is thrown due to user error, it should be an exception and thrown in release binaries. But if it is thrown due to bugs in the engine code, and it's not leaving us in a fatal state, then I'd expect a debug assert which will be encountered by anyone building Stride locally from git.

Ethereal77: I agree with you. But think that Stride is divided into NuGet packages that can be consumed in isolation. Even if you tipically use the engine as a whole and that includes `Stride.Graphics`, you can also use it and consume it on its own. It is also divided that way on purpose, as you can see from the projects being separated into groups like `Core` (which originally I think it was named "SiliconStudio"), `Engine` (was "Xenko"?), `Editor` (was "Xenko.Editor"), etc.

What I consider the foundations of the engine proper is `Stride.Games` and `Stride.Engine`. If there is a bug there, I agree it should be fixed there. There you can even use internal methods in `Stride.Graphics` that have no validation or just `Debug.Assert`s. That's what they have the `[InternalsVisibleTo]` attribute for.

But for `Stride.Core.X`, `Stride.Graphics`, and some others, I like to view them as consumer-facing APIs that can be used on its own. This is just my opinion, of course.

Eideren:
@DockFrankenstein the inconsistent behavior issue you linked has been commented on yesterday, any thoughts on what they said ?

ferafiks:
@Eideren Sorry, I'm not sure which issue you are talking about

Eideren:
This one https://github.com/libsdl-org/SDL/issues/13149

ferafiks:
Okay I read it. Seems like this is a "sort of" intended behavior. I'm gonna have to look into this more and hopefully create a better, more proper fix

sources/engine/Stride.Graphics/OpenGL/SwapChainGraphicsPresenter.OpenGL.cs
```diff
@@ -77,8 +79,15 @@ protected override void ResizeBackBuffer(int width, int height, PixelFormat form
             // Manually update the texture
             backBuffer.OnDestroyed();
 
+            var fastList = DestroyChildrenTextures(backBuffer);
```
Ethereal77: Nitpick: Call variables with a name describing what they are for, not what they are (i.e. `destroyedTextures` for example).

sources/engine/Stride.Graphics/OpenGL/SwapChainGraphicsPresenter.OpenGL.cs
```diff
@@ -90,8 +99,36 @@ protected override void ResizeDepthStencilBuffer(int width, int height, PixelFor
             // Manually update the texture
             DepthStencilBuffer.OnDestroyed();
 
+            var fastList = DestroyChildrenTextures(DepthStencilBuffer);
```
Ethereal77: Same as above

sources/engine/Stride.Graphics/OpenGL/SwapChainGraphicsPresenter.OpenGL.cs
```diff
@@ -90,8 +99,36 @@ protected override void ResizeDepthStencilBuffer(int width, int height, PixelFor
...
+        }
+
+        /// <summary>
+        /// Calls <see cref="Texture.OnDestroyed"/> for all children of the specified texture
+        /// </summary>
+        /// <param name="parentTexture">Specified parent texture</param>
+        /// <returns>A list of the children textures which were destroyed</returns>
+        private FastList<Texture> DestroyChildrenTextures(Texture parentTexture)
+        {
+            var fastList = new FastList<Texture>();
```
Ethereal77: Same as above

sources/engine/Stride.Graphics/OpenGL/GraphicsOutput.OpenGL.cs
```diff
@@ -3,14 +3,66 @@
...
 {
     public partial class GraphicsOutput
     {
+        /// <summary>
+        /// Initializes a new instance of <see cref="GraphicsOutput" />.
+        /// </summary>
+        /// <param name="adapter">The adapter.</param>
+        /// <param name="displayIndex">Index of the output.</param>
+        /// <exception cref="System.ArgumentNullException">output</exception>
+        public GraphicsOutput(GraphicsAdapter adapter, int displayIndex)
```
Ethereal77: As `GraphicsOutput` is a class intended to be created internally and only enumerated by the user, prefer `internal` here. Also, for consistency with the DirectX version.

sources/engine/Stride.Graphics/OpenGL/SwapChainGraphicsPresenter.OpenGL.cs
ferafiks: This code is copied from [the Direct3D implementation](https://github.com/stride3d/stride/blob/master/sources/engine/Stride.Graphics/Direct3D/SwapChainGraphicsPresenter.Direct3D.cs#L270). I would also prefer if it was named better, but I want to keep it as close to the main graphics API, to make it clearer in the future, if someone wanted to rework rendering

sources/engine/Stride.Graphics/OpenGL/SwapChainGraphicsPresenter.OpenGL.cs
ferafiks: Same as above

sources/engine/Stride.Graphics/OpenGL/SwapChainGraphicsPresenter.OpenGL.cs
ferafiks: Same as above

ferafiks:
After a comment from @Eideren, I decided to look into how the engine handles resizing to figure out why I had to do the original hack in the first place. I already explained the issue [in the discord server](https://discord.com/channels/500285081265635328/500292388820877322/1412106997913882644), but the gist is: previously, we incorrectly assumed what `SizeChanged` and `Resized` events in SDL did based on how they behaved in windows.

Here is the correct description:
- `SizeChanged` - when the window changes its size in any way (user resizing, via code, or some other way by the desktop environment)
- `Resized` - when the window is manually resized by the user (note: alt+enter doesn't trigger it)

References: [ [forum post](https://forums.libsdl.org/viewtopic.php?p=38342) ] [ [sdl documentation](https://wiki.libsdl.org/SDL2/SDL_WindowEventID) ]

sources/engine/Stride.Graphics/OpenGL/SwapChainGraphicsPresenter.OpenGL.cs
Ethereal77: Then it should be changed there also. I'll note it for later, as it is me who is rewriting it for Silk.NET.

sources/engine/Stride.Graphics/OpenGL/SwapChainGraphicsPresenter.OpenGL.cs
Ethereal77: Same as above 😁

sources/engine/Stride.Graphics/OpenGL/SwapChainGraphicsPresenter.OpenGL.cs
Ethereal77: Noted

sources/engine/Stride.Games/SDL/GameFormSDL.cs
```diff
@@ -66,7 +65,7 @@ public GameFormSDL(string text) : base(text)
         /// <summary>
         /// Occurs when [user resized].
         /// </summary>
-        public event EventHandler<EventArgs> UserResized;
+        public event EventHandler<EventArgs> SizeChanged;
```
Ethereal77: You have changed the name of the event, but not the comment.

Ethereal77:
Do you think there's something that could be done to the SDL backend to better convey the difference? Or to better document it?
I think maybe the SDL backend is in need of some love by someone that knows about it.

In the future, when Silk supports SDL3, it should be rewritten or improved.

sources/engine/Stride.Games/SDL/GameFormSDL.cs
ferafiks: I didn't touch the summaries, because I thought they had something to do with documentation maybe? I'm not sure why it is even needed to say that the "User Resized" occurs when "User Resized" occurs

ferafiks:
@Ethereal77 I'm not sure. I think just changing the names of the events is enough. We could perhaps write a better summary comment for those

sources/engine/Stride.Games/SDL/GameFormSDL.cs
Ethereal77: They do have something to do with documentation indeed 😁. The API docs are constructed from these comments.

The documentation is pretty lacking in most of Stride. Usually, when you see something like that `when [user resized]` is because it was auto-generated. But the summary is there not to say the same you can already tell from the name of the method/property/field, but to explain it a bit in simple terms, or even point to important things using `<remarks>` tags, etc.

Anyway, my comment was not to say you must write a comprehensive documentation for the event (even if that would be nice), but at least you adjust the comment if you have renamed the event, or if you have discovered (as you said in other reply) that the meaning of the event is not the same as it was previously assumed.

Ethereal77: For events like these I tend to write something along the lines of:
```
Occurs when the window is resized, either by the user or as a result of the system chaning the size of the window (for example, when entering or exiting the full-screen mode).
```
Note this is just an example. I have not the context to know if that summary matches what the `SizeChanged` is representing.

Eideren:
Thanks !

PR #2848, opened 2025-07-22
fix: Bepu and InputSystem update order

Eideren:
# PR Details
Set the update order for the `InputSystem` to a non-zero value, it will now always run before other systems with unset order.
Previously, users may inadvertently introduce a one-frame input latency to their scripts by creating the `ScriptSystem` before the `InputSystem`.
Also provides a way for users to insert systems that may run after the `InputSystem` but before the `ScriptSystem` without touching those two systems or creating a custom game implementation.

Given the above, moved the execution point of `Bepu` earlier in the frame, before `ScriptSystem` runs to make sure interpolated bodies update their positions before users may read them in update, but after `InputSystem` otherwise users would be reading last frame's input when they implement `Components.ISimulationUpdate`.

The above will reduce/eliminate apparent stuttering with user script when moving objects based on physics object transforms.

## Related Issue
None afaict.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.BepuPhysics/Stride.BepuPhysics/Definitions/SystemsOrderHelper.cs
+++ b/sources/engine/Stride.BepuPhysics/Stride.BepuPhysics/Definitions/SystemsOrderHelper.cs
@@ -1,19 +1,23 @@
 // Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net)
 // Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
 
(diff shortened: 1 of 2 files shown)


Doprez:
LGTM 

On a side note, would it make sense to create a PR tag for minor behaviour changes in code? I don't see this breaking anything but adding this with a specific note to releases may help existing users who may have done something specific with `GameSystem` order.

Doprez (approved):

Eideren:
@Doprez Good idea, added Behavioral-Change

PR #2855, opened 2025-08-02
[UI] Add gap support to layout panels

### LINKED ISSUE #2852: Gaps in layout components
When laying out UI forms, I want to add some space between the UI elements. To achieve this I currently have to add a margin to each child component of a stack panel or grid. Alternatively, I add empty columns and rows with the wanted space in the grid. This causes many additional configurations in the UI page and is frustrating when many changes or reorganizations are necessary.

I want to have a gap functionality at Grids and Stack Panels similar to the `gap` in CSS at grid and flex layouts. So I only have to set the row and column gap, and my child components have space between them.

As described below, a much more complex solution is to add many margins in all child components or add empty grid rows and columns.

**Additional context**

Example use cases could be a
- main menu, where I want some space between the buttons
  Image
- settings menu with space between the labels and the fields (EditTexts, Checkboxes, Selects …) and the rows
  Image

Normally I would say that the documentation should be updated. But there is currently no manual describing the UI components. This could be another issue.

laske185:
# PR Details

Introduced gap functionality to Grid, UniformGrid and StackPanel similar to the `gap` in CSS.
Add `ColumnGap`, `RowGap`, and `LayerGap` properties in GridBase for Grid and UniformGrid to manage spacing between elements. Extend the StackPanel to include a gap property.

- Grid: image
- UniformGrid: image
- Stackpanel: image

## Related Issue

Closes #2852

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.UI/Panels/StackPanel.cs
+++ b/sources/engine/Stride.UI/Panels/StackPanel.cs
@@ -73,6 +73,8 @@ namespace Stride.UI.Panels
         /// </summary>
         private int indexElementMaxScrolling;
 
+        private float gap = 0f;
+
         /// <summary>
         /// Gets or sets the value indicating if the <see cref="StackPanel"/> children must be virtualized or not.
         /// When children virtualization is activated, hidden children's measurement, arrangement and draw are avoided.
@@ -134,6 +136,36 @@ namespace Stride.UI.Panels
             }
         }
 
+        /// <summary>
+        /// Gets or sets the gap between child elements in virtual pixels along the stacking orientation.
+        /// </summary>
+        /// <userdoc>The gap between child elements in virtual pixels along the stacking orientation.</userdoc>
+        [DataMember]
+        [Display(category: LayoutCategory)]
+        public float Gap
+        {
+            get { return gap; }
+            set
+            {
+                if (gap != value)
+                {
+                    gap = Math.Max(0, value);
+                    InvalidateMeasure();
+                }
+            }
+        }
+
+        /// <summary>
+        /// Calculates the total gap size based on the number of children.
+        /// </summary>
+        /// <param name="childCount">The number of children</param>
+        /// <returns>The total gap size</returns>
+        private float CalculateTotalGapSize(int childCount)
+        {
+            if (childCount <= 1) return 0f;
+            return gap * (childCount - 1);
+        }
+
         private enum ScrollRequestType
         {
             AbsolutePosition,
@@ -167,10 +199,15 @@ namespace Stride.UI.Panels
             {
                 --indexElement;
                 accumulatedSize += GetSafeChildSize(indexElement, scrollAxis);
+                
+                // Add gap size between elements
+                if (indexElement < Children.Count - 1)
+                    accumulatedSize += gap;
             }
 
-            // calculate the size taken by all elements if proportional
-            return accumulatedSize / (Children.Count - indexElement) * Children.Count;
+            // calculate the size taken by all elements if proportional, including gaps
+            var elementSize = accumulatedSize / (Children.Count - indexElement);
+            return elementSize * Children.Count + CalculateTotalGapSize(Children.Count);
         }
 
         protected override void OnLogicalChildRemoved(UIElement oldElement, int index)
@@ -217,14 +254,19 @@ namespace Stride.UI.Panels
             foreach (var child in children)
                 child.Measure(childAvailableSizeWithMargins);
 
-            // calculate the stack panel desired size
+            // calculate the stack panel desired size, including gaps
             var desiredSize = Vector3.Zero;
+            var childCount = 0;
             foreach (var child in children)
             {
                 desiredSize[accumulatorIndex] += child.DesiredSizeWithMargins[accumulatorIndex];
                 desiredSize[maximizeIndex1] = Math.Max(desiredSize[maximizeIndex1], child.DesiredSizeWithMargins[maximizeIndex1]);
                 desiredSize[maximizeIndex2] = Math.Max(desiredSize[maximizeIndex2], child.DesiredSizeWithMargins[maximizeIndex2]);
+                childCount++;
             }
+
+            // Add total gap size to the stacking dimension
+            desiredSize[accumulatorIndex] += CalculateTotalGapSize(childCount);
 
             return desiredSize;
         }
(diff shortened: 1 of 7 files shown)


sources/engine/Stride.UI/Panels/StackPanel.cs
```diff
@@ -423,6 +478,11 @@ public void ScrollOf(float offsetToApply)
                 {
                     newElementIndex += Math.Sign(offsetToApply);
                     previousElementAccumulatedSize += newElementSize;
+                    
+                    // Add gap size
+                    if (gap > 0 && newElementIndex >= 0 && newElementIndex < Children.Count)
```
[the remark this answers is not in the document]
laske185: It is faster to check that the gap is not zero.

sources/engine/Stride.UI/Panels/StackPanel.cs
```diff
@@ -736,6 +809,10 @@ private void UpdateScrollPosition(float newScrollPosition)
                 {
                     var elementSize = GetSafeChildSize(firstElementIndex, axis);
                     currentSize += elementSize;
+                    
+                    // Add gap size
+                    if (firstElementIndex > 0)
```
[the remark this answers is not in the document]
laske185: There is no difference. The greater-than zero check is easier.

sources/engine/Stride.UI/Panels/UniformGrid.cs
```diff
@@ -84,19 +84,49 @@ public int Layers
...
+            var totalGapSize = new Vector3(
+                CalculateTotalGapSize(0, Columns),
+                CalculateTotalGapSize(1, Rows),
+                CalculateTotalGapSize(2, Layers));
+            
+            var availableForCells = availableSizeWithoutMargins - totalGapSize;
+            var availableForOneCell = new Vector3(
+                availableForCells.X / gridSize.X, 
+                availableForCells.Y / gridSize.Y, 
+                availableForCells.Z / gridSize.Z);
```
Eideren: `CalculateTotalGapSize` could take and return vectors, and the per component division could be a vector / vector

sources/engine/Stride.UI/Panels/UniformGrid.cs
```diff
@@ -84,19 +84,49 @@ public int Layers
...
             {
                 // compute the size available for the child depending on its spans values
                 var childSpans = GetElementSpanValuesAsFloat(child);
-                var availableForChildWithMargin = Vector3.Modulate(childSpans, availableForOneCell);
+                
+                // Calculate available size including gaps for spanned elements
+                var availableForChildWithMargin = new Vector3(
+                    CalculateSpannedSize(availableForOneCell.X, (int)childSpans.X, 0),
+                    CalculateSpannedSize(availableForOneCell.Y, (int)childSpans.Y, 1),
+                    CalculateSpannedSize(availableForOneCell.Z, (int)childSpans.Z, 2));
```
Eideren: `CalculateSpannedSize` could take and return vectors

sources/engine/Stride.UI/Panels/UniformGrid.cs
```diff
@@ -106,25 +136,46 @@ protected override Vector3 MeasureOverride(Vector3 availableSizeWithoutMargins)
...
+            var totalGapSize = new Vector3(
+                CalculateTotalGapSize(0, Columns),
+                CalculateTotalGapSize(1, Rows),
+                CalculateTotalGapSize(2, Layers));
+            
+            var availableForCells = finalSizeWithoutMargins - totalGapSize;
+            finalForOneCell = new Vector3(
+                availableForCells.X / gridSize.X, 
+                availableForCells.Y / gridSize.Y, 
+                availableForCells.Z / gridSize.Z);
```
Eideren: Same here

sources/engine/Stride.UI/Panels/UniformGrid.cs
```diff
@@ -106,25 +136,46 @@ protected override Vector3 MeasureOverride(Vector3 availableSizeWithoutMargins)
...
             {
                 // compute the final size of the child depending on its spans values
                 var childSpans = GetElementSpanValuesAsFloat(child);
-                var finalForChildWithMargin = Vector3.Modulate(childSpans, finalForOneCell);
+                
+                // Calculate final size including gaps for spanned elements
+                var finalForChildWithMargin = new Vector3(
+                    CalculateSpannedSize(finalForOneCell.X, (int)childSpans.X, 0),
+                    CalculateSpannedSize(finalForOneCell.Y, (int)childSpans.Y, 1),
+                    CalculateSpannedSize(finalForOneCell.Z, (int)childSpans.Z, 2));
```
Eideren: And here

Eideren (commented):
Sorry for the delay, this is great, just one minor nitpick - I understand that some part of those are there to mirror's how the previous logic was setup, but prefer vector operation over per component operation.
```cs
var x = new Vector3(v.X/v2.X, v.Y/v2.Y, v.Z/v2.Z);
// into
var x = v / v2;
```
I'll mark areas where this could be improved below.

sources/engine/Stride.UI/Panels/UniformGrid.cs
laske185: Done.

sources/engine/Stride.UI/Panels/UniformGrid.cs
laske185: `CalculateSpannedSize` accepts now two vectors for the cell size and gaps.

sources/engine/Stride.UI/Panels/UniformGrid.cs
laske185: Done.

sources/engine/Stride.UI/Panels/UniformGrid.cs
laske185: Use simplified `CalculateSpannedSize` too.

laske185:
@Eideren thank you for your review. I applied all suggestions.

Eideren:
Thanks for going through it, just one question
https://github.com/stride3d/stride/blob/3a6ab0496e30cbed1c27f319cf6221fd92c5df6c/sources/engine/Stride.UI/Panels/UniformGrid.cs#L93-L97
This doesn't seem right to me, but I might not fully understand the method, are you sure you want to always return a size >= Vector3.One here ?

laske185:
There was an issue after refactoring what I fixed now. 

The function calculates the size of a spanned element, including the gaps between the columns, rows, and layers. Originally the function has returned the size of the element in one dimension when the span is less than or equal to one, because there are no gaps to add in this case. I restored this behavior by normalizing the span parameter to make all dimensions at least one.

Eideren:
Thanks !

PR #2866, opened 2025-08-21
feat: Moved to generics

Doprez:
# PR Details

This PR shows a POC that would reduce the amount of required casts by users. I don't want to merge this but I was curious on what the thoughts would be on something like this as using object causes a ton of casting and required checks to be usable. Ideally I would redo this PR but with the `Obsolete` attribute. The only variable that gets in the way is `CollectionChanged` as I dont know what the non-obsolete variable would be called.

~I also still need to verify that there wont be weird checks in the GameStudio layers as I assume they may not like generics as much as objects.~ Works fine.

## Related Issue

No related issues. I ran into this today trying to use tracking collections for a modular system I am experimenting with as seen in the code snippet below:
```
    private void WornItemsChanged(object? sender, TrackingCollectionChangedEventArgs e)
    {
        if (e.Action == NotifyCollectionChangedAction.Add)
        {
            var newItem = e.Item;
            if (e.Item is GameItem item && !item.HasDescriptor<WearableDescriptor>())
            {
                _logger.Error($"Only wearable items can be worn. Item: {item.Name}");

                // Remove the item if it is not wearable
                var key = e.Key as string;
                var oldItem = e.OldItem == null ? null : e.OldItem as GameItem;
                WornItems[key] = oldItem;
            }
        }
    }
```

## Types of changes

## Checklist


Doprez:
No issues with a simple example.

{264313F5-AFCE-4B43-B7F4-328448DA2DE6}

```csharp
public class TestComponent : StartupScript
{
    public TrackingCollection<int> Numbers { get; set; } = [];
    public TrackingHashSet<int> UniqueNumbers { get; set; } = new TrackingHashSet<int>();
    public TrackingDictionary<int, string> NumberNames { get; set; } = new TrackingDictionary<int, string>();
}
```

Doprez:
For context on if I were to make this real I would just change the existing `TrackingCollectionChangedEventArgs` to inherit from `TrackingCollectionChangedEventArgs<T,T>` like `TrackingCollectionChangedEventArgs<object, object>` and set it as obsolete for the people who already use this functionality so they can migrate over gradually.

sources/core/Stride.Core/Collections/TrackingCollectionChangedEventArgs.cs
```diff
@@ -59,3 +59,58 @@ public TrackingCollectionChangedEventArgs(NotifyCollectionChangedAction action,
     /// </value>
     public bool CollectionChanged { get; }
 }
+
+public class TrackingCollectionChangedEventArgs<TKey, TValue>
```
Kryptos-FR: It should still inherit from `EventArgs`  by convention.

Kryptos-FR: Let's also *seal* if by default.

Kryptos-FR (commented):
I'm no against it if there are no regression.

I don't think we can do a transition with an duplicated obsolete and non-obsolete property. That would be messy. So it has to be a breaking change IMO.

Doprez:
Im ok with it being a breaking change. I think it would be fairly easy to tell people how to migrate as they just need to define their types but I just dont want to wait for this to release 😆 

I was also thinking of making a seperate event args for Lists since they shouldnt need to define 2 types and would simplify the `Collection` and `HashSet` event definitions.

Doprez:
I went ahead with the implementation that separates the definition between collections and keyed collections. This was a lot nicer to work with at a higher level and just looked cleaner overall.

Would this be able to be pushed in to main as a revision breaking change or would it have to wait for the major/minor release?

sources/core/Stride.Core/Collections/ITrackingCollectionChanged.cs
```diff
@@ -1,13 +1,15 @@
-// Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net) and Silicon Studio Corp. (https://www.siliconstudio.co.jp)
-// Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
```
Eideren: Looks like your IDE ate the header

sources/core/Stride.Core/Collections/TrackingCollection.cs
```diff
@@ -12,38 +12,38 @@ namespace Stride.Core.Collections;
 /// </summary>
 /// <typeparam name="T">The type of elements in the collection.</typeparam>
 [DataSerializer(typeof(ListAllSerializer<,>), Mode = DataSerializerGenericMode.TypeAndGenericArguments)]
-public class TrackingCollection<T> : FastCollection<T>, ITrackingCollectionChanged
+public class TrackingCollection<T> : FastCollection<T>, ITrackingCollectionChanged<T>
 {
-    private EventHandler<TrackingCollectionChangedEventArgs>? itemAdded;
-    private EventHandler<TrackingCollectionChangedEventArgs>? itemRemoved;
+    private EventHandler<TrackingCollectionChangedEventArgs<T>>? _itemAdded;
+    private EventHandler<TrackingCollectionChangedEventArgs<T>>? _itemRemoved;
```
Eideren: Wrong naming convention

sources/core/Stride.Core/Collections/TrackingCollectionChangedEventArgs.cs
```diff
@@ -1,32 +1,17 @@
-// Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net) and Silicon Studio Corp. (https://www.siliconstudio.co.jp)
-// Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
-
```
Eideren: I'm starting to worry that you aren't feeding your IDE correctly

sources/core/Stride.Core/Collections/TrackingDictionary.cs
```diff
@@ -18,27 +18,27 @@ namespace Stride.Core.Collections;
...
-public class TrackingDictionary<TKey, TValue> : IDictionary<TKey, TValue>, IDictionary, ITrackingCollectionChanged
+public class TrackingDictionary<TKey, TValue> : IDictionary<TKey, TValue>, IDictionary, ITrackingKeyedCollectionChanged<TKey, TValue>
     where TKey : notnull
 {
     private readonly Dictionary<TKey, TValue> innerDictionary;
 
-    private EventHandler<TrackingCollectionChangedEventArgs>? itemAdded;
-    private EventHandler<TrackingCollectionChangedEventArgs>? itemRemoved;
+    private EventHandler<TrackingKeyedCollectionChangedEventArgs<TKey, TValue>>? _itemAdded;
+    private EventHandler<TrackingKeyedCollectionChangedEventArgs<TKey, TValue>>? _itemRemoved;
```
Eideren: Yep, same here

sources/core/Stride.Core/Collections/TrackingHashSet.cs
```diff
@@ -12,26 +12,26 @@ namespace Stride.Core.Collections;
...
 /// <typeparam name="T">The type of elements in the hash set.</typeparam>
-public class TrackingHashSet<T> : ISet<T>, IReadOnlySet<T>, ITrackingCollectionChanged
+public class TrackingHashSet<T> : ISet<T>, IReadOnlySet<T>, ITrackingCollectionChanged<T>
 {
     private readonly HashSet<T> innerHashSet = [];
 
-    private EventHandler<TrackingCollectionChangedEventArgs>? itemAdded;
-    private EventHandler<TrackingCollectionChangedEventArgs>? itemRemoved;
+    private EventHandler<TrackingCollectionChangedEventArgs<T>>? _itemAdded;
+    private EventHandler<TrackingCollectionChangedEventArgs<T>>? _itemRemoved;
```
Eideren: Yup

Eideren (commented):
Looks good, just minor stuff. Have you looked at the samples solution if there is any remnant over there as well ?

Eideren:
I'll change this to draft for organizational purposes, feel free to change it back when ready :)

sources/core/Stride.Core/Collections/TrackingCollection.cs
Doprez: Ah my mistake, I thought we were going to use the Microsoft standards for new changes if we worked in them. Ill switch them back to the lower camel case structure.

Doprez:
I should have resolved all of the issues brought up before. I also have some of my time back for the next couple of weeks so I should be a bit more active than I was.

Kryptos-FR:
Given that we just released 4.3, I think it would make sense to start a `future` (or `5.0`) branch which includes breaking changes. We need to maintain 4.3 for minor changes.

Alternatively, we could have a `master-4.3`, but we would need to make that one the default branch.

Doprez:
Im ok with that, this doesnt "fix" an issue and is just a QOL change so its not urgent. 

I think we talked about a `future` branch before for the IOC changes as well. Did a branch get created for that? I definitely would consider this a breaking change for anyone using these event based lists so at least a 4.4 would probably make sense.

Eideren:
Added it to a clear milestone for now

xen2:

PR #2871, opened 2025-08-28
fix: removed double brackets in error log string

MEEMexe:
# PR Details

If a physicscomponent attached to an entity has no collidershapes, there is an error log from the physicscomponent. It should log the name of the entity in the message, but there are two brackets to much in the error string which results in "[PhysicsComponent]: Error: Entity {Entity.Name} has a PhysicsComponent without any collider shape." which dosn't include the real name of the entity.

## Related Issue

I didn't create an issue because it's a simple error and fixed it right away when i saw it.
To reproduce: add a e.g. RigidBodyComponent to an Entity but don't add a collidershape and run the game. You will then see the wrong error log.

## Types of changes

- [ x] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [ x] All new and existing tests passed.

### DIFF
--- a/sources/engine/Stride.Physics/Engine/PhysicsComponent.cs
+++ b/sources/engine/Stride.Physics/Engine/PhysicsComponent.cs
@@ -651,7 +651,7 @@ namespace Stride.Engine
 
             if (ColliderShapes.Count == 0 && ColliderShape == null)
             {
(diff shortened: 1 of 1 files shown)


Eideren:
Thanks !

PR #2873, opened 2025-09-01
feat: new LinePlaneIntersection method

Kreblc3428:
# PR Details

## Types of changes

## Checklist

### DIFF
--- a/sources/core/Stride.Core.Mathematics/CollisionHelper.cs
+++ b/sources/core/Stride.Core.Mathematics/CollisionHelper.cs
@@ -359,32 +359,31 @@ public static class CollisionHelper
     /// Determines whether there is an intersection between a Line and a <see cref="Plane"/>.
     /// </summary>
     /// <param name="plane">The plane to test.</param>
-    /// <param name="pt1">The first point of the line to test.</param>
-    /// <param name="pt2">The second point of the line to test.</param> 
-    /// <param name="point">When the method completes, contains the point of intersection,
+    /// <param name="point1">The first point of the line to test.</param>
+    /// <param name="point2">The second point of the line to test.</param> 
+    /// <param name="intersection">When the method completes, contains the point of intersection,
     /// or <see cref="Vector3.Zero"/> if there was no intersection.</param>
     /// <returns>Whether the two objects intersected.</returns>
-    public static bool LinePlaneIntersection(Plane plane, Vector3 pt1, Vector3 pt2, out Vector3 point)
+    public static bool LinePlaneIntersection(Plane plane, Vector3 point1, Vector3 point2, out Vector3 intersection)
     {
         //Source: Mathematics for 3D Game Programming and Computer Graphics (3rd ed) by Eric Lengyel
         //Reference: Page 93, 99
-        
-        Vector3 N = plane.Normal;
-        Vector3 V = pt2 - pt1;
-        float D = -plane.D;
-        Vector3 S = pt1;
 
-        float denominator = Vector3.Dot(N, V);
-        
-        if (denominator == 0)
+        Vector3 normal = plane.Normal;
+        Vector3 directionVector = point2 - point1;
+
+        float denominator = Vector3.Dot(normal, directionVector);
+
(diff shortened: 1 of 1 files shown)


Kreblc3428:
@dotnet-policy-service agree

sources/core/Stride.Core.Mathematics/CollisionHelper.cs
```diff
@@ -355,6 +355,39 @@ public static float DistanceSphereSphere(ref readonly BoundingSphere sphere1, re
...
+    /// <summary>
+    /// Determines whether there is an intersection between a Line and a <see cref="Plane"/>.
+    /// </summary>
+    /// <param name="plane">The plane to test.</param>
+    /// <param name="pt1">The first point of the line to test.</param>
+    /// <param name="pt2">The second point of the line to test.</param> 
+    /// <param name="point">When the method completes, contains the point of intersection,
+    /// or <see cref="Vector3.Zero"/> if there was no intersection.</param>
+    /// <returns>Whether the two objects intersected.</returns>
+    public static bool LinePlaneIntersection(Plane plane, Vector3 pt1, Vector3 pt2, out Vector3 point)
```
Ethereal77: Other methods in this `CollisionHelper` class uses names such as `vertex1`, `point`, etc.  Maybe there could be more descriptive names for arguments `pt1` and `pt2`? Something like `point1`, `point2`, `intersection`, or something along those lines.

sources/core/Stride.Core.Mathematics/CollisionHelper.cs
```diff
@@ -355,6 +355,39 @@ public static float DistanceSphereSphere(ref readonly BoundingSphere sphere1, re
...
+        //Reference: Page 93, 99
+        
+        Vector3 N = plane.Normal;
+        Vector3 V = pt2 - pt1;
+        float D = -plane.D;
+        Vector3 S = pt1;
+
+        float denominator = Vector3.Dot(N, V);
+        
+        if (denominator == 0)
```
Ethereal77: Floating point comparison against zero is usually a bad idea because of precission issues. Better to use relational operators and/or compare against an _epsilon_ value. Look at how it is done in `ClosestPointPointTriangle()`:
```cs
float d1 = Vector3.Dot(ab, ap);
float d2 = Vector3.Dot(ac, ap);

if (d1 <= 0.0f && d2 <= 0.0f)
{
    result = vertex1; //Barycentric coordinates (1,0,0)
    return;
}
```

sources/core/Stride.Core.Mathematics/CollisionHelper.cs
```diff
@@ -355,6 +355,39 @@ public static float DistanceSphereSphere(ref readonly BoundingSphere sphere1, re
...
+        {
+            point = Vector3.Zero;
+            return false;
+        }
+        
+        float t = -(Vector3.Dot(N, S) + D) /denominator;
+
+        point = S + (t * (V));
```
Ethereal77: This one is a minor nitpick.

I know this may come from mathematical formulas where variables are named `S`, `t`, `N`, or `V`. But in code it helps immensely (specially when debugging or reading the code) to have more descriptive names.

Also, if you have variables that only store the value of some other and are never modified (like `float D = -plane.D`), why not use `-plane.D` directly?

Ethereal77 (commented):
Nice contribution. Is this not somewhat similar to `RayIntersectsPlane()` method? That method expects a ray, and this one two points on a line (that could also be used to create a ray), but it is essentially the same thing, no?

Anyway, just left some comments.

Kreblc3428:
> Nice contribution. Is this not somewhat similar to `RayIntersectsPlane()` method? That method expects a ray, and this one two points on a line (that could also be used to create a ray), but it is essentially the same thing, no?
> 
> Anyway, just left some comments.

Ok I think I made the changes you wanted, thanks.

Ethereal77 (approved):
Nice job. LGTM

Eideren:
Thanks !

PR #2875, opened 2025-09-04
fix: incorrect absolute trigger on FBX model import

Doprez:
# PR Details

There was an issue with materials being added into the asset manager with FBX files due to the path starting with a "/". this replaces the slash with an underscore.

## Related Issue

None reported in GitHub. Logs from Tebjan in Discord:
```
ArgumentException: Asset location [/obj/geo1/quickmawterial4/matnet/principledshader] must be relative and not absolute (not start with '/') (Parameter 'item')
   at Stride.Core.Assets.PackageAssetCollection.CheckCanAdd(AssetItem item) in C:\BuildAgent\work\b5f46e3c4829a09e\sources\assets\Stride.Core.Assets\PackageAssetCollection.cs:line 361
   at Stride.Core.Assets.PackageAssetCollection.Add(AssetItem item) in C:\BuildAgent\work\b5f46e3c4829a09e\sources\assets\Stride.Core.Assets\PackageAssetCollection.cs:line 127
   at Stride.Assets.Presentation.Templates.AssetTemplateGenerator.Run(AssetTemplateGeneratorParameters parameters) in C:\BuildAgent\work\b5f46e3c4829a09e\sources\editor\Stride.Assets.Presentation\Templates\AssetTemplateGenerator.cs:line 52
```

## Types of changes

## Checklist

### DIFF
--- a/sources/tools/Stride.Importer.3D/MeshConverter.cs
+++ b/sources/tools/Stride.Importer.3D/MeshConverter.cs
@@ -1171,7 +1171,7 @@ namespace Stride.Importer.ThreeD
             {
                 var lMaterial = scene->MMaterials[i];
                 // Replace slashes with an underscore to indicate it is not absolute to Strides asset system.
-                var materialName = materialNames[(IntPtr)lMaterial].Replace('/', '_'); ;
+                var materialName = materialNames[(IntPtr)lMaterial].Replace('/', '_');
 
                 materials.Add(materialName, ProcessMeshMaterial(scene, lMaterial));
             }


tebjan:
Ah, thanks a lot. However, I think it should replace all slashes when the name is retrieved from the asset source file. Stride might assume that the slashes are folders in the asset library...

Doprez:
No problem, this should be even simpler then and removes a check which is always a plus!

tebjan:
This looks great now, thanks!

PR #2885, opened 2025-09-11
perf: soft revert of the performance regressions introduced in #1469

Eideren:
# PR Details
Calls to `Utilities.CopyMemory` mapping to `System.Buffer.MemoryCopy` were swapped for `Unsafe.CopyBlockUnaligned` in #1469 without clarification.
Those two methods are not equivalent, `CopyBlockUnaligned` is safer but significantly slower compared to `MemoryCopy` as it does not have any fast path based on the allocation sizes.

As for the safety of the unaligned version, under x86/64 and recent ARM, reading or writing data outside of their natural alignment do not cause any issues besides slowdown. I've ensured that other platforms, like ARM7 and connected devices' memory, keep using the unaligned version.

The vast majority of our codebase ensures that fields and allocations are aligned, so using operations that expect to operate on aligned memory provides a significant performance improvement.

Scenes with a significant amount of draw calls rely on memcopy to fill in graphics commands and shader parameters. 
In one of the heavier sections I have at my disposal, frames can take up to 23 millisecond to prepare on the CPU side, introducing this change cuts it down to 14 millisecond.
Here's the [Profiler sessions.zip](https://github.com/user-attachments/files/22282460/Profile.session.zip)

More information in 
https://github.com/dotnet/runtime/issues/1650#issuecomment-573907109
https://github.com/dotnet/runtime/issues/80068#issuecomment-1376031943

## Related Issue
N/A

## Types of changes

## Checklist

### DIFF
--- a/sources/core/Stride.Core/Utilities.cs
+++ b/sources/core/Stride.Core/Utilities.cs
@@ -34,6 +34,50 @@ namespace Stride.Core;
 /// </summary>
 public static class Utilities
 {
+    // MUST BE A METHOD AND AGGRESSIVELY INLINED, OTHERWISE THE JIT WILL NOT ELIMINATE THE BRANCH
+    [MethodImpl(MethodImplOptions.AggressiveInlining)] 
+    private static bool IsUnalignedSafe() =>
+        RuntimeInformation.ProcessArchitecture is Architecture.X64 or Architecture.X86 or Architecture.Arm64;
+
+    /// <inheritdoc cref="AlignmentFallbackDoc"/>
+    /// <inheritdoc cref="Unsafe.CopyBlock(ref byte, ref readonly byte, uint)"/>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static unsafe void CopyWithAlignmentFallback(ref byte destination, ref readonly byte source, uint byteCount)
+    {
+        if (IsUnalignedSafe())
+            fixed (void* src = &source, dst = &destination)
+                Buffer.MemoryCopy(src, dst, byteCount, byteCount);
+        else
+            Unsafe.CopyBlockUnaligned(ref destination, in source, byteCount);
+    }
+
+    /// <inheritdoc cref="AlignmentFallbackDoc"/>
+    /// <inheritdoc cref="Unsafe.CopyBlock(ref byte, ref readonly byte, uint)"/>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static unsafe void CopyWithAlignmentFallback(void* destination, void* source, uint byteCount)
+    {
+        if (IsUnalignedSafe())
+            Buffer.MemoryCopy(source, destination, byteCount, byteCount);
+        else
+            Unsafe.CopyBlockUnaligned(destination, source, byteCount);
+    }
+
+    /// <summary>
+    /// Zero out memory at the address provided
+    /// </summary>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static unsafe void Clear(void* startAddress, uint byteCount)
+    {
+        // Span swaps between InitBlockUnaligned and _ZeroMemory depending on the size
+        new Span<byte>(startAddress, (int)byteCount).Clear();
+    }
+
+    /// <remarks>
+    /// Some of the architecture dotnet runs on do not support arbitrary unaligned reads or writes,
+    /// use this instead of other memcopy if you aren't sure whether the pointers you passed in are aligned.
+    /// </remarks>
+    static void AlignmentFallbackDoc() { }
+
     /// <summary>
     /// Allocate an aligned memory buffer.
     /// </summary>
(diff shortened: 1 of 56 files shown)


sources/core/Stride.Core/Utilities.cs
```diff
@@ -34,6 +34,50 @@ namespace Stride.Core;
 /// </summary>
 public static class Utilities
 {
+    // MUST BE A METHOD AND AGGRESSIVELY INLINED, OTHERWISE THE JIT WILL NOT ELIMINATE THE BRANCH
+    [MethodImpl(MethodImplOptions.AggressiveInlining)] 
+    private static bool IsUnalignedSafe() =>
```
Ethereal77, line 39: What if it is a `static readonly` field? Theoretically, doesn't the JIT evaluate those only once and uses them like a `const` later on?

sources/core/Stride.Core/Utilities.cs
```diff
@@ -34,6 +34,50 @@ namespace Stride.Core;
...
+        RuntimeInformation.ProcessArchitecture is Architecture.X64 or Architecture.X86 or Architecture.Arm64;
+
+    /// <inheritdoc cref="AlignmentFallbackDoc"/>
+    /// <inheritdoc cref="Unsafe.CopyBlock(ref byte, ref readonly byte, uint)"/>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static unsafe void CopyWithAlignmentFallback(ref byte destination, ref readonly byte source, uint byteCount)
+    {
+        if (IsUnalignedSafe())
+            fixed (void* src = &source, dst = &destination)
+                Buffer.MemoryCopy(src, dst, byteCount, byteCount);
```
Ethereal77, line 49: Why not use `Span<T>.CopyTo()` in these cases. [It seems](https://source.dot.net/System.Private.CoreLib/R/8b6f60fcf144a356.html) it does a `Buffer.Memmove` internally, the same as `Buffer.MemoryCopy` ([link](https://source.dot.net/System.Private.CoreLib/R/9335864fe015e780.html)).

sources/core/Stride.Core/Utilities.cs
Eideren: You would think, but nope, here's disasm:
```asm
; Method C:CopyWithAlignmentFallback2(byref,byref,uint):float (FullOpts)
G_M000_IG01:                ;; offset=0x0000
       sub      rsp, 40

G_M000_IG02:                ;; offset=0x0004
       test     byte  ptr [(reloc 0x7ffb0f9b41d8)], 1
       je       SHORT G_M000_IG08

G_M000_IG03:                ;; offset=0x000D
       cmp      byte  ptr [(reloc 0x7ffb0f6fb090)], 0
       jne      SHORT G_M000_IG06

G_M000_IG04:                ;; offset=0x0016
       vmovss   xmm0, dword ptr [reloc @RWD00]

G_M000_IG05:                ;; offset=0x001E
       add      rsp, 40
       ret      

G_M000_IG06:                ;; offset=0x0023
       vxorps   xmm0, xmm0, xmm0

G_M000_IG07:                ;; offset=0x0027
       add      rsp, 40
       ret      

G_M000_IG08:                ;; offset=0x002C
       mov      rcx, 0x7FFB0F9B4168
       call     CORINFO_HELP_GET_NONGCSTATIC_BASE
       jmp      SHORT G_M000_IG03
RWD00  	dd	3F800000h		;         1
```
While this one is:
```asm
; Method C:CopyWithAlignmentFallback2(byref,byref,uint):float (FullOpts)
G_M000_IG01:                ;; offset=0x0000

G_M000_IG02:                ;; offset=0x0000
       vxorps   xmm0, xmm0, xmm0

G_M000_IG03:                ;; offset=0x0004
       ret      
; Total bytes of code: 5
```
Godbolt shows similar overhead, so this is not a config issue.

sources/core/Stride.Core/Utilities.cs
Ethereal77: Also, for cases where we are sure the memory is aligned (like those where we control the type and ensure proper layouts), you also have the option of `Unsafe.CopyBlock`

Eideren: I chose Buffer.MemoryCopy because it's more straightforward than creating two spans and using `Span.CopyTo` between them. But as you said, they map to the same method, so if span does not provide anything else compared to buffer I think that this is kind of bikeshedd-y ?

Eideren: While `Unsafe.CopyBlock` is an intrinsic, it is not optimized for the wide array of ranges one might have when making copies

sources/core/Stride.Core/Utilities.cs
Ethereal77: Maybe one of those things that it does when doing PGO and tiering?
Anyway, not that important.

sources/core/Stride.Core/Utilities.cs
Ethereal77: For these utility methods, it's enough to use `Buffer.MemoryCopy`, yes.
My point is at those call sites where we could be using a `Span`. Maybe we have more of those in the future, as more of the API is converted to use/accept spans.

Eideren: Yep

Issue #2427, opened 2024-09-04
[GameStudio] Crash whenever the user changes a struct value in the property grid


Eideren:
**Release Type**: Latest built in **debug**

**Describe the bug**
Let's say you have a 
```cs
public class MyComp : StartupScript
{
        public List<Test2> MyField = new List<Test2>();
        [DataContract] public struct Test2{ public bool b; }
}
```
Adding a new item and setting its `b` field to true in the property grid crashes the gamestudio.

---

Why ? An `AssetMemberNode` instance runs its Update function, note the `NotifyContentChanging` and `NotifyContentChanged` below:
https://github.com/stride3d/stride/blob/ba36a7e03d6ced082996a6628b80fe7edaf0cacb/sources/presentation/Stride.Core.Quantum/MemberNode.cs#L102-L124

`NotifyContentChanging` calls `OnPropertyChanged` on an `AssetNodeViewModel` instance with name `b` and value `false` further down the stack:
https://github.com/stride3d/stride/blob/ba36a7e03d6ced082996a6628b80fe7edaf0cacb/sources/presentation/Stride.Core.Presentation/ViewModels/ViewModelBase.cs#L240-L260
Notice the hashset insertion wrapped inside a `DEBUG` scope.

After this, `NotifyContentChanged` from the first snippet runs, creating a new `AssetNodeViewModel` with name `b` and value `true`, then immediately calls https://github.com/stride3d/stride/blob/ba36a7e03d6ced082996a6628b80fe7edaf0cacb/sources/presentation/Stride.Core.Presentation/ViewModels/ViewModelBase.cs#L266-L289 on that new instance failing the hashset test since it's an all new instance.

Eideren:
Debugger screenshot for more context in order of execution:
This is the call to -Changing
![image](https://github.com/user-attachments/assets/6d264477-cacd-41d8-accc-dafb39be7f11)
Here, the creation of the new `AssetNodeViewModel`
![image](https://github.com/user-attachments/assets/f4f8aace-6854-42bf-8cee-acd1041d86ab)
And the call to -Changed on that new `AssetNodeViewModel`
![image](https://github.com/user-attachments/assets/f4351ed7-67f9-40fd-9674-46f7cae82d21)

PR #2886, opened 2025-09-12
fix: Presentation, remove false-positive prone exception

Eideren:
# PR Details
See #2427 for the basics.
I did not change it into an assert as this would block execution every time someone change properties in the UI editor with a debugger attached, making for a poor experience.
Lastly, I could not find any straightforward path to specialize this check to avoid false positives as this scope is too low level.
Every time this one triggered for me was for a false positive, given that, I believe removing it altogether is preferable.

## Related Issue
Fix: #2427

## Types of changes

## Checklist

### DIFF
--- a/sources/presentation/Stride.Core.Presentation/ViewModels/ViewModelBase.cs
+++ b/sources/presentation/Stride.Core.Presentation/ViewModels/ViewModelBase.cs
@@ -13,10 +13,6 @@ namespace Stride.Core.Presentation.ViewModels;
 /// </summary>
 public abstract class ViewModelBase : INotifyPropertyChanging, INotifyPropertyChanged, IDestroyable
 {
-#if DEBUG
-    private readonly List<string> changingProperties = [];
-#endif
-
     /// <summary>
     /// A collection of property names that are dependent. For each entry of this collection, if the key property name is notified
     /// as being changed, then the property names in the value will also be notified as being changed.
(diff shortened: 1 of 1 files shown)


Kryptos-FR:
Historically it was added to figure out hard-to-detect bugs especially regarding the property grid and the graph/node view models.

Instead of crashing in OnPropertyChanged, we could break the debugger (if already attached). That way only debugging sessions are affected, and one can easily comment out that line if necessary.

Eideren:
Changing properties in the UI editor triggers this codepath, changing some properties in user scripts do too, the two of them are false positives. Given how prevalent this is, if we expect users to comment out the assert, I feel like it kind of defeats the purpose of having the assert in the first place, no ? Shouldn't we do it the other way around, only enable it on our side if we ever fall in a case where this could help given that the latter is exceptionally rare in comparison ?

I understand that it would be preferable for us to have this case covered, but users are bound to come in and comment on this assert every time they fall onto it, or worse, they keep it in just in case as well and become increasingly frustrated every time they manipulate properties.

Kryptos-FR:
It only happens in debug. Official release doesn't have debug. It was never meant to be found by users but to be used internally.

Users compiling the engine themselves should also do it in release mode. Only people working on the engine or editor should compile in debug mode. And we want to find those bugs in such case.

I could be wrong but I seem to remember that structs have been supported in the property grid for a long time. Does the same bug happens in earlier version of Stride (or even Xenko)? It could be the symptom of a regression.

Eideren:
Users do run the engine in debug, it is the default mode when following the instructions to build the engine, and we tend to have more users building the engine locally than other engines.

It's not just writing to structs, the field needs to be inside a container for the assert to occur as I mention in the issue linked. This issue has been part of the engine since `3.1.0.1-beta02` as you can see in the duplicate to the linked issue.

Let's say we change it to an assert, is that actually going to be useful for debugging when all it has done as of now is report false positives - it seems more likely to do the opposite, have someone waste their time looking into that assert when it's actually working as intended. We should either rework the check to avoid false positives, or remove it entirely.

xen2:
Agree, if it happens and bother normal execution, it should be gone.
Let's add more specific checks as necessary on future situations (if/when they arise).

PR #2887, opened 2025-09-12
perf: Improve UI batch performance

Eideren:
# PR Details
`DrawBatchPerTextureAndPass` is called multiple times per frame and per instance,
https://github.com/stride3d/stride/blob/4903592434d89a37ad5ec08b06f95db3ffa3a4d5/sources/engine/Stride.Graphics/BatchBase.cs#L461-L467
`GetTemporaryBuffer()` here will retrieve the buffer it just used in a previous call when that buffer may be in use by the GPU causing a stall when calling `MapSubresource()` below:
https://github.com/stride3d/stride/blob/4903592434d89a37ad5ec08b06f95db3ffa3a4d5/sources/engine/Stride.Graphics/BatchBase.cs#L504

This PR cleans up the logic and ensures that temporary buffers are only released on the next `Begin` of the batcher to make sure that they are not still being used.

Performance difference goes from ~0.083ms to ~0.0013ms per call to MapSubresource. This method is called around 24 times per frame in the scene I ran, leading to a reduction of 1.96ms per frame. I expect more complex UIs would show an even larger difference.

As you may have noticed, someone commented on this performance issue. It is now fixed so I removed the comment they left.

## Related Issue
N/A

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Graphics/BatchBase.cs
+++ b/sources/engine/Stride.Graphics/BatchBase.cs
@@ -416,7 +416,7 @@ namespace Stride.Graphics
             DrawBatchPerTextureAndPass(sprites, offset, count);
         }
 
-        private void DrawBatchPerTextureAndPass(ElementInfo[] sprites, int offset, int count)
+        private unsafe void DrawBatchPerTextureAndPass(ElementInfo[] sprites, int offset, int count)
         {
             while (count > 0)
             {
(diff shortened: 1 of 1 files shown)


Eideren:
Replaced the `SetData` workaround with an actual fix, and updated the description to reflect those changes.

PR #2898, opened 2025-09-17
feat: Hierarchical Model Importer

### LINKED ISSUE #2553: Hierarchical FBX Importer
### Summary
Our community has voiced concerns about the current handling of FBX files within our engine. The main issue is that FBX files are imported as a single model instead of being split into multiple entities, respecting the original hierarchy as structured in their 3D package. This discrepancy hinders workflow efficiency and limits creative flexibility, impacting both game developers and artists.

### Proposal
We aim to develop an enhancement for our engine that introduces a much-needed feature: the option to split an FBX file into multiple model assets based on its internal hierarchy. When this toggle is activated, our engine will not only split the model but also automatically generate a prefab that mirrors the original hierarchy found in the FBX file. This approach will significantly streamline the asset integration process, fostering a more efficient and intuitive workflow for our users.

### Objectives
- **Develop a Toggle Feature**: Users can choose how FBX files are imported—either as a single model or split into hierarchical entities.
- **Automatic Prefab Creation**: To further automate the process, the engine will generate a prefab that preserves the original hierarchy of the FBX file.
- **Documentation**: Provide updated documentation to ensure users can easily adapt to and leverage the new feature.

### Funding: $700,- USD
https://opencollective.com/stride3d/projects/hierarchical-fbx-importer

noa7:
This PR adds an “Split hierarchy” workflow for model import: when enabled, Stride imports one Model per mesh (named after the mesh) and automatically generates a Prefab that mirrors the FBX node hierarchy. When disabled, the import produces a single combined Model (no “(All)” suffix).

What’s in this PR-

Split hierarchy checkbox in Add asset → 3D model dialog.

Remembers last value per project.

Behavior (per mode)

Split hierarchy = ON

Imports N Model assets (one per mesh) named:
"<ModelName>-<MeshName>" (sanitized; auto-uniquified with -1, -2, …).

Creates a Prefab named "<ModelName> Prefab" that replicates the FBX hierarchy (one Entity per node, correct parent/child relationships).

Each Entity that hosts a mesh gets a ModelComponent referencing the corresponding per-mesh Model.

Each per-mesh Model keeps only the material(s) it actually uses\

Split hierarchy = OFF

Imports a single Model named exactly "<ModelName>"


noa7:
> @noa7 please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> ```
> 
> > Options:
> > 
> > 
> > ```
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 
> > ```
> > ```
> 

@dotnet-policy-service agree

Eideren:
Looks like materials are not handled properly: every individual object is set to the same material.

The prefab's hierarchy is almost right, there is an additional `RootNode` created and a weird parenting issue, see the following comparison:
image

[Here's the file I used for testing](https://github.com/user-attachments/files/22404071/Untitled.zip) - this is not CC0, do not use in projects.

noa7:
Thanks for sharing the test file. I’ve checked in updates, and both issues are now fixed:

1. The prefab in Stride now exactly replicates the FBX hierarchy as seen in Blender- no extra "RootNode"
2. Material assignments on individual sub-meshes remain intact after import

@Eideren

Eideren:
We're back to the previous issue with the materials, we need to only have the materials that are assigned to an object on that object, not every material that was contained in the original scene.

Also, the pivot position for individual models is not retained, they all have their pivot set to the origin instead of wherever it actually is in the file.

noa7:
are you on latest?  it imports fine my side

image

Eideren:
The vertices are at the right position in world space, but each mesh's pivot point is not right below the model, like it is in the file, it is set to the origin. 
For example, if you *only* place the bookends model in the world, setting its entity's position to 0,0,0 should center the object around the origin
image
image
The prefab entity for a given model should have its position moved to match wherever the model is in the imported file, the model asset itself should not have any transformation applied.

noa7:
Makes sense, thanks for pointing out, This is now fixed.  

With this latest-

1. Only the material applied on mesh are visible in inspector 
2. The relative positiong of each item in prefab based of entity data in scene by storing each mesh's TRS data at its nodeinfo object that prefab is constructed with. And each mesh pivot in scene matches to that in source model

Capture
Capture1

@Eideren

sources/editor/Stride.Assets.Presentation/Templates/ModelFromFileTemplateGenerator.cs
```diff
@@ -144,7 +156,111 @@ protected override IEnumerable<AssetItem> CreateAssets(AssetTemplateGeneratorPar
...
+                var assets = importer.Import(file, importParameters)
+                    .Select(x => new AssetItem(UPath.Combine(parameters.TargetLocation, x.Location), x.Asset))
+                    .ToList();
+
+                var baseName = file.GetFileNameWithoutExtension();
+
+                if (splitHierarchy)
+                {
+                    var entityInfo = importer.GetEntityInfo(file, parameters.Logger, importParameters);
+                    if (entityInfo != null && (entityInfo.Models?.Count ?? 0) > 0)
```
Eideren: Can be simplified `if (entityInfo?.Models?.Count > 0)`

sources/editor/Stride.Assets.Presentation/Templates/ModelFromFileTemplateGenerator.cs
```diff
@@ -144,7 +156,111 @@ protected override IEnumerable<AssetItem> CreateAssets(AssetTemplateGeneratorPar
...
+                                if (item?.Asset is ModelAsset modelAsset)
+                                {
+                                    TrimModelAssetToSingleMaterialByIndex(modelAsset,
+                                        entityInfo.Models[i].OriginalMaterialIndex);   // <- the index from step 1
+                                }
+                            }
+
+
+                            // No combined "(All)" model when splitting; Prefab is the combined representation
+                            AssetItem allModelAsset = null;
```
Eideren: Unused

sources/editor/Stride.Assets.Presentation/Templates/ModelFromFileTemplateGenerator.cs
```diff
@@ -144,7 +156,111 @@ protected override IEnumerable<AssetItem> CreateAssets(AssetTemplateGeneratorPar
...
+
+                            for (int i = 0; i < entityInfo.Models.Count; i++)
+                            {
+                                var item = perMeshModels[i];
+                                if (item?.Asset is ModelAsset modelAsset)
+                                {
+                                    TrimModelAssetToSingleMaterialByIndex(modelAsset,
+                                        entityInfo.Models[i].OriginalMaterialIndex);   // <- the index from step 1
+                                }
+                            }
```
Eideren: Looks like splitting into a new loop is unnecessary, you could just append that logic at line 212 ? And then remove `perMeshAssets` ?

sources/editor/Stride.Assets.Presentation/Templates/ModelFromFileTemplateGenerator.cs
```diff
@@ -160,5 +276,185 @@ protected override IEnumerable<AssetItem> CreateAssets(AssetTemplateGeneratorPar
...
+        {
+            if (asset?.Materials == null) return;
+            if (keepIndex < 0 || keepIndex >= asset.Materials.Count) return;
+
+            var keep = asset.Materials[keepIndex];
+            asset.Materials.Clear();
+            asset.Materials.Add(keep);
+        }
+
+        private static AssetItem BuildPrefabForSplitHierarchy(string baseName, EntityInfo entityInfo, IList<AssetItem> perMeshModels, AssetItem allModelAsset, UDirectory targetLocation)
```
Eideren: Annotate return type with `CanBeNull`, remove `AssetItem allModelAsset` parameter

sources/editor/Stride.Assets.Presentation/Templates/ModelFromFileTemplateGenerator.cs
```diff
@@ -144,7 +156,111 @@ protected override IEnumerable<AssetItem> CreateAssets(AssetTemplateGeneratorPar
...
+                        assets.RemoveAt(idx);
+
+                        // Assign a fresh Id to the single model to avoid any Id collisions
+                        ((ModelAsset)old.Asset).Id = AssetId.New(); // <<—— ensure unique Id
+
+                        var uniqueFile = MakeUniqueFileName(baseName, assets);
+                        var renamed = new AssetItem(UPath.Combine(parameters.TargetLocation, uniqueFile), old.Asset);
+                        assets.Insert(idx, renamed);
+                    }
+                }
```
Eideren: Shouldn't the case for `splitHierarchy == false` be left untouched, what is this fixing exactly ?

sources/editor/Stride.Assets.Presentation/Templates/ModelFromFileTemplateGenerator.cs
```diff
@@ -160,5 +276,185 @@ protected override IEnumerable<AssetItem> CreateAssets(AssetTemplateGeneratorPar
...
+
+        private static void TrimModelAssetToSingleMaterialByIndex(ModelAsset asset, int keepIndex)
+        {
+            if (asset?.Materials == null) return;
+            if (keepIndex < 0 || keepIndex >= asset.Materials.Count) return;
+
+            var keep = asset.Materials[keepIndex];
+            asset.Materials.Clear();
+            asset.Materials.Add(keep);
+        }
```
Eideren: A single model should be able to have more than one material assigned to it, this is not the source of the issue obviously, something else earlier in the call chain doesn't handle it appropriately. Here an example where I simply assigned a second material to the bookend mesh:
image
[Here's the file](https://github.com/user-attachments/files/22453001/Untitled.zip)

sources/editor/Stride.Assets.Presentation/Templates/ModelFromFileTemplateGenerator.cs
```diff
@@ -160,5 +276,185 @@ protected override IEnumerable<AssetItem> CreateAssets(AssetTemplateGeneratorPar
...
+                {
+                    var meshInfo = entityInfo.Models[m];
+                    if (string.IsNullOrEmpty(meshInfo.NodeName))
+                        continue;
+
+                    if (!nodeNameToIndex.TryGetValue(meshInfo.NodeName, out var nodeIndex))
+                        continue;
+
+                    var modelItem = (m >= 0 && m < perMeshModels.Count) ? perMeshModels[m] : null;
+                    if (modelItem?.Asset is ModelAsset)
```
Eideren, line 330: In which case is this if test useful ?

sources/editor/Stride.Assets.Presentation/Templates/ModelFromFileTemplateGenerator.cs
```diff
@@ -160,5 +276,185 @@ protected override IEnumerable<AssetItem> CreateAssets(AssetTemplateGeneratorPar
...
+
+                // Iterate meshes in the same order as perMeshModels
+                for (int m = 0; m < entityInfo.Models.Count; m++)
+                {
+                    var meshInfo = entityInfo.Models[m];
+                    if (string.IsNullOrEmpty(meshInfo.NodeName))
+                        continue;
+
+                    if (!nodeNameToIndex.TryGetValue(meshInfo.NodeName, out var nodeIndex))
+                        continue;
```
Eideren: Shouldn't this be an exception ? If this fail we would silently ignore one of the models ?

sources/editor/Stride.Assets.Presentation/Templates/ModelFromFileTemplateGenerator.cs
```diff
@@ -160,5 +276,185 @@ protected override IEnumerable<AssetItem> CreateAssets(AssetTemplateGeneratorPar
...
+
+                        // Host entity is the node entity by default
+                        var host = entities[nodeIndex];
+
+                        // If host already has a ModelComponent, spawn a child entity for this additional mesh
+                        if (host.Get<ModelComponent>() != null)
+                        {
+                            if (!extraChildCountByNode.TryGetValue(nodeIndex, out var k))
+                                k = 0;
+                            k++;
```
Eideren: rename `k` to something more explicit

sources/editor/Stride.Assets.Presentation/Templates/ModelFromFileTemplateGenerator.cs
```diff
@@ -160,5 +276,185 @@ protected override IEnumerable<AssetItem> CreateAssets(AssetTemplateGeneratorPar
...
+
+            root ??= entities[0];
+
+            // Heuristic: collapse trivial wrapper root (same name/base or "RootNode") with a single child
+            var firstNode = entityInfo.Nodes[0];
+            var firstName = firstNode.Name ?? string.Empty;
+            bool looksLikeWrapper =
+                string.IsNullOrWhiteSpace(firstName)
+                || firstName.Equals(baseName, StringComparison.OrdinalIgnoreCase)
+                || firstName.Equals("RootNode", StringComparison.OrdinalIgnoreCase);
```
Eideren: Redudant `firstName`, you're already testing for null in `looksLikeWrapper`, change to
```
bool looksLikeWrapper =
    string.IsNullOrWhiteSpace(firstNode.Name)
    || firstNode.Name.Equals(baseName, StringComparison.OrdinalIgnoreCase)
    || firstNode.Name.Equals("RootNode", StringComparison.OrdinalIgnoreCase);
```

Eideren (commented):
You haven't thoroughly gone through the LLM's work there it seems, there's a fairly large amount of redundant logic, and comments that are clearly targeted at someone interacting with an LLM, not at describing the logic being implemented. 

I'm okay with reviewing *someone's* work, not going over what an LLM spit out, that is your burden to carry - and you're definitely capable of that, I have seen you write better stuff than this.

I haven't gone through the whole thing yet, but do go over your entire PR and clean things up, there's a lot of low hanging fruits I have not marked.

You could also look into using/enabling nullable, there's a bunch of null checks and tests that are redundant or definitely should throw in the null case.

sources/editor/Stride.Assets.Presentation/Templates/ModelFromFileTemplateGenerator.cs
noa7: fixed

sources/editor/Stride.Assets.Presentation/Templates/ModelFromFileTemplateGenerator.cs
noa7: fixed, no longer needed

sources/editor/Stride.Assets.Presentation/Templates/ModelFromFileTemplateGenerator.cs
noa7: fixed with new mesh to material mapping

sources/editor/Stride.Assets.Presentation/Templates/ModelFromFileTemplateGenerator.cs
noa7: fixed

sources/editor/Stride.Assets.Presentation/Templates/ModelFromFileTemplateGenerator.cs
noa7: it was only there to seperate account for (all) model when in prefab and the legacy flow, but since thats not needed so removed

sources/editor/Stride.Assets.Presentation/Templates/ModelFromFileTemplateGenerator.cs
noa7: Fixed, by storing material to mesh mapping at entityinfo level as ground truth

sources/editor/Stride.Assets.Presentation/Templates/ModelFromFileTemplateGenerator.cs
noa7: fixed, enityinfo child count will always match imported items

sources/editor/Stride.Assets.Presentation/Templates/ModelFromFileTemplateGenerator.cs
noa7: fixed

sources/editor/Stride.Assets.Presentation/Templates/ModelFromFileTemplateGenerator.cs
noa7: fixed, renamed it to counter

sources/editor/Stride.Assets.Presentation/Templates/ModelFromFileTemplateGenerator.cs
noa7: fixed

noa7:
> You haven't thoroughly gone through the LLM's work there it seems, there's a fairly large amount of redundant logic, and comments that are clearly targeted at someone interacting with an LLM, not at describing the logic being implemented.
> 
> I'm okay with reviewing _someone's_ work, not going over what an LLM spit out, that is your burden to carry - and you're definitely capable of that, I have seen you write better stuff than this.
> 
> I haven't gone through the whole thing yet, but do go over your entire PR and clean things up, there's a lot of low hanging fruits I have not marked.
> 
> You could also look into using/enabling nullable, there's a bunch of null checks and tests that are redundant or definitely should throw in the null case.

Got it. I was focusing on getting it fully functional before moving on to cleanup and optimization, since the prefab generation feature is a bit tricker than I anticipated. I’ve removed any mechnical commets.  Committed the latest version with fixes and a cleaner codebase @Eideren

sources/editor/Stride.Assets.Presentation/Templates/ModelFromFileTemplateGenerator.cs
```diff
@@ -123,9 +133,11 @@ protected override IEnumerable<AssetItem> CreateAssets(AssetTemplateGeneratorPar
             var importAnimations = parameters.Tags.Get(ImportAnimationsKey);
             var importSkeleton = parameters.Tags.Get(ImportSkeletonKey);
             var skeletonToReuse = parameters.Tags.Get(SkeletonToUseKey);
+            var splitHierarchy = parameters.Tags.Get(SplitHierarchyKey);   // <-- you read it here
```
Eideren: Remove this comment

sources/editor/Stride.Assets.Presentation/Templates/ModelFromFileTemplateGenerator.cs
```diff
@@ -136,15 +148,85 @@ protected override IEnumerable<AssetItem> CreateAssets(AssetTemplateGeneratorPar
 
             foreach (var file in files)
             {
-                // TODO: should we allow to select the importer?
```
Eideren: No reason to delete this comment

sources/editor/Stride.Assets.Presentation/Templates/ModelFromFileTemplateGenerator.cs
```diff
@@ -154,11 +236,195 @@ protected override IEnumerable<AssetItem> CreateAssets(AssetTemplateGeneratorPar
...
+            }
+
+            prefab.Hierarchy.RootParts.Add(root);
+
+            prefab.Id = AssetId.New(); // ensure unique Id
+            var prefabUrl = new UFile($"{baseName} Prefab");
+            return new AssetItem(UPath.Combine(targetLocation, prefabUrl), prefab);
+        }
+
+        private static string SanitizePart(string s)
```
Eideren: `string?` for both parameter and return type

sources/editor/Stride.Assets.Presentation/Templates/ModelFromFileTemplateGenerator.cs
```diff
@@ -154,11 +236,195 @@ protected override IEnumerable<AssetItem> CreateAssets(AssetTemplateGeneratorPar
...
+
+
+            string sourceName = Path.GetFileNameWithoutExtension(asset.Source);
+            string assetName = assetItem.Location.ToString();
+            
+            string underlyingMeshName= assetName.Substring(sourceName.Length+1);
+            var underlyingModel=entityInfo.Models.Where(C=>C.MeshName==underlyingMeshName).FirstOrDefault();
+            var nodeContainingMesh=entityInfo.Nodes.Where(c=>c.Name== underlyingModel.NodeName).FirstOrDefault();
+
+            var materialIndices=entityInfo.NodeNameToMaterialIndices?.Where(c=>c.Key== nodeContainingMesh.Name)?.FirstOrDefault().Value;
```
Eideren: `NodeNameToMaterialIndices` is never null, the first `?` is unnecessary. A `Where` Linq call never returns null so the second `?` is unnecessary as well, `Where(x).FirstOrDefault()` can be simplified like mentioned above but ... it is a dictionary ... you can replace the `Where` with a try get, but should you even do a try get, wouldn't this being null be a bug, in which case it should just be a direct access which would throw when it doesn't find the key.

This method is very fragile, would it even work properly after the making the names unique ?

sources/editor/Stride.Assets.Presentation/Templates/ModelFromFileTemplateGenerator.cs
```diff
@@ -154,11 +236,195 @@ protected override IEnumerable<AssetItem> CreateAssets(AssetTemplateGeneratorPar
...
+        
+            if(materialIndices?.Count()< 1)
```
Eideren: `Count()` is the Linq method, use the property `Count` instead. Also, this will not return if `materialIndices` is null, which would cause the execution to throw at line 267 in that case.

Eideren:
Throws in `ResetMaterialsOnPrefabItems` when importing into a folder, that method definitely required a second cleanup pass anyway.
How about you replace those strings and pass references or ids around, this goes for other methods too. So many string manipulation and lookups, it's no wonder the whole thing throws if we so much as look at it.

noa7:
No more string manipulations by storing source mesh name variable in ModelAsset, checked in latest @Eideren

Eideren:
Most of my comments have not been addressed, moving this to draft while you take care of those. Feel free to set it to ready once it is actually ready for review.

noa7:
> Most of my comments have not been addressed, moving this to draft while you take care of those. Feel free to set it to ready once it is actually ready for review.

Can you tell which ones specifically, I gone through all and made  necessary cahnges!

Eideren:
Those listed below the following message https://github.com/stride3d/stride/pull/2898#issuecomment-3323404313

noa7:
> Those listed below the following message [#2898 (comment)](https://github.com/stride3d/stride/pull/2898#issuecomment-3323404313)

Its importing fine for me in a folder on latest and have removed strig manipulations?

Eideren:
Yes, I got that, but you haven't gone through the review I made below the message I linked here https://github.com/stride3d/stride/pull/2898#issuecomment-3323404313
This stuff:
image

noa7:
> Yes, I got that, but you haven't gone through the review I made below the message I linked here [#2898 (comment)](https://github.com/stride3d/stride/pull/2898#issuecomment-3323404313) This stuff: 

I really can't see that list my side of github :)

image

Eideren:
Ah, okay, my bad, that's on me. Give me a sec.

sources/editor/Stride.Assets.Presentation/Templates/ModelFromFileTemplateGenerator.cs
```diff
@@ -154,11 +236,189 @@ protected override IEnumerable<AssetItem> CreateAssets(AssetTemplateGeneratorPar
...
         }
+
+        private static void ResetMaterialsOnPrefabItems(AssetItem assetItem,  EntityInfo entityInfo)
+        {
+            ModelAsset asset = assetItem?.Asset as ModelAsset;
+            if (asset == null || asset.Materials==null)
+                return;
+
+            var underlyingModel=entityInfo.Models.Where(C=>C.MeshName==asset.MeshName).FirstOrDefault();          
+            var nodeContainingMesh=entityInfo.Nodes.Where(c=>c.Name== underlyingModel.NodeName).FirstOrDefault();
```
Eideren: `Where(x).FirstOrDefault()` can be simplified to `FirstOrDefault(x)`, but if `nodeContainingMesh` or `underlyingModel` are null, the line below would throw, so best use `First(x)`.

sources/editor/Stride.Assets.Presentation/Templates/ModelFromFileTemplateGenerator.cs
```diff
@@ -154,11 +236,189 @@ protected override IEnumerable<AssetItem> CreateAssets(AssetTemplateGeneratorPar
...
+            for (int i = 0; i < asset.Materials.Count; i++)
+            {
+                if (materialIndices.Contains(i))
+                {
+                    (materialsToApply??=new List<ModelMaterial>()).Add(asset.Materials[i]);
+                }
+            }
+
+            asset.Materials.Clear();
+            materialsToApply?.ForEach(_mat => asset.Materials.Add(_mat));
```
Eideren: This is more complicated than necessary, do reverse for loop and remove from `asset.Materials` when `materialIndices.Contains(i) == false` instead

Eideren (commented):
Well, I did mention `ResetMaterialsOnPrefabItems` requiring a cleanup, it hasn't changed much, I'll leave these few comments for now

sources/editor/Stride.Assets.Presentation/Templates/ModelFromFileTemplateGenerator.cs
noa7: Done!

sources/editor/Stride.Assets.Presentation/Templates/ModelFromFileTemplateGenerator.cs
noa7: Fixed

sources/editor/Stride.Assets.Presentation/Templates/ModelFromFileTemplateGenerator.cs
noa7: Done!

sources/editor/Stride.Assets.Presentation/Templates/ModelFromFileTemplateGenerator.cs
noa7: Fixed, The FBX as format doesn’t enforce unique mesh names. Adding an extra identifier simply ensures uniqueness. It should work fine, and if there is edge case realted to this not taken into account and might fail, should be easy to take care of!

sources/editor/Stride.Assets.Presentation/Templates/ModelFromFileTemplateGenerator.cs
noa7: Fixed

sources/editor/Stride.Assets.Presentation/Templates/ModelFromFileTemplateGenerator.cs
noa7: Fixed

noa7: only to account for edge case if theres is no material imported for mesh

sources/editor/Stride.Assets.Presentation/Templates/ModelFromFileTemplateGenerator.cs
noa7: Fixed

sources/editor/Stride.Assets.Presentation/Templates/ModelFromFileTemplateGenerator.cs
noa7: For every mesh in model there would always be a fbx node  and storing mapping directly at node level instead of mesh is to account for multiple materials on mesh

sources/tools/Stride.Importer.3D/MeshConverter.cs
```diff
@@ -67,6 +75,47 @@ private void ResetConversionData()
...
+                // Otherwise: remember the full output name; we will try to match a mesh by name later
+                keptMeshNameHint = name;
+            }
+            catch
+            {
+                keptMeshIndex = -1;
+                keptMeshNameHint = null;
+            }
+        }
+
```
Eideren: I can tell that this is not going to work, far too fragile, and it doesn't, as soon as a user changes the name of a model, the model changes to another or to the whole model instead.

You can add more properties to the `ModelAsset` if needed.

Eideren (commented):
Can you test your PR more thoroughly, started reviewing again and found another issue on the first thing I checked.

noa7:
Removed the filename-based mesh selection and made it an explicit part of the asset data. The chosen mesh now comes from ModelAsset.MeshName, which is passed via ModelAssetCompiler > ImportThreeDCommand to MeshConverter.KeepOnlyMeshByName. The old DecideKeptMeshIndexFromOutput logic is removed @Eideren

Eideren:
Did you test your changes ?

noa7:
Yes tested with the room model file

Eideren:
Well, not sure what you did test, but I can still repro the naming issue I mentioned

noa7:
@Eideren can you point which naming issue?

Eideren:
The one I mentioned there as I said https://github.com/stride3d/stride/pull/2898#discussion_r2392252314 - Changing the name of a model asset in the editor changes the actual mesh that gets imported

noa7:
My mistake, overlooked that part. I’ve checked in updates, storing mesh ID as a key pointer in ModelAsset, which is set only once at time of creation and remains unchanged on renaming etc. @Eideren

Eideren:
If I select this option from the context menu, the model has a bunch of new entries created for its materials instead of just the ones used
image

noa7:
Fixed. It still reimports the mesh upon update from source but will not add extra materials @Eideren

Eideren:
If a model has multiple materials assigned it ends up being split across two models
[Untitled.zip](https://github.com/user-attachments/files/22901074/Untitled.zip) - assigned a different material for one of the cushion on the bed.
See what I mean by saying you should test your PR more thoroughly, we shouldn't be having this much back and forth.

noa7:
Can you tell which mesh to look for in structure? And you mean if material has multiple materials its only applying first one?

Eideren:
There are two cushions on the bed, one of them is of the same material as the bed, while the other has a different one assigned to it, both of them are part of the bed model in the original fbx file.
When imported as a split, the bed gets separated into two model asset, one for the bed and cushion of material A and another with only the other cushion with material B

noa7:
Fixed material mapping mismatch between Stride and Assimp.
Stride importer has one material per mesh, while Assimp exports multiple submeshes for multi-material meshes is what makes this tricky. Updated the mesh conversion logic to merge these correctly so materials now display correctly. @Eideren

Eideren:
There's another fairly obvious issue you've introduced with the latest changes, didn't even have to do anything this time around, import your model and test things out properly.

noa7:
Just tested, its loading fine for me, can you tell whats the problem? @Eideren 
image

Eideren:
It's visible on your screen, half of the furniture is not on the right spot in the prefab

noa7:
I see, its going in circle, if I try to match exact materials than meshs dont position correctly if set mesh then materials dont set up correctly, ideally classes should have set up keeping multiple material in mind, i can try some more but not sure. If its acceptable to have only one material material imported like it is now, it will require only more additional stepon users part to assign the extra material on mesh in user interface?

Eideren:
If it's not a bug but a limitation in the design, sure, but that's not really the case here. You'll figure it out I'm sure !

noa7:
Think I got it working, here is side by side of stride import vs blender import @Eideren 

stride-import
blender-import

Eideren:
Looks like it's almost ready for me to review the code, just have to fix two edge case I found:
- Importing a model as *not* split adds a `(All)` suffix to the asset name.
- Importing a model, then assigning a new material to the file, saving it, going back to the game studio and doing right click > `Update Selected Assets From Source`, does not create a new entry under the list of materials in the model asset.

You can use this new file to test the second issue: [Untitled.zip](https://github.com/user-attachments/files/23290808/Untitled.zip)
Import the version you already have, make a backup of that file, then overwrite it with the one above and `Update Selected Assets From Source`. `Untitled-Object_0_1` should go from one to five material entries.

noa7:
> Looks like it's almost ready for me to review the code, just have to fix two edge case I found:
> 
> * Importing a model as _not_ split adds a `(All)` suffix to the asset name.
> * Importing a model, then assigning a new material to the file, saving it, going back to the game studio and doing right click > `Update Selected Assets From Source`, does not create a new entry under the list of materials in the model asset.
> 
> You can use this new file to test the second issue: [Untitled.zip](https://github.com/user-attachments/files/23290808/Untitled.zip) Import the version you already have, make a backup of that file, then overwrite it with the one above and `Update Selected Assets From Source`. `Untitled-Object_0_1` should go from one to five material entries.

- Removed the suffix "All".

The problem with second issue is update pipeline and the initial creation pipeline are different. Even in existing import system (apart from split-hierarchy), “Update from Source” only refreshes the existing model asset — it does not pass through ModelFromFileTemplateGenerator, so no new material assets are generated on update. The initial import pipeline uses ModelFromFileTemplateGenerator, which handles material creation and asset registration.

Changing this behavior may take a larger architectural update to unify the import and update logic. May be for now, we can keep the current behavior as-is and create a separate task or future update for this. Please advice. @Eideren

Eideren:
>  so no new material assets are generated on update

Yep, I did notice that it doesn't create a *new material asset*, but it still creates a *new material entry* on *the model asset*, which is why I'm not asking you to support the former, but only the latter.

noa7:
Just so I understand, by entry you mean additional empty or default
material slots on model one for each material in source mesh ?

Eideren:
The entries highlighted here
image

noa7:
Its the same reason, the sub model assets with split hierarchy are not real items with independent source makes them different, the material slots are set in ModelFromFileTemplateGenerator. May be I can find something to show up materials and make it work but not sure how far Updating such virtual assets individually this way will work

noa7:
Checked in update to best accommodate this, it should show empty slots based on asset but I am not sure how much it would hold as updating material this way can lead to potential problems. The structure of imported file after changing materials is different in terms of how mesh and fbx nodes are connected to sync back with data original file in terms of matching which mesh and material to current comes with all sorts of problems to account all ways it changed, naming, order, hierarchy or any other way user can tinker file in 3d model tool. Updating entire model (split hierarchy) from source works as part wise match to what exiting is not needed!

@Eideren

Eideren:
Welp, it does add new entries for new material, but it doesn't remove ones that are not part of the model anymore

noa7:
I'm not sure what more can be done here, updating hierarchical import from source once imported is not a right and will not work, it has to be done manually

xen2:

Issue #2895, opened 2025-09-16
[Launcher] Make .NET version with white colour


VaclavElias:
**Release Type**: Official Release

**Version**: Latest

**Platform(s)**: Windows

**Describe the bug**

1. The visibility of .NET version is bad, almost invisible, it should be white or lighter colour
2. It isn't .NET Core anymore, it should be just **.NET 10** or **.NET 8**

<img width="958" height="645" alt="Image" src="https://github.com/user-attachments/assets/88f5a516-4f45-4bfc-a7fa-bb224c29c374" />

**To Reproduce**
1. Launch the launcher

**Expected behavior**
The text should be more visible. It is good to advertise that we are on .NET 8 or .NET 10.

**Screenshots**
Above.

**Log and callstacks**
N/a

**Additional context**
N/a

AIFSKATE:
I will try to solve this.

Kryptos-FR:
It's only like that because there is a single version and hence the drop down is disabled.

PR #2899, opened 2025-09-18
chore: Improve launcher framework combobox (#2895)

AIFSKATE:
# PR Details

1. The reason this line is barely visible is because it's the default setting when IsEnabled is set to false. The easiest way to fix this is to change the check condition from 1 to 0.
2. I changed the version description from .NET Core XX to .NET XX. And because the issue says "It isn't .NET Core anymore, it should be just .NET 10 or .NET 8," I also deleted an if branch.

image

## Related Issue

#2895 

## Types of changes

## Checklist


AIFSKATE:
> @AIFSKATE please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> ```
> 
> > Options:
> > 
> > 
> > ```
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 
> > ```
> > ```
> 

@dotnet-policy-service agree

VaclavElias:
Thanks @AIFSKATE , visually it looks definitelly better. I will let others to review your updates.

MsEpsilon (approved):

Eideren:
@VaclavElias was your intent to have the combo box be interactible when there is just one framework to choose from, or keep it non-interactible but slightly more visible than it currently is ? I think keeping it non-interactive is better as it conveys to the user that there is nothing for them to change there. Otherwise users might press on the combo box just to check if there is anything else they could set it to when that box doesn't have anything else but the default value.

VaclavElias:
I wanted only better visibility which is very good to promote .NET version. See the current state. If the combo box isn't needed that would be even better, but that would be better to solve in a separate issue. Which version even allowed to select multiple frameworks?
 
image

Eideren:
> Which version even allowed to select multiple frameworks?

Pre net6.0 I think ? The launcher should support every engine version as long as feasible, so best leave that combo box in and have it slightly more opaque if possible.

VaclavElias:
I see, so the reason it wasn't visible was because there was only 1 item. What about, if there is only 1 item, it should be only a text?

sources/launcher/Stride.Launcher/ViewModels/FrameworkConverter.cs
```diff
@@ -13,10 +13,8 @@ public override object Convert(object value, [NotNull] Type targetType, object p
             var frameworkFolder = (string)value;
 
             var framework = NuGetFramework.ParseFolder(frameworkFolder);
-            if (framework.Framework == ".NETFramework")
-                return $".NET {framework.Version.ToString(3)}";
-            else if (framework.Framework == ".NETCoreApp")
-                return $".NET Core {framework.Version.ToString(2)}";
+            if (framework.Framework == ".NETCoreApp")
+                return $".NET {framework.Version.ToString(2)}";
```
Kryptos-FR: This is required for older versions that supported multiple frameworks.

sources/launcher/Stride.Launcher/Views/LauncherWindow.xaml
```diff
@@ -487,7 +487,7 @@
                       </i:Interaction.Behaviors>
                     </Button>
                   </Grid>
-                  <ComboBox Margin="0,4" Name="FrameworkSelector" IsEnabled="{Binding ActiveVersion.Frameworks.Count, Converter={sskk:Chained {sskk:IsGreater}, Parameter1={sskk:Double 1}}}" ItemsSource="{Binding ActiveVersion.Frameworks}" SelectedItem="{Binding ActiveVersion.SelectedFramework}" SelectionChanged="FrameworkChanged">
+                  <ComboBox Margin="0,4" Name="FrameworkSelector" IsEnabled="{Binding ActiveVersion.Frameworks.Count, Converter={sskk:Chained {sskk:IsGreater}, Parameter1={sskk:Double 0}}}" ItemsSource="{Binding ActiveVersion.Frameworks}" SelectedItem="{Binding ActiveVersion.SelectedFramework}" SelectionChanged="FrameworkChanged">
```
Kryptos-FR, line 490: That's kind of a useless check since it can never be negative. The reason it was greyed out was because there is no need to select anything when only one framework is available.

I'd rather not display anything in that case, i.e. change `IsEnabled` to `IsVisible`.

Kryptos-FR (changes requested):

sources/launcher/Stride.Launcher/Views/LauncherWindow.xaml
VaclavElias: What if we display plain text (in white) when there's only one item? It would still be helpful to show the .NET version so that those unfamiliar with Stride can see it.

The behavior would remain the same when multiple items are available, but if there's only one item, the combo box wouldn't be displayed, just plain white text instead.

Kryptos-FR: That's also a good option.

xen2:

Issue #2154, opened 2024-02-10
Stride's new library projects target uap10.0.16299 which causes build issues until manually fixed


SeleDreams:
**Release Type**: Official Release

**Version**: 4.2.0.2067

**Platform(s)**: Windows

**Describe the bug**
When a code library project is generated via the stride editor, the project will have uap10.0.16299 as one of its targets causing build issues.

**To Reproduce**
1. Open Stride
2. Create a code library project

**Expected behavior**
The generated project should only target dotnet8

**Screenshots**
![image](https://github.com/stride3d/stride/assets/16335601/1cb023a0-6fde-4fb8-857c-371671d42d06)

Kryptos-FR:
To add more details, the template has this line:

```
<#@ include file="..\Common.TargetFrameworks.targets.t4" #>
```

Which in turns resolves to:

```csproj
<TargetFrameworks><#= string.Join(";", ((IEnumerable<Stride.Core.Assets.SolutionPlatform>)Properties.Platforms).Select(x => x.TargetFramework).Distinct()) #></TargetFrameworks>
```

I feel like a library project (which also applies to the Game project generated by the full template) should only reference `net8.0` because it is supposed to be cross-platform. If users ever need specific stuff (like for Android or iOS) they can add these additional targets manually later.

Also the `Properties.Platforms collection doesn't seem to be up to date as it is still trying to add the `uap` targets which don't exist anymore on our nuget packages.

PR #2902, opened 2025-09-21
fix: Adjust target frameworks to only include available targets when creating new code library

Acissathar:
# PR Details

Currently trying to add a new code library project via the stride editor, results in an error:

image

Rather than try to build everything, only do what's currently available. Adding a new library now compiles as expected with no errors without having UWP or Android workloads for example.

## Related Issue

#2154

## Types of changes

## Checklist


Acissathar:
@dotnet-policy-service agree

Eideren:
Thanks !

PR #2914, opened 2025-09-30
fix: Instantiate() behavior for Prefab and Entity references

Eideren:
# PR Details
Prefab instantiation clones the prefab itself and returns only its content, changed it so that prefab clones its list of entities and returns that. 
There doesn't seem to be that much of a difference initially, but let's dive deeper:

When cloning an object through `EntityCloner`, the engine, for a given property it will do one of two things
- Either create a whole new object based on the source object of that property
- or just pass the reference of the source to the clone.
It does so depending on the type of the object and some other misc logic.

For objects that are fully cloned, like our Prefab, the system keeps track of the newly created object to swap any other references it finds in component properties that point to the source object by our newly created clone of that object.

Meaning that if the prefab contains a component that has a prefab property pointing to that same prefab, after cloning, the property would now point to the newly created clone of the prefab instead of the original prefab. 
This may seem like expected behavior, until you consider the fact that `Instantiate()` does not return a new prefab, but a list of entities instead.

### Entity References
The other part of this PR is to ensure that references in components to `Entity` that lay outside the hierarchy are not cloned. Instantiation is about cloning a hierarchy under an entity, not every entity ever referenced in the components of said entity.

I would have liked to include `EntityComponent` references too, but I couldn't fit them in yet, it requires a larger amount of change I'm not as confident wouldn't break users game, but at least this time around it will throw an exception in that case instead of duplicating the reference.
The reason it's not as simple is because `EntityComponent` do not have a specific serializer, we would need one that derive from `DataContentSerializerWithReuse`, the serializer for `TransformComponent` derives from a simple `DataSerializer` for example.

Also refactored a test which used a goto for no good reason. Don't get me wrong, I like gotos, but using it here just doesn't make sense.

## Related Issue
Couldn't find any.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Engine/Engine/Design/EntityCloner.cs
+++ b/sources/engine/Stride.Engine/Engine/Design/EntityCloner.cs
@@ -48,7 +48,7 @@ namespace Stride.Engine.Design
         }
 
         /// <summary>
-        /// Instantiate the content of the prefab provided.
+        /// Instantiates the content of the prefab provided.
         /// <see cref="Prefab.Entities"/>, children <see cref="Entity"/> and their <see cref="EntityComponent"/> will be cloned.
         /// Other assets will be shared.
         /// </summary>


Eideren:
Another nasty issue I just found which would be fixed by this PR, when your prefabs contain a circular reference like in the following case:
```cs
Prefab#0:
	Entity#0:
		Component#0:
			MyPrefab: ref to Prefab#0
```
Instantiating this prefab returns the following hierarchy
```cs
Entity#1:
	Component#1:
		MyPrefab: ref to Prefab#1

```
`Prefab#1` is a new instance of the same prefab like outlined in the main post - and its content is ...
```cs
Prefab#1:
	Entity#1:
		...
```
Surprise ! The very same entity that was just instantiated and is now part of the scene !
Meaning that if you now instantiate Prefab#1 you are actually cloning the object that is part of the scene, with its component and properties altered. 
Its state will differ compared to spawning Prefab#0 even though they both came from the same prefab.

VaclavElias:
How do you even find these issues? 🙂

Eideren:
That's my secret ;)

Kryptos-FR:
I'll have to check manually for that one. Reading from my phone doesn't compute in my head.

Just to be sure, this doesn't break nested prefabs and archetypes (at design time)? Or copy/pasting a prefab in the asset explorer?

Eideren:
@Kryptos-FR Have a look at the new test I setup, might make more sense to read through that than my nonsense :)

Eideren:
> this doesn't break nested prefabs and archetypes (at design time)? Or copy/pasting a prefab in the asset explorer?

Forgot to answer, nope, it does not :P

Do you have time to take a quick look at this one ?

Kryptos-FR:
I'll try early next week, it's late today.

sources/engine/Stride.Engine/Engine/Design/EntityCloner.cs
```diff
@@ -51,7 +54,7 @@ private static HashSet<object> ClonedObjects()
         /// </summary>
         /// <param name="prefab">The prefab to clone.</param>
         /// <returns>A cloned prefab</returns>
-        public static Prefab Clone(Prefab prefab)
+        public static List<Entity> Instantiate(Prefab prefab)
```
Kryptos-FR, line 57: Maybe update the documentation since the method name has changed.

sources/engine/Stride.Engine/Engine/Design/EntityCloner.cs
```diff
@@ -132,12 +136,17 @@ private static T Clone<T>(HashSet<object> clonedObjects, TryGetValueFunction<obj
                 cloneSerializerSelector = new SerializerSelector(true, false, "Default", "Clone");
             }
 
+            if (entitySerializerSelector == null)
```
Kryptos-FR: nit: simplify with the `??=` operator.

sources/engine/Stride.Engine.Tests/TestEntity.cs
```diff
@@ -270,4 +289,12 @@ public abstract class CustomEntityComponentBase : EntityComponent
     {
         public Action<CustomEntityComponentEventArgs> Changed;
     }
+
+    [DataContract]
+    public class EntityComponentWithPrefab : EntityComponent
+    {
+        public required Prefab Prefab { get; set; }
+        public required Entity ExternalEntityRef { get; set; }
+        /*public required TransformComponent ExternalComponentRef { get; set; }*/
```
Kryptos-FR, line 312: Add a comment around the commented out code to explain why it is kept around (maybe reference the issue/PR id as well).

Kryptos-FR (commented):
Looks ok to me.

However I discovered a regression in the editor regarding copy/pasting nested prefabs. That regression doesn't seem related to this fix (I could reproduce it with earlier version), but it didn't occur in Stride 4.0. I'll investigate more to find where the regression was introduced and try to apply your fix to a version without the regression to figure out if there are other potential regressions hidden by that one.

Eideren:
Took care of the review - I think we can merge this fix regardless of the regression you mentioned ? If this specific change ends up compromising that regression we can always modify it through the PR for that regression ?

Kryptos-FR:
Yes, it looks the bug is related to copy/paste processors and unlikely to be affected by this change.

sources/engine/Stride.Engine/Engine/Design/EntityCloner.cs
```diff
@@ -48,12 +48,12 @@ private static HashSet<object> ClonedObjects()
         }
 
         /// <summary>
-        /// Clones the specified prefab.
-        /// <see cref="Entity"/>, children <see cref="Entity"/> and their <see cref="EntityComponent"/> will be cloned.
+        /// Instantiate the content of the prefab provided.
```
Kryptos-FR: nit: `Instantiate**s**` method comments are in third-person by convention

Kryptos-FR (approved):
Aside from the latest nit-picking, looks good to me.

Eideren:
Done, thanks for the review @Kryptos-FR !

Kryptos-FR:
I have created #2953 to track the regression I mentioned earlier (which is not caused by this PR).

Issue #2903, opened 2025-09-23
GameStudio crashes when multi-selecting entities in specific scene (ArgumentNullException in IsPrimitiveType / Nullable.GetUnderlyingType)


dawnmichal:
**Release Type**: Official Release

**Version**: 4.2.1.2442

**Platform(s)**: Windows 11

**Describe the bug**
GameStudio crashes when I select multiple entities at once in the scene editor/hierarchy. Selecting the same entities one by one does not crash. The crash appears related to the property/inspector UI trying to determine a primitive type for a null Type/nullable value.

**Expected behavior**
Inspector shows a combined selection (or disables mismatched properties) without crashing.

**Log and callstacks**
[CrashReport.txt](https://github.com/user-attachments/files/22502374/CrashReport.txt)

**Additional context**
- This does not occur when selecting entities individually.
- It seems to depend on the specific scene/content. Likely triggered when one of the selected entities has a component/property of type System.Type or a Nullable<T> that is currently null, and the inspector checks IsPrimitiveType without a null guard.
- Unrelated build errors about missing audio assets were present in the log (Audio/Vehicles/Engine01, Audio/Vehicles/skid), but they don’t appear to cause the crash.

Acissathar:
I can recreate this pretty consistently with a New Game template. Specifically, it seems to happen when multi-selecting the materials, as the Tessellation node property has a value of null. I am unable to recreate it with any other object/asset/entity types though.

A simple if type is null return false seems to solve the issue, but given its very specifically (for me at least) this Tessellation node property and I'm not experienced enough with this code base to know what ramifications might be, I wasn't sure if that was worth diving into more rather than just the general null guard bandaid.

Notably, running that guard locally seems to solve the problem though. It still shows the property grid as expected and doesn't seem to cause any further errors.

I don't mind throwing it into a quick PR, but thought I'd make a comment first.

luca-domenichini:
Same here for me! Using previous released version is OK 4.2.0.2381

Acissathar:
I believe this issue can be closed as fixed

PR #2915, opened 2025-09-30
fix: Add null check for IsPrimitiveType to prevent crash when selecting multiple entities in GameStudio

Acissathar:
# PR Details

Attempting to select multiple entities may crash GameStudio due to an ArgumentNullException during the IsPrimitiveType check, as type is sometimes null. Selecting the entities individually does not appear to cause the issue.

I can recreate this pretty consistently with a New Game template specifically when multi-selecting the materials, as the Tessellation node property has a value of null. I am otherwise unable to recreate it with any other object/asset/entity types.

Given its specifically (for me at least) this Tessellation node property, I'm not entirely sure if this is more of a band-aid that is hiding something that should be dived into (like why is Tessellation node property null at this point) or not.

Running that guard locally seems to solve the problem though, and it still shows the property grid as expected.

## Related Issue

#2903 

## Types of changes

## Checklist

### DIFF
--- a/sources/assets/Stride.Core.Assets.Quantum/AssetNodeContainer.cs
+++ b/sources/assets/Stride.Core.Assets.Quantum/AssetNodeContainer.cs
@@ -48,6 +48,9 @@ public class AssetNodeContainer : NodeContainer, IPrimitiveTypeFilter
     
     public virtual bool IsPrimitiveType(Type type)
     {
+        if (type is null)
+            return false;
+
         if (Nullable.GetUnderlyingType(type) is { } underlyingType)
             type = underlyingType;
 


Acissathar:
As a follow up, I was finally able to recreate it with other entity types, such as Skeletons and Models, so I don't think it's specific to just Tessellation which makes me feel a little better about the more blanket return false.

Eideren:
Given that this method does not specify that it accepts null, it should be fixed at the callsite of the method instead, and handled appropriately there.
The log specifies that it throws at 
https://github.com/stride3d/stride/blob/ce72d512b5f9346f9fd2fc9efb64f6df538e43b2/sources/presentation/Stride.Core.Presentation.Quantum/ViewModels/NodeViewModel.cs#L355

Eideren:
Looks good to me but not really my area - pinging @Kryptos-FR if he has time take a quick look ?

sources/presentation/Stride.Core.Presentation.Quantum/ViewModels/NodeViewModel.cs
```diff
@@ -352,7 +352,7 @@ protected virtual object GetNodeValue()
             {
                 currentValue = nodePresenter.Value;
             }
-            else if (nodePresenter.Factory.IsPrimitiveType(nodePresenter.Value?.GetType()))
+            else if (nodePresenter.Value != null && nodePresenter.Factory.IsPrimitiveType(nodePresenter.Value.GetType()))
```
Kryptos-FR: Good catch. I have a different fix in my ` xplat-editor` branch (see 70b50bbcc2c8276ef8e0be1d9d7d3be20c80a794):
```suggestion
            else if (nodePresenter.Factory.IsPrimitiveType(nodePresenter.Value?.GetType() ?? nodePresenter.Type))
```

Can you check if that also solves the issue (and don't introduce a different regression)? I'd prefer to have that version if it also works as it seems more correct to me.

To be honest, I think both fixes are equivalent. I can't think of a case where that would make a difference. But who knows?

Acissathar: @Kryptos-FR I ran it through the same tests I did of the original and this change handles them. I also ran it through my experimental yes to all prompt, which previously would also duplicate the issue, and this change handles it as well.

Granted for context my test is largely just continually ctrl + click selecting assets in the asset view along with tabbing to shift focus, as that was the only way I was able to replicate it (outside my experimental feature which did it every time). 

Not sure if it was recreated other ways, but that's also how the original issue was written up, so I think this handles the issue.

Kryptos-FR (approved):

Eideren:
Thanks !

PR #2930, opened 2025-10-13
feat: Bepu, getter to retrieve the impulse and force applied by a given constraint

Eideren:
# PR Details
See title. Could be used to detach a constraint when under too much strain, dynamically measure a constraint to adjust other constraint strengths, play different straining sound effect based on those values, ...

## Related Issue
None.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.BepuPhysics/Stride.BepuPhysics/Constraints/_ConstraintComponentBase.cs
+++ b/sources/engine/Stride.BepuPhysics/Stride.BepuPhysics/Constraints/_ConstraintComponentBase.cs
@@ -53,6 +53,23 @@ public abstract class ConstraintComponentBase : EntityComponent
     /// <remarks> May not be attached if it is not in a scene, when not <see cref="Enabled"/>, when any of its target is null, not in a scene or in a different simulation </remarks>
     public abstract bool Attached { get; }
 
+    /// <summary>
+    /// Returns the squared sum of all impulses this constraint applied on the last tick
+    /// </summary>
+    /// <remarks>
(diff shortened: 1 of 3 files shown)


sources/engine/Stride.BepuPhysics/Stride.BepuPhysics/Constraints/_ConstraintComponentBase.cs
```diff
@@ -53,6 +53,23 @@ protected BodyComponent? this[int i]
     /// <remarks> May not be attached if it is not in a scene, when not <see cref="Enabled"/>, when any of its target is null, not in a scene or in a different simulation </remarks>
     public abstract bool Attached { get; }
 
+    /// <summary>
+    /// Returns the squared sum of all impulses this constraint applied on the last tick
```
Ethereal77, line 57: Just a nit: The doc says "`Returns the squared sum of all...`", but your implementation above does a `MathF.Sqrt()`. Is that correct?

Eideren: Good catch, I intended to replicate bepu's api on that front, but ended up forcing a sqrt as I did not see any cases were one would prefer a squared version. I'll make a PR in a bit to fix this.

Ethereal77: Normally squared versions are nice for checking relative magnitudes (i.e. checking if not zero, or comparing `A > B`, or comparing against a precomputed threshold). In those cases, it is sufficient to get the squared magnitude and save a square root.
Anyway, I'm not very into physics simulations these days, so you may know better 😀.

PR #2945, opened 2025-10-23
refactor: Use CollectionsMarshal.SetCount to resize lists

azeno:
# PR Details

Makes code easier to read (as it was before a122c71dc2fcfcd1d46c367c14a92471f0f57ecf). The [SetCount ](https://github.com/dotnet/runtime/blob/1d1bf92fcf43aa6981804dc53c5174445069c9e4/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/CollectionsMarshal.cs#L85) implementation looks very similar to the previously used Resize method of FastList. For value types it just sets the internal count, for reference types it clears the pointers. Both what was done before as well.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Engine/Rendering/Compositing/ForwardRenderer.cs
+++ b/sources/engine/Stride.Engine/Rendering/Compositing/ForwardRenderer.cs
@@ -471,16 +471,7 @@ namespace Stride.Rendering.Compositing
         private void ResolveMSAA(RenderDrawContext drawContext)
         {
             // Resolve render targets
-            if (currentRenderTargetsNonMSAA.Count < currentRenderTargets.Count)
-            {
-                currentRenderTargetsNonMSAA.EnsureCapacity(currentRenderTargets.Count);
-                while (currentRenderTargetsNonMSAA.Count != currentRenderTargets.Count)
-                    currentRenderTargetsNonMSAA.Add(null);
-            }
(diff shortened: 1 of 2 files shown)


Kryptos-FR:
Does it zero the new data? If not that's a behavioral change. I can only assume that the code that was explicitly clearing the data was here for a reason.

And if it is not needed, it should be documented in the code why.

Ethereal77:
It seems that if it is going to shrink, and the type is a reference type or contains references, it clears. [You can see it here](https://github.com/dotnet/runtime/blob/1d1bf92fcf43aa6981804dc53c5174445069c9e4/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/CollectionsMarshal.cs#L98-L101).

If it needs to grow the list, it does a `new T[]` (which always is zeroed out), then a copy. [Here](https://github.com/dotnet/runtime/blob/1e09fc169a2c4d0c54c483967b845b03d11215d5/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/List.cs#L111-L116).

For value types that don't contain references, it does nothing to them (so they can expose stale data).

Kryptos-FR:
On second read it looks fine. Thanks.

azeno:
What @Ethereal77 said - for value types stale data could be observed. But only the third case (RenderSystem) in this commit is about value types and when we read a few lines further down each entry in the re-sized list gets written (no read).

Eideren:
Thanks !

Issue #2944, opened 2025-10-22
VirtualButton.Mouse does not have MouseWheelDelta


Acissathar:
**Release Type**: Official Release/GitHub

**Version**: 4.2.1.2442/master

**Platform(s)**: All

**Describe the bug**
VirtualButton.Mouse currently supports all mouse fields (left, right, middle click, X and Y position delta, extended buttons) other than the MouseWheelDelta.

I could not find anything on github or discord on why this was done, so it may just be an oversight? You can manually grab the MouseWheelDelta through the input manager, but similar to the X and Y position deltas in the VirtualButton, I feel MouseWheelDelta can also be passed through.

**To Reproduce**
N/A

**Expected behavior**
MouseWheelDelta is reachable through the Virtual Buttons.

**Screenshots**
N/A

**Log and callstacks**
VirtualButton.Mouse.cs in Stride.Input

**Additional context**
This seems to really only need a couple lines, one adding the field with the constructor for the name and ID, and then another in the switch statement to grab the delta from the InputManager. Locally this seems to work just fine in a little test project, but thought I'd make an issue for visibility incase there is some context on why this was done (and if so, so that its at least put out there)

Also not entirely sure if "bug" is more so accurate than "enhancement" as I feel it kind of fits in both camps.

PR #2946, opened 2025-10-23
fix: Add mouse wheel delta to virtual button

Acissathar:
# PR Details

This is sort of a mix of a feat/fix given that no one seems to have really mentioned it (github or discord), so not sure what to really mark it as. Currently, Mouse WheelDelta can only be grabbed directly from the InputManager rather than passed through VirtualButton like all other Mouse fields. 

This change adds Mouse WheelDelta as a new field on the VirtualButton.Mouse, so that you can use the full range of the mouse through VirtualButtons.

Using it like so returns values just as if it were a GamePad thumbstick axis for example:
                new VirtualButtonBinding(InputActions.FIELD_CAM_ZOOM_BACKWARD, VirtualButton.Mouse.WheelDelta),

I haven't run into any issues with my local fork, albeit I'm still new to Stride so not entirely sure of _all_ scenarios worth testing beyond using it directly as a normal input.

## Related Issue

[2944](https://github.com/stride3d/stride/issues/2944)

## Types of changes

## Checklist

### DIFF
--- a/sources/editor/Stride.Core.Assets.Editor/ViewModel/AssetCollectionViewModel.cs
+++ b/sources/editor/Stride.Core.Assets.Editor/ViewModel/AssetCollectionViewModel.cs
@@ -32,7 +32,7 @@ using Stride.Core.Presentation.Interop;
 using Stride.Core.Presentation.Services;
 using Stride.Core.Translation;
 using Stride.Core.Presentation.ViewModels;
-using Stride.Core.Presentation.Windows;
+using System.Collections;
 
 namespace Stride.Core.Assets.Editor.ViewModel
 {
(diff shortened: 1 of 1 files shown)


sources/editor/Stride.Core.Assets.Editor/ViewModel/AssetCollectionViewModel.cs
Kryptos-FR, line 1: Changes on this file seem unrelated.

Acissathar: Ah good catch! Must have accidentally merged that into this branch

Kryptos-FR:
LGTM but if another maintainer or contributor can test it you before merging that would be nice. I unfortunately don't have that time.

Eideren:
Thanks !

PR #2952, opened 2025-10-26
fix: Regression in mesh bounds calculation introduced through #2858

### LINKED ISSUE #2948: The model thumbnails and asset preview are not working

**Version**: 4.2.1.2485

**Platform(s)**: Windows

**Describe the bug**
The model thumbnails are not showing up for me in the Asset View. The models render fine in 3D, in the scene, and in front of the camera… but their thumbnails are not visible, and the asset preview in the bottom right corner also doesn’t work. It worked in previous versions, and I haven’t changed the models. I also tried deleting the bin and obj folders so that the asset loading phase would run again… but without success.

johang88:
Fixes #2948

Changed ComputeBounds helper to be class instead of struct as it contained stateful data that was not retained due to copy by value semantics.

## Related Issue

https://github.com/stride3d/stride/issues/2948

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Graphics/VertexBufferHelper.cs
+++ b/sources/engine/Stride.Graphics/VertexBufferHelper.cs
@@ -241,7 +241,7 @@ public readonly struct VertexBufferHelper
         IConverter<Color, TDest>,
         ISemantic 
         where TDest : unmanaged
-        where TReader : IReader<TDest>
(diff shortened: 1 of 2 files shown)


Eideren:
Thanks for that, I meant to swap these to allow ref structs but I got sidetracked and never got back to it it seems.

Issue #2942, opened 2025-10-22
Script Prioritization has significant overhead


Eideren:
**Release Type**: Local release build

**Version**: Latest as of writing

**Platform(s)**: Any

**Describe the bug**
`Scheduler.Run()` is called every frame to update both sync and async scripts.
For example, all SyncScripts are shoved into that object every frame from here:
https://github.com/stride3d/stride/blob/04dc30647a240a3a6aed582de8e8f00132fb9a82/sources/engine/Stride.Engine/Engine/Processors/ScriptSystem.cs#L95-L104

`Scheduler.Run()` contains the following line:
https://github.com/stride3d/stride/blob/04dc30647a240a3a6aed582de8e8f00132fb9a82/sources/core/Stride.Core.MicroThreading/Scheduler.cs#L148

[that method](https://github.com/stride3d/stride/blob/master/sources/core/Stride.Core/Collections/PriorityNodeQueue.cs#L165) is *definitely* not O1, its execution can take a significant amount of time, *longer than each scripts it actually returns take to run*.

Here's a profiling session showing this behavior:
https://share.firefox.dev/47iWokp
Generated with [Ultra](https://github.com/xoofx/ultra) and visualized through firefox's profiler
A screenshot of the callsite in question:
<img width="855" height="709" alt="Image" src="https://github.com/user-attachments/assets/12936424-f527-4698-8d7d-a734ab17a28c" />

**To Reproduce**
N/A.

**Expected behavior**
Doesn't take as long.

**Additional context**
We might want to swap from a priority node queue to something providing O1 access, and may want to avoid enqueuing and dequeuing syncscripts everyframe, unnecessary overhead.

PR #2986, opened 2025-11-28
fix: ScriptSystem performance

Eideren:
# PR Details

Used a non-synthetic private scene to test these changes out. The scene is comprised of around 20k `SyncScript` of different types and priorities.
Out of the box `ScriptSystem.Update` takes ~9.5ms with min/max of 9.0 and 10.4.
With this change, it takes ~4.7ms with min/max of 4.6 and 5.25. 

Profile capture: https://share.firefox.dev/49LWPGK

Previously, scheduling any script was `log N`, so total cost of `N log N`. It now is `log N` only for each unique priority scheduled. Otherwise scheduling is `O1`

The implementation uses a `Deque` to provide ordered insertion at the start, end and removal from the start.
Then we have a `PriorityQueue` which sorts these `Deque` based on their priority to consume the most prioritized queue first.
And finally a `Dictionary` for `O1` subscription to a particular `Deque`.
Then we have a couple of minor details to further amortize the cost of all of this.

## Related Issue
fix: #2942

## Types of changes

## Checklist
- [ ] I have added tests to cover my changes. - there already are tests covering these areas

### DIFF
--- a/sources/core/Stride.Core.MicroThreading/Scheduler.cs
+++ b/sources/core/Stride.Core.MicroThreading/Scheduler.cs
@@ -28,7 +28,7 @@ public class Scheduler : IDisposable
     private bool isDisposed;
     private int runRecursion;
     private readonly Lock bucketsLock = new();
-    private readonly PriorityQueue<ExecutionQueue> sortedPriorities = new(PrioritiesComparer.Shared);
+    private readonly PriorityQueue<ExecutionQueue> sortedPriorities = new(NonNullPrioritiesComparer.Shared);
     private readonly Dictionary<long, ExecutionQueue> buckets = new();
     private readonly Dictionary<long, ExecutionQueue> emptyBuckets = new();
     private readonly Stack<ExecutionQueue> bucketPool = new();
@@ -434,7 +434,7 @@ public class Scheduler : IDisposable
             if (schedulerEntry.CurrentQueue is { } deque)
             {
                 schedulerEntry.CurrentQueue = null;
-                deque.Deque.RemoveAt(deque.Deque.BinarySearch(schedulerEntry, DequeComparer.Shared));
+                deque.Deque.RemoveAt(deque.Deque.BinarySearch(schedulerEntry, new NonNullBinarySearchComparer()));
                 if (deque.Deque.Count == 0)
                 {
                     emptyBuckets.Add(deque.Priority, deque);
@@ -442,18 +442,6 @@ public class Scheduler : IDisposable
                     sortedPriorities.Remove(deque);
                 }
             }
-        }
-    }
-
-    private class DequeComparer : IComparer<SchedulerEntry>
-    {
-        public static DequeComparer Shared = new();
-        
-        public int Compare(SchedulerEntry? x, SchedulerEntry? y)
-        {
-#pragma warning disable CS8602 // Dereference of a possibly null reference.
-            return x.BinarySearchHelper.CompareTo(y.BinarySearchHelper);
-#pragma warning restore CS8602 // Dereference of a possibly null reference.
         }
     }
 
@@ -475,13 +463,27 @@ public class Scheduler : IDisposable
             return CallbackNodePool.TryPop(out var node) ? node : new MicroThreadCallbackNode();
     }
 
-    private class PrioritiesComparer : IComparer<ExecutionQueue>
+    private struct NonNullBinarySearchComparer : IComparer<SchedulerEntry>
     {
-        public static PrioritiesComparer Shared = new();
+        public int Compare(SchedulerEntry? x, SchedulerEntry? y)
+        {
+            // On purpose, nulls are not allowed in the collection
+#pragma warning disable CS8602 // Dereference of a possibly null reference.
+            return x.BinarySearchHelper.CompareTo(y.BinarySearchHelper);
+#pragma warning restore CS8602 // Dereference of a possibly null reference.
+        }
+    }
+
+    private class NonNullPrioritiesComparer : IComparer<ExecutionQueue>
+    {
+        public static NonNullPrioritiesComparer Shared = new();
         
         public int Compare(ExecutionQueue? x, ExecutionQueue? y)
         {
+            // On purpose, nulls are not allowed in the collection
+#pragma warning disable CS8602 // Dereference of a possibly null reference.
             return x.Priority.CompareTo(y.Priority);
+#pragma warning restore CS8602 // Dereference of a possibly null reference.
         }
     }
 
--- a/sources/core/Stride.Core.MicroThreading/SchedulerEntry.cs
+++ b/sources/core/Stride.Core.MicroThreading/SchedulerEntry.cs
@@ -10,11 +10,47 @@ namespace Stride.Core.MicroThreading;
 /// </summary>
 internal class SchedulerEntry
 {
-    public int BinarySearchHelper;
-    public Scheduler.ExecutionQueue? CurrentQueue;
-    public Scheduler.ExecutionQueue? PreviousQueue;
+    /// <summary>
+    /// A hint to accelerate searches in collections when inserting at the edges,
+    /// should always have a value greater than the one at n-1.
+    /// </summary>
+    internal int BinarySearchHelper;
+
+    /// <summary>
+    /// The queue this is scheduled to run on, or null if it hasn't been scheduled
+    /// </summary>
+    internal Scheduler.ExecutionQueue? CurrentQueue;
+
+    /// <summary>
+    /// The queue this was previously scheduled to run on, or null if it has never been scheduled
+    /// </summary>
+    /// <remarks>
+    /// Provides additional acceleration for entries that are frequently scheduled
+    /// </remarks>
+    internal Scheduler.ExecutionQueue? PreviousQueue;
+
+    /// <summary>
+    /// The action this entry will run
+    /// </summary>
+    /// <remarks>This is mutually exclusive with <see cref="MicroThread"/>, where <see cref="MicroThread"/> would run in its stead when set</remarks>
     public Action? Action;
+
+    /// <summary>
+    /// The microthread whose <see cref="MicroThread.Callbacks"/> this entry will run
+    /// </summary>
+    /// <remarks>This is mutually exclusive with <see cref="Action"/>, where <see cref="MicroThread"/> would run in its stead when set</remarks>
     public MicroThread? MicroThread;
+
+    /// <summary>
+    /// An object you can attach to this entry for any purpose
+    /// </summary>
     public object? Token;
+
+    /// <summary>
+    /// The profiling key to use while executing this entry
+    /// </summary>
+    /// <remarks>
+    /// <see cref="MicroThreadProfilingKeys.ProfilingKey"/> will be used when this is null
+    /// </remarks>
     public ProfilingKey? ProfilingKey;
 }
--- a/sources/core/Stride.Core/Collections/Dequeue.cs
+++ b/sources/core/Stride.Core/Collections/Dequeue.cs
@@ -169,19 +169,31 @@ public sealed class Deque<T> : IList<T>, System.Collections.IList
         return -1;
     }
 
-    public int BinarySearch(T item, IComparer<T>? comparer = null)
+    /// <summary>Perform a binary search over this <see cref="Deque{T}"/> to find <paramref name="value"/> using the default comparer.</summary>
+    /// <remarks>Result is undefined when the items in this <see cref="Deque{T}"/> are unordered.</remarks>
+    /// <returns>
+    /// The zero-based index of <paramref name="value" /> in this sorted <see cref="Deque{T}"/>, if <paramref name="value" /> is found; otherwise,
+    /// a negative number that is the bitwise complement of the index of the next element that is larger than <paramref name="value" /> or,
+    /// if there is no larger element, the bitwise complement of <see cref="Deque{T}.Count" />.
+    /// </returns>
+    public int BinarySearch(T value) => BinarySearch(value, Comparer<T>.Default);
+
+    /// <summary>Perform a binary search over this <see cref="Deque{T}"/> to find <paramref name="value"/> using <paramref name="comparer"/>.</summary>
+    /// <exception cref="T:System.ArgumentNullException">
+    /// <paramref name="comparer" /> is <see langword="null" />.</exception>
+    /// <inheritdoc cref="BinarySearch"/>
+    public int BinarySearch<TComparer>(T value, TComparer comparer) where TComparer : IComparer<T>, allows ref struct
     {
-        comparer ??= Comparer<T>.Default;
         if (WrapsAround(offset, Count, out var splitA, out var splitB))
         {
-            int i = splitA.BinarySearch(item, comparer);
+            int i = splitA.BinarySearch(value, comparer);
             if (i >= 0)
                 return i;
             if (~i < splitA.Length)
                 return i;
             
             int left = Count - (Capacity - offset);
-            i = splitB[..left].BinarySearch(item, comparer);
+            i = splitB[..left].BinarySearch(value, comparer);
             if (i >= 0)
                 return splitA.Length + i;
             
@@ -189,7 +201,7 @@ public sealed class Deque<T> : IList<T>, System.Collections.IList
         }
         else
         {
-            return splitA.BinarySearch(item, comparer);
+            return splitA.BinarySearch(value, comparer);
         }
     }
 
--- a/sources/engine/Stride.Engine/Engine/Processors/ScriptSystem.cs
+++ b/sources/engine/Stride.Engine/Engine/Processors/ScriptSystem.cs
@@ -18,7 +18,7 @@ namespace Stride.Engine.Processors
     /// </summary>
     public sealed class ScriptSystem : GameSystemBase
     {
-        internal const long UpdateBit = 1L << 32;
+        private const long UpdateBit = 1L << 32;
         internal static readonly Logger Log = GlobalLogger.GetLogger("ScriptSystem");
 
         /// <summary>


sources/core/Stride.Core.MicroThreading/SchedulerEntry.cs
```diff
@@ -8,32 +8,14 @@ namespace Stride.Core.MicroThreading;
 /// <summary>
 /// Either a microthread or an action with priority.
 /// </summary>
-internal struct SchedulerEntry : IComparable<SchedulerEntry>
+internal class SchedulerEntry
```
Ethereal77, line 11: I haven't tested this, but just out of curiosity:
How does this change from `struct` to `class` fare from the garbage/GC standpoint? In your extreme sample scene with many scripts, doesn't this result in thousands of `SchedulerEntry` instances?

sources/core/Stride.Core/Collections/Dequeue.cs
```diff
@@ -65,7 +68,12 @@ public Deque(int capacity)
     {
         if (capacity < 1)
             throw new ArgumentOutOfRangeException(nameof(capacity), "Capacity must be greater than 0.");
+
+        if ((capacity & (capacity - 1)) != 0)
```
Ethereal77: You can now do `int.IsPow2()` for this

sources/core/Stride.Core/Collections/Dequeue.cs
```diff
@@ -57,6 +58,8 @@ internal sealed class Deque<T> : IList<T>, System.Collections.IList
     /// </summary>
     private int offset;
 
+    private int mask;
```
Ethereal77, line 65: Maybe add a little comment indicating what this `mask` is for, even if basic.
Something like: `A bitmask used to constrain indices to the size of the Deque`, or something along those lines.

sources/core/Stride.Core.MicroThreading/SchedulerEntry.cs
Eideren: It does, yes. Previous implementation wrapped this struct inside of a priority node, so the memory overhead is about the same per instance between both implementation. In this new implementation, those are not pooled, so they will be GCed once the script or microthread are collected. If creating a script/microthread did not allocate otherwise, I would worry, but given that it does, here it's just another allocation on top of the base one which is not nearly as problematic.

Right now it was swapped into a class to mark it as unscheduled once it runs, which is required to skip an unnecessary lookup when unscheduling it outside of the scheduler loop. We could do without and bare one binary search on every syncscript removal, or build additional logic around this to conditionally skip it when ran after a scheduler run, but that's a bit too fragile. I'll check monday if there's anything else I could do on that side.

Eideren:
Looked into making the SchedulerEntry a struct, not sure it's worth it given that we *have* to pay the cost of an additional hashset lookup whenever we schedule, process or unschedule one. We're kind of limited in what we can setup given that the Scheduler is re-entrant and can schedule actions while processing scheduled actions.
You can see the changes required for that over here: https://github.com/stride3d/stride/pull/2986/commits/cac41bbde628852422330219bf44ce09a17a6de4
I'll move back to using a pool instead.

sources/core/Stride.Core/Collections/Dequeue.cs
Eideren: Done

sources/core/Stride.Core/Collections/Dequeue.cs
Eideren: Done

Eideren:
This PR is ready for review

xen2:
LGTM!

sources/core/Stride.Core/Collections/Dequeue.cs
```diff
@@ -157,6 +169,30 @@ public int IndexOf(T item)
         return -1;
     }
 
+    public int BinarySearch(T item, IComparer<T>? comparer = null)
```
Kryptos-FR: It would be nice to document every public API.

sources/core/Stride.Core.MicroThreading/SchedulerEntry.cs
```diff
@@ -8,32 +8,13 @@ namespace Stride.Core.MicroThreading;
...
 {
+    public int BinarySearchHelper;
```
Kryptos-FR: Document public properties and fields.

Kryptos-FR (commented):
LGTM. If possible improve the documentation on some types.

VaclavElias:
@Kryptos-FR, are you testing copilot? You would like copilot to create xml documentation?

Kryptos-FR:
> @Kryptos-FR, are you testing copilot? You would like copilot to create xml documentation?

I was trying to have it make a review but apparently I can't. Thanks for doing it.

My other review comments were directed at the original author but if copilot can do it, why not?

VaclavElias:
I do it this way, click the icon on the top right to request a review and select copilot. You can request for sure also review but can you see there aslo copilot? 
image

VaclavElias:
We need to update these instructions https://github.com/stride3d/stride/blob/master/.github/copilot-instructions.md, and include  to always suggest xml doc for public APIs if they are missing, or needs to be updated or corrected.

VaclavElias:
Copilot can't do it on existing PR at the moment.. so I am listing it below. @Eideren, use if any of those are helpful..

```
    /// <summary>
    /// Base class for scripts that are updated synchronously every frame on the main thread.
    /// </summary>
    /// <remarks>
    /// Sync scripts are batched and executed in priority order during <see cref="Stride.Engine.Processors.ScriptSystem.Update(Stride.Games.GameTime)"/>.
    /// </remarks>
    public abstract class SyncScript : StartupScript
    {
        /// <summary>
        /// Called every frame to update the script.
        /// </summary>
        public abstract void Update();

        /// <summary>
        /// Notifies the system that the script priority has changed and triggers rescheduling within the update batch.
        /// </summary>
        /// <remarks>
        /// This ensures the script moves to the correct priority bucket for synchronous updates.
        /// </remarks>
        protected internal override void PriorityUpdated();
    }
``` 

```
    /// <summary>
    /// Base class for scripts that run a startup action before their update loop begins.
    /// </summary>
    public abstract class StartupScript : ScriptComponent
    {
        /// <summary>
        /// Called before the script enters its update loop.
        /// </summary>
        /// <remarks>
        /// This method is scheduled by <see cref="Stride.Engine.Processors.ScriptSystem"/> based on the script’s priority.
        /// </remarks>
        public abstract void Start();

        /// <summary>
        /// Notifies the system that the script priority has changed and should be rescheduled if necessary.
        /// </summary>
        /// <remarks>
        /// Scripts changing their priority should call or rely on this to update scheduling buckets accordingly.
        /// </remarks>
        protected internal virtual void PriorityUpdated() { /* default implementation in base */ }
    }
```

```
    /// <summary>
    /// Manages the lifecycle and execution of scripts in a game: startup, synchronous updates, and asynchronous micro threads.
    /// </summary>
    public sealed class ScriptSystem : GameSystemBase
    {
        /// <summary>
        /// Initializes a new instance of the <see cref="ScriptSystem"/> class.
        /// </summary>
        /// <param name="registry">The service registry.</param>
        public ScriptSystem(IServiceRegistry registry);

        /// <summary>
        /// Gets the scheduler used to run micro threads and scheduled script actions.
        /// </summary>
        public Scheduler Scheduler { get; }

        /// <summary>
        /// Schedules all pending script starts, batches synchronous script updates by priority,
        /// and runs the scheduler for the current frame.
        /// </summary>
        /// <param name="gameTime">Time information for the current frame.</param>
        /// <remarks>
        /// - Startup scripts are scheduled according to their priority.
        /// - Synchronous scripts are grouped by priority and executed in batches to reduce scheduling overhead.
        /// - Asynchronous scripts run on micro threads via the scheduler.
        /// </remarks>
        public override void Update(GameTime gameTime);

        /// <summary>
        /// Returns an awaiter that completes on the next frame boundary.
        /// </summary>
        /// <returns>An awaiter to resume on the next frame.</returns>
        public ChannelMicroThreadAwaiter<int> NextFrame();

        /// <summary>
        /// Adds a new micro thread to the scheduler.
        /// </summary>
        /// <param name="microThreadFunction">The micro thread function to execute.</param>
        /// <param name="priority">Lower values run the associated micro thread sooner.</param>
        /// <returns>The created and scheduled <see cref="MicroThread"/>.</returns>
        public MicroThread AddTask(Func<Task> microThreadFunction, long priority = 0);

        /// <summary>
        /// Registers a script for execution and schedules its startup and (if applicable) asynchronous execution.
        /// </summary>
        /// <param name="script">The script component to add.</param>
        /// <remarks>
        /// - Startup scripts are scheduled to run <see cref="StartupScript.Start"/> with their priority.
        /// - Async scripts create a micro thread for <c>Execute()</c>.
        /// - Sync scripts will be batched by priority and executed during <see cref="Update"/>.
        /// </remarks>
        public void Add(ScriptComponent script);

        /// <summary>
        /// Removes a script from execution and unschedules any pending actions associated with it.
        /// </summary>
        /// <param name="script">The script component to remove.</param>
        public void Remove(ScriptComponent script);

        /// <summary>
        /// Replaces a live script instance with a new instance and flags it for live reload startup execution.
        /// </summary>
        /// <param name="oldScript">The current script instance.</param>
        /// <param name="newScript">The new script instance.</param>
        public void LiveReload(ScriptComponent oldScript, ScriptComponent newScript);

        /// <summary>
        /// A bit that indicates the “Update” scheduling band for scripts.
        /// </summary>
        /// <remarks>
        /// Script priorities can be combined with this bit (by OR) to ensure sync updates are separated from other work.
        /// </remarks>
        internal const long UpdateBit = 1L << 32;
    }
```

```
    /// <summary>
    /// Coordinates the execution of micro threads and scheduled actions according to priority.
    /// </summary>
    /// <remarks>
    /// - Uses per-priority execution queues and a priority heap to select the next work item.
    /// - Provides channel-based frame synchronization and exception propagation options.
    /// </remarks>
    public class Scheduler : IDisposable
    {
        /// <summary>
        /// Executes scheduled work until the current frame’s queues are drained, yields frame channels, and finalizes completed micro threads.
        /// </summary>
        /// <remarks>
        /// Reentrant-safe. On reentry it will avoid re-pooling unused priority buckets until the outermost call returns.
        /// </remarks>
        public void Run();

        /// <summary>
        /// Creates a new micro thread associated with this scheduler.
        /// </summary>
        /// <returns>A new <see cref="MicroThread"/> that can be started and scheduled.</returns>
        public MicroThread Create();

        /// <summary>
        /// Returns a task that completes when all provided micro threads have completed.
        /// </summary>
        /// <param name="microThreads">The micro threads to wait for completion.</param>
        /// <returns>A task that completes when all micro threads have completed.</returns>
        public Task WhenAll(params MicroThread[] microThreads);

        // Note: Public events (MicroThreadStarted/Ended/CallbackStart/CallbackEnd) should retain their existing docs or add them if missing.

        // Internal methods and types introduced by this PR are not part of the public API surface and thus don’t need public XML docs.
    }
```

```
    public readonly struct MicroThreadYieldAwaiter
    {
        /// <summary>
        /// Gets a value indicating whether the awaiter has completed.
        /// </summary>
        /// <remarks>
        /// Returns <c>true</c> if the micro thread has finished or if the scheduler currently has no entries scheduled.
        /// </remarks>
        public bool IsCompleted { get; }
    }
```

```
    public class MicroThread
    {
        /// <summary>
        /// Gets or sets the scheduling priority of this micro thread.
        /// </summary>
        /// <remarks>
        /// - Lower values run earlier; higher values run later.
        /// - Changing priority while scheduled will reschedule the micro thread at the new priority.
        /// - Priority can be combined with system-specific flags (e.g. <see cref="Stride.Engine.Processors.ScriptSystem.UpdateBit"/>).
        /// </remarks>
        public long Priority { get; set; }

        /// <summary>
        /// Starts the micro thread by scheduling its function at the specified <paramref name="scheduleMode"/>.
        /// </summary>
        /// <param name="microThreadFunction">The micro thread function to run.</param>
        /// <param name="scheduleMode">
        /// Whether to schedule at the front or back of the priority queue for the current priority band.
        /// </param>
        /// <exception cref="ArgumentNullException">Thrown when <paramref name="microThreadFunction"/> is null.</exception>
        public void Start(Func<Task> microThreadFunction, ScheduleMode scheduleMode = ScheduleMode.Last);

        /// <summary>
        /// Runs the micro thread’s queued callbacks and yields appropriately within the scheduler.
        /// </summary>
        /// <remarks>
        /// This method is intended to be called by the scheduler and will process pending continuations.
        /// </remarks>
        public Task Run();

        // Other existing public members should retain or get XML docs if missing in source (not shown here).
    }
```

```
 /// <summary>
    /// A double-ended queue (deque), which provides O(1) indexed access, O(1) removals from the front and back,
    /// amortized O(1) insertions to the front and back, and O(N) insertions and removals anywhere else (with the
    /// operations getting slower as the index approaches the middle).
    /// </summary>
    /// <typeparam name="T">The type of elements contained in the deque.</typeparam>
    /// <remarks>
    /// - Capacity is always a power of two and grows by rounding up to the next power of two.
    /// - Indices are stable and contiguous from 0 to <see cref="Count"/> - 1.
    /// - The implementation uses a circular buffer and bit-masked index arithmetic for speed.
    /// </remarks>
    public sealed class Deque<T> : IList<T>, System.Collections.IList
    {
        /// <summary>
        /// Initializes a new instance of the <see cref="Deque{T}"/> class with the specified capacity.
        /// </summary>
        /// <param name="capacity">The initial capacity. Must be a power of two greater than <c>0</c>.</param>
        /// <exception cref="ArgumentOutOfRangeException">Thrown when <paramref name="capacity"/> is less than 1.</exception>
        /// <exception cref="InvalidOperationException">Thrown when <paramref name="capacity"/> is not a power of two.</exception>
        public Deque(int capacity) { /* existing code */ }

        /// <summary>
        /// Gets or sets the total number of elements the internal data structure can hold without resizing.
        /// </summary>
        /// <remarks>
        /// - Setting <see cref="Capacity"/> to a value less than <see cref="Count"/> throws.
        /// - The value must be a power of two; otherwise an <see cref="InvalidOperationException"/> is thrown.
        /// - When increased, existing elements are preserved in logical order.
        /// </remarks>
        /// <exception cref="InvalidOperationException">Thrown if the new capacity is not a power of two or less than <see cref="Count"/>.</exception>
        public int Capacity { get; set; }

        /// <summary>
        /// Gets the number of elements contained in the <see cref="Deque{T}"/>.
        /// </summary>
        public int Count { get; private set; }

        /// <summary>
        /// Performs a binary search for <paramref name="item"/> in the deque assuming elements are sorted according to <paramref name="comparer"/>.
        /// </summary>
        /// <param name="item">The item to locate.</param>
        /// <param name="comparer">
        /// The comparer used to order elements. If <c>null</c>, <see cref="Comparer{T}.Default"/> is used.
        /// </param>
        /// <returns>
        /// The zero-based index of <paramref name="item"/> in the deque if found; otherwise, a negative number that is the bitwise complement
        /// of the index of the next element that is larger than <paramref name="item"/> or, if there is no larger element, the bitwise complement
        /// of <see cref="Count"/>.
        /// </returns>
        /// <remarks>
        /// This method functions like <see cref="Array.BinarySearch{T}(T[], T)"/> semantics adapted for the deque’s split circular buffer.
        /// Elements must already be sorted by <paramref name="comparer"/>.
        /// </remarks>
        public int BinarySearch(T item, IComparer<T>? comparer = null);

        /// <summary>
        /// Adds an item to the end of the deque.
        /// </summary>
        /// <param name="value">The element to insert.</param>
        public void AddToBack(T value);

        /// <summary>
        /// Adds an item to the front of the deque.
        /// </summary>
        /// <param name="value">The element to insert.</param>
        public void AddToFront(T value);

        /// <summary>
        /// Removes and returns the item at the end of the deque.
        /// </summary>
        /// <returns>The former last element.</returns>
        /// <exception cref="InvalidOperationException">Thrown when the deque is empty.</exception>
        public T RemoveFromBack();

        /// <summary>
        /// Removes and returns the item at the front of the deque.
        /// </summary>
        /// <returns>The former first element.</returns>
        /// <exception cref="InvalidOperationException">Thrown when the deque is empty.</exception>
        public T RemoveFromFront();

        /// <summary>
        /// Inserts the elements of a collection into the deque at the specified <paramref name="index"/>.
        /// </summary>
        /// <param name="index">The zero-based index at which the new elements should be inserted.</param>
        /// <param name="collection">The collection whose elements should be inserted into the deque.</param>
        /// <exception cref="ArgumentOutOfRangeException">Thrown when <paramref name="index"/> is less than 0 or greater than <see cref="Count"/>.</exception>
        /// <remarks>
        /// - If needed, capacity grows to the next power of two that can fit <paramref name="collection"/> and existing elements.
        /// - Insertion near ends is optimized; insertion near middle is O(N).
        /// </remarks>
        public void InsertRange(int index, IEnumerable<T> collection);

        // Note: Other IList<T> members already have typical semantics; ensure they have or retain XML docs elsewhere if exposed.
    }
```

Kryptos-FR:
> I do it this way, click the icon on the top right to request a review and select copilot. You can request for sure also review but can you see there aslo copilot?

I don't have that option. Copilot is not listed for me.

sources/core/Stride.Core/Collections/Dequeue.cs
Eideren: Done

sources/core/Stride.Core.MicroThreading/SchedulerEntry.cs
Eideren: Done

sources/core/Stride.Core.MicroThreading/Scheduler.cs
```diff
@@ -312,42 +354,142 @@ public async Task WhenAll(params MicroThread[] microThreads)
...
-            schedulerEntry.Value.SchedulerCounter = nextCounter;
+    private class DequeComparer : IComparer<SchedulerEntry>
+    {
+        public static DequeComparer Shared = new();
+        
+        public int Compare(SchedulerEntry? x, SchedulerEntry? y)
+        {
+#pragma warning disable CS8602 // Dereference of a possibly null reference.
+            return x.BinarySearchHelper.CompareTo(y.BinarySearchHelper);
+#pragma warning restore CS8602 // Dereference of a possibly null reference.
```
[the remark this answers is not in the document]
Eideren: On purpose; the collection does not allow nulls, the type is private and checking for null is unnecessary overhead.

sources/core/Stride.Core.MicroThreading/Scheduler.cs
```diff
@@ -312,42 +354,142 @@ public async Task WhenAll(params MicroThread[] microThreads)
...
+            return CallbackNodePool.TryPop(out var node) ? node : new MicroThreadCallbackNode();
+    }
+
+    private class PrioritiesComparer : IComparer<ExecutionQueue>
     {
-        lock (ScheduledEntries)
+        public static PrioritiesComparer Shared = new();
+        
+        public int Compare(ExecutionQueue? x, ExecutionQueue? y)
         {
```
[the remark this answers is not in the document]
Eideren: On purpose; the collection does not allow nulls, the type is private and checking for null is unnecessary overhead.

Eideren:
@VaclavElias Well, it somehow missed every single cases that should actually be documented aside from the binary search ... :D

VaclavElias:
I guees, that's all we can get from it today 🤣.

PR #2987, opened 2025-11-30
feat: Port Stride.Navigation native code to C# using DotRecast

Jklawreszuk:
# PR Details

PR ports C++ code to manageable .NET code using the DotRecast library. Additionally, I simplified/removed many code fragments in the module. 
The benefits I see from this changes:
- More maintainable code - entire module is now written in one language = more contributors will undertand the code
  - Also this is another step towards eliminating the need to install the MSVC compiler.
- I removed the need to the use of pointers and manual allocation/dealloc

PR was tested on the TopDownRPG sample.

## Related Issue

https://github.com/stride3d/stride/issues/1394

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Navigation/Processors/RecastNavigationMesh.cs
+++ b/sources/engine/Stride.Navigation/Processors/RecastNavigationMesh.cs
@@ -37,11 +37,7 @@ namespace Stride.Navigation.Processors
         /// <param name="coord">The tile coordinate</param>
         public bool RemoveTile(Point coord)
         {
-            if (!tileCoordinates.Contains(coord)) 
-                return false;
-
-            tileCoordinates.Remove(coord);
-            return navmesh.RemoveTile(coord);
+            return tileCoordinates.Remove(coord) && navmesh.RemoveTile(coord);
         }
 
         /// <summary>


VaclavElias:
I started a Copilot reriew, if it helps at all..

sources/engine/Stride.Navigation/NavigationBuilder.cs
```diff
@@ -0,0 +1,192 @@
...
+    private RcPolyMeshDetail meshDetail;
+    private DtMeshData navmeshData;
+    
+    public GeneratedData BuildNavmesh(ref Vector3[] vertices, ref int[] indices)
+    {
+        GeneratedData ret = result;
+        ret.Success = false;
+
+        RcVec3f bmin = new (buildSettings.BoundingBox.Minimum.X,buildSettings.BoundingBox.Minimum.Y,buildSettings.BoundingBox.Minimum.Z);
+        RcVec3f bmax = new (buildSettings.BoundingBox.Maximum.X,buildSettings.BoundingBox.Maximum.Y,buildSettings.BoundingBox.Maximum.Z);
```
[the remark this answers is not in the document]
Jklawreszuk: Fair point, I changed it.

sources/engine/Stride.Navigation/Processors/RecastNavigationMesh.cs
```diff
@@ -3,62 +3,45 @@
...
+        public void RemoveTile(Point coord)
         {
-            if (!tileCoordinates.Contains(coord))
-                return false;
+            if (!tileCoordinates.Contains(coord)) 
+                return;
 
             tileCoordinates.Remove(coord);
-            return Navigation.RemoveTile(navmesh, coord);
+            navmesh.RemoveTile(coord);
```
Eideren: Could you comment on this change in return type ?

sources/engine/Stride.Navigation/Processors/RecastNavigationMesh.cs
```diff
@@ -3,62 +3,45 @@
...
+using DotRecast.Detour;
 using Stride.Core.Mathematics;
 
 namespace Stride.Navigation.Processors
 {
     /// <summary>
     /// Recast native navigation mesh wrapper
     /// </summary>
-    public unsafe class RecastNavigationMesh : IDisposable
+    public class RecastNavigationMesh(NavigationMesh navigationMesh)
```
Eideren: Is it no longer required to dispose this type, ownership of the NavigationMesh is handler someplace else or no longer requires disposal ?

sources/engine/Stride.Navigation/Processors/RecastNavigationMesh.cs
Jklawreszuk: The logic behind this was that the consuming method (AddOrReplaceTile) did not take the return value into account in any way, so I changed it into void type.

Jklawreszuk: But now that I think about it, and considering that this is a public method, I should actually undo it.

Eideren:
Thanks !

PR #2994, opened 2025-12-04
refactor: Remove unused tools, Update Stride.PublicApiCheck

Jklawreszuk:
# PR Details

PR removes following tools:
- Stride.DebugTools - I am not fully aware of the purpose of the project. It appears to be a unfinished WPF library with references to not existing classes from 2014.
- Stride.RemoteShaderCompiler - contains a non-existent class ShaderCompilerServer that indirectly refers to the Stride removed/disabled feature for connecting to remote devices (probably a feature dedicated for mobile devices) by listening to specific Stride port(?)
- Stride IOS Relay - project whose main goal is to run a script called "stride-io-relay.py".

What this script does:

- Monitors USB connections using usbmuxd (the iOS USB daemon)
- When an iOS device connects, the script:
    - Opens device - port 31255 via usbmuxd
    - Opens host - port 31254 via normal TCP
    - Bi-directionally relays all data between the two sockets
- Continues doing this as long as the device stays connected

This is likely due to limitations of the .NET Framework at the time, and since we focus our efforts on the latest and most modern versions of .NET, it is no longer necessary. Connecting to mobile devices is available from both the dotnet cli and Rider, Visual Studio, etc.

- I also updated Stride.PublicApiCheck to net 10 and added a comment on how to use it. I didn't know we had a tool for checking breaking changes 😅

## Types of changes

## Checklist


Kryptos-FR:
I wasn't aware of most of those tools 😅. I'll have a look at the WPF debug tool to see whether there is something to salvage from it.

edit: found a reference to this project from an old Paradox (not even Xenko) archive from 11 years ago. It's safe to say that it was likely a very early attempt at having some kind of editing/instrospection capabilities at runtime.

Issue #1129, opened 2021-08-27
[Build] Need to document build system


jasonswearingen:
**Is your feature request related to a problem? Please describe.**
Extremely difficult to learn how the build/nuget system works, or debug issues when they arise.

**Describe the solution you'd like**
Document how both the dev-time MSBuild system works, and how the runtime Nuget import system works.

**Describe alternatives you've considered**
- Looking through the Stride.sln
   - Very difficult because of how VS invisibly utilizes the build system.  I have found that certain files like ```SharedAssemblyInfo.cs``` contain config info but do not know how this is actually utilized in the build system.
- Debug build breaks
  - Very difficult because of how msbuild rewrites the output folder locations.  I see that various ```.csproj``` contain references to the build system files by including lines such as ``` <Import Project="..\..\targets\Stride.Core.props" />```  but how does one go about debugging that?  
     - For example ```Stride.Core.props``` contains the line ```<StrideAssemblyProcessorOptions Condition="'$(StrideAssemblyProcessorOptions)' == ''">--auto-notify-property --auto-module-initializer --serialization</StrideAssemblyProcessorOptions>```  What does this ```StrideAssemblyProcessor``` do, why is it here in this global config, and how does one go about testing/debugging it?
- Debug runtime nuget import breaks
  - In my Net6 port of Stride, when I try to run a game project from the Stride.GameStudio app, I get an error with no ability to break on exception occuring:
```
[C:\Program Files\dotnet\sdk\6.0.100-preview.7.21379.14\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.targets(536,5)]: Error: The "CreateAppHost" task failed unexpectedly.
System.IO.FileLoadException: Could not load file or assembly 'Microsoft.NET.HostModel, Version=6.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'. Could not find or load a specific file. (0x80131621)
File name: 'Microsoft.NET.HostModel, Version=6.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'
 ---> System.IO.FileLoadException: Could not load file or assembly 'Microsoft.NET.HostModel, Version=6.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'.
[...]

PR #2997, opened 2025-12-06
feat: Build system analysis and improvements

Kryptos-FR:
# PR Details

Our build system is complex and a high barrier of entry for contributors (old and new). The goal of this PR is to document how it currently works and to propose a few areas of improvements.

Claude Sonnet AI was used to create the initial documentation. Now we need to review it, fix any mistakes and decide how to move from there.

Note: the prompts used are documented in the commit messages.

## Related Issue

#1129, stride3d/stride-docs#346 

## Types of changes

## Checklist

- [ ] ~~I have added tests to cover my changes~~.
- [ ] ~~All new and existing tests passed~~.
- [ ] ~~**I have built and run the editor to try this change out.**~~

### DIFF
--- a/build/docs/02-platform-targeting.md
+++ b/build/docs/02-platform-targeting.md
@@ -1,0 +1,540 @@
+# Platform Targeting in Stride
+
+This document explains how Stride handles building for multiple platforms (Windows, Linux, macOS, Android, iOS, UWP).
+
+## Platform Detection
+
+Stride determines the target platform through a combination of:
+1. **TargetFramework** (primary method for mobile/UWP)
+2. **OS Detection** (for desktop platforms)
+3. **RuntimeIdentifier** (for cross-compilation)
+
+### Detection Logic
+
+```xml
+<!-- Default: Windows -->
+<StridePlatform>Windows</StridePlatform>
+
+<!-- Linux: via OS check or RID -->
+<StridePlatform Condition="'$([MSBuild]::IsOSPlatform(Linux))' Or '$(RuntimeIdentifier.StartsWith(linux))'">
+  Linux
+</StridePlatform>
+
+<!-- macOS: via RID -->
+<StridePlatform Condition="$(RuntimeIdentifier.StartsWith('osx'))">
+  macOS
+</StridePlatform>
+
+<!-- UWP: via TargetFramework -->
+<StridePlatform Condition="'$(TargetFramework)' == 'uap10.0.16299'">
+  UWP
+</StridePlatform>
+
+<!-- Android: via TargetFramework -->
+<StridePlatform Condition="'$(TargetFramework)' == 'net10.0-android'">
+  Android
+</StridePlatform>
+
+<!-- iOS: via TargetFramework -->
+<StridePlatform Condition="'$(TargetFramework)' == 'net10.0-ios'">
+  iOS
+</StridePlatform>
+```
+
+## TargetFramework Mapping
+
+| Platform | TargetFramework | Notes |
+|----------|----------------|-------|
+| **Windows** | `net10.0` or `net10.0-windows` | `net10.0-windows` only when Windows-specific APIs needed |
+| **Linux** | `net10.0` | Same TFM as Windows, differentiated by OS |
+| **macOS** | `net10.0` | Same TFM as Windows, differentiated by OS |
+| **Android** | `net10.0-android` | Mobile-specific TFM |
+| **iOS** | `net10.0-ios` | Mobile-specific TFM |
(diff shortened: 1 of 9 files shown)


Jklawreszuk:
Shouldnt this be dedicated in the [docs ](https://github.com/stride3d/stride-docs) repo ?

VaclavElias:
Thank you @Kryptos-FR , this is indeed important. The most important part is the review part, which can be done only by those who understand Stride build process. It will be most likely you, xen2 and Eideren, maybe others. Once you do first round clean up, I can join and read and see if the docs make me understand the build process better 🤣.

Probably, it needs to be here in this repo till it is in the good shape, so AI can easily access Stride source code if needed. Once we are happy with the content, we can move it here https://doc.stride3d.net/latest/en/contributors/engine/index.html, and merge/restructure it with the existing content, including this one https://doc.stride3d.net/latest/en/contributors/contribution-workflow/index.html

build/docs/02-platform-targeting.md
```diff
@@ -0,0 +1,540 @@
...
+# Cross-compile for Windows
+dotnet build -f:net10.0 -r:win-x64
+
+# Android
+dotnet build -f:net10.0-android
+```
+
+### From macOS
+
+macOS can build:
```
Jklawreszuk, line 358: I'm not sure about that. Of course, you can RUN the project, but in order to build it, someone should make PR with built native dependencies.

Kryptos-FR:
> Shouldnt this be dedicated in the [docs ](https://github.com/stride3d/stride-docs) repo ?

While we work on it, it is easier if it is in the main repo. Especially if we want to use AI to help us.

build/docs/02-platform-targeting.md
Kryptos-FR: One of the few hallucinations the AI made. I'll have to tell it that currently only build from Windows is fully supported.

I'll wait for more review comments and then do another pass. Keep them coming :)

xen2:
Thanks for the initiative!

A lot of stuff we did in the past to work around very disparate build between Windows/iOS/Android.
Some examples:
- the .ssdeps system because each system was behaving differently when it came to copying content, libraries and native libs
- our own concept of "StridePlatform"
- our own system to deal with native libraries copies in a somewhat unified way
- I think in some cases transitive dependencies were not properly copied under some platforms so we had to deal with that (I think that's what the .ssdeps system was for?)

Since then, .NET Core arrived and over time they unified many things (incl. iOS/Android).
Also they introduced a lot of new mechanism (i.e. Directory.props, StridePlatform => use TFM and `#if ANDROID`)
And many nuget packages now have their own target files included automatically to add behavior so we could rely on existing libraries/packages for some specific needs.
So we could possibly get rid of a **lot** (if not most) of old target/rules!

Parts that will require more specific research/discussions (to possibly find a better/more up-to-date approach), so I would delay touching them (or deal with it as a separate topic) and focus on the rest for the main build system rewrite issue:
- Native lib build
- Per-Graphics API build

Another thing to look for is for any use of a target/props starting with an underscore, i.e. this: https://github.com/stride3d/stride/blob/4aba569074393fb38329a07a82d5001c8c7523e6/sources/targets/Stride.Core.targets#L212-L214
They are supposedly internal and we shouldn't rely on those as they can change.
Most of them were done this way as there was no workaround, but since then there might be an alternative approach or they might have exposed an entry point to fix it.

Kryptos-FR:
I have started to work on transforming our build system into internal SDKs so that the projects are properly recognized by the standard tooling (e.g. the C# DevKit in VS Code) and it's also easier to maintain (at the same time simplify and remove obsolete conf).

I have a WIP branch here: https://github.com/Kryptos-FR/stride/tree/feature%2Fstride-sdk. It's still very early.

xen2:
Awesome! 
Let me know if some obsolete stuff is difficult to port, we could try to get rid of it before switching to new system.

Ethereal77:
Btw, I remember having read that MSBuild 18 now allows to call Tasks without the hard requirement of `netstandard2.0` by allowing them to live out-of-process. It is an opt-in config that you need to specify in the project file when defining the Task.
Can this be useful to also simplify some of the build system regarding the `AssemblyProcessor` and `Stride.Core.Tasks`?

Edit: I found it. [Here](https://learn.microsoft.com/en-us/dotnet/core/whats-new/dotnet-10/sdk#use-net-msbuild-tasks-with-net-framework-msbuild) you can read about that.

Kryptos-FR:
@xen2 I guess this PR can be closed now that we have merged the rewrite.

Issue #2964, opened 2025-11-09
Task: Remove .vs workspace folder (samples/Templates/Packs/MaterialPackage/.vs)


VaclavElias:
**Release Type**: GitHub source current branch

**Version**: Latest commit at time of report

**Platform(s)**: Windows

**Describe the bug**
Visual Studio local workspace artifacts under `samples/Templates/Packs/MaterialPackage/.vs` are present in the repository even though `.gitignore` contains an entry to exclude `.vs/`. These are per-user and should not be versioned. 

**To Reproduce**
1. Clone the repository
2. Navigate to `samples/Templates/Packs/MaterialPackage/.vs`.
3. Observe that `.vs` workspace files are tracked in source control.

**Expected behavior**
The `.vs` directory should be absent from version-controlled content.

**Screenshots**
Not applicable (listing of tracked .vs path is sufficient).

**Log and callstacks**
Not applicable.

**Additional context**
Repository clean up.

D4rkDuck:
Hey 👋 

I would like to contribute, is this Issue still relevant?

Eideren:
yep

D4rkDuck:
@Eideren I believe this Issue can be closed now

PR #3011, opened 2025-12-14
chore: removed expectedly ignored vs directory

D4rkDuck:
# PR Details

Manually removed the folder .vs folder containing 2 files.

## Related Issue

#2964 

## Types of changes

## Checklist


D4rkDuck:
@dotnet-policy-service agree

D4rkDuck:
Related to testing out these changes:

I assume these changes do not require testing, but I nonetheless built the GameStudio and started the MaterialShader Sample successfully. I also tested whether creating a new .vs folder with any file would add it back, even though the gitignore shouldn't allow that. It did not, so is now working as intended.
I would actually not really know if there are any specific areas, that could be affected by the removal of a .vs folder. Could a more knowledgeable maintainer let me know where stride might be affected by such changes for learning purposes?

Kryptos-FR:
There shouldn't be any are affected since a .vs folder is local to a given developer. It's only needed by Visual Studio and doesn't influence the build.

Eideren:
Thanks !

PR #3037, opened 2026-01-19
fix: Rollback breaking changes made to the pixel format public API

Eideren:
# PR Details
Users shouldn't have to fix syntactic issues when upgrading versions, rolling back related changes.
Feel free to setup an upgrader if you want to keep your changes in @Ethereal77 and I'll close this PR. 
Also, are there any other breaking changes in the public API @Ethereal77 ?

## Types of changes

## Checklist

### DIFF
--- a/sources/editor/Stride.Assets.Presentation/AssetEditors/SpriteEditor/Services/SpriteEditorImageCache.cs
+++ b/sources/editor/Stride.Assets.Presentation/AssetEditors/SpriteEditor/Services/SpriteEditorImageCache.cs
@@ -92,7 +92,7 @@ namespace Stride.Assets.Presentation.AssetEditors.SpriteEditor.Services
                     using (var texTool = new TextureTool())
                     {
                         texImage = texTool.Load(filePath, false);
-                        texTool.Decompress(texImage, texImage.Format.IsSRgb);
+                        texTool.Decompress(texImage, texImage.Format.IsSRgb());
                         if (texImage.Format == PixelFormat.R16G16B16A16_UNorm)
                             texTool.Convert(texImage, PixelFormat.R8G8B8A8_UNorm);
                         var image = texTool.ConvertToStrideImage(texImage);
--- a/sources/editor/Stride.Editor/Thumbnails/StaticThumbnailCommand.cs
+++ b/sources/editor/Stride.Editor/Thumbnails/StaticThumbnailCommand.cs
@@ -51,7 +51,7 @@ namespace Stride.Editor.Thumbnails
             using (var texImage = texTool.Load(image, Parameters.SRgb))
             {
                 // Rescale image so that it fits the thumbnail asked resolution
-                texTool.Decompress(texImage, texImage.Format.IsSRgb);
+                texTool.Decompress(texImage, texImage.Format.IsSRgb());
                 texTool.Resize(texImage, thumbnailSize.X, thumbnailSize.Y, Filter.Rescaling.Lanczos3);
 
                 // Save
--- a/sources/engine/Stride.Assets.Tests/TexturePackerTests.cs
+++ b/sources/engine/Stride.Assets.Tests/TexturePackerTests.cs
@@ -748,7 +748,7 @@ namespace Stride.Assets.Tests
 
             var source = Image.New2D(width, height, 1, PixelFormat.R8G8B8A8_UNorm);
 
-            Assert.Equal(source.TotalSizeInBytes, PixelFormat.R8G8B8A8_UNorm.SizeInBytes * width * height);
+            Assert.Equal(source.TotalSizeInBytes, PixelFormat.R8G8B8A8_UNorm.SizeInBytes() * width * height);
             Assert.Equal(1, source.PixelBuffer.Count);
 
             Assert.Equal(1, source.Description.MipLevels);
@@ -777,7 +777,7 @@ namespace Stride.Assets.Tests
 
             var source = Image.New2D(width, height, 1, PixelFormat.R8G8B8A8_UNorm);
 
-            Assert.Equal(source.TotalSizeInBytes, PixelFormat.R8G8B8A8_UNorm.SizeInBytes * width * height);
+            Assert.Equal(source.TotalSizeInBytes, PixelFormat.R8G8B8A8_UNorm.SizeInBytes() * width * height);
             Assert.Equal(1, source.PixelBuffer.Count);
 
             Assert.Equal(1, source.Description.MipLevels);
--- a/sources/engine/Stride.Assets/Skyboxes/SkyboxGenerator.cs
+++ b/sources/engine/Stride.Assets/Skyboxes/SkyboxGenerator.cs
@@ -146,7 +146,7 @@ namespace Stride.Assets.Skyboxes
             if (textureSize < 64) textureSize = 64;
 
             // TODO: Add support for HDR 32bits
-            var filteringTextureFormat = skyboxTexture.Format.IsHDR ? skyboxTexture.Format : PixelFormat.R8G8B8A8_UNorm;
+            var filteringTextureFormat = skyboxTexture.Format.IsHDR() ? skyboxTexture.Format : PixelFormat.R8G8B8A8_UNorm;
 
             //var outputTexture = Texture.New2D(graphicsDevice, 256, 256, skyboxTexture.Format, TextureFlags.ShaderResource | TextureFlags.UnorderedAccess, 6);
             using (var outputTexture = Texture.New2D(context.GraphicsDevice, textureSize, textureSize, true, filteringTextureFormat, TextureFlags.ShaderResource | TextureFlags.RenderTarget, 6))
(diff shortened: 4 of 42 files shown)


Ethereal77:
Yes, there are more breaking changes. Some time ago, we discussed if this PR (and others of the same kind) would more appropriately merge into a `future` or `5.0` branch, and that it was acceptable then to break things. That's the why 😢

I can create upgraders to ease the pain, but I'm not very knowledgeable in how the upgraders work. Any idea where to start looking?

Eideren:
@Ethereal77 Makes sense, I would rather we leave it as a last resort for syntax though, the editor does not handle loading uncompilable projects well enough right now. And for non-programming people, this breaking change makes a project effectively unusable.

Syntax changes are unfortunately not really handled well afair, have a look at how `XenkoToStrideRenameHelper` is used in the engine. Not sure how feasible making an upgrader for this case would be though, given how fragile those regex would have to be.

Ethereal77:
> I would rather we leave it as a last resort for syntax though, the editor does not handle loading uncompilable projects well enough right now.

If this is a breaking change so big to create that kind of friction, I propose what I already proposed (and some people agreed on), to create a separate branch for "future Stride", where all these kind of changes can go (like, for example, the cross-platform editor, the OpenGL deprecation, the new D3D12-only editor, the new shader compilation pipeline, etc.)
Then, merge those changes back to `master` only at specific tagged versions that are deemed to have "acceptable" breaking changes.

Even though I understand what difficulties can imply these syntax changes, I think it would be a pitty to leave out some of these. C# 14's "Extension everything" in particular can be very valuable to modernize the codebase and allow some very interesting things.

Edit: On a second thought, maybe instead of an upgrader, this could be an analyzer? You've seen that the change basically is removing a `()` from the extension invocation. There are some renames (like `IsVideo` to `IsVideoFormat`), but those we can duplicate and annotate with `[Obsolete]`.

> And for non-programming people, this breaking change makes a project effectively unusable.

Get what I'm about to say not as if they are less worthy than "programming people", but just as my opinion as someone that has been following gamedev topics since before editor-engines (like Unity), and based on the context that currently Stride requires far more coding understanding than other engines:

I can't talk for non-programming people, but I think Stride may not be the best engine for people that do not want to touch and understand the underlying code. Ideally, everything should just work forever without problems, but retrocompatibility usually imposes a very hard limit to evolution in complex projects, and in a game engine, which is a very complex project and one that needs to be always at the very edge of efficiency, it does more so.

Having said that, the best advice I can give to non-programming people that want to create a complex project touching little code with Stride is to stick to a version that works for them and not moving versions (or at least, only minor revisions, bug fixes, etc.)

> ...have a look at how `XenkoToStrideRenameHelper` is used in the engine.

Correct me if I'm wrong, but upgraders and renamers are not just for assets? (I mean, asset names/types, serialized properties, type changes and conversions, etc.)
If that's the case, an upgrader cannot upgrade code. Can it?

---

Anyway, sorry for the wall of text. 😁
Please understand that it’s a bit frustrating for me to see some of my changes undone, as with clearer planning and more alignment among us this could have been a smoother process.

Eideren:
I also would rather we break users project if we have some critical issue that needs to be addressed, method calls being swapped into properties with the same functionality is not that.
I would advocate against these kinds of changes even if we had a future branch. If we had other syntax breaking changes that would actually provide some functionality or fix some critical issues, why not, but none of the changes we introduce in the future will prevent a project from being built, the opengl one for example will fallback to the appropriate API.

I'm also not saying that we shouldn't use extension everything, just that we can only use it in case where we wouldn't introduce a breaking change.

> I can't talk for non-programming people, but I think Stride may not be the best engine for people that do not want to touch and understand the underlying code. 

Almost no games are made solo, the editor is supposed to be for non-programmers to work with programmers, and we've had terrible adoption specifically because we have thousands of things breaking all the time, let's not add one more thing.

`XenkoToStrideRenameHelper` goes through c# file content to swap namespaces usages

> Please understand that it’s a bit frustrating for me to see some of my changes undone, as with clearer planning and more alignment among us this could have been a smoother process.

I totally get that, I wouldn't want to have my work undone either, you've done some incredible work for the engine, I very much respect and am glad of the many weeks of work you put into these changes.

I think that's just the limitation of how we're organized, we barely have the time to review, actually getting into those kinds of conversations seems very unlikely.

Ethereal77:
> I'm also not saying that we shouldn't use extension everything, just that we can only use it in case where we wouldn't introduce a breaking change.

I agree with this half-way. In my opinion, having new features implemented in the future with "extension everything" while we still have things using the old syntax is kind of inelegant, and I foresee questions by users regarding this.

> Almost no games are made solo, the editor is supposed to be for non-programmers to work with programmers, and we've had terrible adoption specifically because we have thousands of things breaking all the time, let's not add one more thing.

Well, of course not. But in this projects not made solo, the responsibility to know the syntax changes (in scripts) fall precisely in the programmers, no? Anyway, I understand that there will be cases of artists or designers that would want to edit scripts themselves and feel confused about syntax errors where previously there were none.
 
> `XenkoToStrideRenameHelper` goes through c# file content to swap namespaces usages

If that's the case, I'll see if it is possible to craft something that removes the `()` from those extension calls. Maybe that is enough to circumvent the problem.

> I totally get that, I wouldn't want to have my work undone either, you've done some incredible work for the engine, I very much respect and am glad of the many weeks of work you put into these changes.
> 
> I think that's just the limitation of how we're organized, we barely have the time to review, actually getting into those kinds of conversations seems very unlikely.

I appreciate that. Thank you.
I think it would be wise if we at least stablish some kind of protocol when integrating changes into the engine, to avoid potential conflicts and regressions, to avoid having to do rollbacks, etc.
I insisted in my PR that it needed to be tested and used in real use cases at least some time before considering it safe to merge. I have been trying to get the tests green most of the time, but those are synthetic tests.
Maybe something to discuss in the next meeting?

Anyway, you decide if you merge the rollback now (if urgent), or wait until a (possible?) upgrader can be done.

Eideren:
Let's leave this as is for now until after you've looked into whether an upgrader would be feasible. An alternative option would be to change the nomenclature of those properties and mark the extension methods as obsolete.

Ethereal77:
I have been trying different things. What is your opinion on a "two-stage obsoletion" strategy? Let me explain:

* Any extension method that now conflicts with the new syntax is marked with, for example:

      [Obsolete("BlockSize() is obsolete and will be deprecated in a future version. It is to be replaced by the extension property BlockSizeInBytes. You can try the new syntax by importing the Experimental namespace.")]

  Note it has a different name for now. More on that later.

* At the same time, we have the conflicting extensions in another class in an `Stride.Graphics.Experimental` namespace.
   To avoid ambiguity and allow to use the new syntax immediately, we use a different (albeit readable) name temporarily, i.e., `BlockSize` -> `BlockSizeInBytes`, or `BlockWidth` -> `BlockPixelWidth`, etc.

* Then, in the next version of Stride, we effectively remove the old extension method syntax, move the new syntax out of `Stride.Graphics.Experimental`, and do the following:

       [Obsolete("BlockSizeInBytes will be renamed to its previous name (BlockSize) in the next version. This is a temporary workaround to allow both syntaxes to coexist during the transition period.")]
       public int BlockSizeInBytes => /* Whatever */;
       public int BlockSize => /* Whatever */;

What do you think?

Eideren:
Yep, sounds good to me

Eideren:
Leaving this as a draft while Ethereal takes care of the above, will close it as soon as his opens.

xen2:
With the latest changes I did to package upgrader (it properly restores old version before doing actual upgrade), we can use Roslyn API to upgrade user code properly (since it can resolve symbols on the old version before update):

https://github.com/stride3d/stride/blob/master/sources/engine/Stride.Assets/StridePackageUpgrader.cs#L184-L199

I will try to add a roslyn skeleton we can use for such upgrades.

I believe it's much more preferable to use Roslyn rather than having a lot of obsolete API, backward-compat code, multiple API versions to maintain and avoiding much-needed API cleanliness upgrades.

Since we use C# and Roslyn we have this huge advantage over other languages/ecosystem to be able to do this kind of user-code upgrade so I would leverage on it so that all this obsolete stuff is dealt with inside upgraders rather than polluting engine codebase/API.

Note: I am talking about simple cases like basic renames, turning a method into a property or vice-versa, etc.

xen2:
BTW, I was saying that if we actually want to keep those new API changes. Otherwise, fine to go back to old version if we're OK with that long term.

xen2:
OK, please hold, I am doing a proper framework to easily do roslyn-based code upgrade.

xen2:
Added support for those.
I tested it on a 4.1 sample where I added some code with `PixelFormat.IsHDR()` and it worked properly.
I also tested it with the new `AssetCompiler.exe upgrade` path and it worked as well.

https://github.com/stride3d/stride/commit/49f299192d34d432b49cf924f41606311d3866e2

Since it uses Roslyn and properly analyze type, it can easily parse stuff such as `objA.Test.IsHDR()` and figure out if `objA.Test` is really a PixelFormat and IsHDR() extension method was available with proper `using`.
So it won't upgrade invalid code by mistake (like a dumb regex would do).

Ethereal77:
Oh, my! Sorry, I totally forgot about this. I had a local branch ready with the `[Obsolete]` solution I talked about above. Should I discard it entirely?

Issue #3030, opened 2026-01-11
Feat: 2D Plattformer Template


D4rkDuck:
# Is your feature request related to a problem? Please describe.
Stride has a lot of Sample Projects, but is lacking in the 2D department. I have checked Discord and saw some messages in the past asking for 2D Projects to try in Stride. With the addition of this template I believe the starting experience for developers aiming to create 2D projects can be improved.

# Describe the solution you'd like
I would like to contribute a simple 2D-Plattformer template (similiar to the 3D one), that can be selected when launching Stride and choosing a project-type.

# Additional context/Features

The envision the project to roughly contain following features:
- Character
  - Character Controller
  - Sprite + Animations (existing/provided or free assets-online)
- Camera (either static within a small room or following character in a bigger room)
- Collectibles (Coins)
  - Sound (collecting Coins)
- Scene (simple room with few plattforms)

_Possible Additions_ (Nice-to-Haves)
- Effects (f.e. Dust on jump/turning direction from left to right, ...)
- UI

D4rkDuck:
If everything seems to be in order, getting assigned to the Issue, as I plan to contribute this myself, would be appreciated :)

Eideren:
Done

PR #3042, opened 2026-01-28
feat: Add a new template: 2D Platformer

D4rkDuck:
# PR Details

This PR adds a new template to the stride launcher.
The template adds very basic functionality with no specialized code architecture, making it perfect for newbies to get into creating a new 2D project.
This template uses new assets from two new CC0 licensed asset packs:

- [Camelot](https://kevins-moms-house.itch.io/camelot) - Knight and Coin
- [4 Season platformer](https://kevins-moms-house.itch.io/four-seasons-platformer-tileset) - Plattforms

## Related Issue

[Feat: 2D Platformer Template #3030](https://github.com/stride3d/stride/issues/3030)

## Types of changes

## Checklist

When rebuilding the solution there are 3 error messages related to tests. I believe these are unrelated to my changes. Stride.GameStudio launches with no problems.

image

## Visuals

Here are screenshots of the launcher and launched game:

image

image

### DIFF
--- a/samples/Templates/Platformer2D/Platformer2D/Platformer2D.Game/Gameplay/CollectCoin.cs
+++ b/samples/Templates/Platformer2D/Platformer2D/Platformer2D.Game/Gameplay/CollectCoin.cs
@@ -29,7 +29,6 @@ public class CollectCoin : AsyncScript
                 // Disable Sprite
                 Entity.Get<SpriteComponent>().Enabled = false;
                 sfxInstance = SoundEffect?.CreateInstance();
-                sfxInstance?.Stop();
                  
                 // Play Sound
                 sfxInstance?.Play();


Ethereal77:
Is the PR done? Or still WIP?

D4rkDuck:
> Is the PR done? Or still WIP?

Done and ready to be reviewed.

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/Gameplay/CoinRotation.cs
```diff
@@ -0,0 +1,39 @@
...
+
+namespace _2DPlatformer.Gameplay;
+
+public class CoinRotation : SyncScript
+{
+    [Display("Coin Sprite Sheet")]
+    public SpriteSheet CoinSprites { get; set; }
+    private SpriteFromSheet spriteComponent;
+    
+    private double animationTimer = 0f;
```
Ethereal77: Not very important, but for the range of this variable (berween 0 and 1 or less), a `float` would be enough.

`double` is slower (not much on current hw, but still), and 2x the size, and the extra precission here is not very important.

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/Gameplay/CoinRotation.cs
```diff
@@ -0,0 +1,39 @@
...
+    private readonly float animationInterval = 1f / 12f;
```
Ethereal77: If `12` is the number of frames of the sprite sheet, do not hardcode that value. Get it from the sprite animation itself.

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/Gameplay/CoinRotation.cs
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    public override void Update()
+    {
+        animationTimer += Game.UpdateTime.Elapsed.TotalSeconds;
+        
+        if (animationTimer >= animationInterval)
+        {
+            animationTimer = 0f;
+            spriteComponent.CurrentFrame = (spriteComponent.CurrentFrame + 1) % 12;
```
Ethereal77: Again, do not hardcode `12` if it is the number of frames.

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/Gameplay/CoinRotation.cs
```diff
@@ -0,0 +1,39 @@
...
+        spriteComponent = Entity.Get<SpriteComponent>().SpriteProvider as SpriteFromSheet;
+    }
+
+    public override void Update()
+    {
+        animationTimer += Game.UpdateTime.Elapsed.TotalSeconds;
+        
+        if (animationTimer >= animationInterval)
+        {
+            animationTimer = 0f;
```
Ethereal77: Instead of setting `animationTimer` to 0, subtract `animationInterval`, i.e.:

    animationTimer -= animationInterval;

That way, if you accumulated a bit over the interval, it is not lost, and animation pacing is more stable.

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/Gameplay/CollectCoin.cs
```diff
@@ -0,0 +1,64 @@
...
+{
+    private StaticColliderComponent staticColliderComponent;
+    
+    [Display("Collect Coin")]
+    public Sound SoundEffect { get; set; }
+    private SoundInstance sfxInstance;
+
+    public override async Task Execute()
+    {
+
```
Ethereal77: nit: Empty line

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/Gameplay/CollectCoin.cs
```diff
@@ -0,0 +1,64 @@
...
+    
+        while(Game.IsRunning)
+        {
+            // Waits for a Collision with the Character
+            var collision = await staticColliderComponent.NewCollision();
+            var otherCollider = staticColliderComponent == collision.ColliderA ? collision.ColliderB : collision.ColliderA;
+            
+            if (otherCollider.Entity.Get<PlayerController>() != null)
+            {
+
```
Ethereal77: nit: Empty line

samples/Templates/Platformer2D/Platformer2D/Platformer2D.Game/Gameplay/CollectCoin.cs
```diff
@@ -0,0 +1,64 @@
...
+            var collision = await staticColliderComponent.NewCollision();
+            var otherCollider = staticColliderComponent == collision.ColliderA ? collision.ColliderB : collision.ColliderA;
+            
+            if (otherCollider.Entity.Get<PlayerController>() != null)
+            {
+                
+                // Disable Sprite
+                Entity.Get<SpriteComponent>().Enabled = false;
+                sfxInstance = SoundEffect?.CreateInstance();
+                sfxInstance?.Stop();
```
Ethereal77: Is this really needed? Does it not istantiate stopped?
Asking to anyone more knowledgeable on the audio part.

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/Gameplay/CollectCoin.cs
```diff
@@ -0,0 +1,64 @@
...
+                var task = Task.Run(() => WaitMethod());
+                task.Wait(TimeSpan.FromSeconds(3));
+            }
+        }
+    }
+    
+    /// <summary>
+    /// Wait Method, for playing the Sound and then deactivating the object.
+    /// </summary>
+    private void WaitMethod()
```
Ethereal77: Maybe a better name than `WaitMethod` that better describes its intent.
Also, the comment says it plays a sound, but the method only removes the coin entity.

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/Gameplay/CollectCoin.cs
```diff
@@ -0,0 +1,64 @@
...
+using System.Threading.Tasks;
+using Stride.Core.Mathematics;
+using Stride.Input;
+using Stride.Engine;
+using Stride.Audio;
+using Stride.Physics;
+using System.Diagnostics;
+using System.Threading;
+using _2DPlatformer.Player;
+using SharpFont.PostScript;
```
Ethereal77: ??

Please cleanup unused `using`s.

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/PlayerController.cs
```diff
@@ -0,0 +1,162 @@
...
+
+public class PlayerController : SyncScript
+{
+    private CharacterComponent characterComponent;
+    
+    [Display("Character Sprite Sheet")]
+    public SpriteSheet CoinSprites { get; set; }
+    private SpriteFromSheet spriteComponent;
+    
+    private double animationTimer = 0f;
```
Ethereal77: Same as in `CoinRotation`: This should be a `float` preferably.

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/PlayerController.cs
```diff
@@ -0,0 +1,162 @@
...
+    private readonly float animationInterval = 1f / 10f;
```
Ethereal77: Same as in `CoinRotation`: Do not hardcode `10` if it is the number of frames in the sprite sheet.

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/PlayerController.cs
```diff
@@ -0,0 +1,162 @@
...
+    private int runFrame = 0;
+    private int jumpFrame = 0;
+    
+    private readonly float moveSpeed = 5f;
```
Ethereal77: If this is an actual constant value (not assigned in the constructor, and not modified), make it a `const` instead.

    private const float MOVE_SPEED = 5.0f;  // units per second?

Also, it is good practice to indicate (in comments for example) the units to give a bit of context to the value. In this case, units per second? meters per second?

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/PlayerController.cs
```diff
@@ -0,0 +1,162 @@
...
+    /// <summary>
+    /// Handles both the Input of the Player via Keyboard (WAS- & Arrow-Keys) and the resulting Animation at the end according to input
+    /// </summary>
+    private void HandleInput()
+    {
+         if (Input.HasKeyboard)
+            {
+                var moveDirection = Vector3.Zero;
+                bool isMoving = false;
+
```
Ethereal77: nit: Avoid too much empty space unless it is for clearly separating types or groups of methods, for example.

If it is to separate steps inside a method, maybe it is more clear if you split up the method or use a comment as separator.

Not talking about this one specifically. Just an advice.

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/PlayerController.cs
```diff
@@ -0,0 +1,162 @@
...
+                {
+                    isMoving = true;
+                    if (isFacingRight)
+                    {
+                        isFacingRight = false;
+                        characterComponent.Orientation = Quaternion.RotationY(MathUtil.DegreesToRadians(180));
+                    }
+                    
+                    moveDirection  = -Vector3.UnitX * moveSpeed;
+
```
Ethereal77: nit: Empty line

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/PlayerController.cs
```diff
@@ -0,0 +1,162 @@
...
+        spriteComponent = Entity.Get<SpriteComponent>().SpriteProvider as SpriteFromSheet;
+    }
+
+    public override void Update()
+    {
+         HandleInput();
+    }
+    
+    /// <summary>
+    /// Handles both the Input of the Player via Keyboard (WAS- & Arrow-Keys) and the resulting Animation at the end according to input
```
Ethereal77: Comment does not mention D (WASD).

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/PlayerController.cs
```diff
@@ -0,0 +1,162 @@
...
+                    }
+                    moveDirection = Vector3.UnitX * moveSpeed;
+                }
+                
+                // Up (Jump)
+                if (Input.IsKeyDown(Keys.W) && characterComponent.IsGrounded || Input.IsKeyDown(Keys.Up) && characterComponent.IsGrounded)
+                {
+                    characterComponent.Jump();
+                }
+
```
Ethereal77: nit: Double empty line

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/PlayerController.cs
```diff
@@ -0,0 +1,162 @@
...
+                
+                characterComponent.SetVelocity(moveDirection);
+                HandleAnimation(isMoving);
+            }
+    }
+    
+    
+    private void HandleAnimation(bool isMoving)
+    {
+
```
Ethereal77: nit: Empty line

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/PlayerController.cs
```diff
@@ -0,0 +1,162 @@
...
+    /// <summary>
+    /// When there is no input, play idle-Animation. Which loops frames 0-3.
+    /// </summary>
+    private void PlayIdleAnimation()
+    {
+        jumpFrame = 0;
+        
+        if (animationTimer >= animationInterval)
+        {
+            animationTimer = 0f;
```
Ethereal77: Same as in `CoinRotation`: Subtract `animationInterval` instead of setting `animationTimer` to 0.

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/PlayerController.cs
```diff
@@ -0,0 +1,162 @@
...
+            spriteComponent.CurrentFrame = (spriteComponent.CurrentFrame + 1) % 4;
```
Ethereal77: It there is going to be many hardcoded frame indices or ranges of frames, instead of spreading magic numbers all over the file, consider putting readable constants at the top of the file. Something like:

    private const int IDLE_FRAMES_START = 0;
    private const int IDLE_FRAMES_END = 3;
    private const int IDLE_FRAMES_COUNT = 1 + IDLE_FRAMES_END - IDLE_FRAMES_START;

That way, the code serves as documentation for itself when you use those values, and you only need to modify those consts in case you change the animation or the sprite sheet.

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/PlayerController.cs
```diff
@@ -0,0 +1,162 @@
...
+    /// When there is movement to either side, play run-animation. Which loops frames 4-19
+    /// </summary>
+    private void PlayRunAnimation()
+    {
+        jumpFrame = 0;
+        if (animationTimer >= animationInterval)
+        {
+            animationTimer = 0f;
+            runFrame = (runFrame + 1) % 16;
+            spriteComponent.CurrentFrame = 4 + runFrame;
```
Ethereal77: Read what I wrote for `PlayIdleAnimation()`.

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/PlayerController.cs
```diff
@@ -0,0 +1,162 @@
...
+    /// When the character is not grounded, play jump-animation. Which only plays frames 20-27 once and then resets when the character is grounded.
+    /// </summary>
+    private void PlayJumpAnimation()
+    {
+        if (animationTimer >= animationInterval)
+        {
+            animationTimer = 0f;
+            spriteComponent.CurrentFrame = 21 + jumpFrame;
+            
+            if (jumpFrame < 7 - 1)
```
Ethereal77: Read what I wrote for `PlayIdleAnimation()`.

samples/Templates/Platformer2D/Platformer2D/Resources/coin.png
Ethereal77, line 1: Why the underscore in the file name?

samples/Templates/Platformer2D/Platformer2D/Resources/lancelot.png
Ethereal77, line 1: Why the underscore in the file name?

samples/Templates/2DPlatformer/_2DPlatformer/Resources/skybox_texture_ldr.dds
Ethereal77, line 1: If the scene is not rendering a skybox (i.e., you have removed the skybox component, and the sky lighting, etc.), remove the resources that are not being used.

samples/Templates/Platformer2D/Platformer2D/Resources/spring.png
Ethereal77, line 1: Why the underscore in the file name?

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/_2DPlatformer.csproj
```diff
@@ -0,0 +1,18 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net10.0-windows</TargetFrameworks>
```
Ethereal77: The `Game` project is supposed to be crossplatform so it can be referenced from platform-specific projects (in this case, the `.Windows` one).
So this should not need the `-windows` part in the TFM.

Ethereal77 (changes requested):
I have made some comments on things you can improve.
Sorry for all the nitpicking, it is to mantain a common code style along Stride codebase.

Nice work overall!

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/Gameplay/CollectCoin.cs
D4rkDuck: I agree, I believe I tried a different approach here first and forgot to rename and document the new behaviour.

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/PlayerController.cs
D4rkDuck: Done

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/PlayerController.cs
D4rkDuck: This is intended, as there is no behaviour for pressing D / down-arrow. No ducking-feature available.

Nikyria: But D is for moving right..?

D4rkDuck: Woops! Ofc! WAD it is supposed to be

samples/Templates/Platformer2D/Platformer2D/Resources/coin.png
D4rkDuck: These are the original names from the asset packs, I'll rename it

samples/Templates/Platformer2D/Platformer2D/Resources/lancelot.png
D4rkDuck: These are the original names from the asset packs, I'll rename it

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/Gameplay/CoinRotation.cs
D4rkDuck: This is basically the value I use for adjusting the animations speed. Every 1/12 passed seconds the next frame is played in this case.
Closer to 1.0 would be a slow(er) animation.

I'll document this part and set it to const rather than readonly.

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/PlayerController.cs
D4rkDuck: Same as in `CoinRotation`:
This is basically the value I use for adjusting the animations speed. Every 1/10 passed seconds the next frame is played in this case.
Closer to 1.0 would be a slow(er) animation.

I'll document this part and set it to const rather than readonly.

samples/Templates/2DPlatformer/_2DPlatformer/Resources/skybox_texture_ldr.dds
D4rkDuck: Removed that sneaky asset.

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/_2DPlatformer.csproj
D4rkDuck: I am unfamiliar with setting platform specifics, should the whole property group be completely removed or need some settings to be changed?
Does this also affect the folder _2DPlatformer.Windows?

Ethereal77: No, just change `net10-windows` to `net10.0` and that's it!
`_2DPlatformer.Windows` does indeed need the `-windows` part because that project is platform-specific (i.e., specific to Windows, hence its name)

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/Gameplay/CoinRotation.cs
```diff
@@ -28,12 +24,12 @@ public override void Start()
...
     {
-        animationTimer += Game.UpdateTime.Elapsed.TotalSeconds;
+        animationTimer += (float) Game.UpdateTime.Elapsed.TotalSeconds;
         
         if (animationTimer >= animationInterval)
         {
-            animationTimer = 0f;
-            spriteComponent.CurrentFrame = (spriteComponent.CurrentFrame + 1) % 12;
+            animationTimer -= animationInterval;
+            spriteComponent.CurrentFrame = (spriteComponent.CurrentFrame + 1) % COIN_FRAMES_END;
```
Ethereal77: Sorry for the confusion. When I talked about not using a hardcoded `12` for the number of frames, what I meant is that instead of specifying a fixed 12 in the script, you query and store the actual number of frames in the animation, if possible.

If not, then a `const` is good enough, I suppose.

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/PlayerController.cs
```diff
@@ -137,9 +133,9 @@ private void PlayRunAnimation()
         jumpFrame = 0;
         if (animationTimer >= animationInterval)
         {
-            animationTimer = 0f;
-            runFrame = (runFrame + 1) % 16;
-            spriteComponent.CurrentFrame = 4 + runFrame;
+            animationTimer -= animationInterval;
+            runFrame = (runFrame + 1) % RUN_FRAME_END;
```
Ethereal77: Does the animation still work okay?

I ask because in my example I wrote the `RUN_FRAME_COUNT` (computed as `1 + RUN_FRAME_END - RUN_FRAME_START`) because you want to cycle for just the frames of the running animation. Let me explain with an example:
* The animation starts at frame 2, ends at frame 6, so there are actually `1 + 6 - 2` = 5 frames.
* Let's say `runFrame` now is at frame 4 (the final running frame), and the next frame is 5. Remember `runFrame` cycles between 0 to 4, and back to 0.
* Next, to compute the actual frame index from this `runFrame` of 5, you want it to compute it as `start + (frame % count)`, in this case that would be `2 + 5 % 5` = 2 (i.e., the running start frame again).
* If you do the modulo `% RUN_FRAME_END` like you are doing now, you'll show frames past the final frame for the running animation (i.e., you'll show frame 7, which does not belong to "running".

I hope all of that is clear enough. Tell me if you have any doubt, and I'll try rephrasing it some other way.

D4rkDuck: Just some clarifications beforehand:


I am using different frameCounters for each idle, running and jumping. This might not be the best way to do it, but I had problems using the same one for all and the resulting calculations 😅 
I think the confusion you see from your correct calculations stem from my run animation not actually starting at 0 and ending at 16, but rather start at 4 and end at 20 (that's why the 4 is added at the end and lead to a max value of 20 = last run frame) as seen in the screenshot. I might be doing this a totaly weird way way, so let me know if there is a better approach. Had a rough time coming from Unity where I would create Animation clips from frames and just play those not really caring about the individual frame count.

# Log

image

Now that I look at the log, I think I should reset the run/idle frame count to 0 like I do with jump to always start the animation at the beginning 🤔 

# Sprite Asset

image

D4rkDuck: I have pushed changes to the animation code, this mainly concerns the resetting of frameCounters to ensure they all start at their first intended frame (idle = 0, run = 4, jump = 20)

Ethereal77: Yeah, you are right. I misread and thought you were using just one counter. That has been the confusion.

Using three counters is valid. A bit overkill maybe, because when an animation is playing, the counters for the others are not used, and when returning to those other animations they begin playing from their start frame anyway.

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/Gameplay/CoinRotation.cs
```diff
@@ -0,0 +1,35 @@
...
+using Stride.Engine;
+using Stride.Graphics;
+using Stride.Rendering.Sprites;
+
+namespace _2DPlatformer.Gameplay;
+
+public class CoinRotation : SyncScript
+{
+    [Display("Coin Sprite Sheet")]
+    public SpriteSheet CoinSprites { get; set; }
```
Eideren: This property is unused

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/Gameplay/CoinRotation.cs
```diff
@@ -0,0 +1,35 @@
...
+    private SpriteFromSheet spriteComponent;
+    
+    private float animationTimer = 0f;
+    // Animation Spped: Every 1/12 passed seconds the next frame will be played. Closer to 1.0 is a slower Animation.
+    private const float animationInterval = 1f / 12f;
+    private const int COIN_FRAMES_END = 12;
+    
+    public override void Start()
+    {
+        spriteComponent = Entity.Get<SpriteComponent>().SpriteProvider as SpriteFromSheet;
```
Eideren: Best practice would be to setup a required reference; `public required SpriteComponent Sprite { get; init; }` and then rename your `SpriteFromSheet spriteComponent` into `spriteSheet` to better convey what this field represents.

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/Gameplay/CollectCoin.cs
```diff
@@ -0,0 +1,53 @@
...
+{
+    private StaticColliderComponent staticColliderComponent;
+    
+    [Display("Collect Coin")]
+    public Sound SoundEffect { get; set; }
+    private SoundInstance sfxInstance;
+
+    public override async Task Execute()
+    {
+        staticColliderComponent = Entity.Get<StaticColliderComponent>();
```
Eideren: Same idea as the other, best setup a required property

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/Gameplay/CollectCoin.cs
```diff
@@ -0,0 +1,53 @@
...
+
+namespace _2DPlatformer.Gameplay;
+
+public class CollectCoin : AsyncScript
+{
+    private StaticColliderComponent staticColliderComponent;
+    
+    [Display("Collect Coin")]
+    public Sound SoundEffect { get; set; }
+    private SoundInstance sfxInstance;
```
Eideren: this should be marked as nullable

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/Gameplay/CollectCoin.cs
```diff
@@ -0,0 +1,53 @@
...
+                Entity.Get<SpriteComponent>().Enabled = false;
+                sfxInstance = SoundEffect?.CreateInstance();
+                sfxInstance?.Stop();
+                 
+                // Play Sound
+                sfxInstance?.Play();
+                 
+                // Disable Coin Object
+                var task = Task.Run(() => RemoveEntityAfterWaiting());
+                task.Wait(TimeSpan.FromSeconds(3));
```
Eideren: This would convey the logic more clearly I think:
```cs
await Task.Delay(TimeSpan.FromSeconds(3));
Entity.Scene = null;
```

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/PlayerController.cs
```diff
@@ -0,0 +1,174 @@
...
+using System.Collections.Generic;
+
+namespace _2DPlatformer.Player;
+
+public class PlayerController : SyncScript
+{
+    private CharacterComponent characterComponent;
+    
+    [Display("Character Sprite Sheet")]
+    public SpriteSheet CoinSprites { get; set; }
```
Eideren: Unused

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/PlayerController.cs
```diff
@@ -0,0 +1,174 @@
...
+using System.Windows.Input;
+using Stride.Graphics;
+using Stride.Rendering.Sprites;
+using System.Collections.Generic;
+
+namespace _2DPlatformer.Player;
+
+public class PlayerController : SyncScript
+{
+    private CharacterComponent characterComponent;
```
Eideren: Same as the others, this should be a required property

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/PlayerController.cs
```diff
@@ -0,0 +1,174 @@
...
+    
+    [Display("Character Sprite Sheet")]
+    public SpriteSheet CoinSprites { get; set; }
+    private SpriteFromSheet spriteComponent;
```
Eideren: Same here

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/PlayerController.cs
```diff
@@ -0,0 +1,174 @@
...
+         HandleInput();
+    }
+    
+    /// <summary>
+    /// Handles both the Input of the Player via Keyboard (WAD- & Arrow-Keys) and the resulting Animation at the end according to input
+    /// </summary>
+    private void HandleInput()
+    {
+         if (Input.HasKeyboard)
+            {
```
Eideren: Incorrect spacing/increment

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/PlayerController.cs
```diff
@@ -0,0 +1,174 @@
...
+                    moveDirection = Vector3.UnitX * MOVE_SPEED;
+                }
+                
+                // Up (Jump)
+                if (Input.IsKeyPressed(Keys.W) && characterComponent.IsGrounded || Input.IsKeyPressed(Keys.Up) && characterComponent.IsGrounded)
+                {
+                    characterComponent.Jump();
+                }
+                characterComponent.SetVelocity(moveDirection);
+                HandleAnimation(isMoving);
```
Eideren: This should be called inside of `Update`, it doesn't depend on the input whether the user has a keyboard. `HandleInput` could return `isMoving` which is then passed to `HandleAnimation` for example

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/PlayerController.cs
```diff
@@ -0,0 +1,174 @@
...
+                    if (!isFacingRight)
+                    {
+                        isFacingRight = true;
+                        characterComponent.Orientation = Quaternion.RotationY(MathUtil.DegreesToRadians(0));
+                    }
+                    moveDirection = Vector3.UnitX * MOVE_SPEED;
+                }
+                
+                // Up (Jump)
+                if (Input.IsKeyPressed(Keys.W) && characterComponent.IsGrounded || Input.IsKeyPressed(Keys.Up) && characterComponent.IsGrounded)
```
Eideren: Can be simplified into `if ((Input.IsKeyPressed(Keys.W) || Input.IsKeyPressed(Keys.Up)) && characterComponent.IsGrounded)`

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/PlayerController.cs
```diff
@@ -0,0 +1,174 @@
...
+    
+    /// <summary>
+    /// When there is movement to either side, play run-animation. Which loops frames 4-19
+    /// </summary>
+    private void PlayRunAnimation()
+    {
+        if (animationTimer >= animationInterval)
+        {
+            animationTimer -= animationInterval;
+            spriteComponent.CurrentFrame = IDLE_FRAME_END + runFrame;
```
Eideren: Did you mean to use RUN_FRAME_END ?

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/PlayerController.cs
D4rkDuck: yeah I was using just one at first, but ran into problems. Maybe just my bad math skills 😅

Ethereal77:
As a last step after having done all of that (to avoid too much hassle), but before calling it done, I would propose to rename the template and its related files and projects to `Platformer2D` instead of `_2DPlatformer`. It is pretty unconventional to name programming things starting with a digit, and imho the underscore is ugly.

VaclavElias:
> As a last step after having done all of that (to avoid too much hassle), but before calling it done, I would propose to rename..

This makes sense. I had to do the same in the toolkit.. many things are now Something2D..

D4rkDuck:
Alright, but before I create a new project with the same settings: Is there an easy way to renaming?

Because I assume with projects there are always some kind of references set somewhere that relate to the name and I can't just go through each file and `_2DPlatformer` -> `Platformer2D`, same with file names right?

So what I'm saying is, it's easiest to just create a new project with Platformer2D as it's name + namespace.

Ethereal77:
Whatever is easier for you. For these kind of things, I usually do this:
1. Close VS.
2. Do a bulk rename of directories and files.
3. Here I do a commit so the file renaming is registered in Git (even if they still contain incorrect data).
4. Then I do a search and replace in all text files (code files: `*.cs`, project files: `*.sln`, `*.csproj`, and Stride files: `*.sdsl`, `*.sdfx`, `*.sdpkg`, and the `*.sd*` asset files).
5. Then I try to compile to verify that all is still working as intended.
6. Finally, I do the commit that includes all this replacing, and done!

This way, the history of the files is correctly preserved.

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/Gameplay/CoinRotation.cs
D4rkDuck: I wasn't familiar with `required`. From what I read, I need to init Sprite on object creation. In my case I have Prefabs of the Coins in the Scene and never Instantiate the Prefabs via Code.
So required would block setting the Sprite. Or am I misunderstanding/missing the way to use it here?

Eideren: This engine has an analyzer that runs when build games which lets users know when a `required` property of a scene or prefab object is unset. Meaning that if a designer forgets to set your `Sprite` property in the property grid in the game studio, a warning or error will be displayed on build mentioning that said property should be set.

It's fairly straightforward to reason about; you should add this keyword whenever you have a variable that must be set for the component to work appropriately. When the value of a property is *required* for that thing to do its job.

I'll try and find the time to document this behavior in https://doc.stride3d.net/latest/en/manual/scripts/serialization.html

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/PlayerController.cs
D4rkDuck: No, since I have 1 spritesheet with all animations, I used `IDLE_FRAME_END` instead of a magic number as discussed with @Ethereal77 before.
This could theoretically also be re-named to RUN_FRAME_START, but in turn will make it more confusing in the IDLE Animation since this is used for module on the Idle Animation and the basis of Starting the Run Animation. (Run Animation comes after the Idle Animation in the Sheet)

Eideren: Right, I missed the fact that the sheet is shared between all animations

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/Gameplay/CoinRotation.cs
D4rkDuck: Ah I see, if I mark something as required via Script that I expect to be a Component I can select it via Editor.
I guess this ensures that the component is attached to the object since the developer needs to manually select it without getting build errors as you mentioned.
Something to get used to, but feels safer. Which I would expect is the core idea of using this?

Eideren: You can select it in the editor whether it has required or not, but yes, the idea is that required means that you can figure out before even running your game whether running it would result in your logic throwing null reference exceptions when accessing properties that are not properly assigned

samples/Templates/2DPlatformer/_2DPlatformer/_2DPlatformer.Game/Gameplay/CollectCoin.cs
D4rkDuck: By this you mean the sfxInstance?
Is using a `?` after the Classname enough or is a `#nullable annotation context` required?
Sorry for the proably very basic questions, I am not very experienced in using C# 😅

Eideren: No biggie, yep, `private SoundInstance? sfxInstance;`

D4rkDuck:
@Eideren so I pushed the code changes, but I guess the required fields set via the editor are saved in a different file. Looking through the files like scene and prefabs I am unable to see where exactly those values are saved (what has been selected for the required fields). Any idea? The only difference I see between my local project and the template is a difference in UUIDs

Eideren:
@D4rkDuck Maybe you forgot to reload the code or something broke during reload, but here is the required property highlighted in the editor on your coin prefab
image
And the warnings when building the game
image

D4rkDuck:
> @D4rkDuck Maybe you forgot to reload the code or something broke during reload, but here is the required property highlighted in the editor on your coin prefab image And the warnings when building the game image

Yes I commented about 20minutes ago that the required fields are missing haha. How would I get the changes? I am testing the code in a local project and then copy stuff into the template-repo. So just copying the code from the .cs files isn't enough for the component selection in the editor.

Eideren:
Ah, right, my bad, I didn't understand your issue. All objects and properties in prefabs and scenes are saved in their respective `*.sdprefab` and `*.sdscene` files in your `2DPlatformer\Assets` directory.
Once you've assigned the properties in the editor, you can take those files and overwrite those in your template

D4rkDuck:
> Ah, right, my bad, I didn't understand your issue. All objects and properties in prefabs and scenes are saved in their respective `*.sdprefab` and `*.sdscene` files in your `2DPlatformer\Assets` directory. Once you've assigned the properties in the editor, you can take those files and overwrite those in your template

hmm what exactly is needed to be overwritten? Taking just the content of the coin prefab and scene is not enough and seems to break the project (assets can't be found and scene doesn't exist). I assume that is because of conflicting IDs set in those files (main ID at the top and all ref ids). So the simplest would be: 
1. Create a `Platformer2D` project from the template
2. Set references for required fields
3. Copy and replace whole files from the local project into the template folder (not sure what the boundaries are, as in what files actually need to be replaced and which not)

Eideren:
Right, you could copy/paste the whole asset directory, best to open the project directly in the sample directory and correct it there though, which I did and pushed in this PR to avoid wasting your time ;)

D4rkDuck:
> Right, you could copy/paste the whole asset directory, best to open the project directly in the sample directory and correct it there though, which I did and pushed in this PR to avoid wasting your time ;)

Ah I see, new things learned. Thanks!

samples/Templates/Platformer2D/Platformer2D/Platformer2D.Game/Gameplay/CollectCoin.cs
Eideren: It is not, stripped from latest commit.

Eideren:
Finally took the time to merge it in, thanks a bunch @D4rkDuck

Issue #2916, opened 2025-09-30
Add a yes/no to all for multiple asset import


Acissathar:
**Is your feature request related to a problem? Please describe.**
Importing multiple assets into Stride can be a bit of a pain, especially coming from Unity where drag and drop handles asset copying for you.

While Stride does support this with prompts, it can very quickly add up as you're prompted with a dialogue box to copy the object to resources for every single asset, an additional one for the location (if choosing yes), and lastly, one more potential box if it already exists in that location.

Depending on the number of assets, this can quickly become a hassle to have to click 3+ times just to get them imported. 

**Describe the solution you'd like**
A simple Yes to all / No to all during the first dialogue prompt, along with a Yes to all for the overwrite prompt.

This reduces it to a maximum of 3 prompts (needs to be copied, location, overwrite) no matter the amount of assets being imported.

**Describe alternatives you've considered**
The alternative is just manually click Yes/No, which is the current behaivor.

**Additional context**
I threw together a quick concept idea in [my fork](https://github.com/Acissathar/stride/tree/feat-Add-Yes-To-All-for-asset-import) that seems to solve it. Granted I've only really tested FBX, and it also relies on [issue 2903](https://github.com/stride3d/stride/issues/2903) being fixed as it will trigger the IsPrimitiveType check with a null value (as Stride auto selects everything imported), but I think this could be useful to have as a compromise between Unity's auto copy without prompt and Stride's prompt (and directory selection) before copy.

PR #3052, opened 2026-02-04
feat: Add Yes/No to all for batch asset import

Acissathar:
# PR Details

This change modifies the asset import to help with the flow of importing more than one item at a time. Importing multiple assets into Stride can be a bit of a pain, especially coming from Unity where drag and drop handles asset source copying for you.

While Stride does support multiple asset imports, it can very quickly add up as you're prompted with a dialogue box to copy the object to resources for every single asset, an additional one for the location (if choosing yes), and lastly, one more potential box if it already exists in that location.

Depending on the number of assets, this can quickly become a hassle to have to click 3+ times per asset just to get them imported.

The intent of this is to allow a user to pick once and handle it for all assets, or individually if still desired.

When picking a different location to copy the resources to, this makes the assumption (if you hit yes to all) that you want everything copied to that location. Otherwise, hitting Yes will prompt per asset as it currently does.

## Related Issue

#2916 

## Types of changes

## Checklist

There is currently a failure around BuildResult and an assembly reference in SampleTestFixture, but since this is mentioned in another PR, I don't believe its related to these changes.

### DIFF
--- a/sources/editor/Stride.Core.Assets.Editor/ViewModel/AssetCollectionViewModel.cs
+++ b/sources/editor/Stride.Core.Assets.Editor/ViewModel/AssetCollectionViewModel.cs
@@ -630,7 +630,7 @@ namespace Stride.Core.Assets.Editor.ViewModel
                     {
                         var message = Tr._p("Message", "Source file '{0}' is not inside of your project's resource folders, do you want to copy it?").ToFormat(file.FullPath);
 
-                        var copyResult = await Dialogs.MessageBoxAsync(message, files.Count > 1 && i != files.Count - 1 ? copyPromptWithToAllButtons : dialogDefaultButtons, MessageBoxImage.Warning);
+                        var copyResult = await Dialogs.MessageBoxAsync(message, i != files.Count - 1 ? copyPromptWithToAllButtons : dialogDefaultButtons, MessageBoxImage.Warning);
 
                         if (copyResult is DialogClosed or DialogNo)
                         {
@@ -667,7 +667,7 @@ namespace Stride.Core.Assets.Editor.ViewModel
                             {
                                 var message = Tr._p("Message", "The file '{0}' already exists, it will get overwritten if you continue, do you really want to proceed?").ToFormat(finalPath);
 
-                                var copyResult = await Dialogs.MessageBoxAsync(message, files.Count > 1 && i != files.Count - 1 ? overwritePromptWithToAllButtons : dialogDefaultButtons, MessageBoxImage.Warning);
+                                var copyResult = await Dialogs.MessageBoxAsync(message, i != files.Count - 1 ? overwritePromptWithToAllButtons : dialogDefaultButtons, MessageBoxImage.Warning);
 
                                 overwriteAll = copyResult is DialogYesToAll;
 
--- a/sources/editor/Stride.Core.Assets.Editor/ViewModel/AssetCollectionViewModel.cs
+++ b/sources/editor/Stride.Core.Assets.Editor/ViewModel/AssetCollectionViewModel.cs
@@ -1,13 +1,10 @@
 // Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net) and Silicon Studio Corp. (https://www.siliconstudio.co.jp)
 // Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
-using System;
-using System.Collections.Generic;
 using System.Collections.Specialized;
 using System.IO;
-using System.Linq;
-using System.Threading;
-using System.Threading.Tasks;
 using System.Threading.Tasks.Dataflow;
+using Stride.Core;
+using Stride.Core.Annotations;
 using Stride.Core.Assets.Analysis;
 using Stride.Core.Assets.Editor.Components.AddAssets;
 using Stride.Core.Assets.Editor.Components.Properties;
@@ -19,8 +16,6 @@ using Stride.Core.Assets.Editor.View.Behaviors;
 using Stride.Core.Assets.Editor.ViewModel.Progress;
 using Stride.Core.Assets.Templates;
 using Stride.Core.Assets.Tracking;
-using Stride.Core;
-using Stride.Core.Annotations;
 using Stride.Core.Diagnostics;
 using Stride.Core.Extensions;
 using Stride.Core.IO;
@@ -30,9 +25,9 @@ using Stride.Core.Presentation.Core;
 using Stride.Core.Presentation.Dirtiables;
 using Stride.Core.Presentation.Interop;
 using Stride.Core.Presentation.Services;
-using Stride.Core.Translation;
 using Stride.Core.Presentation.ViewModels;
-using System.Collections;
+using Stride.Core.Presentation.Windows;
+using Stride.Core.Translation;
 
 namespace Stride.Core.Assets.Editor.ViewModel
 {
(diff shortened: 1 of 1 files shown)


Acissathar:
Marked as a draft as I have a fair bit of test cases I want to run through locally before saying its ready, but I likely won't get to them until later this week/weekend so wanted to throw it up incase anyone had thoughts / nitpicks.

Test cases I plan to run through (think this covers all scenarios)

Single File Import

- [x] 	Single File Yes - Clicking Yes on a single file copies it to resources and creates the asset.
- [x]  Single File No - Clicking No on a single file skips it without copying, but still creates an asset.
- [x]  Single File In Resources - File already in resources skips the copy dialog and creates the asset directly.

Multiple Files - Basic Flow
- [x] Multiple Yes to All - Clicking "Yes to all" on the first file copies all remaining files to the same directory without additional prompts.
- [x] Multiple No to All - Clicking "No to all" on the first file aborts the entire import process immediately.
- [x] Multiple Yes Each - Clicking Yes individually for each file prompts for directory selection for each file separately.
- [x] Multiple Mixed Yes/No - Mixing Yes and No responses copies only the files where Yes was selected, skipping others.
- [x] Multiple Yes Then Yes to All - Clicking "Yes to all" after some individual Yes responses copies remaining files to the same directory without further prompts.
- [x] Multiple Dialog Close - Closing the dialog with X button skips the current file and continues prompting for remaining files.

Mixed File Locations
- [x] Mixed Yes to All - Files already in resources are imported directly while outside files prompt once with "Yes to all" copying all to same directory.
- [x] Mixed Sequential - Only files outside resources prompt for copy dialog while in-resource files import silently.
Overwrite Scenarios
- [x] Single Overwrite Yes - Overwriting a single existing file shows both copy and overwrite dialogs with Yes/No only.
- [x] Multiple Overwrite Yes to All - "Yes to all" on copy and overwrite dialogs overwrites all files with two total prompts.
- [x] Multiple Overwrite No - Clicking No on any overwrite dialog aborts the entire import and returns gracefully.
- [x] Multiple Overwrite Mixed - Clicking "Yes to all" on overwrite after individual confirmations overwrites remaining files automatically.

Edge Cases
- [x] Directory Picker Cancel - Canceling the directory picker uses the default resource folder location.
- [x] Large File Count - Importing 10+ files with "Yes to all" completes efficiently in a single prompt.

Button Index Verification
- [x] Button Mapping Validation - All buttons (0=Close, 1=Yes, 2=No, 3=Yes to all, 4=No to all) trigger their corresponding actions correctly.

Acissathar:
Ran through all my test cases and everything seems to work as expected and consistent with previous behavior where applicable, should be ready!

sources/editor/Stride.Core.Assets.Editor/ViewModel/AssetCollectionViewModel.cs
```diff
@@ -587,39 +582,103 @@ [new FilePickerFilter("") { Patterns = [file.GetFileExtension()]}],
...
 
-                    string finalPath = await GetAssetCopyDirectory(directory, file);
+                        if (copyResult is DialogYes or DialogYesToAll)
+                        {
+                            finalPath = await GetAssetCopyDirectory(directory, file);
+                        }
+                    }
+                    else
+                    {
+                        // If "Yes to all" we're going to assume they want to use the same directory as the initial file.
```
ds5678, line 657: Is this a valid assumption? It sounds like it would flatten any hierarchy during the copy.

sources/editor/Stride.Core.Assets.Editor/ViewModel/AssetCollectionViewModel.cs
```diff
@@ -587,39 +582,103 @@ [new FilePickerFilter("") { Patterns = [file.GetFileExtension()]}],
...
+            [
+                Tr._p("Button", "Yes"),
+                Tr._p("Button", "No"),
+                Tr._p("Button", "Yes to all")
+            ]
+            :
+            [
+                Tr._p("Button", "Yes"),
+                Tr._p("Button", "No")
+            ], 1, 2);
```
ds5678: This would produce mildly undesirable behavior when the user reaches the last file, by including the "to all" buttons.

sources/editor/Stride.Core.Assets.Editor/ViewModel/AssetCollectionViewModel.cs
Acissathar: I might be misunderstanding, but I believe this assumption is valid as this is consistent with the current behavior. If you select No to the default location, and specify a directory in the window that pops up, it puts it right where you tell it to without sub folder generation vs saying Yes to the default, which puts it in Resources/subfolder

sources/editor/Stride.Core.Assets.Editor/ViewModel/AssetCollectionViewModel.cs
```diff
@@ -587,39 +582,99 @@ [new FilePickerFilter("") { Patterns = [file.GetFileExtension()]}],
...
+                    {
                         continue;
+                    }
+
+                    if (!yesToAll)
+                    {
+                        var message = Tr._p("Message", "Source file '{0}' is not inside of your project's resource folders, do you want to copy it?").ToFormat(file.FullPath);
 
-                    var message = Tr._p("Message", "Source file '{0}' is not inside of your project's resource folders, do you want to copy it?").ToFormat(file.FullPath);
+                        var copyResult = await Dialogs.MessageBoxAsync(message, files.Count > 1 && i != files.Count - 1 ? copyPromptWithToAllButtons : dialogDefaultButtons, MessageBoxImage.Warning);
```
ds5678: Checking `i != files.Count - 1` makes checking `files.Count > 1` unnecessary.

sources/editor/Stride.Core.Assets.Editor/ViewModel/AssetCollectionViewModel.cs
```diff
@@ -587,39 +582,99 @@ [new FilePickerFilter("") { Patterns = [file.GetFileExtension()]}],
...
                         Directory.CreateDirectory(Path.GetDirectoryName(finalPath));
                         if (File.Exists(finalPath))
                         {
-                            message = Tr._p("Message", "The file '{0}' already exists, it will get overwritten if you continue, do you really want to proceed?").ToFormat(finalPath);
+                            if (!overwriteAll)
+                            {
+                                var message = Tr._p("Message", "The file '{0}' already exists, it will get overwritten if you continue, do you really want to proceed?").ToFormat(finalPath);
 
-                            copyResult = await Dialogs.MessageBoxAsync(message, MessageBoxButton.YesNo, MessageBoxImage.Warning);
+                                var copyResult = await Dialogs.MessageBoxAsync(message, files.Count > 1 && i != files.Count - 1 ? overwritePromptWithToAllButtons : dialogDefaultButtons, MessageBoxImage.Warning);
```
ds5678: Checking `i != files.Count - 1` makes checking `files.Count > 1` unnecessary.

sources/editor/Stride.Core.Assets.Editor/ViewModel/AssetCollectionViewModel.cs
```diff
@@ -587,39 +582,99 @@ [new FilePickerFilter("") { Patterns = [file.GetFileExtension()]}],
...
-                    string finalPath = await GetAssetCopyDirectory(directory, file);
+                        if (copyResult is DialogYes or DialogYesToAll)
+                        {
+                            finalPath = await GetAssetCopyDirectory(directory, file);
+                        }
+                    }
+                    else
+                    {
+                        // If "Yes to all" we're going to assume they want to use the same directory as the initial file.
+                        finalPath = Path.Combine(Path.GetDirectoryName(finalPath), file.GetFileName());
```
ds5678, line 658:
```suggestion
                        finalPath = Path.Join(Path.GetDirectoryName(finalPath), file.GetFileName());
```

Acissathar: I'm not opposed to the change personally, my only hang up is that the rest of Stride uses Path.Combine. I could not find any uses of Path.Join in the code base, so I'm hesitant to deviate from that consistency.

ds5678: Path.Join is newer, faster, and likely exhibits behavior closer to coder intentions. The primary difference between the two is:
* Path.Join is essentially concatenation.
* Path.Combine checks if later parameters are rooted and excludes earlier parameters if they are.

sources/editor/Stride.Core.Assets.Editor/ViewModel/AssetCollectionViewModel.cs
```diff
@@ -552,7 +547,7 @@ private string ComputeNamespace(DirectoryBaseViewModel directory)
         private async Task<string> GetAssetCopyDirectory(DirectoryBaseViewModel directory, UFile file)
         {
             var path = directory.Path;
-            var message = Tr._p("Message", "Do you want to place the resource in the default location ?");
+            var message = Tr._p("Message", "Do you want to place the resource in the default location?");
```
Kryptos-FR: While I understand why you might want to fix that, it will break existing translations since the key string will be different. Either revert that change or update all the impacted .po and .pot files.

sources/editor/Stride.Core.Assets.Editor/ViewModel/AssetCollectionViewModel.cs
Kryptos-FR: While @ds5678 remark is valid, let's keep Path.Combine for now. We can explore Path.Join as a separate cleanup task throughout the whole codebase.

Kryptos-FR:
> Multiple Overwrite No - Clicking No on any overwrite dialog aborts the entire import and returns gracefully.

Is that what we want? Shouldn't "No" skip the overwrite of the current file but continue processing? And then we would need an "No To All" for the overwrite case as well.

I genuinely don't know. I do feel like answering No to a single question shouldn't abort everything. Maybe we need a Cancel button to abort the whole import at any point.

sources/editor/Stride.Core.Assets.Editor/ViewModel/AssetCollectionViewModel.cs
Acissathar: Sounds good! I completely missed this change, it might have just been picked up by some auto cleanup. I'll revert it for now

Acissathar:
> > Multiple Overwrite No - Clicking No on any overwrite dialog aborts the entire import and returns gracefully.
> 
> Is that what we want? Shouldn't "No" skip the overwrite of the current file but continue processing? And then we would need an "No To All" for the overwrite case as well.
> 
> I genuinely don't know. I do feel like answering No to a single question shouldn't abort everything. Maybe we need a Cancel button to abort the whole import at any point.

Running through it again, I agree that a single No shouldn't cancel everything, but rather just continue in line and ask. Since the message boxes have an X, I think using that as cancel would work as an abort? I'm not super familiar with standard UI/UX, so if an explicit cancel button should be added that shouldn't be any trouble either.

At a minimum though, I'll refactor this to not end early if the user explicitly presses No (and add in No to All).

Issue #3021, opened 2025-12-27
feat: Font compilation support for Linux


Jklawreszuk:
**Is your feature request related to a problem? Please describe.**
Currently, Stride Asset Compiler does not support FileFont compilation on Linux

**Describe the solution you'd like**
Currently FileFontProvider relies on DirectWrite - windows only dependency, contributor willing to help needs to replace that with any popular glyph/font rendering library.

**Additional context**

Here are some questions that can aid in the description of the feature request:

- What are the use cases?
- What kind of options or settings are expected or wanted?
- What would the (pseudo) code look like?
- Does the feature affect behavior in the editor?
  - If so, how should this look in the editor?
- Is there a component or data structure involved?
- What kind of documentation is needed or needs to be updated?
- Should there be an example project?

Issue #2584, opened 2024-12-26
Improve SDF Font Outlines Rendering for Better Compatibility with Various Font Styles


ourabigdev:
## **Release Type**  
GitHub (master branch)  

## **Version**  
Master branch  

## **Platform(s)**  
Windows

## **Describe the Problem**  
This issue pertains to the implementation of text outline support, introduced in [PR #2582](https://github.com/stride3d/stride/pull/2582). The feature is functional and usable, but there are some rendering issues when working with certain font styles. Specifically:  
- Fonts with rounder shapes tend to render well even with larger outline thicknesses.  
- Fonts with sharp or thin features may display visible gaps between outline offsets as the outline thickness grows.  

While the current implementation is a good starting point, further refinements are necessary to ensure consistent rendering across all font styles.  

## **Steps to Reproduce**  
1. Use the master branch containing the changes from [PR #2582](https://github.com/stride3d/stride/pull/2582).  
2. Set up a `TextBlock` with a font of your choice.  
3. Assign the font type as `SpriteFontType.SDF`.  
4. Apply an `OutlineThickness` value greater than 0 to the `TextBlock`.  
5. Test with both round and sharp/thin fonts.  
6. Observe:  
   - Round fonts render outlines correctly.  
   - Sharp/thin fonts may exhibit visible gaps as the outline thickness increases.  

## **Expected Behavior**  
Outlines for all fonts (round, sharp, or thin) should render smoothly without gaps or artifacts, regardless of the outline thickness.  

## **Screenshots**  
https://github.com/user-attachments/assets/a58dd2c8-356d-4cac-b699-00acbc8773e8

## **Additional Context**  
This implementation of text outline support is a functional addition to the renderer, introduced in [PR #2582](https://github.com/stride3d/stride/pull/2582). While it works well for most cases, improvements are needed to address edge cases, especially for fonts with sharp or thin features.  

### Next Steps  
- Explore a method to dynamically adjust kernel sizes based on the font style to mitigate gaps.  
- Investigate shader-based solutions for smoother rendering.  
- Open to further testing and feedback to refine the feature.  

This is a significant step forward for text rendering in Stride, and I will continue to work on improving this feature. Feedback and suggestion and contributions are welcome!

PR #3055, opened 2026-02-04
feat: Dynamic Runtime Rasterized MSDF Sprite Font

yuechen-li-dev:
# PR Details

This PR adds an experimental runtime MSDF font path to Stride’s sprite font system.

Screenshot 2026-02-04 003111

### Introduction

Stride’s current sprite font rasterization options each have trade-offs. Dynamically generating **MSDF glyphs on-demand** is the common approach in modern engines (e.g., Unity/Unreal): it combines the benefits of Stride’s **runtime rasterized** fonts (large character sets like CJK) with **offline SDF/MSDF** advantages (smooth edges at high resolution, easy scaling) with minimal downsides.

### Implementation

Dynamic MSDF generation differs substantially from Stride’s existing static SDF pipeline.

- Uses **SharpFont** (already a Stride dependency) to extract **glyph outlines** on-demand when glyphs are requested at runtime.
- Generates **MSDF textures asynchronously** via a bounded **Channel + worker pool**, then uploads into the font atlas.
- The MSDF rasterizer is intentionally **modular / swappable** to keep the pipeline future-proof.

#### Rasterizer backends

Two managed, cross-platform MSDFGen ports are included (no P/Invoke):

- [Remora.MSDFGen](https://github.com/Remora/Remora.MSDFGen/): lighter/older, fewer features
- [MSDFGen-Sharp / msdfgen.core](https://github.com/ExtraBinoss/MSDFGen-Sharp): newer, more features, slightly better results on some fonts

Also included:
- Two debug rasterizers:
  - a simple pattern output backend
  - an outline diagnostic backend

### Notes / limitations

- "Default Size" parameter is used for baking, keep it relatively large (64+) for best results. 
- Some fonts (especially those with **overlapping / self-intersecting contours**) can produce repeatable MSDF artifacts (notably in punctuation/CJK). Current workaround: preprocess fonts in **FontForge** (Remove Overlap / Simplify / Correct Direction). This keeps runtime lightweight and avoids bundling heavy geometry cleanup.
- **Game Studio “Play” vs running from Visual Studio** can differ in dependency deployment. Ensure the MSDF backend assemblies are present in the runtime output; if needed, run the game via VS at least once so dependencies get copied/resolved. Otherwise, fonts may render in the editor but not in-game.
- **Thumbnail preview** for runtime MSDF fonts is currently blank; I wasn’t able to find a clean fix yet.

### Testing

- Verified glyph generation + atlas upload in the editor.
- Verified runtime behavior when launched from Visual Studio.
- Used debug backends to validate atlas upload/sampling independently of MSDF generation.

### Core files:

- `RuntimeSignedDistanceFieldSpriteFont.cs`
- `SharpFontOutlineExtractor.cs`
- `MsdfGenCoreRasterizer.cs` (MSDFGen-Sharp backend)
- `RemoraMsdfRasterizer.cs`
- Pipeline glue (MsdfGenerationPipeline.cs, etc.)

Thanks. This took a lot longer than I thought it would.

#### Comparison
Existing Runtime Rasterized Font:
image
New Runtime MSDF Font:
image

## Related Issue

#2584

## Types of changes

## Checklist


yuechen-li-dev:
@dotnet-policy-service agree

Eideren:
We already have an offline SDF option - but I understand that the reason you implemented this feature is because non-latin character sets tend to be very large and require a high resolution. Offline SDF we currently have wouldn't work because it would have to cover multiple thousands of characters with a fairly high resolution, potentially blowing out build sizes.

Definitely cool to see, now, is there a particular reason why this PR specifically sidesteps using most of the logic already setup for offline SDF ?

Also, as it stands, the limitations seems a bit like a show stopper. Our engine already has enough quirk as it is, best to fix those before we merge this in

yuechen-li-dev:
Thanks for the thoughtful review - I agree with the motivation you outlined (CJK + large character sets + resolution/build-size constraints is the core reason for doing this at runtime).

### Why this approach doesn’t reuse most of the offline SDF pipeline

Stride’s current offline SDF path is essentially a font compiler wrapper around an external tool (`msdfgen.exe`) to generate bitmap plus some atlas bookkeeping called through DirectWrite. In other words, the “real” work (outline extraction, contour cleanup/orientation, MSDF generation, shape normalization) happens inside the external binary, not inside Stride’s runtime code, as we can see in [`SignedDistanceFieldFontImporter.cs`](https://github.com/stride3d/stride/blob/master/sources/engine/Stride.Assets/SpriteFont/Compiler/SignedDistanceFieldFontImporter.cs) here.

For runtime MSDF, we need a very different shape of solution:
- **no external exe** (shipping an msdfgen tool / invoking processes at runtime isn’t viable, especially cross-platform)
- **async generation + caching**, because glyph generation is heavy and needs to happen incrementally under gameplay timing constraints
- **incremental atlas updates** while the game is running

So there isn’t much logic to reuse without effectively rewriting the offline pipeline into a reusable in-process library anyway, which is what this PR does, but oriented around runtime constraints.

This approach lets us keep the pipeline cross-platform and embeddable (no msdfgen.exe / no DirectWrite-only path), and it would cleanly resolve #3021 (because no platform specific font compilation is required, as atlas generation happens at runtime using managed portable C# libraries, rather than an external Windows-specific toolchain).

Finally, because of the modular architecture, this approach unlocks future extensions, such as MTSDF alpha map based effects (outlines, glow, rich text effects) which can be done by evolving just the rasterizer/shader instead of reworking the full asset pipeline again.

To address notes and limitations I listed:
- This implementation is fully functional. The only real issue is where the font thumbnail for Runtime SDF fonts in the Game Editor do not render correctly. Attempts to isolate a fix for this specific UI element currently impact standard font thumbnails; given it is purely cosmetic, it has been left as-is to preserve the integrity of the runtime rendering.
- For optimal quality and performance, it is recommended to set a relatively large default size in the asset properties. This defines the "bake" resolution; however, the font remains fully resizable within TextBlocks without loss of clarity.
- Like the existing offline MSDF implementation, certain fonts with overlapping or self-intersecting contours may display artifacts. This is a characteristic of the MSDF algorithm needing a clearly defined "inside" and "outside" and not an issue with this implementation. If artifacts appear, preprocessing the font file to simplify contours once before import is an easy solution. Integrating a SkiaSharp based preprocessing step was considered but omitted to avoid unnecessary project complexity for a niche edge case.

This is from release version of stride to demonstrate these artifacts on offline SDF fonts: 
Font used: https://fonts.google.com/noto/specimen/Noto+Sans+SC, the static semi-bold version from the zip file. 
image
Fixed after preprocessing:
image

Eideren:
Thank you for the clarification, the source for msdfgen isn't part of this repository, but we do maintain our own version through https://github.com/stride3d/msdfgen - which could have been used as the base for a c# port or integration, but doesn't matter too much now, even more so with the fixes you've introduced.

I'll try to find the time to take a look at it this week end, thanks for the work.

sources/engine/Stride.Assets/SpriteFont/SpriteFontAssetCompiler.cs
```diff
@@ -201,6 +217,40 @@ protected override Task<ResultStatus> DoCommandOverride(ICommandContext commandC
...
+
+                var runtimeSdfType = (RuntimeSignedDistanceFieldSpriteFontType)Parameters.FontType;
+
+                var sdfFont = FontDataFactory.NewRuntimeSignedDistanceField(
+                    runtimeSdfType.Size,
+                    Parameters.FontSource.GetFontName(),
+                    Parameters.FontSource.Style,
+                    runtimeSdfType.PixelRange,
+                    runtimeSdfType.Padding,
+                    useKerning: false,
```
Eideren, line 236: Is there a reason why kerning is always off, on this call site and all other usage I could find

sources/engine/Stride.Assets/SpriteFont/SpriteFontAssetCompiler.cs
```diff
@@ -201,6 +217,40 @@ protected override Task<ResultStatus> DoCommandOverride(ICommandContext commandC
...
+                : base(url, description, assetFinder)
+            {
+            }
+
+            protected override Task<ResultStatus> DoCommandOverride(ICommandContext commandContext)
+            {
+                // M1 NOTE:
+                // This is a scaffolding step. We serialize a functional font so the pipeline works end-to-end.
+                // In M3/M4, this will be replaced by a real Runtime MSDF font object.
+                commandContext.Logger.Warning("Runtime SDF font is currently scaffolded in M1 (temporary). It will behave like a runtime raster font until the MSDF runtime generator is implemented.");
```
Eideren: This seems either unrelated or too loosely documented, what do you mean by M1 and co. ?

sources/engine/Stride.Graphics/Font/RuntimeMSDF/DummyTestRasterizer.cs
```diff
@@ -0,0 +1,133 @@
...
+using Stride.Core.Mathematics;
+using Stride.Graphics.Font;
+
+namespace Stride.Graphics.Font.RuntimeMsdf
+{
+    /// <summary>
+    /// Dummy MSDF rasterizer that generates simple test patterns.
+    /// Use this to isolate pipeline issues from MSDF generation issues.
+    /// </summary>
+    public sealed class DummyTestRasterizer : IGlyphMsdfRasterizer
```
Eideren, line 11: Could you set it up as a test in `Stride.Graphics.Tests` ?

sources/engine/Stride.Graphics/Font/RuntimeMSDF/MsdfGenCoreRasterizer.cs
```diff
@@ -0,0 +1,283 @@
...
+        /// </summary>
+        private static void EdgeColoringInkTrap(Shape shape, double angleThreshold)
+        {
+            const double crossThreshold = 0.05; // sin(~3 degrees) for detecting corners
+            
+            foreach (var contour in shape.Contours)
+            {
+                if (contour.Edges.Count == 0) continue;
+
+                EdgeColor[] colors = { EdgeColor.CYAN, EdgeColor.MAGENTA, EdgeColor.YELLOW };
```
Eideren, line 157: This could be a static array, or a stack allocated one

sources/engine/Stride.Graphics/Font/RuntimeMSDF/MsdfGenCoreRasterizer.cs
```diff
@@ -0,0 +1,283 @@
...
+                    // Find a color that doesn't conflict with neighbors
+                    EdgeColor chosen = colors[0];
+                    foreach (var c in colors)
+                    {
+                        if (c != prevColor && c != nextColor)
+                        {
+                            chosen = c;
+                            break;
+                        }
+                    }
```
Eideren, line 184: You only need to test the two other colors afaict ?

sources/engine/Stride.Graphics/Font/RuntimeMSDF/OutlineDiagnosticRasterizer.cs
```diff
@@ -0,0 +1,216 @@
...
+using Stride.Core.Mathematics;
+using Stride.Graphics.Font;
+
+namespace Stride.Graphics.Font.RuntimeMsdf
+{
+    /// <summary>
+    /// Diagnostic rasterizer that visualizes the outline data to identify extraction issues.
+    /// Draws the outline directly without MSDF to see if the geometry is correct.
+    /// </summary>
+    public sealed class OutlineDiagnosticRasterizer : IGlyphMsdfRasterizer
```
Eideren, line 12: This could also be setup as a test

sources/engine/Stride.Graphics/Font/RuntimeMSDF/SharpFontOutlineExtractor.cs
```diff
@@ -0,0 +1,129 @@
...
+            if (face == null) return false;
+
+            try
+            {
+                face.LoadChar(charCode, loadFlags, LoadTarget.Normal);
+            }
+            catch (FreeTypeException)
+            {
+                return false;
+            }
```
Eideren: Does it throw only in the case where the character is not present in the font ? If not this should be surfaced in some way I think ? You could push it to a global logger `internal static readonly Logger Log = GlobalLogger.GetLogger(nameof(SharpFontOutlineExtractor));`

sources/engine/Stride.Graphics/Font/RuntimeMSDF/SharpFontOutlineExtractor.cs
```diff
@@ -0,0 +1,129 @@
...
+        {
+            var result = new GlyphOutline();
+            GlyphContour currentContour = null;
+            Vector2 lastPoint = Vector2.Zero;
+
+            // FreeType automatically closes contours, so we don't need to add closing segments
+            var funcs = new OutlineFuncs(
+                moveTo: (ref FTVector to, IntPtr user) =>
+                {
+                    currentContour = new GlyphContour { IsClosed = true };
```
Eideren: `IsClosed` is never used and always true, why is that ?

sources/engine/Stride.Graphics/Font/RuntimeMSDF/SharpFontOutlineExtractor.cs
```diff
@@ -0,0 +1,129 @@
...
+            ft.Decompose(funcs, IntPtr.Zero);
+
+            return result;
+        }
+
+        private static Vector2 ConvertVector(FTVector v)
+        {
+            // FreeType outline points are in 26.6 fixed point format
+            // Even though FTVector uses Fixed16Dot16 type, the actual values are 26.6
+            // This is a quirk of the SharpFont wrapper that's included in Stride.
```
Eideren: Might want to create an issue about this on the repo ? https://github.com/stride3d/SharpFont

sources/engine/Stride.Graphics/Font/CharacterBitmapRgba.cs
```diff
@@ -0,0 +1,135 @@
...
+    /// <summary>
+    /// An RGBA bitmap representing a glyph (4 bytes per pixel).
+    /// Intended for runtime MSDF font (stored in RGB, alpha optional).
+    /// </summary>
+    internal sealed class CharacterBitmapRgba : IDisposable
+    {
+        private readonly int width;
+        private readonly int rows;
+        private readonly int pitch;
+        private readonly IntPtr buffer;
```
Eideren, line 19: This doesn't have to be an IntPtr, or have the class as disposable.
You should replace this with a `Color[]`, the only non native usage is in `CommandList.UpdateSubResource` which has two signature for `ReadOnlySpan<T>`

sources/engine/Stride.Graphics/Font/CharacterBitmapRgba.cs
```diff
@@ -0,0 +1,135 @@
...
+            if (width != 0 && rows != 0)
+            {
+                buffer = MemoryUtilities.Allocate(checked(pitch * rows), 1);
+            }
+        }
+
+        /// <summary>
+        /// Allocates an RGBA bitmap and copies data from a source buffer with the given pitch.
+        /// </summary>
+        public unsafe CharacterBitmapRgba(IntPtr srcRgba, int width, int rows, int srcPitchBytes)
```
Eideren, line 51: Unused

sources/engine/Stride.Graphics/Font/FontCacheManagerMsdf.cs
```diff
@@ -0,0 +1,223 @@
...
+            {
+                glyph.IsUploaded = false;
+                if (glyph.Owner != null)
+                    glyph.Owner.IsBitmapUploaded = false;
+            }
+                      
+            cachedGlyphs.Clear();
+            packer.Clear(cacheTextures[0].ViewWidth, cacheTextures[0].ViewHeight);
+
+            System.Diagnostics.Debug.WriteLine("MSDF ClearCache()");
```
Eideren: Did you mean to leave this one in ?

sources/engine/Stride.Graphics/Font/FontSystem.cs
```diff
@@ -79,6 +82,7 @@ public void Unload()
             // TODO possibly save generated characters bitmaps on the disk
             FontManager.Dispose();
             FontCacheManager.Dispose();
+            FontCacheManagerMsdf?.Dispose();
```
Eideren: Redundant `?` operator, or mark the property as `CanBeNull`

sources/engine/Stride.Graphics/Font/FontManager.Msdf.cs
```diff
@@ -0,0 +1,61 @@
...
+            FontStyle fontStyle,
+            float size,
+            char character,
+            out GlyphOutline outline,
+            out GlyphOutlineMetrics metrics,
+            LoadFlags loadFlags = LoadFlags.NoBitmap | LoadFlags.NoHinting)
+        {
+            return TryGetGlyphOutline(fontFamily, fontStyle, new Vector2(size, size), character, out outline, out metrics, loadFlags);
+        }
+    }
```
Eideren: Not sure how necessary splitting the class is for two methods, 'should be fine to leave it inside the base FontManager

sources/engine/Stride.Graphics/Font/FontCacheManagerMsdf.cs
```diff
@@ -0,0 +1,223 @@
...
+            }
+
+            if (!packer.Insert(bitmap.Width + atlasPad * 2, bitmap.Rows + atlasPad * 2, ref subrect))
+            {
+                if (!EnsureSpaceFor(bitmap.Width, bitmap.Rows, atlasPad))
+                    throw new InvalidOperationException("MSDF glyph does not fit in cache even after eviction.");
+
+                if (!packer.Insert(bitmap.Width + atlasPad * 2, bitmap.Rows + atlasPad * 2, ref subrect))
+                    throw new InvalidOperationException("MSDF cache allocation failed unexpectedly after eviction.");
+            }
```
Eideren, line 84: Shouldn't this be taken care of in the glyph thread instead ? Doesn't seem like it necessarily depends on the mainthread's context

sources/engine/Stride.Graphics/Font/RuntimeSignedDistanceFieldSpriteFont.cs
```diff
@@ -0,0 +1,653 @@
...
+                DrainUploads(commandList);
+
+            // 3) Upload
+            if (spec.IsBitmapUploaded && cacheRecords.TryGetValue(key, out var handle))
+            {
+                // If evicted/cleared, this will flip false and we’ll reupload next draw
+                if (!handle.IsUploaded)
+                {
+                    spec.IsBitmapUploaded = false;
+                    cacheRecords.TryRemove(key, out _); ;
```
Eideren: Duplicate `;`

sources/engine/Stride.Graphics/Font/RuntimeSignedDistanceFieldSpriteFont.cs
```diff
@@ -0,0 +1,653 @@
...
+                            // CPU SDF build
+                            var sdf = generator.Generate(item.Input, item.Params);
+
+                            // hand off to render thread for GPU upload
+                            readyForUpload.Enqueue((item.Key, sdf));
+                        }
+                        catch
+                        {
+                            // If generation fails, we just allow rescheduling later.
+                        }
```
Eideren, line 282: We should be a bit more explicit here, which kind of exceptions does it make sense to catch, which should we send to be logged by some logger ?

sources/engine/Stride.Graphics/Font/RuntimeSignedDistanceFieldSpriteFont.cs
```diff
@@ -0,0 +1,653 @@
...
+                }
+                else if (outline?.Bounds.Width > 0 && outline.Bounds.Height > 0)
+                {
+                    w = Math.Max(1, (int)MathF.Ceiling(outline.Bounds.Width));
+                    h = Math.Max(1, (int)MathF.Ceiling(outline.Bounds.Height));
+                }
+
+                if (w > 0 && h > 0)
+                {
+                    var input = (GlyphInput)new OutlineInput(outline, w, h);
```
Eideren: OutlineInput derives from GlyphInput, pretty sure you can omit the cast

sources/engine/Stride.Graphics/Font/RuntimeSignedDistanceFieldSpriteFont.cs
```diff
@@ -0,0 +1,653 @@
...
+
+                    if (workChannel.Writer.TryWrite(new WorkItem(key, input, p)))
+                        return;
+                }
+                // If outline path can't be scheduled (no dims / queue full), fall back to coverage below.
```
Eideren: May fall down when w == 0 and h == 0, does that state make sense for everything going down lower ?
Also, you set `FullMode = BoundedChannelFullMode.Wait`, so when the queue is full, TryWrite will return false and the system will generate a coverage-based sdf instead ... ?

sources/engine/Stride.Graphics/Font/RuntimeSignedDistanceFieldSpriteFont.cs
```diff
@@ -0,0 +1,653 @@
...
+            spec.Glyph.Subrect = handle.InnerSubrect;
+            spec.Glyph.BitmapIndex = bitmapIndex;
+            spec.IsBitmapUploaded = true;
+
+            cacheRecords[key] = handle;
+            cache.NotifyGlyphUtilization(handle);
+
+        }
+
+        private void DrainUploads(CommandList commandList, int maxUploadsPerFrame = 8)
```
Eideren, line 514: Eight new characters visible per frame seems excessively small, is the upload cost that significant ? I would prefer that count to be configurable and larger by default. The property for that could live in one of the font-related systems

sources/engine/Stride.Graphics/Font/RuntimeSignedDistanceFieldSpriteFont.cs
```diff
@@ -0,0 +1,653 @@
...
+        }
+
+        private static GlyphKey MakeKey(char c, DistanceFieldParams p) => new(c, p.PixelRange, p.Pad);
+
+        // --- Generator input: discriminated union (coverage today, outline later) ---
+        private abstract record GlyphInput;
+
+        private sealed record CoverageInput(
+            byte[] Buffer,
+            int Length,
```
Eideren: Length is unused

sources/engine/Stride.Graphics/Font/RuntimeSignedDistanceFieldSpriteFont.cs
```diff
@@ -0,0 +1,653 @@
...
+                        new DistanceFieldSettings(
+                            PixelRange: p.PixelRange,
+                            Padding: p.Pad,
+                            Width: o.Width,
+                            Height: o.Height),
+                        msdfEncode),
+
+                    _ => throw new ArgumentOutOfRangeException(nameof(input)),
+                };
+        }
```
Eideren: Move inner type definitions to the end of the file

sources/engine/Stride.Graphics/Font/RuntimeSignedDistanceFieldSpriteFont.cs
```diff
@@ -0,0 +1,653 @@
...
+                        return GetGlyph(commandList, DefaultCharacter.Value, in fontSize, uploadGpuResources, out fixScaling);
+
+                    return null;
+                }
+            }
+
+            // 2) Schedule async SDF generation (only once per char)
+            EnsureSdfScheduled(key, spec);
+            if (commandList != null)
+                DrainUploads(commandList);
```
Eideren, line 152: `DrainUploads` is called for every char that is drawn on screen, I don't think that's expected on your part.

sources/engine/Stride.Graphics/Font/RuntimeSignedDistanceFieldSpriteFont.cs
```diff
@@ -0,0 +1,653 @@
...
+        {
+            FontType = SpriteFontType.SDF;
+        }
+
+        public override bool IsCharPresent(char c)
+        {
+            return FontManager != null && FontManager.DoesFontContains(FontName, Style, c);
+        }
+
+        protected override Glyph GetGlyph(CommandList commandList, char character, in Vector2 fontSize, bool uploadGpuResources, out Vector2 fixScaling)
```
Eideren, line 109: This method is especially slow for something that is called for every character that gets drawn on screen, for loaded characters, you have one concurrent dictionary read, one concurrent dictionary write, one linked list removal and another add (which, granted, those last two aren't necessarily that slow).

From what I can see, you can reduce it to that but only for each unique character, any character that was already drawn this frame could be amortized into a single dictionary tryget

sources/engine/Stride.Graphics/Font/RuntimeSignedDistanceFieldSpriteFont.cs
```diff
@@ -0,0 +1,653 @@
...
+            }
+        }
+
+        private void EnsureSdfScheduled(GlyphKey key, CharacterSpecification spec)
+        {
+            // Already uploaded? nothing to do.
+            if (spec.IsBitmapUploaded) return;
+
+            // Ensure worker infrastructure is alive (safe even if already started)
+            EnsureWorkersStarted();
```
Eideren, line 305: Isn't the call from the `FontSystem` setter enough ?

sources/engine/Stride.Graphics/Font/RuntimeSignedDistanceFieldSpriteFont.cs
```diff
@@ -0,0 +1,653 @@
...
+
+        // Runtime SDF glyph cache key (future-proof for multiple ranges/modes)
+        private readonly ConcurrentDictionary<GlyphKey, CharacterSpecification> characters = [];
+        private readonly ConcurrentDictionary<GlyphKey, FontCacheManagerMsdf.MsdfCachedGlyph> cacheRecords = [];
+
+        [DataMemberIgnore]
+        internal FontManager FontManager => FontSystem?.FontManager;
+
+        [DataMemberIgnore]
+        internal FontCacheManagerMsdf FontCacheManagerMsdf => FontSystem?.FontCacheManagerMsdf;
```
Eideren: These can be private afaict ?

sources/engine/Stride.Graphics/Font/RuntimeSignedDistanceFieldSpriteFont.cs
```diff
@@ -0,0 +1,653 @@
...
+                // Apply padding offset
+                if (spec.Bitmap != null && spec.Glyph.XAdvance != 0)
+                {
+                    spec.Glyph.Offset -= new Vector2(p.Pad, p.Pad);
+                }
+
+                EnsureSdfScheduled(key, spec);
+
+            }
+        }
```
Eideren, line 196: Inconsistent linebreaks

sources/engine/Stride.Graphics/Font/RuntimeSignedDistanceFieldSpriteFont.cs
```diff
@@ -0,0 +1,653 @@
...
+            var pitch = bmp.Pitch;
+
+            int len = pitch * rows;
+            var srcCopy = ArrayPool<byte>.Shared.Rent(len);
+            try
+            {
+                unsafe
+                {
+                    System.Runtime.InteropServices.Marshal.Copy((IntPtr)bmp.Buffer, srcCopy, 0, len);
+                }
```
Eideren: Unsafe scope is unnecessary here, so is the cast

sources/engine/Stride.Graphics/Font/RuntimeSignedDistanceFieldSpriteFont.cs
```diff
@@ -0,0 +1,653 @@
...
+
+                var input = (GlyphInput)new CoverageInput(srcCopy, len, width, rows, pitch);
```
Eideren: Cast is unnecessary

sources/engine/Stride.Graphics/Font/RuntimeSignedDistanceFieldSpriteFont.cs
```diff
@@ -0,0 +1,653 @@
...
+
+            // Already scheduled? bail.
+            if (!inFlight.TryAdd(key, 0)) return;
+
+            var p = new DistanceFieldParams(key.PixelRange, key.Pad, DefaultEncode);
+
+            // Prefer outline-based MSDF generation when available.
+            // We still rely on the bitmap path to populate glyph metrics today, but MSDF uses the outline.
+            if (FontManager != null &&
+                FontManager.TryGetGlyphOutline(FontName, Style, Size, key.C, out var outline, out _))
```
Eideren: Shouldn't we warn users somewhere that the font they chose does not provide glyph outline, maybe as part of the build steps for a Runtime SDF asset ?

Eideren (changes requested):
Looks really good, I was able to repro the missing characters when building a game, haven't had time to look into it yet, but we can't quite merge before this is fixed.
Still, I have a fair amount of things to run by you.

yuechen-li-dev:
> Looks really good, I was able to repro the missing characters when building a game, haven't had time to look into it yet, but we can't quite merge before this is fixed. 

Yeah, my findings on this issue are that the atlas generation/upload for runtime SDF is async, and the existing font thumbnail compiler is not async aware and since it is only called once to generate the thumbnail, and since the fonts aren't ready when it is called, it returns nothing. I'm not sure what the clean way to resolve it would be, and I've experimented a lot on this.

I might be a little busy this weekend, so I'll address these gradually over the next week. Thanks for the thorough review.

sources/engine/Stride.Assets/SpriteFont/SpriteFontAssetCompiler.cs
yuechen-li-dev: These are personal milestones for me to keep track of project development. Since MVP Is complete, this can and will be removed.

sources/engine/Stride.Graphics/Font/RuntimeMSDF/DummyTestRasterizer.cs
yuechen-li-dev: These two debug rasterizers are for visual debug feedbacks. I can set up some unit test to test outline extraction/MSDF generation functionality without having to manually switch out rasterizers.

sources/engine/Stride.Graphics/Font/RuntimeSignedDistanceFieldSpriteFont.cs
yuechen-li-dev: The upload per frame is not "how many new characters can be displayed/rendered per frame"(if it is, then I would agree 8 character per frame would be excessively small), it is "how many unique glyphs MSDF texture can we bake into the font atlas per frame". With the current architecture, each unique character only needs to be baked into the atlas once per game session, and not once per character displayed, to be rendered at any scale in game afterwards.

This limit will probably never be hit under normal gameplay scenarios, especially for Latin character based languages as there are only 26 characters total + punctuation. This is a guardrail against the worst case scenario: If a player tries to paste a huge block of 500+ unique CJK characters into the game, it ensure the upload queue won't suddenly be overloaded, for which I think 8 per frame (baking 240-480 unique glyphs per second is plenty)

sources/engine/Stride.Graphics/Font/RuntimeSignedDistanceFieldSpriteFont.cs
yuechen-li-dev: That was something I was unsure about when coding this, but it really looks like all fonts would always have a glyph outline and the extraction process is fairly direct. I will set up some unit tests for this, but ideally, I would like to see if it is possible to include an open source .ttf font file （I would suggest Noto Sans CJK Regular for coverage） so the unit tests aren't reliant on checking what font does the tester have installed on their computer.

sources/engine/Stride.Graphics/Font/FontSystem.cs
yuechen-li-dev: Fixed.

sources/engine/Stride.Graphics/Font/FontManager.Msdf.cs
yuechen-li-dev: This is mostly for easy reviewability more than anything else. I'll go ahead and leave it in then.

sources/engine/Stride.Graphics/Font/RuntimeSignedDistanceFieldSpriteFont.cs
yuechen-li-dev: Fixed.

sources/engine/Stride.Graphics/Font/RuntimeSignedDistanceFieldSpriteFont.cs
yuechen-li-dev: Fixed.

sources/engine/Stride.Graphics/Font/RuntimeSignedDistanceFieldSpriteFont.cs
yuechen-li-dev: Fixed.

sources/engine/Stride.Graphics/Font/RuntimeMSDF/SharpFontOutlineExtractor.cs
yuechen-li-dev: SharpFont has been a large source of frustration for this project as it is not thread safe, so outline extraction has to be done on main thread in sync while MSDF generation is async and is what I considered to be the current biggest bottleneck for this implementation. 

Since even upstream SharpFont has been abandoned for 8 years now, I would recommend moving towards a managed native C# FreeType2 port like [FreeTypeSharp](https://github.com/ryancheung/FreeTypeSharp) as the next logical upgrade instead of continuing to patch up SharpFont.

sources/engine/Stride.Graphics/Font/RuntimeSignedDistanceFieldSpriteFont.cs
yuechen-li-dev: Fixed.

sources/engine/Stride.Graphics/Font/RuntimeSignedDistanceFieldSpriteFont.cs
Eideren: I actually got that part, but saw how significant the delay was before characters are shown on screen and expected that to come from this limit. 8 per frame should be good enough like you say, the delay I saw doesn't match, it likely comes from the actual building of each character's msdf.

sources/engine/Stride.Graphics/Font/RuntimeSignedDistanceFieldSpriteFont.cs
Eideren: Sounds good, we have a test setup for japanese over here
https://github.com/stride3d/stride/blob/master/sources/engine/Stride.Graphics.Tests/TestDynamicSpriteFontJapanese.cs 
Which uses a font bundled with the repo:
https://github.com/stride3d/stride/blob/master/sources/data/tests/Fonts/SourceHanSans-Light.otf
Feel free to look at how that one is setup to build yours, as long as the font is compatible with MIT it'll be fine.

sources/engine/Stride.Assets/SpriteFont/SpriteFontAssetCompiler.cs
yuechen-li-dev: The initial basis of this project is an exact copy of the runtime rasterized font that was already in the system. 

https://github.com/stride3d/stride/blob/master/sources/engine/Stride.Assets/SpriteFont/SpriteFontAssetCompiler.cs

And kerning was set as off for it, so I just copied.

sources/engine/Stride.Assets/SpriteFont/SpriteFontAssetCompiler.cs
yuechen-li-dev: I just had the warning to label it as an experimental feature.

sources/engine/Stride.Graphics/Font/FontManager.Msdf.cs
yuechen-li-dev: Merged and removed partial.

sources/engine/Stride.Graphics/Font/RuntimeSignedDistanceFieldSpriteFont.cs
yuechen-li-dev: You don't HAVE to leave it in, but it's recommended to do so as keeping them both is much safer and carries zero performance penalties as it is a simple null check if already initialized, and it will prevent crashes if for whatever reason the worker isn't available in EnsureSdfScheduled. Classic lazy initialization pattern.

sources/engine/Stride.Graphics/Font/RuntimeSignedDistanceFieldSpriteFont.cs
yuechen-li-dev: I should have formatted my comments better: DrainUploads is the step that copies the CPU generated MSDF texture into the GPU atlas and disposed of the CPU bitmap. It's not called for every character that's drawn on screen, but once for each unique character that is uploaded to the atlas.

sources/engine/Stride.Graphics/Font/RuntimeSignedDistanceFieldSpriteFont.cs
yuechen-li-dev: Yeah, MSDF generation is VERY heavy. You can try running one of my commits with bitmap outlined SDF before I implemented async if you are curious.

sources/engine/Stride.Graphics/Font/RuntimeSignedDistanceFieldSpriteFont.cs
yuechen-li-dev: Very good catch, you are right, I think you helped me figured out the reason behind the repeated unexplained upload of empty character that I've seen in debug. Refactored the EnsureSDFScheduled method so it should be fixed now.

sources/engine/Stride.Graphics/Font/RuntimeSignedDistanceFieldSpriteFont.cs
yuechen-li-dev: Fixed, removed during method refactor.

sources/engine/Stride.Graphics/Font/RuntimeSignedDistanceFieldSpriteFont.cs
yuechen-li-dev: Removed and fixed.

sources/engine/Stride.Graphics/Font/FontCacheManagerMsdf.cs
yuechen-li-dev: Removed, and fixed.

sources/engine/Stride.Graphics/Font/FontCacheManagerMsdf.cs
yuechen-li-dev: Fixed. Moved the safety check to the main file as it is more efficient and won't throw.

sources/engine/Stride.Graphics/Font/RuntimeMSDF/SharpFontOutlineExtractor.cs
yuechen-li-dev: Had a much more complicated method previously to build outline previously before I found FontSharp's builtin method that automatically closes the outline. Fixed.

sources/engine/Stride.Graphics/Font/CharacterBitmapRgba.cs
yuechen-li-dev: I made another branch to use Color[] without the dispose, however, it doesn't seem to display anything. This one is probably one of these changes that requires a lot of debugging and I suggest that we leave it until later.

Eideren:
Changing this to draft while you work on it, let me know when you're ready for another review :)

yuechen-li-dev:
> Changing this to draft while you work on it, let me know when you're ready for another review :)

Yeah, sorry, I don't think I will be able to work on this for quite some time as the direct results from my work on Stride had led to some major projects that are taking priority right now, so unfortunately, I would not be able to work on this for a while. Feel free to poke around the code and continue my work, the main concern from me that I'm honestly unsure what to do about is that the thumbnail generation is not async aware, and I don't want to rewrite the entire font pipeline to fit async in when it doesn't really need it since async kinda infects everything, and of course, harden the entire MSDF pipeline via unit tests, which should be fairly simple with LLMs.

Thanks for the support, I'll circle back to this when I have time.

sources/engine/Stride.Graphics/Font/FontManager.cs
```diff
@@ -307,10 +307,11 @@ private void LoadFontInMemory(string fontPath)
                 return;
 
             // Load the font from the database
-            using var fontStream = contentManager.OpenAsStream(fontPath);
-            // create the font data from the stream
-            var newFontData = new byte[fontStream.Length];
-            fontStream.ReadExactly(newFontData);
+            using (var fontStream = contentManager.OpenAsStream(fontPath))
+            {
```
Ethereal77, line 308: This change is not needed. Also, it adds a brace that is never closed?

sources/engine/Stride.Assets/SpriteFont/SpriteFontAssetCompiler.cs
```diff
@@ -201,6 +213,37 @@ protected override Task<ResultStatus> DoCommandOverride(ICommandContext commandC
             }
         }
 
+        internal class RuntimeSignedDistanceFieldFontCommand : AssetCommand<SpriteFontAsset>
```
Ethereal77, line 217: Just a little thought: Is not this file a bit full of classes already? Would not make more sense to split the internal classes to their own files to ease discovery?

sources/engine/Stride.Graphics/Font/FontManager.cs
yuechen-li-dev: Weird. I just merged master into this feature branch from VS and took incoming change to get this branch sync'd since I have a little bit of time to work on this. Give me a min, let me look it over.

sources/engine/Stride.Graphics/Font/RuntimeMSDF/SharpFontOutlineExtractor.cs
Ethereal77: Another alternative worth considering is [SixLabors Fonts](https://github.com/SixLabors/Fonts.git) (from the creators of ImageSharp). Btw, it also does text layouting

sources/engine/Stride.Graphics/Font/CharacterBitmapRgba.cs
Ethereal77: If there is a bug somewhere when using `Color[]` it would be nice to know if it is from your code or from Stride.

Also, these "leave until later" tend to be forgotten once a PR is merged, and is a big sorce of technical debt. Imho, better to pinpoint the culprit and at least create an issue than to sweep it under the rug.

sources/engine/Stride.Graphics/Font/FontManager.cs
```diff
@@ -360,13 +356,46 @@ private void BuildBitmapThread()
                         RenderBitmap(character, fontFace);
                     }
 
-                DequeueRequest:
+DequeueRequest:
 
-                    // update the generated cached data
+// update the generated cached data
```
Ethereal77: Indentation

sources/engine/Stride.Graphics/Font/FontSystem.cs
```diff
@@ -61,7 +64,7 @@ public void Load(GraphicsDevice graphicsDevice, IDatabaseFileProviderService fil
         /// <param name="defaultSize">The default font size in pixels.</param>
         /// <param name="style">The font style.</param>
         /// <returns>A <see cref="SpriteFont"/> instance if the font is registered; otherwise, <c>null</c>.</returns>
-        public SpriteFont? LoadRuntimeFont(string fontName, float defaultSize = 16f, FontStyle style = FontStyle.Regular)
+        public SpriteFont LoadRuntimeFont(string fontName, float defaultSize = 16f, FontStyle style = FontStyle.Regular)
```
Ethereal77: The XML docs still says `otherwise, <c>null</c>`. Why was the return type changed from `SpriteFont?` to non-null `SpriteFont`?

sources/engine/Stride.Graphics/Font/RuntimeSignedDistanceFieldSpriteFont.cs
```diff
@@ -0,0 +1,645 @@
...
+
+namespace Stride.Graphics.Font
+{    /// <summary>
+     /// A dynamic font that asynchronously generates multi-channel signed distance mapping for glyphs as needed, enabling sharp, smooth edges and resizability.
+     /// </summary>
+    [ReferenceSerializer, DataSerializerGlobal(typeof(ReferenceSerializer<RuntimeSignedDistanceFieldSpriteFont>), Profile = "Content")]
+    [ContentSerializer(typeof(RuntimeSignedDistanceFieldSpriteFontContentSerializer))]
+    [DataSerializer(typeof(RuntimeSignedDistanceFieldSpriteFontSerializer))]
+
+    internal sealed class RuntimeSignedDistanceFieldSpriteFont : SpriteFont
```
Ethereal77: Blank line between attributes and class, Comments on the same line as the brace

sources/engine/Stride.Graphics/Font/RuntimeSignedDistanceFieldSpriteFont.cs
```diff
@@ -0,0 +1,645 @@
...
+
+        private readonly record struct DistanceEncodeParams(float Bias, float Scale);
+        private readonly record struct DistanceFieldParams(int PixelRange, int Pad, DistanceEncodeParams Encode);
+
+        private readonly record struct GlyphKey(char C, int PixelRange, int Pad);
+
+        // Keep today’s encoding behavior explicit and centralized.
+        private static readonly DistanceEncodeParams DefaultEncode = new(Bias: 0.4f, Scale: 0.5f);
+
+        private DistanceFieldParams GetDfParams()
```
Ethereal77: Try to not use abbreviations: `GetDfParams()` -> `GetDistanceFieldParams()` or something alike. Reads better in callstacks and is apparent what it does at a quick glance

sources/engine/Stride.Graphics/Font/RuntimeSignedDistanceFieldSpriteFont.cs
Ethereal77: `PerformPendingUploads()`? At least to me "drain" sounds like discarding

VaclavElias:
> Another alternative worth considering is [SixLabors Fonts](https://github.com/SixLabors/Fonts.git) (from the creators of ImageSharp). Btw, it also does text layouting

@Ethereal77, not sure if the license is ok here https://github.com/SixLabors/Fonts/blob/main/LICENSE, and how it would affect commercial games.

sources/engine/Stride.Graphics/Font/RuntimeSignedDistanceFieldSpriteFont.cs
Ethereal77: Just an idea: Maybe a "warming" API could be introduced so at least some amount of common characters are already baked and ready for the first time they are needed

sources/engine/Stride.Graphics/Font/RuntimeSignedDistanceFieldSpriteFont.cs
Ethereal77, line 1: I think this file has too many responsibilities and is too long. For contrast, `SignedDistanceFieldSpriteFont` is just the serializable definition of a SDF sprite font. This file is not only that, but also has infrastructure for asynchronous atlas building, EDT transform computation, and maybe other things.

I would split those parts to the systems in charge of managing fonts, or to components specialized in each of those tasks that can as an added benefit be tested in isolation (and maybe even used in some other part where they may be beneficial)

Ethereal77:
> > Another alternative worth considering is [SixLabors Fonts](https://github.com/SixLabors/Fonts.git) (from the creators of ImageSharp). Btw, it also does text layouting
> 
> @Ethereal77, not sure if the license is ok here https://github.com/SixLabors/Fonts/blob/main/LICENSE, and how it would affect commercial games.

I think they changed the license some time ago to allow free use for open-source projects, and to be also free if the library is consumed transitively (i.e., through Stride). The paid license would only apply if your project is closed-source and you are using the library directly. Correct me if I'm wrong

VaclavElias:
> ..and to be also free if the library is consumed transitively (i.e., through Stride). The paid license would only apply if your project is closed-source and you are using the library directly. Correct me if I'm wrong

I believe you are right.

sources/engine/Stride.Graphics/Font/RuntimeSignedDistanceFieldSpriteFont.cs
yuechen-li-dev: Fixed after partial class split.

sources/engine/Stride.Graphics/Font/RuntimeSignedDistanceFieldSpriteFont.cs
yuechen-li-dev: Should be fixed after the partial class split.

sources/engine/Stride.Graphics/Font/FontManager.cs
yuechen-li-dev: Fixed.

sources/engine/Stride.Graphics/Font/RuntimeSignedDistanceFieldSpriteFont.cs
yuechen-li-dev: To avoid a major architecture rewrite, I've split the file into 3 partial classes: the font file, the generator/rasterizer seams, and the fallback bitmap based SDF builder. If there is need, I think there could be a bigger split later on, but the seams are already pretty clean.

sources/engine/Stride.Assets/SpriteFont/SpriteFontAssetCompiler.cs
yuechen-li-dev: I don't think I've added much to this file other than what is needed to implement runtime MSDF, and I'd like to keep the scope of this PR narrow to just feature implementation cleanly right now. Refactoring/cleaning up existing file will probably have to be another pass.

sources/engine/Stride.Graphics/Font/RuntimeMSDF/SharpFontOutlineExtractor.cs
yuechen-li-dev: I think a full refactor of this is outside the scope of this PR, so it will have to be done in a separate PR during another spike. I already have an idea of how I wanted to approach this, in case someone wants to follow up after me if I get too busy.

So for this, only outline extraction is needed. So, I think a lighter pure Csharp library like [Typography.OpenFont](https://github.com/LayoutFarm/Typography/tree/master/Typography.OpenFont) is probably better for this application. The idea is, write a Csharp interface, route the existing SharpFontOutlineExtractor against it, then build a TypographyOpenFontOutlineExtractor, test it out, confirm it works, build unit tests to lock it in, and then only remove the SharpFontOutlineExtractor when it is safe.

sources/engine/Stride.Graphics/Font/RuntimeMSDF/SharpFontOutlineExtractor.cs
yuechen-li-dev: I haven't seen it throw any exceptions even when the character is not present in the font actually, when I type a CJK character into say, Arial, it just returns a blank character.

sources/engine/Stride.Graphics/Font/FontSystem.cs
yuechen-li-dev: Good catch. I think I removed it too hastily without looking at it too carefully. File should be properly nullable now.

Kryptos-FR:
That's also my take. Our project is open source so our direct consumption is fine. For users of the engine, it falls under transitive dependency which is also covered by the opensource license.

We just need to be careful and monitor any change in licensing terms.

xen2:
FYI I have merged #3129 and #3131 today.

Question: can this new msdfgen C# implementation be used as a replacement for msdfgen.exe at asset compilation time too?
That would be one less dependency blocking us the way for cross-platform asset compilation.

yuechen-li-dev:
> FYI I have merged #3129 and #3131 today.
> 
> Question: can this new msdfgen C# implementation be used as a replacement for msdfgen.exe at asset compilation time too? That would be one less dependency blocking us the way for cross-platform asset compilation.

Easily. You can just call the rasterizer, which is designed to be reusable, or implemented it with the library. It's one of the key blockers before implementing something like rounded corner rectangles in the UI, which relies very heavily on SDF.

https://github.com/yuechen-li-dev/stride/blob/runtime-msdf-font/sources/engine/Stride.Graphics/Font/RuntimeMSDF/MsdfGenCoreRasterizer.cs

xen2:
> > FYI I have merged #3129 and #3131 today.
> > Question: can this new msdfgen C# implementation be used as a replacement for msdfgen.exe at asset compilation time too? That would be one less dependency blocking us the way for cross-platform asset compilation.
> 
> Easily. You can just call the rasterizer, which is designed to be reusable, or implemented it with the library. It's one of the key blockers before implementing something like rounded corner rectangles in the UI, which relies very heavily on SDF.
> 
> https://github.com/yuechen-li-dev/stride/blob/runtime-msdf-font/sources/engine/Stride.Graphics/Font/RuntimeMSDF/MsdfGenCoreRasterizer.cs

Great to hear that!
Do you think it could also be added as part of this PR?

yuechen-li-dev:
> > > FYI I have merged #3129 and #3131 today.
> > > Question: can this new msdfgen C# implementation be used as a replacement for msdfgen.exe at asset compilation time too? That would be one less dependency blocking us the way for cross-platform asset compilation.
> > 
> > 
> > Easily. You can just call the rasterizer, which is designed to be reusable, or implemented it with the library. It's one of the key blockers before implementing something like rounded corner rectangles in the UI, which relies very heavily on SDF.
> > https://github.com/yuechen-li-dev/stride/blob/runtime-msdf-font/sources/engine/Stride.Graphics/Font/RuntimeMSDF/MsdfGenCoreRasterizer.cs
> 
> Great to hear that! Do you think it could also be added as part of this PR?

It is already part of the PR as this is a necessary functionality for this feature to work. After this, I think msdfgen.exe can just be removed if the old static SDF font was migrated to the new rasterizer or removed altogether.

xen2:

yuechen-li-dev:
> 👋 Apologies for the noise here.
> 
> This PR was closed unintentionally by a bad force-push to `master` on my side, which auto-closed a batch of PRs. GitHub won't let me (or you) reopen it, since the close is pinned to the now-rewritten commit.
> 
> I've contacted GitHub Support to try reopening these in bulk, but no guarantees. If this change is still relevant, it's a good opportunity to sync your branch with the latest master and open a fresh PR (your code is intact).
> 
> Sorry again for the disruption! 🙏

No worries, this one is honestly ready for review again, just missing like a couple of tests I think. I've just been so busy that I haven't got the time to post that it's mostly done.

But yeah, to give you guys a preview of what I've been working on lately:

https://github.com/yuechen-li-dev/oct/blob/main/internal/prometheus/shaders/sdslv/production/reduction/softmax_fused.sdslv

Should have some interesting results soon.

Issue #3054, opened 2026-02-04
Tooltip + hover icon mismatch


metalgearsloth:
**Release Type**: Official Release

**Version**: 4.3.0.2507 but haven't bisected it.

**Platform(s)**: Windows

**Describe the bug**
If I hover the edge of the transform buttons, no tooltip shows up but it plays the hover colour change. Then when I mouseover the centre of the icon the tooltip will finally show up. Ideally the tooltip shows up if you hover the icon anywhere.

**Expected behavior**
The icon bounds and tooltip bounds to be aligned.

**Screenshots**
<img width="175" height="85" alt="Image" src="https://github.com/user-attachments/assets/3f4a1dfb-6e95-449e-a1f7-b668f4211c20" />

I can get a video to describe what I mean if required.

Acissathar:
This can be closed

PR #3059, opened 2026-02-06
fix: Update Transformation Tooltip gizmo to have transparent background property so hit registration matches highlighting

Acissathar:
# PR Details

The hover highlight and tooltip for the transform gizmos are not consistent as the tooltip will only occur when directly over the actual icon itself, while the highlight occurs when on the button.

Example:
InitialTooltipIssue

The tooltip was attached to the Border element inside the DataTemplate, causing a mismatch where hovering over the ListBoxItem's padding area triggered the highlight but didn't show the tooltip. This change moves the tooltip from the Border element into the ListBoxItem container style so that it applies to the entire item.

With change:

InitialTooltipFixedf

## Related Issue

#3054 

## Types of changes

## Checklist

### DIFF
--- a/sources/editor/Stride.Assets.Presentation/AssetEditors/EntityHierarchyEditor/Views/EntityHierarchyEditorView.xaml
+++ b/sources/editor/Stride.Assets.Presentation/AssetEditors/EntityHierarchyEditor/Views/EntityHierarchyEditorView.xaml
@@ -30,9 +30,6 @@
         <Setter Property="Width" Value="28"/>
         <Setter Property="Height" Value="28"/>
         <Setter Property="Padding" Value="0"/>
-        <Setter Property="ToolTip" Value="{Binding Converter={sd:EnumToTooltip}}"/>
-        <Setter Property="sd:ToolTipHelper.Status" Value="{Binding Editor.Status, Source={x:Static sd:SessionViewModel.Instance}}"/>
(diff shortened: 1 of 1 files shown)


sources/editor/Stride.Assets.Presentation/AssetEditors/EntityHierarchyEditor/Views/EntityHierarchyEditorView.xaml
```diff
@@ -44,8 +47,7 @@
         <Setter Property="ItemTemplate">
           <Setter.Value>
             <DataTemplate>
-              <Border Width="28" Height="28"
-                      ToolTip="{Binding Converter={sd:EnumToTooltip}}" sd:ToolTipHelper.Status="{Binding Editor.Status, Source={x:Static sd:SessionViewModel.Instance}}" ToolTipService.ShowOnDisabled="True">
+              <Border Width="28" Height="28">
```
Kryptos-FR: It seems to me, the only missing property was `Background="Transparent"` which is required for hit testing.

I prefer that solution because if we change the `DataTemplate` here to have something else around it or to add margin or padding, we will have the tooltip apply to the whole `ListBoxItem` instead of the right content.

Acissathar: Made the tweak!

Issue #3065, opened 2026-02-10
Feature Request: Support re-orderable list property values in GameStudio


Acissathar:
**Is your feature request related to a problem? Please describe.**
Currently Lists in the GameStudio editor are one and done. Once you add an item, its place is cemented and cannot be adjusted without removing and re-adding all items to the desired order.

**Describe the solution you'd like**
It looks like GameStudio at one point _did_ support re-ordering (or attempted to) in ReorderCollectionItemViewModel, however, it is all commented out with "Fix Me". 

I was not able to find any messages or issues on why/what was the original issue.

**Describe alternatives you've considered**
Alternatives are a code only approach where GameStudio doesn't matter, leaving the order as is, and/or manually adjusting it by adding/removing items.

**Additional context**
I've been messing around with it locally, but wanted to get some feels out on if anyone knows what the original problem was / why it was gutted. 

Also does the Editor rewrite make this unnecessary / not worth pursuing at the moment?

- What are the use cases?
Largely just an Editor convenience.

Kryptos-FR:
If I remember correctly, the main issues are with Undo/Redo and the Archetype system.

PR #3067, opened 2026-02-12
Feat: Restore re-orderable lists and add reordable arrays in GameStudio

Acissathar:
# PR Details

Restore re-orderable list functionality in the GameStudio editor, along with adding support for Arrays and restrict dictionaries. Most of this code was already present in the engine, it was just commented out with "FIXME" without any info on _what_ was broken.

Through local testing, the issue seemed to really just be that the re-ordering was attached to the Parent rather than the actual items of the list. This was a minor tweak, and supporting arrays just needed a small helper to store the indexes we're looking to swap.

Dictionaries I think are out of the scope of this PR, as I believe the Quantum system will need a decent amount of changes to account for needing to store a positional reference.

## Related Issue

#3065 

## Types of changes

## Checklist

In order to have an array or list be re-ordable, it needs the [MemberCollection(CanReorderItems = true)] attribute.


Acissathar:
Draft because I haven't really put it through any wringers, just a super basic test.

Test script:

    public class ReorderTest : ScriptComponent
    {
        // Can re-order
        [MemberCollection(CanReorderItems = true)]
        public List<int> Items { get; set; } = new List<int>();

        // Cannot re-order
        public List<int> NonReorderableItems { get; set; } = new List<int>();

        // Can re-order
        [MemberCollection(CanReorderItems = true)]
        public int[] ItemsArray { get; set; } = new int[4];

        // Cannot re-order
        [MemberCollection(CanReorderItems = true)]
        public Dictionary<int, int> AnotherItemDictionary { get; set; } = new Dictionary<int, int>();

        // Cannot re-order
        [MemberCollection(CanReorderItems = true)]
        public int SingleItem { get; set; }
    }
    
And GameStudiobehaivor:
        
ReorderTest

Acissathar:
Per the linked issue, I tested Undo/Redo and Archetypes as mentioned by @Kryptos-FR as potential issues with reorderable collections in the past.

Undo/Redo applies to the list element and changes done to the elements (in order of operation):

ListUndoRedo

Archetypes I don't have much practical experience with, just a general overview and assumption on how it works based on experience in Unity:

Re-ordering the derived / child asset seems to work as expected without messing with the archetype / parent:

Base Material:
image

Derived Asset changing order:
image

However, changing it on the parent / archetype asset when the child has their own changes, does cause an issue and duplicates elements rather than shifting the order:

image

I suppose the question then becomes, what _should_ it do? If the Archetype has Layers 0, 1, and 2 and on the Derived asset we swap layers 1 and 2 (0 - 2 - 1), then go on the Archetype and swap 0 and 1 (1 - 0 - 2), what should be reflected? With prefabs, we leave overridden child changes alone when changing the base/parent, but since we're swapping elements, I don't think that really works here because then we'd copy the unchanged layer (0) but since 1 was already moved, we'd have (1 - 2 -1) which seems wrong too.

Maybe if _any_ elements are re-ordered in the given section, we ignore any changes on the Archetype on the derived asset?

I'll be gone this weekend for a trip so won't be able to take a stab at the archetypes part for a bit, so keeping it in a draft, but wanted to leave some more info.

Acissathar:
Coming back to this, this is what Unity does for reference:

Base Object:
image

Variant, swapping element 1 and 2:
image

Now go and change element 1 and 4 on the base:
image

Variant leaves the overridden values in place, and matches all non-overrides with the base:
image

Even though this ends up "duplicating" one of the elements, I suppose given the nature of the problem that's probably the most reasonable compromise.

Are there any hang ups with going forward with this as the solution here too (in addition to fixing the mismatch in element numbers)?

Eideren:
The default behavior When serializing a collection to yaml is to associate an id to each item in a collection. Those ids are shared between the archetype and its children. Not sure what Unity does, but in our case changing the order should not break, as the value to id mapping is retained when moving the item up or down the list.
Even when removed, the id still exist on the yaml's side within the child's collection, its value is just marked with a [~(Deleted)](https://github.com/stride3d/stride/blob/master/sources/assets/Stride.Core.Assets/Yaml/CollectionWithIdsSerializerBase.cs#L22) tag and is ignored when deserialized. See [TestCollectionIdsSerialization.cs](https://github.com/stride3d/stride/blob/master/sources/assets/Stride.Core.Assets.Tests/Yaml/TestCollectionIdsSerialization.cs#L197) for example.
There may be an issue preventing this behavior though, perhaps the id to value mapping fails when re-ordering on the parent as it is not notified of the change in ordering, might want to validate that this works as expected.

xen2:

Issue #2480, opened 2024-10-08
Build for Android platform fails when using Graphic Profile other than Direct3D 9.2 or Direct3D 9.3


Feralnex:
**Release Type**: Official Release

**Version**: v4.2.0.2232

**Platform(s)**: Project build for Android platform

**Describe the bug**
If You have selected Default Graphic Profile for android platform other than Direct3D 9.2 or Direct3D 9.3 and You have any Sprite sheet in Assets folder, then when You try to build project for Android platform You've got null reference exceptions and build fails.

**To Reproduce**
1. Create project for Android platform
2. Add Sprite sheet to Assets (e.g. created from png)
3. Set Default Graphic Profile to any other than Direct3D 9.2 or Direct3D 9.3 under Rendering in GameSettings (or override only for Android platform)
4. Try to build project for Android platform (Debug/Release doesn't matter)
5. Failed to build the project

**Expected behavior**
Be able to use Graphic Profiles other than Direct3D 9.2 or Direct3D 9.3 only and successfully build the project.

**Screenshots**
![Image](https://github.com/user-attachments/assets/a158e8d8-cf7d-42a5-a167-bacd82d9c9b6)
![Image](https://github.com/user-attachments/assets/c162a68c-ff4d-4163-9639-5d76d0038406)

**Log and callstacks**
[StrideAndroidPlatformBuildFailed.txt](https://github.com/user-attachments/files/17296199/StrideAndroidPlatformBuildFailed.txt)

**Additional context**
Builds just fine when using Direct3D 9.2 or Direct3D 9.3

PR #3070, opened 2026-02-16
[fix] Fixed PvrttTexLib implementation which caused Android game projects build to fail

Feralnex:
# PR Details

First of all we had incorrect P/Invoke signature in PVRTexture.cs which caused null reference exception.
Second we retrieved wrong native format when mapping stride PixelFormat to EPVRTPixelFormat and it resulted in wrong memory allocation when creating PVRTexture which caused crash when using this library - Internal CLR error (0x80131506).

## Related Issue

In this issue https://github.com/stride3d/stride/issues/2480 when Graphics Profile was higher than Direct3D 9.3 in Android game projects and any of the textures had Compress checked in Parameters then we select PvrttTexLib for the compression which was broken till now. PvrttTexLib is selected when output format is set to ETC1_* or ETC2_* if I remember correctly.

image

## Types of changes

## Checklist

### DIFF
--- a/sources/tools/Stride.TextureConverter/Backend/TexLibraries/PvrttTexLib.cs
+++ b/sources/tools/Stride.TextureConverter/Backend/TexLibraries/PvrttTexLib.cs
@@ -124,8 +124,10 @@ namespace Stride.TextureConverter.TexLibraries
             EPVRTVariableType pixelType = RetrieveNativePixelType(image.Format);
             libraryData.Header = new PVRTextureHeader(format, image.Height, image.Width, image.Depth,
                 image.MipmapCount, imageArraySize, imageFaceCount, colorSpace, pixelType);
+            // Initializing a new native texture, allocating memory.
             libraryData.Texture = new PVRTexture(libraryData.Header, IntPtr.Zero);
 
+            // Copying TexImage data into the native texture allocated memory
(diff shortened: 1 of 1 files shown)


xen2:
Since we're dropping OpenGL and OpenGL ES in next release, it is likely not necessary anymore?

Feralnex:
> Since we're dropping OpenGL and OpenGL ES in next release, it is likely not necessary anymore?

Is it taken into account to also use different compression library on mobile devices while at it?

xen2:
As far as i remember, all those formats are relic from early mobile OpenGL ES not supporting DXC and each having their own format.
Any Vulkan-compliant device should have access to the standard compression format (which simplifies a lot our code since we don't need to support all those custom formats anymore).

Still, I will merge this one for now since the library is still there in `master`, but it might be removed later if not necessary anymore.

Feralnex:
> As far as i remember, all those formats are relic from early mobile OpenGL ES not supporting DXC and each having their own format. Any Vulkan-compliant device should have access to the standard compression format (which simplifies a lot our code since we don't need to support all those custom formats anymore).
> 
> Still, I will merge this one for now since the library is still there in `master`, but it might be removed later if not necessary anymore.

Thanks for clarifying. I like the direction in which Stride development is going ^^

PR #3078, opened 2026-02-28
fix: null checking in IndexBufferBinding equals operator

johang88:
# PR Details

The `==` would throw on null arrguments causing a crash in MeshRenderFeature if no index buffer was set in the draw data.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Graphics/IndexBufferBinding.cs
+++ b/sources/engine/Stride.Graphics/IndexBufferBinding.cs
@@ -87,7 +87,7 @@ namespace Stride.Graphics
 
         public static bool operator ==(IndexBufferBinding left, IndexBufferBinding right)
         {
-            if (left is null && right is null)
+            if (ReferenceEquals(left, right))
                 return true;
 
             return left?.Equals(right) ?? false;


sources/engine/Stride.Graphics/IndexBufferBinding.cs
```diff
@@ -87,6 +87,12 @@ public override bool Equals(object obj)
 
         public static bool operator ==(IndexBufferBinding left, IndexBufferBinding right)
         {
+            if (ReferenceEquals(left, right))
```
Ethereal77: As those checks are already done at `Equals(IndexBufferBinding)`, maybe this should just be `left?.Equals(right)`. That was the idea I had at least, just forgot the `?`. Sorry about that

johang88: you mean something like this instead?
```
public static bool operator ==(IndexBufferBinding left, IndexBufferBinding right)
{
	if (left is null && right is null)
		return true;

	return left?.Equals(right) ?? false;
}
```

Ethereal77: No, what I mean is that `Equals(IndexBufferBinding other)` is already checking for `other is null` and for `ReferenceEquals(this, other)`, so for the `operator ==` a simple `return left?.Equals(right) ?? false` will suffice.
That way the comparison logic is just in one place (`Equals`).

johang88: I might be tired but won't this fail then as it wont ever enter Equals and return false?
```
IndexBufferBinding test = null;
if (test == null)
{
   test = new();
}
```

Ethereal77: The one who is sleepy is me, apparently 😁
You are entirely right. Forget my last comment; the comparison against `null` is needed in `operator ==`.

johang88: :D I'll tweak it to remove the duplicate reference comparison.

sources/engine/Stride.Graphics/IndexBufferBinding.cs
```diff
@@ -87,7 +87,10 @@ public override bool Equals(object obj)
 
         public static bool operator ==(IndexBufferBinding left, IndexBufferBinding right)
         {
-            return left.Equals(right);
+            if (left is null && right is null)
```
ds5678:
```suggestion
            if (ReferenceEquals(left, right))
```
This is more efficient than a double null check.

johang88: Makes sense

Eideren:
Thanks !

PR #3079, opened 2026-02-28
perf: use ArrayPool for temporary buffers where useful

steveberdy:
# PR Details

> Low priority

I noticed some unnecessary allocations in classes that deal with asset management. Usually these buffers are around the same size, so I opted for an array pool. 

These arrays are not read past their needed length, are not used after returning, and aren't large enough to allocate any arrays that aren't already allocated in the shared allocator, so array pools shouldn't cause any issues here.

note: also used ReadExactly instead of Read for a stream read as per CA2022.

## Types of changes

## Checklist

- [x] All new and existing tests passed. (I ran all tests, but some unrelated tests were failing)

### DIFF
--- a/sources/core/Stride.Core.Serialization/Streaming/ContentChunk.cs
+++ b/sources/core/Stride.Core.Serialization/Streaming/ContentChunk.cs
@@ -2,6 +2,7 @@
 // Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net) and Silicon Studio Corp. (https://www.siliconstudio.co.jp)
 // Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
 
+using System.Buffers;
 using System.Diagnostics;
 using System.Runtime.CompilerServices;
 #if !USE_UNMANAGED
@@ -93,7 +94,7 @@ public sealed class ContentChunk
             var chunkBytes = MemoryUtilities.Allocate(Size);
 
             var bufferCapacity = Math.Min(8192u, (uint)Size);
-            var buffer = new byte[bufferCapacity];
+            var buffer = ArrayPool<byte>.Shared.Rent((int)bufferCapacity);
 
             var count = (uint)Size;
             fixed (byte* bufferStart = buffer) // null if array is empty or null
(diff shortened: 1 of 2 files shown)


sources/core/Stride.Core.Serialization/Streaming/ContentChunk.cs
```diff
@@ -109,9 +110,10 @@ public unsafe IntPtr GetData(DatabaseFileProvider fileProvider)
                     count -= read;
                 } while (count > 0);
             }
+            ArrayPool<byte>.Shared.Return(buffer);
 #else
                 var bytes = new byte[Size];
-                stream.Read(bytes, 0, Size);
+                stream.ReadExactly(bytes, 0, Size);
```
Ethereal77, line 116: I think you can just use `stream.ReadExactly(bytes)` when the offset is zero and the length is the exact buffer length.

steveberdy: Yeah, I could do that. There are hundreds/thousands of mini performance and reliability improvements I could make on this repo just from analyzers alone.

sources/core/Stride.Core.Serialization/Streaming/ContentChunk.cs
```diff
@@ -109,9 +110,10 @@ public unsafe IntPtr GetData(DatabaseFileProvider fileProvider)
                     count -= read;
                 } while (count > 0);
             }
+            ArrayPool<byte>.Shared.Return(buffer);
```
xen2, line 113: Maybe we want a try/finally to make sure it's released?

Kryptos-FR: The doc specifically warns against using try/finally for rented arrays.

https://learn.microsoft.com/en-us/dotnet/standard/unsafe-code/best-practices#recommendations-19

https://github.com/dotnet/runtime/discussions/48257

steveberdy: In this case it's fine since it won't be used after return. But there's no reason to use try-finally here.

xen2:
I had to revert this because TestLoadAndSave failed (new test infra being merged in master will catch this)
Feel free to reopen a new PR once this is fixed.

steveberdy:
@xen2 Thanks for pointing this out. I'll look into it when I get a chance

PR #3085, opened 2026-03-07
feat: Refactor old build system into a proper Sdk

Kryptos-FR:
# PR Details

Replaces the legacy build system (`sources/targets/` + scattered `.props`/`.targets` files) with three MSBuild SDK packages under `sources/sdk/`. All 112 projects now use `<Project Sdk="Stride.Sdk">` (or `Stride.Sdk.Editor` / `Stride.Sdk.Tests`).

## Motivation

The old build system was a web of 20+ `.props` and `.targets` files imported via relative paths in every `.csproj`. This made project files verbose, fragile to move, and hard to reason about. The new SDK approach follows .NET conventions: build logic lives in versioned SDK packages, and project files only declare what makes them unique.

## What changed

**New SDK packages** (`sources/sdk/`):
| Package | Purpose |
|---------|---------|
| `Stride.Sdk` | Base SDK — platform detection, target frameworks, graphics API multi-targeting, assembly processor, native dependencies |
| `Stride.Sdk.Editor` | Composes `Stride.Sdk` + editor framework properties |
| `Stride.Sdk.Tests` | Composes `Stride.Sdk.Editor` + xunit infrastructure, test launchers, asset compilation |

**Simplified project files** — 184 `.csproj` files modified with a net reduction of ~2,100 lines. A typical project went from 30+ lines of imports and boilerplate to a clean SDK-style file.

**Deleted legacy files:**
- `sources/targets/` (20 files, ~1,500 lines) — the old build system
- Obsolete `build/*.props` and `build/*.targets` files

**Solution filters** — `.slnf` files replace separate `.sln` files for Android/iOS builds.

**CI updated** — all workflows now build the SDK packages first via a shared `build-sdk` action.

## How to review

- **Start with** `sources/sdk/Stride.Sdk/Sdk/Sdk.props` and `Sdk.targets` to understand what the SDK provides
- **Skim a few `.csproj` diffs** to see the simplification (e.g., `Stride.Core.csproj`, `Stride.Engine.csproj`)
- **Detailed guide:** `build/docs/SDK-GUIDE.md` explains the SDK hierarchy, property evaluation order, and development workflow

## Related Issue

Build system modernization — no prior issue.

## Types of changes

## Checklist


Kryptos-FR:
~~Some tests are failing. I'm investigating.~~ I haven't tried yet to run and open the GameStudio with these changes.

@xen2 this likely affects the TeamCity builds, though I tried to make sure that using Stride.build should continue to work.

build/Stride.iOS.sln
```diff
@@ -1,485 +0,0 @@
...
-EndProject
-Global
-	GlobalSection(SharedMSBuildProjectFiles) = preSolution
-		..\sources\engine\Stride.Shared\Refactor\Stride.Refactor.projitems*{b33e576f-2279-4bfc-a438-d9b84343b56b}*SharedItemsImports = 13
-		..\sources\engine\Stride.Shared\Refactor\Stride.Refactor.projitems*{c121a566-555e-42b9-9b0a-1696529a9088}*SharedItemsImports = 4
-		..\sources\shared\Stride.Core.ShellHelper\Stride.Core.ShellHelper.projitems*{e8b3553f-a79f-4e50-b75b-acee771c320c}*SharedItemsImports = 4
-		..\sources\engine\Stride.Shared\Refactor\Stride.Refactor.projitems*{fb06c76a-6bb7-40be-9afa-fec13b045fb5}*SharedItemsImports = 4
-	EndGlobalSection
-	GlobalSection(SolutionConfigurationPlatforms) = preSolution
-		Debug|iPhone = Debug|iPhone
```
xen2, line 141: FYI, I remember Xamarin.iOS had issues if the Platform was not "iPhone" (and "iPhoneSimulator" to use the simulator), which is why we couldn't reuse `Stride.Runtime.sln`.

I suspect this is not a problem anymore with `net10-android/ios` and all the recent improvements/unification, so let's keep it deleted and see what happens when we try to make iOS/Android supported again.

SUMMARY.md
```diff
@@ -0,0 +1,182 @@
+# Session Summary - Stride SDK Migration
```
xen2: This file should be removed.

sources/tools/Stride.VisualStudio.Package.Tests/Stride.VisualStudio.Package.Tests.csproj
```diff
@@ -1,16 +1,13 @@
-<Project>
-  <Import Project="..\..\targets\Stride.props" />
+<Project Sdk="Stride.Sdk.Editor">
```
xen2: I assume we don't want a `Stride.Sdk.Editor.Tests` since it's too different from our other typical unit tests projects, being a test for our VS SDK package?

sources/core/Stride.Core.IO/Stride.Core.IO.csproj.backup
```diff
@@ -0,0 +1,39 @@
+<Project>
```
xen2: File to remove

sources/core/Stride.Core.MicroThreading/Stride.Core.MicroThreading.csproj.backup
```diff
@@ -0,0 +1,27 @@
+﻿<Project>
```
xen2: File to remove

sources/core/Stride.Core.Serialization/Stride.Core.Serialization.csproj.backup
```diff
@@ -0,0 +1,37 @@
+﻿<Project>
```
xen2: File to remove

sources/core/Stride.Core.Tests/Stride.Core.Tests.csproj.backup
```diff
@@ -0,0 +1,30 @@
+<Project>
```
xen2: File to remove

sources/core/Stride.Core/Stride.Core.csproj.backup
```diff
@@ -0,0 +1,99 @@
+<Project>
```
xen2: File to remove (I won't do others, but basically all backup files)

sources/engine/Stride.Graphics.Regression/Scripts/RuniOSTest.sh
```diff
@@ -4,7 +4,7 @@ MDTOOL=/Applications/MonoDevelop.app/Contents/MacOS/mdtool
 MTOUCH=/Developer/MonoTouch/usr/bin/mtouch
 
 # build
-$MDTOOL -v build -t:Build "-c:Release|iPhone" Stride.iOS.sln
+$MDTOOL -v build -t:Build "-c:Release|iPhone" Stride.iOS.slnf
```
xen2, line 7: Platform name might need adjustment (but this file is quite outdated so let's see about it later)

sources/core/Stride.Core/Stride.Core.csproj
```diff
@@ -1,23 +1,17 @@
...
     <LangVersion>latest</LangVersion>
     <Nullable>enable</Nullable>
     <StrideCodeAnalysis>true</StrideCodeAnalysis>
+    <StrideRuntime>true</StrideRuntime>
   </PropertyGroup>
   
   <PropertyGroup>
     <StrideAssemblyProcessor>true</StrideAssemblyProcessor>
     <StrideAssemblyProcessorOptions>--auto-module-initializer --serialization</StrideAssemblyProcessorOptions>
-    <StrideBuildTags>*</StrideBuildTags>
```
xen2, line 19: FYI `StrideBuildTags` was used by the `Stride.ProjectGenerator` (used by `update_solutions.bat`).
I assume it's now all obsolete since we use `.slnf` files.
So probably`Stride.ProjectGenerator` and `update_solutions.bat` should be deleted?

sources/sdk/Stride.Build.Sdk.Tests/Sdk/Sdk.props
```diff
@@ -0,0 +1,93 @@
...
+    <!-- Custom output paths for test projects -->
+    <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
+    <AppendRuntimeIdentifierToOutputPath>false</AppendRuntimeIdentifierToOutputPath>
+    <BaseIntermediateOutputPath>obj\</BaseIntermediateOutputPath>
+
+    <!-- Allow unsafe code (common in tests) -->
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+
+    <!-- Enable assembly processor for test projects -->
+    <StrideAssemblyProcessor Condition="'$(StrideAssemblyProcessor)' == ''">true</StrideAssemblyProcessor>
```
xen2: From what I understood, Sdk.props is inserted before our project properties.
So I think it doesn't need to check if this user property is already defined? (or it should be moved to Targets file).
Same goes for everything checking existing properties from .csproj in .props file (except `TargetFramework` and all the properties derived from it,  since this it is passed as a global parameter in inner builds)

sources/sdk/Stride.Build.Sdk/Sdk/Stride.GraphicsApi.InnerBuild.targets
```diff
@@ -145,9 +186,13 @@
...
-  <!-- ==================================================================
-       Rewrote _WalkEachTargetPerFramework using @(_InnerBuildProjects)
-  -->
+  <!-- ================================================================ -->
+  <!-- Override _WalkEachTargetPerFramework for NuGet Pack              -->
+  <!-- ================================================================ -->
+  <!-- Original: NuGet.Build.Tasks.Pack.targets
+       Standard NuGet uses _ProjectsWithTFM (one per TFM). We override
+       to use @(_InnerBuildProjects) which includes TFM x API combos.
+       This ensures all graphics API variants get included in the package. -->
```
xen2: Maybe we can add a small comment that once https://github.com/NuGet/Home/issues/10556 is fixed, we could remove some of those targets.

sources/sdk/Stride.Build.Sdk/Sdk/Sdk.props
```diff
@@ -0,0 +1,48 @@
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <!--
+    ================================================================
+    Stride.Sdk - Early Property Definitions (Sdk.props)
```
xen2, line 5: Should we rename it to `Stride.Internal.Sdk` or something similar?

Rationale is, this is used for Stride build itself.
But we might want to keep `Stride.Sdk` available for user project (even thought it might not be needed as you mentioned, PackageReference + targets seems to do a good enough job).

sources/sdk/Stride.Build.Sdk.Tests/Sdk/Sdk.props
```diff
@@ -0,0 +1,93 @@
...
+  <PropertyGroup>
+    <!-- Compute test output directory using StrideRoot (set by sources/Directory.Build.props) -->
+    <_StrideTestOutputDir Condition="'$(StrideRoot)' != ''">$(StrideRoot)bin\Tests\$(MSBuildProjectName)\$(StridePlatform)\</_StrideTestOutputDir>
+    <_StrideTestOutputDir Condition="'$(StrideGraphicsApiDependent)' == 'true'">$(_StrideTestOutputDir)$(StrideGraphicsApi)\</_StrideTestOutputDir>
+
+    <!-- Set output path (use custom path if not already set) -->
+    <OutputPath Condition="'$(OutputPath)' == ''">$(_StrideTestOutputDir)</OutputPath>
+
+    <!-- Set intermediate output path -->
+    <IntermediateOutputPath Condition="'$(StrideGraphicsApiDependent)' != 'true'">$(BaseIntermediateOutputPath)$(StridePlatform)\$(Configuration)\</IntermediateOutputPath>
```
xen2: Isn't it same problem? `StrideGraphicsApiDependent` can't be accessed in .props, only .targets

build/Stride.build
```diff
@@ -28,6 +28,18 @@ Example of use:
   <Target Name="Help">
     <Message Importance="high" Text="TODO"/>
   </Target>
+
+  <!--
+  Build Stride MSBuild SDK packages (Stride.Sdk, Stride.Sdk.Editor, Stride.Sdk.Tests)
+  into the local NuGet cache. Required before any project using Sdk="Stride.Sdk" can build.
+  -->
+  <Target Name="BuildSdk">
+    <PropertyGroup>
```
xen2: What would the workflow be if we want to build Stride.sln with Visual Studio?
Do we need to open Stride.Sdk.slnx first, build it, then open Stride.sln?
If yes, would there be a way to auto build it if not present (or at least put an error message about what to do), i.e. in Directory.build.targets?

xen2 (commented):
Thanks for working on it!
It was a much needed rewrite of our build system 🏗️ 

I might need to do another pass on the details, but here is what I found so far.
Note: I didn't try to build/run anything yet.

build/Stride.iOS.sln
Kryptos-FR: The build is passing but I don't have an iPhone or macOS desktop to test it out.

sources/tools/Stride.VisualStudio.Package.Tests/Stride.VisualStudio.Package.Tests.csproj
Kryptos-FR: The `Stride.Sdk.Tests` is kind of already a `Stride.Sdk.Editor.Tests`. `Stride.Sdk.Tests` imports `Stride.Sdk.Editor` which imports `Stride.Sdk`.

I'll look into why it is not used for this project. I suspect it is because it wasn't importing the tests .props and .targets before.

sources/core/Stride.Core.IO/Stride.Core.IO.csproj.backup
Kryptos-FR: I already removed them but I forgot to push that commit 😅

sources/sdk/Stride.Build.Sdk.Tests/Sdk/Sdk.props
Kryptos-FR: I'll review that. I had back and forth for these stuff. I think it is checked in both .props and .targets especially because of inner builds.

sources/sdk/Stride.Build.Sdk/Sdk/Sdk.props
Kryptos-FR: Why not? It shouldn't be hard to rename it. I should then rename all 3 Sdks, since they depend on each other.

sources/sdk/Stride.Build.Sdk.Tests/Sdk/Sdk.props
Kryptos-FR: I think it is again because of inner builds. I'll review again the analysis the AI made. And if so, add some notes in the SDK-guide document.

build/Stride.build
Kryptos-FR: For now we do need to buid it first. I think it is hard to have it automatic since the restore phase does need the Sdk to be present. It's a chicken-egg problem.

sources/core/Stride.Core/Stride.Core.csproj
Kryptos-FR: Good point. I'll look into it.

sources/sdk/Stride.Build.Sdk/Sdk/Sdk.props
Kryptos-FR: Renamed to Stride.Build.Sdk

build/Stride.build
Kryptos-FR: I asked AI and it confirms it.

----

The reviewer is right that this is a friction point, but I agree with you that it's essentially unsolvable at the SDK level. Here's why:

**Why auto-building the SDK from Directory.Build.targets won't work**

MSBuild SDK resolution happens **before any targets or props execute.** The sequence is:

1. MSBuild reads `<Project Sdk="Stride.Build.Sdk">`
2. MSBuild looks up `Stride.Build.Sdk` version in `global.json`
3. MSBuild resolves the SDK package from NuGet cache
4. **If not found → immediate failure** (MSBuild error, not a target failure)
5. Only then does it import `Sdk.props` → `.csproj` → `Sdk.targets`

`Directory.Build.targets` runs at step 5. The failure happens at step 3/4. There's no hook to intercept "SDK not found" — it's a hard MSBuild infrastructure error, not something a target can catch.

**What about a better error message?**

Unfortunately, MSBuild already gives a clear error when an SDK isn't found:

    error MSB4236: The SDK 'Stride.Build.Sdk' specified could not be found.

We can't customize that message — it comes from MSBuild internals before any project code runs.

**What about Visual Studio specifically?**

VS does NuGet restore on solution open, but SDK resolution is a prerequisite for restore itself. The SDK must exist in the NuGet cache before VS can even parse the project files. There's no VS extensibility point to intercept this either.

**Practical options for the reviewer's concern**

1. **Documentation** — Make it clear in the repo README / contributor guide that `dotnet build sources/sdk/Stride.Build.Sdk.slnx` must be run once before opening `Stride.sln`. This is the same pattern as many repos with custom SDKs.

2. **The existing `Stride.build approach`** — `MSBuild Stride.build /t:Build` already runs `BuildSdk` first, then builds the solution. This is the command-line workflow.

3. **A setup script** — A `setup.bat`/`setup.ps1` that builds the SDK could reduce friction for first-time contributors. But this is just syntactic sugar over `dotnet build sources/sdk/Stride.Build.Sdk.slnx`.

4. **Pre-built SDK in the repo** — We could check in the `.nupkg` files to `build/packages/`, so the SDK is always available without a build step. The `nuget.config` already points there. This would make "clone and open in VS" work immediately. The downside is committing binary packages (~small, since these are mostly .props/.targets files).

Option 4 is the only one that truly solves the "open Stride.sln in VS without a prior step" problem. The packages are tiny (mostly XML files + assembly processor binaries that are already in `deps/`), so the repo size impact would be minimal.

xen2: Option 4 could be good to have.

We'll have to make sure people don't forget to rebuild them on change. (from top of my head, I can think of 2 options: some warning top of props/targets file in the SDK, or to automatize it, a check for any modified date of props/targets files in SDK folder later than nupkg itself => issue a warning/error).

Also, if doing so, please make sure to add .nupkg to .gitattributes for git LFS (BEFORE doing any `git add`).

build/Stride.iOS.sln
xen2: Yes sorry if it wasn't clear, this was just a FYI so nothing to do.
However the Android build seems broken: https://github.com/stride3d/stride/actions/runs/22866630705/job/66335122196?pr=3085

Kryptos-FR: Android already doesn't build on master

build/Stride.build
xen2: What about another option 5:

Simply keep props/targets in Sdk-like format, but add them as proper Import in the file? (with relative path)
(a bit like we were doing before, except now we can properly put them as first and last line of csproj)

Even if it doesn't look as nice as Sdk, I think it would fit better the intent: import a local target which is in the same repo
But I wonder if there would be any disadvantage? Are all the features going to work fine? Is tooling OK with that?

xen2: @Kryptos-FR what do you think?

xen2:
Except https://github.com/stride3d/stride/pull/3085#discussion_r2964138884 and some testing, is there any blocker?
I would like to merge this soon to avoid dealing too much with conflicts (i.e. on any csproj or creating new projects)

build/Stride.build
Ethereal77: Here we are talking about things needed for developing Stride. So the target user is a developer. What's wrong with having something like a `Setup.ps1` script (or `Setup.bat`) that ensures the SDK is ready, restored, and can even ensure the `AssemblyProcessor` is also up-to-date prior to opening the solution?

In many projects, specially in the C/C++ world (but also other languages), a setup / initialization script is pretty common and expected even.

Kryptos-FR: > What about another option 5:
> 
> Simply keep props/targets in Sdk-like format, but add them as proper Import in the file? (with relative path) (a bit like we were doing before, except now we can properly put them as first and last line of csproj)
> 
> Even if it doesn't look as nice as Sdk, I think it would fit better the intent: import a local target which is in the same repo But I wonder if there would be any disadvantage? Are all the features going to work fine? Is tooling OK with that?

The reason the custom sdk is needed is exactly because it wasn't working like that. We have properties that need to be set before the standard sdk.props is loaded and some set after the sdk.targets is loaded.

Kryptos-FR: I have updated the READMEs to explain that building the sdk is a required first step.

sources/tools/Stride.VisualStudio.Package.Tests/Stride.VisualStudio.Package.Tests.csproj
Kryptos-FR: So this project doesn't seem to follow the usual patterns that we have for other test projects. I think it is ok to keep using the editor sdk instead of the test sdk.

sources/sdk/Stride.Build.Sdk.Tests/Sdk/Sdk.props
Kryptos-FR: I'm making some changes. Review again once my commits are pushed.

sources/sdk/Stride.Build.Sdk.Tests/Sdk/Sdk.props
Kryptos-FR: I'm making some changes. Review again once my commits are pushed.

sources/sdk/Stride.Build.Sdk/Sdk/Stride.GraphicsApi.InnerBuild.targets
Kryptos-FR: Done.

Kryptos-FR:
@xen2 aside from testing, there shoudn't be any blockers.

sources/README.md
```diff
@@ -16,13 +35,10 @@ Folders and projects layout
    High-level serialization and git-like CAS storage system.
 * __Stride.Core.MicroThreading__:
    Micro-threading library based on C# 5.0 async (a.k.a. stackless programming)
-* __Stride.Core.AssemblyProcessor__:
-   Internal tool used to patch assemblies to add various features, such as Serialization auto-generation, various memory/pinning operations, module initializers, etc...
-
```
Ethereal77, line 21: Why remove the mention to `AssemblyProcessor`? It is still needed, isn't it?

Kryptos-FR: ~~This whole section was about steps that are now handled by the Sdk. That said, I might have removed too much. I'll rewrite that README.~~ I'll restore both parts.

And write a new README inside the assembly processor project.

Kryptos-FR: Done.

sources/README.md
```diff
@@ -56,3 +58,38 @@ For detailed SDK documentation, see [SDK-GUIDE.md](../build/docs/SDK-GUIDE.md).
...
 
 * MSBuild SDK packages that provide build logic for all Stride projects. See [Build System](#build-system) above.
+
+## Use in your project
+
+### Source repository ###
+
+There are two options to integrate this repository in your own repository:
+
+* __git subtree__ [documentation](https://github.com/git/git/blob/master/contrib/subtree/git-subtree.txt) and [blog post](http://psionides.eu/2010/02/04/sharing-code-between-projects-with-git-subtree/)
```
Ethereal77: The "psionides.eu" blog is broken, I think. It seems to be gone from the Internet since at least 2015 ([Web Archive](https://web.archive.org/web/20150206075033/http://psionides.eu/2010/02/04/sharing-code-between-projects-with-git-subtree/))

Kryptos-FR: Both links are broken actually. I'll update the doc from git and remove the blog post.

build/Stride.build
xen2: > > What about another option 5:
> > Simply keep props/targets in Sdk-like format, but add them as proper Import in the file? (with relative path) (a bit like we were doing before, except now we can properly put them as first and last line of csproj)
> > Even if it doesn't look as nice as Sdk, I think it would fit better the intent: import a local target which is in the same repo But I wonder if there would be any disadvantage? Are all the features going to work fine? Is tooling OK with that?
> 
> The reason the custom sdk is needed is exactly because it wasn't working like that. We have properties that need to be set before the standard sdk.props is loaded and some set after the sdk.targets is loaded.

Sorry if I wasn't clear, I meant, couldn't we simply import the new files directly?
We would follow the new SDK-like system in how csproj is organized, unlike before (nothing used before the .props files and keep .targets at the very end)

In practice, going from this:
```
<Project Sdk="Stride.SDK>
    <!-- Here is any property or item that we would usually do -->
  <PropertyGroup>
  </PropertyGroup>
  <ItemGroup>
  </ItemGroup>
</Project>
```
To this:
```
<Project>
  <!-- Manually import the SDK's early properties -->
  <Import Project="..\sdk\Stride.Sdk.props" />

    <!-- Here is any property or item that we would usually do -->
  <PropertyGroup>
  </PropertyGroup>
  <ItemGroup>
  </ItemGroup>

  <!-- Manually import the SDK's build targets -->
  <Import Project="..\sdk\Stride.Sdk.targets" />
</Project>
```

All your new SDK props/targets would stay as is.

We would still follow guidelines for SDK (nothing defined before props, targets included at the end).
But we won't have the weird indirection forcing us to use to package nuget as first step of the build (or any update to SDK props/targets).

That will also make it instant to do any change in target file.
No need to recompile/rebuild a separate solution we might forget.
I think it's important because I already feel the pain for every assemblyprocessor change, or nuget package change (for user games), I would like to avoid yet another one during dev if possible.

I think those 2 imports are not big of a deal and it would have the same functionality as SDK (except if something can't be supported?)

Kryptos-FR: It's not recognized properly by .NET tooling. Sdk-like projects require a Sdk declaration at the root.

So you wouldn't be able to work on Stride from VS Code for instance with the C# extension, especially on Linux.

Visual Studio also reverted their previous changes to have different Guid in the .sln file be different for classic .NET Framework project and .NET Sdk ones. The only discrimant now is the Sdk declaration.

This is what triggered my work on this.

Ethereal77:
Hey @Kryptos-FR , is there anything blocking migration to the new `.slnx` format? Does something break if you do `dotnet sln migrate`?

This new SDK refactor seems like a good opportunity to modernize that if possible.

Kryptos-FR:
One drawback is that there is no filter format for .slnx, so either we need to have several .slnx and no .slnf, or we keep .sln along .slnf.

laske185:
slnf works with slnx too.

Kryptos-FR:
I didn't work for me earlier. I will try again.

build/Stride.build
xen2: Understood, thanks for the explanation!
Agree that it would be better if working properly with tooling.

Just one question about another possible option 6:
Could we do a simple build step in Stride.Core.csproj that check if nupkg is less recent than the input files (props/targets), and if yes, build it, and send a <Error> that VS needs to be restarted?

Otherwise, I guess we can go with option 3 or 4.

xen2: Also, about option 4, I wonder if the .nupkg is deterministic? (if new checkout and build, will it generate the same nupkg?)
If not determinstic, it might often get committed by mistake (and lot of meaningless git confict) => option 3 better
Even if deterministic, I am sure we will often encounter git conflict on those files (sometime user change targets without comitting nupkg, etc.)

xen2: Thinking about it, even if nupkg is in our repo, it will deploy them in `%UserProfile%\.nuget\packages\Stride.SDK`, right?
In which case we need to bump package version on each change?
That sound cumbersome (esp. if working on branch where version could be weird, etc.)

If yes, I would like to offer a 7th option:
Make the Stride.SDK packages a very small redirect package that will load props/targets from Stride repo (which we can find from any .csproj in repo, going back up a few folders I think?).

That way, it will stay on v1.0.0.0 almost forever. As a result:
- we could package the .nupkg once and almost never touch it
- we can keep editing the targets in our project easily (same as before)
- almost never any conflicts on the .nupkg itself (only if we change this wrapper/redirect, should be extremely rare)

xen2: Sorry for the spam again in this comment, but thinking more about it:

Seeing those limitations, I still wonder if using a SDK for this situation is the right tool.

It seems SDK is more something that could for example be used for user project (not that we need it though).
But it doesn't seem to match well this use case (create SDK for internal use and use it directly same project). The dev cycle would become weird, and all those workaround options feel off.

I think the important part in your PR was reorganizing the props/targets which is great, and we should def. keep that, but I am less convinced by publishing it as a SDK.
So I would like to find a better way to consume those props/targets.

I quickly tried this with VSCode:
```
<Project>
  <Import Project="Test.props"/>
  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net10.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
  </PropertyGroup>
  <Import Project="Test.targets"/>
</Project>
```
with Test.targets/props being like this:
```
<Project>
  <Import Project="Sdk.targets" Sdk="Microsoft.NET.Sdk" />
</Project>
```

VSCode seemed happy with it (I could F5 run, and projects pane show 1 project loaded -- anything else I should check?)

If we have too much indirection/complex target for VSCode for this to work, I don't mind to have something slightly different, where we just extend without including Sdk.props/targets ourselves.
```
<Project>
  <Import Project="Stride.props"/>
  <Import Project="Sdk.props" Sdk="Microsoft.NET.Sdk" />
  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net10.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
  </PropertyGroup>
  <Import Project="Sdk.targets" Sdk="Microsoft.NET.Sdk" />
  <Import Project="Stride.targets"/>
</Project>
```
(we could even have 2 props/targets, Pre/Post, in case we need both, or we could use Directory.build.props/targets -- but it wouldn't be as nice, so hopefully it's not needed)

Kryptos-FR: I understand the worries and it is a good compromise. I'll work on restoring explict imports while keeping all the other changes. If I make it as a single commit, going back to a full sdk later will be rather simple.

Kryptos-FR: Done

sources/core/Stride.Core/Stride.Core.csproj
Kryptos-FR: Removed

Kryptos-FR:
I think I have handled all feedbacks. Ready for the final review.

edit: Actually going back to manual imports breaks completely in Visual Studio. I need to check what I did.

build/Stride.build
xen2: Great, thanks!

xen2:
Yes, I have the same behavior where it tries to "ugprade" project (because it thinks it's old style project)
After a quick investigation, this happens only with .slnx.

`.sln` was OK because the project type GUID `9A19103F-16F7-4668-BE54-9A1E7A4F7556` was enough for it to know it's new SDK style project but with slnx it uses auto-detection (broken).

xen2:
(note: did some edit)

`Type` can be added to `<Project>` in .slnx and it works (VS2026 doesn't ask for any upgrade)...
...but VS2026 removes them once saving the slnx back on any change (i.e. new project).

I see ~two~ (edit) three possible choices:
- Revert back to `.sln` until they better support round-trip saving
   Maybe not a bad thing until we switch to github runner 100%, since Teamcity doesn't have different build settings per branch (maybe it can, but not configured that way yet)
- Stick to `.slnx` but make sure nobody commit back the `Type` removal
- Create and upload on nuget.org (or local source) a small `Stride.SDK.Empty` or something similar, that basically do nothing. Then VS will properly detect our project as SDK style project.
  We probably never need to update it anymore once published.

xen2:
@Kryptos-FR any preference as for one of those 3 options?

Kryptos-FR:
None of these solutions is perfect. So I tend to revert to .sln for now. It will likely be supported a while longer by Visual Studio. If that changes, we can revisit.

Kryptos-FR:
Note for later: the documentation of the build system is here: https://github.com/dotnet/project-system/blob/main/docs/opening-with-new-project-system.md. I'll see if I need to open an issue for Visual Studio.

xen2:
I have made some changes/fixes, it seems working now.

But can't push back to your source.
In the meantime I will push to stride3d repo (same branch name)

Can you include it (fast forward) and give me permissions for future push?
(note: unit tests are not running until my master merge is included)

edit: hmm, "Maintainers are allowed to edit this pull request." seems checked
But I get this LFS error:
```
batch response: You need Push access to upload Git LFS objects.
error: failed to push some refs to 'github.com:Kryptos-FR/stride.git'
```

Kryptos-FR:
@xen2 done. And you should have received an invitation to contribute to my repo.

xen2:
Tested GameStudio and unit tests, seems OK now!
Thanks again for your contribution

VaclavElias:
Too late but not sure if here is something interesting to see regarding SDKs https://github.com/godotengine/godot/tree/master/modules/mono/editor/Godot.NET.Sdk

PR #3087, opened 2026-03-09
fix: TransformComponent.SetWorld rotation multiplication order

rafzi:
# PR Details

`TransformComponent.SetWorld` has a wrong multiplication order of the quaternions. It first applied the new world rotation, then undid the parent world rotation. In some cases this works, e.g. rotating along the same axis, but not in general.

Added a test which reproduced the issue and then passes with the issue fixed.

## Types of changes

Could be breaking if someone relied on the wrong behavior.

## Checklist


rafzi:
@dotnet-policy-service agree

Eideren:
Looks good, thanks !

Issue #1913, opened 2023-10-08
Make "Add Component" button functional with multiple entities selected


LouisIngenthron:
**Is your feature request related to a problem? Please describe.**
When multiple entities are highlighted in the hierarchy, the "Add Component" button in the property grid does nothing at all.

**Describe the solution you'd like**
The Add Component button should work normally, and whatever component the user selects should be added to **all** highlighted entities.

**Additional context**
N/A

luca-domenichini:
FYI - just pushed a PR to address this annoying issue

PR #3089, opened 2026-03-11
[GameStudio] FIX for #1913 - Make "Add Component" button functional with multiple entities selected

luca-domenichini:
# PR Details

This PR fixes a bug not allowing users to add components to multple entities at once. The popup for component addition was not showing.

## Related Issue

#1913 

## Types of changes

## Checklist

### DIFF
--- a/sources/editor/Stride.Core.Assets.Editor/Quantum/NodePresenters/Keys/AbstractNodeEntryData.cs
+++ b/sources/editor/Stride.Core.Assets.Editor/Quantum/NodePresenters/Keys/AbstractNodeEntryData.cs
@@ -12,21 +12,6 @@ namespace Stride.Core.Assets.Editor.Quantum.NodePresenters.Keys
         public const string AbstractNodeMatchingEntries = nameof(AbstractNodeMatchingEntries);
         public static readonly PropertyKey<IEnumerable<AbstractNodeEntry>> Key = new PropertyKey<IEnumerable<AbstractNodeEntry>>(AbstractNodeMatchingEntries, typeof(AbstractNodeEntryData), new PropertyCombinerMetadata(CombineProperties<AbstractNodeEntry>));
         
-        [Obsolete("Use the generic version of CombineProperties instead, which allows to specify the type of the properties to combine. This method is kept for backward compatibility, but it is recommended to use the generic version instead.")]
-        public static object CombineProperty(IEnumerable<object> properties)
-        {
-            HashSet<AbstractNodeEntry> result;
-            var hashSets = new List<HashSet<AbstractNodeEntry>>();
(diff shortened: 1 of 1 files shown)


luca-domenichini:
@dotnet-policy-service agree

luca-domenichini:
question for maintainers (tagging a random one @VaclavElias)

I changed a public facing method `public static object CombineProperty(IEnumerable<object> properties)` into `public static IEnumerable<TAbstractNodeEntry> CombineProperty<TAbstractNodeEntry>(IEnumerable<object> properties)`, so technically this is a breaking change.

I am open to do the following, as maintainers prefer:
- add original method back plus a `[Obsolete]` attribute explaining why it's a bad idea to use it
- leave all as it is and marking this PR as "breaking change"
- anything else?

Ethereal77:
I'm not a mantainer, but imho (talking from my understanding, I have not checked if this makes sense):
* If the new method operates over several properties, maybe a good idea would be to name it `CombineProperties`, then make `CombineProperty` a special case that just does what it did before (or calls the old `CombineProperty` and returns just the first, and mark it as `[Obsolete]`.
* Change all the calls to the old `CombineProperty` to the new `CombineProperties`.

Does this make sense?

luca-domenichini:
> * If the new method operates over several properties, maybe a good idea would be to name it `CombineProperties`, then make `CombineProperty` a special case that just does what it did before (or calls the old `CombineProperty` and returns just the first, and mark it as `[Obsolete]`.

Actually, it already was `CombineProperty` despite the fact it returned an `IEnumerable`.
Definetely, `CombineProperties` would be better.
I could just add `[Obsolete]` to old method, but I would like to hear from maintainers if that is really needed, since I already changed all calls to the new method, so the old one is to be dropped ASAP. 
I am wondering if removing that method would break someone using it outside of Stride GameStudio, since it belongs to `Stride.Assets.Core.Editor` project.

> * Change all the calls to the old `CombineProperty` to the new `CombineProperties`.

Yes, done 👍.

In the meanwhile, let me refactor with your suggestions..

Ethereal77:
From a very quick search across all the repo, that method is only used in `AbstractNodeEntryData.cs` and `EntityHierarchyData.cs`. Usually, for public APIs, obsoleting is the preferred way, but as this is only used in presentation/editor-related assemblies, let's ask @Kryptos-FR

Kryptos-FR:
> From a very quick search across all the repo, that method is only used in `AbstractNodeEntryData.cs` and `EntityHierarchyData.cs`. Usually, for public APIs, obsoleting is the preferred way, but as this is only used in presentation/editor-related assemblies, let's ask @Kryptos-FR

We can remove the obsolete method. No other code is using it. And if by an extraordinary chance, someone else, somewhere is making another editor that uses our library, then it will be a breaking change for them. I'm ok with that.

luca-domenichini:
done 👍

PR #3093, opened 2026-03-16
fix vs2026 cant compile

Redwarx008:
# PR Details
My VS 2026 wouldn't compile no matter what I did because Windows Desktop couldn't find MSVC 143 (even though I had it installed). So I changed the target platform to 145, and everything worked fine. Since the project is already being compiled using VS2026, it's unnecessary for these projects to depend on 143.

## Types of changes

## Checklist

### DIFF
--- a/sources/native/WindowsProjects/UWP/UWP.vcxproj
+++ b/sources/native/WindowsProjects/UWP/UWP.vcxproj
@@ -43,35 +43,35 @@
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'" Label="Configuration">
     <ConfigurationType>DynamicLibrary</ConfigurationType>
     <UseDebugLibraries>true</UseDebugLibraries>
-    <PlatformToolset>v143</PlatformToolset>
+    <PlatformToolset>v145</PlatformToolset>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|ARM'" Label="Configuration">
     <ConfigurationType>DynamicLibrary</ConfigurationType>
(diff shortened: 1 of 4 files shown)


xen2:
Thanks!
I remember skipping that change at the time because I thought it would force us to bump our C++ redistributable to VS2026 (we would need to upgrade in prerequisite installer).
However, it seems that v145 is still backward compatible with VS 2019 C++ redist, so it looks safe to keep as is.

Kryptos-FR:
Is v145 available in VS 2022 though? If not, it means we are now forcing the use of VS 2026 in order to build Stride. The doc would need update.

VaclavElias:
> Is v145 available in VS 2022 though? If not, it means we are now forcing the use of VS 2026 in order to build Stride. The doc would need update.

But you can build Stride 4.3 only with VS 2026?

xen2:
I thought VS2026 was already a requirement since .NET 10.0 switch (Compiling .NET 10 with Visual Studio 2022 is generally not officially supported)
README.md was updated but I didn't check the doc in detail.

VaclavElias:
Also, should this PR update also this? https://github.com/stride3d/stride/blob/d6c133cfa8ab1dc4a56d9118d89f6bcd1a5850a8/README.md?plain=1#L44-L46

Note, ideally, we need exact wordings e.g. `MSVC v143 - VS 2022 C++ ARM64/ARM64EC build tools (Latest)` as it is in the Visual Studio installer, so it is easy to find/filter. Obviously, the exact wordings is going to work only for EN and that is ok. Also, I believe they changed the wordings at least twice in the last 6 months 🤣 but we can just use the latest.. 

I can do a separate PR for stride-docs..

PR #3095, opened 2026-03-20
Prevent typeless textures from creating srv / rtv on initialization

johang88:
# PR Details

DirectX 11 does not allow Shader Resource Views or Render Target Views to be created for typeless textures. This change prevents their creation upon texture initialization, which would have given an exception when `NativeDevice.CreateShaderResourceView` / `NativeDevice.CreateRenderTargetView` were called. Users can then create views in their desired format using `ToTextureView`. 

Example usage from when I had to interpret the same texture as both SRgb and non SRgb, typeless is required here as creating a non srgb view from an srgb texture is not allowed.

```
        _stagingDiffuse = Texture.New2D(graphicsDevice, padded, padded,
            PixelFormat.R8G8B8A8_Typeless,
            TextureFlags.RenderTarget | TextureFlags.ShaderResource);

        _stagingDiffuseTargetView = _stagingDiffuse.ToTextureView(new()
        {
            Flags = TextureFlags.RenderTarget,
            Format = PixelFormat.R8G8B8A8_UNorm_SRgb,
            Type = ViewType.Single
        });

        _stagingDiffuseCompressView = _stagingDiffuse.ToTextureView(new()
        {
            Flags = TextureFlags.ShaderResource,
            Format = PixelFormat.R8G8B8A8_UNorm,
            Type = ViewType.Single
        });
```

## Types of changes

## Checklist


xen2:
Probably we should do the same for D3D12 (and maybe Vulkan if it works the same)?

johang88:
Good point, I'll have to check how they behave.

xen2:

Issue #3064, opened 2026-02-10
Feature Request: Support palettes in color picker outside recompiling source


Acissathar:
**Is your feature request related to a problem? Please describe.**
The color picker palette in Stride is currently hardcoded in ColorPaletteViewModel.cs, making it impossible to customize without modifying engine source. This limits workflows that rely on specific palettes or shared team color sets.

**Describe the solution you'd like**
Add support for loading palettes from external files or project assets. The editor should allow users to import and switch between palettes. Unity’s Swatch Library system is a good reference:
https://docs.unity3d.com/Manual/SwatchLibraries.html

I think even just a more straight forward "single palette" loading / editor setting field would be enough too, just some way to modify the palette without having to recompile source.

**Describe alternatives you've considered**
Editing the hardcoded palette directly (not maintainable / annoying for multi-project workflow)

Storing palettes externally and copy/pasting colors (slow and potentially error‑prone)

**Additional context**
- What are the use cases?
  - Shared team palettes
  - Project‑specific UI or art‑direction palettes
  - Pixel‑art or stylized workflows needing curated color sets

- What kind of options or settings are expected or wanted?
  - At a minimum, the ability to specify a different palette than the hardcoded one.

- Does the feature affect behavior in the editor?
  - Color picker displays the active palette instead of a fixed list
  - Simple UI control for selecting or loading palettes
- Is there a component or data structure involved?
  - Depending on solution, will need a parser and format.
- What kind of documentation is needed or needs to be updated?
  - Same as above, documentation would just need to be added for the supported file format. 
  - Can look at gimp's GPL format which makes it easy to grab from lospec for example. I believe Krita supports this file format for importing palettes as well.

I don't mind taking a stab at the changes, but wanted to throw it out there first for any design considerations others may have too / what seems most useful.

PR #3100, opened 2026-03-23
feat: Support external GPL color palettes in color picker

SmitBdangar:
C## What this does
Users can now drop a `ColorPalette.gpl` file next to the Stride editor 
executable to customize the color picker palette without recompiling.

## How to use
1. Download any `.gpl` palette from [Lospec](https://lospec.com/palette-list)
2. Rename it to `ColorPalette.gpl`
3. Place it next to the GameStudio executable
4. Restart the editor — your palette will load automatically

If no file is found, the original default palette is used as fallback.

## Changes
- Added `GplPaletteParser.cs` — parses standard GIMP `.gpl` files
- Updated `ColorPaletteViewModel.cs` — lazy-loads from file, falls back to defaults
- Fixed typo: `"Platimium"` → `"Platinum"`

#3064

### DIFF
--- a/samples/Tests/Tests.sln
+++ b/samples/Tests/Tests.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.28010.2026
+# Visual Studio Version 18
+VisualStudioVersion = 18.4.11612.150
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Stride.Samples.Tests", "Stride.Samples.Tests.csproj", "{AA6F38A3-8A04-42AE-A19C-9B0123EC1589}"
 EndProject
--- a/sources/editor/Stride.Core.Assets.Editor/ViewModel/ColorPaletteViewModel.cs
+++ b/sources/editor/Stride.Core.Assets.Editor/ViewModel/ColorPaletteViewModel.cs
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net) and Silicon Studio Corp. (https://www.siliconstudio.co.jp)
 // Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
 using System.Collections.Generic;
+using System.IO;
 
 using Stride.Core.Mathematics;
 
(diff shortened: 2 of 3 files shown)


MsEpsilon:
> 3. Place it next to the GameStudio executable
> 4. Restart the editor — your palette will load automatically

I don't think it should function like that (I really dislike the idea of an hardcoded file - and I think others would agree), I think an other kind of integration would be a better idea, for example in the GameStudio settings?

SmitBdangar:
@dotnet-policy-service agree

SmitBdangar:
@MsEpsilon Thanks for the feedback Id like to move this to GameStudio settings. Could you point me to an existing setting in the codebase I could use as a reference to follow the same pattern

MsEpsilon:
Start by looking at [`sources/editor/Stride.Core.Assets.Editor/Settings/EditorSettings.cs`](https://github.com/stride3d/stride/blob/master/sources/editor/Stride.Core.Assets.Editor/Settings/EditorSettings.cs). But that would implement an pallete setting that would affect all projects.

An other approach would be using palletes as assets, but I'm not sure how that could be implemented in the engine itself, as that would be a breaking change on how the serializer+editor work.

Finally, an approach I'd consider myself is an post-process shader that converts any RGBA color to an color in a given pallete.

Ethereal77:
I agree that a hardcoded file located besides the executable (why the executable and not the project, for example?) is weird.

A more reasonable option would be, like  @MsEpsilon said, an asset type that the color picker could load (or just a pallete file the color picker can load). Note that I believe that if it is not going to be compiled/processed for inclusion in the built game, that should simplify things (should not invoke `AssetCompiler`). But being an asset, it would be nice if it had a preview, an editor, and could be assigned to slots.

Second preferred imho would be the editor settings, but that would be global. Not bad, but not ideal.

> Finally, an approach I'd consider myself is an post-process shader that converts any RGBA color to an color in a given pallete.

While a color grading effect that allowed arbitrary palletes would be a nice addition (we already have color transform shaders  in post-processing), I would refrain from using a shader for something that would be normally done just picking different colors. It seems such a waste of computing power.

MsEpsilon:
Compared to everything else, the post process shader option would be the option that requires least effort to implement.

Ethereal77:
> Compared to everything else, the post process shader option would be the option that requires least effort to implement.

Yes, but costs performance (some, not huge, but...) and is not the solution this OP was talking about in #3064 , that is, for artists to have an easy way **inside the editor** of customizing the palettes in the color picker. As a quality-of-life improvement to the editor, it seems quite reasonable to me

PR #3103, opened 2026-03-24
feat: implement ColorPalette as a proper Stride asset (.sdpalette)

### LINKED ISSUE #3064: Feature Request: Support palettes in color picker outside recompiling source

SmitBdangar:
# PR Details
Reworks the color palette feature as a proper Stride asset (.sdpalette) 
instead of a hardcoded file, addressing feedback from #3100.

## Related Issue
Closes #3064

## Types of changes

## Changes
- Added `ColorPalette.cs` — runtime class with DataContract + ContentSerializer
- Added `ColorPaletteAsset.cs` — design-time .sdpalette asset with AssetCompiler
- Re-used `GplPaletteParser.cs` — parses .gpl files at compile time
- Updated `ColorPaletteViewModel.cs` — uses SetPalette() instead of hardcoded file path
- Added `ColorPalette.sdtpl` — makes Color Palette appear in GameStudio Add Asset menu
- Fixed typo: Platimium → Platinum

## How to use
1. In GameStudio → Asset View → Add Asset → Rendering → Color Palette
2. Set SourceFile to a .gpl file path, or manually define colors in the Colors dictionary
3. The color picker will use the assigned palette automatically

## Checklist
- [ ] I have built the editor to try this change out.

## Notes
Could not run GameStudio locally due to native C++ build tool limitations,
but `Stride.Core.Assets.Editor` builds successfully with 0 errors.
Supersedes #3100
```

### CODE REVIEW

**Comment by Kryptos-FR** [comment]
Please follow the PR template

**Comment by SmitBdangar** [comment]
@Kryptos-FR  Updated the PR description to follow the template, sorry about that

**Comment by MsEpsilon** [comment]
If this PR superseeds your previous, did you consider closing the previous PR?

File: sources/editor/Stride.Core.Assets.Editor/ViewModel/ColorPaletteViewModel.cs [hunk 1]
```diff
@@ -1,24 +1,43 @@
...
-            { "Titanium", new Color3(0.756946965f, 0.727607463f, 0.695207239f) },
-            { "Cobalt", new Color3(0.829103572f, 0.824958926f, 0.812750243f) },
-            { "Platimium", new Color3(0.834934076f, 0.814845027f, 0.783999116f) },
+            { "Gold",      new Color3(1.0f,         0.88565079f,  0.609162496f) },
+            { "Copper",    new Color3(0.979292159f, 0.814900815f, 0.754550145f) },
+            { "Chromium",  new Color3(0.761787833f, 0.765888197f, 0.764724015f) },
+            { "Nickel",    new Color3(0.827766422f, 0.797984931f, 0.74652364f)  },
+            { "Titanium",  new Color3(0.756946965f, 0.727607463f, 0.695207239f) },
+            { "Cobalt",    new Color3(0.829103572f, 0.824958926f, 0.812750243f) },
+            { "Platinum",  new Color3(0.834934076f, 0.814845027f, 0.783999116f) }, // Fixed typo: platimium
```
MsEpsilon: Useless comment...

For an reader this wouldn't matter, but it is enough that you specified this change in your PR message.

sources/editor/Stride.Core.Assets.Editor/ViewModel/ColorPaletteAsset.cs
```diff
@@ -0,0 +1,67 @@
...
+                }
+                else
+                {
+                    colors = Parameters.Colors;
+                }
+
+                var runtime = new ColorPalette { Colors = colors };
+                assetManager.Save(Url, runtime);
+
+                commandContext.Logger.Info($"ColorPaletteAsset compiled: {Url} ({colors.Count} colors)");
```
MsEpsilon: Is logging here really necessary? Compared to other asset types, they don't use logging when they finish asset compilation. Only warnings or errors if it is the case.

That would uselessly fill the build output.

MsEpsilon (changes requested):
Nitpicks, mostly. Except the logging in `DoCommandOverride` that I consider that should be removed.

Also make sure you did compile, run the GameStudio, and ALSO create, save and reload a project with the asset of the type `ColorPaletteAsset` (assuming you didn't by not checking ` I have built the editor to try this change out.`)

MsEpsilon (commented):
Nothing I can comment on, wonder what other persons think about this.

LGTM when I try this out myself in the next hours.

sources/editor/Stride.Core.Assets.Editor/ViewModel/ColorPaletteViewModel.cs
```diff
@@ -1,24 +1,43 @@
...
+        {
+            get
+            {
+                if (_colors is null)
+                    _colors = DefaultColors;
+                return _colors;
+            }
+        }
+
+        public static void SetPalette(ColorPalette? palette)
```
Kryptos-FR, line 36: This method is never called.

sources/editor/Stride.Core.Assets.Editor/ViewModel/ColorPaletteViewModel.cs
```diff
@@ -1,24 +1,43 @@
...
+            { "Chromium",  new Color3(0.761787833f, 0.765888197f, 0.764724015f) },
+            { "Nickel",    new Color3(0.827766422f, 0.797984931f, 0.74652364f)  },
+            { "Titanium",  new Color3(0.756946965f, 0.727607463f, 0.695207239f) },
+            { "Cobalt",    new Color3(0.829103572f, 0.824958926f, 0.812750243f) },
+            { "Platinum",  new Color3(0.834934076f, 0.814845027f, 0.783999116f) },
         };
+
+        private static Dictionary<string, Color3>? _colors;
+
+        public static Dictionary<string, Color3> Colors
```
Kryptos-FR, line 26: The XAML binding to this property won't work as it is a static binding. If the Dictionary is meant to be modified, then:
- the class must derive from ViewModelbase
- the property must raise the PropertyChanged event upon changes
- we need to use an observable collection instead of a dictionary (in such case we can change it to a collection of a tuple of string and Color3).

sources/editor/Stride.Core.Assets.Editor/ViewModel/ColorPaletteAsset.cs
```diff
@@ -0,0 +1,65 @@
...
+using Stride.Core.BuildEngine;
+using Stride.Core.Mathematics;
+using Stride.Core.Serialization.Contents;
+
+namespace Stride.Core.Assets.Editor.ViewModel
+{
+    [AssetDescription(FileExtension, AllowArchetype = false)]
+    [AssetContentType(typeof(ColorPalette))]
+    [AssetFormatVersion("Stride", CurrentVersion, "1.0.0.0")]
+    public sealed class ColorPaletteAsset : Asset
```
Kryptos-FR, line 18: It's the first time we ever have an asset defined within an editor library. I'm not 100% sure it will work properly.

Kryptos-FR, line 18: It is missing a `[DataContract]` attribute.

sources/editor/Stride.Core.Assets.Editor/ViewModel/GplPaletteParser.cs
```diff
@@ -0,0 +1,67 @@
...
+using System.Collections.Generic;
+using System.IO;
+
+using Stride.Core.Mathematics;
+
+namespace Stride.Core.Assets.Editor.ViewModel
+{
+    public static class GplPaletteParser
+    {
+        public static Dictionary<string, Color3>? TryParse(string filePath)
```
Kryptos-FR, line 13: The expected signature for TryParse method is `bool TryParse(string filePath, out Dictionary<string, Color3>? colors)`. Taking into account nullability annotation:

```csharp
bool TryParse(string filePath, [NotNullWhen(true)] out Dictionary<string, Color3>? colors)
```

Kryptos-FR (changes requested):
There are a couple of fix to do before we can check whether it actually works.

sources/editor/Stride.Core.Assets.Editor/ViewModel/GplPaletteParser.cs
Ethereal77: Also, this is not "parsing" actually. This is loading. Maybe a better name would be `TryLoad()`. Parsing implies you give a string, it returns data extracted from that string.

xen2:

PR #3105, opened 2026-03-24
Allow to opt-out of loading shaders from original location

azeno:
# PR Details

This PR adds a property to the `ShaderSourceManager` which allows to opt-out of the system trying to load from the original path first.
We ran into this issue several times in the past that on build machines our app would behave differently or not run at all because it would load shaders in a different state than the app was build with.

## Types of changes

## Checklist


xen2:
From top of my head, it was for either (or both) of those scenarios (dev/desktop only):
- map to real source file (#file and #line info for debugger), otherwise any shader debugger can't trace into original source if there is no actual file
- hot reload

Sounds good to me to have a flag for now, I will eventually get back to it with new shader system anyway.

PR #3109, opened 2026-03-29
fix: Incorrect value returned when calculating nearest tick in sliders with non-zero minimums

Acissathar:
# PR Details

CalculateClosestTick in Slider.cs subtracts the Minimum to convert into a relative offset, does the tick-snapping math in that relative space, but then returns times * step without translating it back to the absolute value. This makes it so that any value you enter is going to have the Minimum subtracted from it when Minimum is a non-zero value, even if this makes the value go below the Minimum.

Adding it back in at the very end makes both the value and minimum be respected:

SliderMinMaxFixed

## Related Issue

#3108 

## Types of changes

## Checklist


Eideren (approved):

xen2:

PR #3115, opened 2026-04-01
Migrate CI/CD from TeamCity to GitHub Actions: release workflows & WARP GPU testing (D3D11)

xen2:
# PR Details

- **Release workflows**:
  Added `release.yml` (engine), `release-launcher.yml` (launcher/installer), `release-vspackage.yml` (VS extension), and `release-unlist.yml` (unlist broken releases from NuGet)
  All release workflows require code signing by default, `production` environment secrets, and `stride-release-managers` team membership.
- **GPU testing with WARP**:
  Added D3D11 software rendering support via `STRIDE_GRAPHICS_SOFTWARE_RENDERING=1` env var, which selects the WARP adapter. Game tests now run on standard GitHub runners without GPU hardware.
  Added WARP reference images (they seem to cover almost half of GPU tests deterministically).
  When running in VS, added launcherConfig.json that default to WARP (but allow to select GPU)
- **CI improvements**:
  Added Game test category to `main.yml`.
  Disabled flaky benchmark test. Removed `Stride.Audio.Tests` graphics API dependency.
  Added `CODEOWNERS` for release workflow protection.
- **Bug fixes**:
  Fixed EditText selection/caret positioning for SDF fonts with non-native TextSize.
  Fixed EditText caret drawing at wrong end of selection.
- **TeamCity removal**:
  Replaced all TeamCity references (badges, packages, build messages) with GitHub Actions equivalents.

## Open question
* Do we even want to keep GPU gold? They are much less predictable and everybody has a different GPU/driver.
  Maybe comparison could be ignored, test result marked "inconclusive" with message "use WARP for proper testing, GPU is only for debugging purpose"?

## Types of changes

## Checklist


.github/workflows/main.yml
```diff
@@ -14,11 +14,23 @@ on:
       - '!**/.gitignore'
       - '!**/*.md'
       - '!crowdin.yml'
+  pull_request:
```
Kryptos-FR, line 17: The initial idea was to not run the full main CI on PR, but instead only individual workflows depending on which files has been touched.

If we decide to run everything all the time, then we should remove the filters on all other workflows. Otherwise, each will be run twice.

xen2: Good point, I was hesitating about that one, but I didn't fully realize about those other workflows.
I understand better how it was intended to work now.

Assuming full CI is fast enough and part of dotnet infra so not a problem to run every PR, do you still think worth to have something smaller for PR?

Here are the pros/cons I can come up with:
- pros: faster, less risk of unrelated previous breakage confusing PR
- cons: slight risk of miss, and need to maintain additional workflow

When adding it, my rationale was that in practice we would want to run it anyway before final merge and it is a bit annoying to do manually (and easy to forget)
Since I am adding GPU tests again with WARP (quite deterministic), we could easily spot and fix those before PR merge (rather than it polluting PR commit and any further commits/PR with bad chekmark and false positive until fixed).

VaclavElias: We could use some labels to trigger specific GitHub Actions or part of it? So for example, the default would be a smaller GitHub Actions and based on the label added (e.g. area-Graphics) it could trigger different GitHub Action or path in GitHub Action. 

cons: Probaby the same as above

Also, not sure if some repeated workflows could be re-used so they are maintained only once.

xen2: Right now I am spending some time to get all the image testing unit test running properly again on GH (incl new platforms like D3D12/Vulkan).
It's quite an effort to get everything working properly.

When we used them in the past, they broke so many time and stayed broken for long stretch of time because it wasn't run every commit. And once it gets broken for a while, it gets very hard and time consuming to trace back where it happened (if it even compiled) and fix back.

So I vote to simply enable the full CI matrix on every master commit and every PR before merge for now.

We can do a few "safe" optimization like running only editor tests for changes on editor folders, but I don't want to take any risk the other way around (i.e. a change in Core/Graphics means a full test of **everything**, incl. graphics unit tests and editor tests, even if breaking them might seem unlikely).
Likewise, I don't a slightly different version of testing workflow as it might not test the exact same thing, and we'll end up having the same problem of having to track down broken commits with the real CI suite.

Kryptos-FR: Sounds good. It's better to be on the safe side.

Issue #2904, opened 2025-09-24
[UI] Dropdown UI Element


MsEpsilon:
As the title specifies, a Drop-down UI element (also known as ComboBox in WinForms) specifically a DropDownList).

Reference image (from MSDN, Windows API documentation)
![Reference image](https://learn.microsoft.com/en-us/windows/win32/controls/images/simplecombo.png)

This is possible to create on scratch, but it creates clutter for example. when I create a settings page with multiple DDs.

I'm creating this issue since I'm considering writing a DropDownList component for `Stride.UI.Controls` (so to discuss how it could be implemented).

Acissathar:
@MsEpsilon Did you by chance ever end up making this component? Think I'm about to run into needing this as well (for largely the same reasons) and remembered seeing this issue a bit ago. Figured I'd check before looking into creating a component as well.

MsEpsilon:
Hi, I only worked on that for one single day, but I stopped working on that as I didn't really understood how Stride's entire UI library works. I went to work on other stuff.

I've actually forgot this thread existed 😅.

I think you can implement it by having a button that shows up a scrollable list when clicked (and I remember Stride has support for scrolling as far as I remember). But having scrollable buttons is an issue of itself I think.

PR #3120, opened 2026-04-04
feat: Add DropDown UI element to built in UI library

Acissathar:
# PR Details

Currently, the built in Stride UI library is missing a drop down element. You can probably throw one together yourself with multiple elements (button that onclick turns on a scrollable list, etc), but that's a bit unwieldy for a general purpose dropdown / combobox.

This is sort of a proof a concept, as I need it for my own project but haven't yet really put it through any real use cases and determined what all is missing with regards to functionality. Super dumb version of a dropdown, just a list with Text elements for the buttons and a SelectedIndex.

At a quick level, it supports the following properties for design:

Screenshot 2026-04-04 173351

Placeholder text is used when the Selected Index is -1, but the selected index can be defaulted to an actual item in the list instead if desired.

There is an event that's fired that can be subscribed to know when the index has changed.

It also supports hiding the scroll view when clicking outside of the list.

DropDownPOC

As mentioned above, I plan to use this in my project so there may be more tweaks which is why I'm throwing it up as a draft PR but also so that its _out there_ if anyone has any functionality they think it should have that its missing, want to use it in the meantime, etc.

## Related Issue

#2904 

## Types of changes

## Checklist


sources/editor/Stride.Assets.Presentation/AssetEditors/UIEditor/Services/UIEditorController.cs
```diff
@@ -424,6 +427,23 @@ private UIComponent GetUIComponent()
...
+            if (root is DropDown dropDown)
+                dropDown.IsInteractable = false;
+
+            foreach (var child in root.VisualChildren.BreadthFirst(e => e.VisualChildren))
+            {
+                if (child is DropDown childDropDown)
+                    childDropDown.IsInteractable = false;
+            }
+        }
+
```
Eideren, line 451: This seems a bit too specific, it's also fairly heavy and allocates quite a bit, things you could look into:
- `UIElement.IsHierarchyEnabled`
- How the scroll for panel works around this issue if it does.

xen2:

PR #3127, opened 2026-04-09
Linux build and CI

xen2:
# PR Details

WIP (opened to be able to test CI)

## Types of changes

## Checklist


sources/tests/xunit.runner.stride/StrideXunitRunner.cs
```diff
@@ -14,6 +14,15 @@ public static class StrideXunitRunner
     // yet and stuff might break.
     public static void Main(string[] _, Action<bool>? setInteractiveMode = null, Action<bool>? setForceSaveImage = null)
     {
+        // Skip Avalonia UI on headless systems (no display).
+        // Tests will be run by dotnet test / xunit adapter instead.
+        if (IsHeadless())
```
Kryptos-FR, line 19: @xen2 I don't think that is correct. The `Main` method is only run when doing `dotnet run`. It is ignored when doing `dotnet test` since the testing framework doesn't try to run the `Main` method, so it shouldn't impact tests run in the CI.

However this breaks the previous behavior when running with `dotnet run` or from Visual Studio was launching the Avalonia UI, esp. for the core tests such as `Stride.Core.Tests` or `Stride.Core.Mathematics.Tests`.

PR #3161, opened 2026-05-08
sdk: auto-detect StrideNativeBuildMode

xen2:
Pick MSVC only when targeting Windows AND the VS C++ tools are loaded
(VCINSTALLDIR set), Clang otherwise.

- Plain `dotnet build` from a regular shell on Windows now works out of the box
- Cross-compile from Windows host (Linux/Android/iOS targets) auto-uses Clang
- Drops the now-redundant `-p:StrideNativeBuildMode=Clang` from many CI workflows (release.yml keeps it, it runs msbuild.exe
  with VS env active)

Note: we'll have to check if MSVC mode is really necessary still at all.

### DIFF
--- a/.github/workflows/build-android.yml
+++ b/.github/workflows/build-android.yml
@@ -89,7 +89,6 @@ jobs:
       - name: Build
         run: |
           dotnet build build\Stride.Android.slnf `
-            -p:StrideNativeBuildMode=Clang `
             -nr:false `
             -p:Configuration=${{ github.event.inputs.build-type || inputs.build-type || 'Debug' }} `
             -p:StridePlatforms=Android `
--- a/.github/workflows/build-ios.yml
+++ b/.github/workflows/build-ios.yml
@@ -57,7 +57,6 @@ jobs:
       - name: Build
         run: |
           dotnet build build\Stride.iOS.slnf `
-            -p:StrideNativeBuildMode=Clang `
             -nr:false `
             -v:m -p:WarningLevel=0 `
             -p:Configuration=${{ github.event.inputs.build-type || inputs.build-type || 'Debug' }} `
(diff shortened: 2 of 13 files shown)


Kryptos-FR:
@xen2 `dotnet build` will fail when used from within a Visual Studio Prompt because dotnet is not able to build C++ projects. It still needs to use msbuild.

>  C:\Projects\Stride\Engine\stride\sources\native\WindowsProjects\WindowsDesktop\WindowsDesktop.vcxproj(35,3): error MSB4278: The imported file "$(VCTargetsPath)\Microsoft.Cpp.Default.props" does not exist and appears to be part of a Visual Studio component. This file may require MSBuild.exe in order to be imported successfully, and so may fail to build in the dotnet CLI.
>  Stride.Audio net10.0 failed with 10 warning(s) (1.7s)

Still, it is a big improvement because now dotnet build does work from a regular console prompt. We need to update our build documentation to explain this behavior.

xen2:
> @xen2 `dotnet build` will fail when used from within a Visual Studio Prompt because dotnet is not able to build C++ projects. It still needs to use msbuild.
> 
> > C:\Projects\Stride\Engine\stride\sources\native\WindowsProjects\WindowsDesktop\WindowsDesktop.vcxproj(35,3): error MSB4278: The imported file "$(VCTargetsPath)\Microsoft.Cpp.Default.props" does not exist and appears to be part of a Visual Studio component. This file may require MSBuild.exe in order to be imported successfully, and so may fail to build in the dotnet CLI.
> > Stride.Audio net10.0 failed with 10 warning(s) (1.7s)
> 
> Still, it is a big improvement because now dotnet build does work from a regular console prompt. We need to update our build documentation to explain this behavior.

Thanks for the feedback, I reviewed all case and I think the 55b468e should fix those.

I also took the opportunity to simplify top-level README.md build instructions and moved details to another file in 6ee2e04.
This also include a workaround so that even if ARM64 build tools are not installed, it works for local dev. VS with 2 simple workload and everything should work out of the box now! (as soon as we recompile that .NET 6 dependency in deps)

PR #3169, opened 2026-05-13
core: drop MicroThreading dep from Stride.Core.Serialization

xen2:
## Summary

Move `ContentManager.LoadAsync` / `ReloadAsync` to extension methods (`ContentManagerAsyncExtensions` in `Stride.Engine`) so `Stride.Core.Serialization` no longer depends on `Stride.Core.MicroThreading`.

`IContentManager.LoadAsync<T>` is removed from the interface and replaced by an extension.

The 12 projects that use MicroThreading types directly now reference `Stride.Core.MicroThreading` explicitly (was transitive via Serialization).

For user code, no changes needed (as long as `Stride.Engine` is referenced).

### DIFF
--- a/sources/assets/Stride.Core.Assets.CompilerApp/Stride.Core.Assets.CompilerApp.csproj
+++ b/sources/assets/Stride.Core.Assets.CompilerApp/Stride.Core.Assets.CompilerApp.csproj
@@ -28,6 +28,7 @@
     <ProjectReference Include="..\..\engine\Stride.BepuPhysics\Stride.BepuPhysics\Stride.BepuPhysics.csproj" />
     <ProjectReference Include="..\..\engine\Stride.SpriteStudio.Offline\Stride.SpriteStudio.Offline.csproj" />
     <ProjectReference Include="..\..\engine\Stride.UI\Stride.UI.csproj" />
+    <ProjectReference Include="..\..\core\Stride.Core.MicroThreading\Stride.Core.MicroThreading.csproj" />
   </ItemGroup>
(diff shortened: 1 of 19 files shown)


Ethereal77:
While I generally like to remove transitive dependencies if they are not needed, I feel like removing an `async` API from such a IO-centric component as `ContentManager` unless you are using `Stride.Engine` is a bit the wrong way to go. I would like the opposite even, more asynchronous support for all IO-related paths.

Having an extension in `Stride.Engine` keeps the functionality for Stride engine uses, but diminishes the usefulness of `Stride.Serialization` instead, imho

Eideren:
@xen2 
> For user code, no changes needed

Are you sure about that ? Afaik you still need to import the namespace the extension method is defined in to call an extension methods, whereas for standard method you don't need to.

xen2:
Good points, let me reconsider a bit some other approaches, maybe I was a bit too hasty with this one.

My rationale:
I wanted to untie some unnecessary dependencies from each other a bit (faster compile, more standalone modules, etc.).
I thought Serialization should have nothing to do with MicroThreading and depend on it.
It was added artificially just for that helper method that could live elsewhere IMHO.
I think this really should live as an extension method, but maybe where I did it is a bad choice
(I hesitated a lot and went back and forth several time trying to decide it)

At first I tried moving this as an extension method in MicroThreading assembly itself. However this adds the reverse dependency: MicroThreading depends on Serialization (though it is probably better than the previous opposite).
However, I ended up going further by totally untying them, maybe that was too much...
I could make sure the extension method is in same namespace as IContentManager rather than `Stride.Engine` (but there's still a few case when the namespace might be missing).

PR #3184, opened 2026-05-18
build: per-checkout NuGet version suffix (disambiguates parallel work…

xen2:
## Problem

Multiple Stride checkouts on one machine (git worktrees / separate clones) all auto-pack first-party packages at same version (i.e. `4.4.0.1`) and clobber each other in the shared `%LocalAppData%/Stride/NugetDev` feed and the NuGet global cache.

## Proposal

A per-machine ledger at `<LocalAppData>/Stride/worktree-ids.txt` assigns each checkout a token — the first to register is `(primary)` (no suffix, builds `4.4.0.1` exactly as today), the rest get `dev1`, `dev2`, …
You can also register it yourself to anything custom.

The token feeds `NuGetVersionSuffix` on both ends:

- **Producer**: appended to `PackageVersion` before the pack pipeline.
- **Consumer**: regex-patched into a generated `SharedAssemblyInfo.Worktree.cs` swapped in for `SharedAssemblyInfo.cs` at compile time, so the `StrideVersion.NuGetVersion` const matches the `.nupkg`.

Resolution runs via an inline `RoslynCodeTaskFactory` task; the steady-state cost is very low (once loaded after first call).

## CI / package builds

Gated off via `$(CI)` / `$(GITHUB_ACTIONS)` / `$(TF_BUILD)`. The gate is on the `<UsingTask>` element itself, so `RoslynCodeTaskFactory` never compiles the task on CI.

## Notes

- Stale entries (folder gone) are pruned before a new registration; freed slots are reused (`dev1` deleted while `dev2` is taken → next checkout becomes `dev1`, not `dev3`).
- Path casing: `OrdinalIgnoreCase` on Windows, `Ordinal` on Linux/macOS.
- UNC entries are skipped from pruning — a network glitch can't silently delete a real entry.
- `StrideWorktreeId` env var / MSBuild property overrides the ledger.
- `dotnet msbuild -t:StrideRegisterWorktree` registers a checkout explicitly without a build.

## Alternative considered

Going the other way: have each locally-created sample remember which Stride checkout created it (a back-reference in the project), and at build time resolve packages directly from that checkout's outputs — bypassing the shared NuGet feed/cache entirely. No version mechanics, no ledger, no CI gate.

Rejected because it ties the sample to a specific folder on disk: moving or renaming the checkout breaks the sample, and it doesn't help externally-distributed samples (GitHub clones, tutorials, samples shared between devs).
Also it forces us to maintain a completely different system with its own class of bugs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -169,3 +169,6 @@ project.lock.json
 
 # Auto-generated local developer settings (Stride.Platform.props bootstrap)
 /build/Stride.Local.props
+
+# Auto-generated per-checkout by Stride.WorktreeVersion.targets (per-checkout NuGet version suffix)
+/sources/shared/SharedAssemblyInfo.Worktree.cs
--- a/sources/sdk/Stride.Build.Sdk/Sdk/Sdk.targets
+++ b/sources/sdk/Stride.Build.Sdk/Sdk/Sdk.targets
@@ -35,6 +35,7 @@
   <Import Project="Stride.AssemblyProcessor.targets" />
   <Import Project="Stride.CodeAnalysis.targets" />
   <Import Project="Stride.PackageInfo.targets" />
+  <Import Project="$(StrideRoot)sources\targets\Stride.WorktreeVersion.targets" Condition="'$(_StrideWorktreeVersionTargetsImported)' != 'true'" />
 
   <!-- ================================================================ -->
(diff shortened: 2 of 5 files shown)


xen2:
Hmm, somehow I pushed this as part of another branch by mistake...
Let me know if you think I should revert it.

Issue #3183, opened 2026-05-17
Need ability to extend to the scene editor


Basewq:
**Is your feature request related to a problem? Please describe.**
Trying to create an extension to the scene editor, eg. spline editor, terrain editor, is difficult due to current code structure.

I have created a [proof of concept project](https://github.com/Basewq/BasewqStrideSplineTest)  of a spline editor, along with a write up of [issues that need to be resolved.](https://github.com/Basewq/BasewqStrideSplineTest/blob/main/EditorIntegrationIssues.md) for proper integration (currently uses reflection to bypass engine issues).

The TLDR is that developers need to be able to register an editor control, and be able to properly take control of the scene in a predictable way.

You can extend `EditorGameMouseServiceBase` and block input with `IsControllingMouse = true`. 
For something like a spline editor, this is necessary for blocking the entity selector service which controls selection/deselection of an entity (or multiple entities) that you click in the scene, however doing so also blocks the editor camera and transform gizmo (which we don't want to block).

---
**Describe the solution you'd like**
Most common solution is to extend some base editor tool class, with configuration like if it is targeted for an entity component or a global tool.

Looking at other engines:

**Unreal Engine** divides scene editing into distinct modes (Select, Landscape, Foliage, etc)
https://dev.epicgames.com/documentation/unreal-engine/level-editor-modes-in-unreal-engine
Not counting their blueprint solution, you need to extend `UEdMode` and register that class
https://dev.epicgames.com/documentation/unreal-engine/API/Editor/UnrealEd/UEdMode
Does not appear to have input blocking, I assume modes are exclusive (or it is done in a different way).

**Unity** tries to be more context based, based on the current entity selection, then select the edit mode:
https://docs.unity3d.com/6000.4/Documentation/ScriptReference/EditorTools.EditorTool.html
https://www.foundations.unity.com/patterns/contextual-tooling
You extend `EditorTool` and define the context (eg. component specific or global). The different editing modes only appear after selecting the correct object (assuming you've made a component specific editor).
It is very IMGUI centric, so things like displaying handles, input handling (and consuming input) is done in IMGUI way.

**Godot** can be context based, based on the current entity ("node") selection, however it does not appear have different edit modes:
https://docs.godotengine.org/en/latest/classes/class_editorplugin.html
You extend `EditorPlugin` and define context with `_handles`
Can block input with `_forward_3d_gui_input`

**O3DE** is also context based:
https://docs.o3de.org/docs/user-guide/programming/components/editor-components/
[...]

PR #3186, opened 2026-05-23
[Editor] Change TransformationGizmo Update to be synchronous and change entity duplication behavior

Basewq:
# PR Details

This serves as a bit of prep work leading to #3183.

The purpose of this PR is to remove the async part of `TransformationGizmo.Update`.
The only two places where async is used is:
1. Keyboard shortcut to change the transform gizmo mode (translate/rotate/scale via `W`/`E`/`R` keys)
2. Hold `Ctrl` -> Drag gizmo to duplicate entities (note: an existing quirk is you must focus in the scene before `Ctrl` key down is detected, which is not fixed in this PR).

My changes and reasoning:
1. No longer await UI invoke. This is an example of `game -> editor -> game`, which relies on the `editor -> game` return to set update the transform gizmo mode (refer to Related Issue section for a bit more context).
It now just treats it as a fire-and-forget request to the editor, which will 'eventually' update the mode back to the game thread.
I'm not really sure if there is a cleaner solution, other than either 'predicting' the mode change (ie. update directly as well), or block the gizmo until the editor 'confirms' the request.

2. Moved the entity duplication logic out of the transform gizmo and into EditorGameEntityTransformService.
This makes the transform gizmo reusable for other things which may or may not have 'duplication' logic (eg. spline tool).
Another change, which might be more debatable, is now the initial duplication is only done on the game side (temp cloned entities), then at the end of the transform the request is sent to the editor.
Doing it this way also means there is only **one** undo/redo transaction, whereas the previous behavior had two transactions (1st transaction on the drag start to generate entities in place, then a 2nd transaction for the position update), ie. you had to undo twice to actually remove the duplicated objects.
One other quirk is when duplicating a large number of entities (eg. use 3rd Person Platformer template and duplicate the entire 'Platform' folder via `Ctrl` + drag), the original behavior has a large initial freeze since it clones at the start.
This change makes the large freeze occur at the end of the transform (along with a progress bar popup if longer than 500ms), however this does introduce a different quirk where the preview entities disappear then the real objects are added afterwards.

3. (New functionality) There is also now the ability to hit Escape key to cancel the transform when you're in the middle of transforming (no transaction goes the editor).

## Related Issue

Prep work for #3183

Regarding further context for change 1:
There is a bit of concern with the current architecture of the 'game services', which may need additional thoughts on.
The issue is the back and forth marshaling between 'WPF UI thread' (ie. where the asset/quantum data live) and the 'game thread' (ie. live scene editor).

There are some 'game services' that contain both the game scene data and the 'view model'/UI thread data (ie. implement `IEditorGameService` and `IEditorGameViewModelService` as a single class), and can end up doing things like `'game thread' -> await InvokeAsync to 'UI thread' -> InvokeAsync to 'game thread'`, which is exactly what the gizmo code was doing.

Regarding further context for change 2:
The future goal is to change classes that inherit `EditorGameMouseServiceBase` to only have sync update methods, so they can ultimately be bundled through some 'edit mode' class and updated in some specified order (and optionally enabled/disabled depending on the mode), rather than the free-for-all approach that's happening now (they will still have the option to register an independent task via ScriptSystem if they need it).

## Types of changes

Not sure if doc change is required or not, `TranslationGizmo`/`ScaleGizmo`/`RotationGizmo` are public, however these gizmos are part of `Stride.Assets.Presentation`, which the general user cannot accidentally stumble into due to the lack of proper plugin integration.

## Checklist

### DIFF
--- a/sources/editor/Stride.Assets.Presentation/AssetEditors/EntityHierarchyEditor/Game/EditorGameEntityTransformService.cs
+++ b/sources/editor/Stride.Assets.Presentation/AssetEditors/EntityHierarchyEditor/Game/EditorGameEntityTransformService.cs
@@ -5,11 +5,6 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
-using Stride.Core.BuildEngine;
-using Stride.Core;
-using Stride.Core.Annotations;
-using Stride.Core.Extensions;
-using Stride.Core.Mathematics;
 using Stride.Assets.Presentation.AssetEditors.EntityHierarchyEditor.Services;
 using Stride.Assets.Presentation.AssetEditors.EntityHierarchyEditor.ViewModels;
 using Stride.Assets.Presentation.AssetEditors.GameEditor;
@@ -17,8 +12,14 @@ using Stride.Assets.Presentation.AssetEditors.GameEditor.Game;
 using Stride.Assets.Presentation.AssetEditors.GameEditor.Services;
 using Stride.Assets.Presentation.AssetEditors.Gizmos;
 using Stride.Assets.Presentation.SceneEditor;
+using Stride.Core;
+using Stride.Core.Annotations;
+using Stride.Core.BuildEngine;
+using Stride.Core.Extensions;
+using Stride.Core.Mathematics;
 using Stride.Editor.EditorGame.Game;
 using Stride.Engine;
+using Stride.Input;
 using Stride.Rendering;
 using Stride.Rendering.Compositing;
 
@@ -38,6 +39,12 @@ namespace Stride.Assets.Presentation.AssetEditors.EntityHierarchyEditor.Game
         private TransformationSpace space;
         private double gizmoSize = 1.0f;
         private bool dynamicSnappingInUse = false;
+
+        private bool isEntityDuplicationInProgress = false;
+        private Entity entityDuplicationPreviousEntityWithGizmo;
+        private IReadOnlyCollection<Entity> entityDuplicationPreviousSelection;
+        private readonly Dictionary<Entity, Entity> entityDuplicationSrcToPreviewEntityMap = [];
+        private readonly Dictionary<AbsoluteId, Entity> entityDuplicationSrcIdToPreviewEntityMap = [];
 
         public EditorGameEntityTransformService([NotNull] EntityHierarchyEditorViewModel editor, [NotNull] IEditorGameController controller)
         {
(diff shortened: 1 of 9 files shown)


sources/editor/Stride.Assets.Presentation/AssetEditors/EntityHierarchyEditor/Game/EditorGameEntityTransformService.cs
```diff
@@ -98,7 +105,14 @@ public Entity EntityWithGizmo
...
 
-        public override IEnumerable<Type> Dependencies { get { yield return typeof(IEditorGameEntitySelectionService); } }
+        public override IEnumerable<Type> Dependencies
+        {
+            get
+            {
+                yield return typeof(IEditorGameEntitySelectionService);
+                yield return typeof(EditorGameModelSelectionService);
+            }
+        }
```
Ethereal77, line 115: I know this is not the only place this is done like this but, why keep doing it like this? I mean, if the types are known at compile time (i.e., not dynamically determined), this is equivalent, even better if the JIT can optimize it more in the future:

Image

Basewq: I can change this if wanted and people don't mind having some in this way and others in the `yield return` code.
I think there is a subtle difference if they do include `base.Dependencies`, ie. `[.. base.Dependencies, typeof(A), typeof(B)]` since it's making an expanded array though the likelihood of an extremely large array through chaining is highly unlikely in practice.

xen2: Yes, I am fine to use new syntax from now on (for compactness & readability).
We can fix the others in separate PR later.
(but fine to merge as is and do everything later too, it's not a deal breaker)

Kryptos-FR:
Does it still work with multi-selection of entities that are not under the same sub-tree?

Basewq:
> Does it still work with multi-selection of entities that are not under the same sub-tree?

Yes, it uses existing code that gets the root level entities from the selection(s) then copies all of them.

For the time being, I've decided to put this on WIP status, as I am switching to another WIP thing, which may or may not use this (or parts of this).

I think the overall idea is still valid:
```
Drag -> create proxies -> if completed -> commit & reconcile proxy/duplicated entities
                       -> if canceled -> remove proxies (no transactions committed)
```
However I feel someone might come up with a cleaner implementation than here.

Basewq:
Some parts may return, eg. transform gizmo cancellation event args.

PR #3189, opened 2026-05-24
Avalonia test runner: UI revamp

xen2:
# Avalonia test runner: UI revamp

Rebuilds the interactive test runner (Avalonia window shown when a Stride test exe is launched with `STRIDE_TESTS_INTERACTIVE=1`).

## Highlights

- **Filter textbox** (auto-focused) + **status chips** (All / Passed / Failed / Skipped / Not run) with live counts.
- **Per-row run buttons:** test mode (gold compare diff) & interactive run (preview only, doesn't close window)
- **Re-run failed** + **Run filtered/selected** with `Ctrl+R` / `F5` shortcuts.
- **Multi-select** in the tree.
- **Save image** checkbox (forces save on success) and **RenderDoc** capture-mode dropdown (Never / On error / Always).
- **Inspect pane**: failure message + stack trace + live-streamed stdout/stderr with `Warning:` / `Error:` tinting.
- **Summary bar** with running pass/fail/skip totals and elapsed time.
- **Keyboard nav**: `/` or `Ctrl+F` to focus filter, `Esc` to clear, `Up`/`Down`/`PageUp`/`PageDown` from the filter to navigate the tree, `Enter` / `Shift+Enter` to run / preview.
- **Dark theme** + **responsive layout** (toolbar wraps, main split flips vertical on narrow / portrait viewports).
- **Gold image preview** + open folder in explorer

A screenshot is worth a thousand words:

image

## Types of changes

## Checklist

### DIFF
--- a/sources/tests/xunit.runner.stride/App.axaml
+++ b/sources/tests/xunit.runner.stride/App.axaml
@@ -1,8 +1,23 @@
 <Application xmlns="https://github.com/avaloniaui"
              xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
              x:Class="xunit.runner.stride.App"
-             RequestedThemeVariant="Light">
+             RequestedThemeVariant="Dark">
   <Application.Styles>
     <FluentTheme />
   </Application.Styles>
+  <Application.Resources>
+    <ResourceDictionary>
+      <!-- Fluent Dark's default Window background is pure black, which is too contrasty for
+           the runner. We define our own theme-aware brush so the color is centralized and
+           still flips automatically if the runner is ever opened in a light variant. -->
+      <ResourceDictionary.ThemeDictionaries>
+        <ResourceDictionary x:Key="Dark">
+          <SolidColorBrush x:Key="WindowBackgroundBrush" Color="#1E1E1E" />
+        </ResourceDictionary>
+        <ResourceDictionary x:Key="Light">
+          <SolidColorBrush x:Key="WindowBackgroundBrush" Color="#F3F3F3" />
+        </ResourceDictionary>
+      </ResourceDictionary.ThemeDictionaries>
+    </ResourceDictionary>
+  </Application.Resources>
 </Application>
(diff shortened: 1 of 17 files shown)


sources/tests/xunit.runner.stride/Views/MainWindow.axaml
```diff
@@ -6,6 +6,7 @@
         mc:Ignorable="d" d:DesignWidth="800" d:DesignHeight="450"
         x:Class="xunit.runner.stride.Views.MainWindow"
         Icon="/Assets/avalonia-logo.ico"
-        Title="XUnit Runner (Stride)">
+        Title="XUnit Runner (Stride)"
+        Background="#1E1E1E">
```
Kryptos-FR: This should use a color from the theme so that if the users prefers a light theme it picks the corresponding background color. Or no color as I think the theme as one by default.

sources/tests/xunit.runner.stride/App.axaml
```diff
@@ -1,7 +1,7 @@
 <Application xmlns="https://github.com/avaloniaui"
              xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
              x:Class="xunit.runner.stride.App"
-             RequestedThemeVariant="Light">
+             RequestedThemeVariant="Dark">
```
Kryptos-FR, line 4: Should probably be `Default` instead so that it follows the system theme.

bnwa9723:
**Go Xen2 !!!**

sources/tests/xunit.runner.stride/App.axaml
xen2: I will keep it as is, dark only for now so that I can focus other things & simplify maintenance for now.
(I chose dark to save my poor eyes during those long coding nights ^^)

PR welcome of course if somebody think it's important.

bnwa9723: That makes sense.

sources/tests/xunit.runner.stride/Views/MainWindow.axaml
xen2: I remember setting it manually because default was ugly
image
I thought it was the default but it was simply because it didn't properly apply the theme one.
Adjusting now

xen2: Fixed in latest version
Somehow I had to add it to App.xaml; `{DynamicResource SolidBackgroundFillColorBase}` didn't work as is.
I wonder if it was really worth the 10+ new lines vs a hardcoded color for this specific dev tool though?

PR #3202, opened 2026-06-02
feat: Include asset name when logging asset-related issues

Eideren:
# PR Details
The different log display do not `ToString()` log messages they receive, instead they extract the `LogType` and `Text`, this PR ensures that the `Text` for AssetLogMessage includes the asset location.
image

The different logger omit certain information depending on the filters, so using `ToString` or introducing a `ToStringShort()` and `ToStringLong` did not feel like the best way to go about solving this.

## Related Issue
None

## Types of changes

## Checklist

### DIFF
--- a/sources/assets/Stride.Core.Assets/Diagnostics/AssetLogMessage.cs
+++ b/sources/assets/Stride.Core.Assets/Diagnostics/AssetLogMessage.cs
@@ -99,6 +99,20 @@ public class AssetLogMessage : LogMessage
 
     public int Character { get; set; }
 
+    public override string Text
+    {
+        get
+        {
+            if (AssetReference?.Location != null)
+                return $"{AssetReference.Location}({Line + 1},{Character + 1}): {base.Text}";
+            return base.Text;
+        }
+        set
+        {
+            base.Text = value;
+        }
+    }
+
     /// <summary>
     /// Gets or sets the message code.
     /// </summary>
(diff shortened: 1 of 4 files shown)


sources/assets/Stride.Core.Assets/Diagnostics/AssetLogMessage.cs
```diff
@@ -99,6 +99,20 @@ public static AssetLogMessage From(Package? package, IReference? assetReference,
...
 
+    public override string Text
+    {
+        get
+        {
+            if (AssetReference?.Location != null)
+                return $"{AssetReference.Location}({Line + 1},{Character + 1}): {base.Text}";
+            return base.Text;
+        }
+        set
```
xen2: Good change!
However I would rather avoid getter/setter discrepancy like this.

Few options:
- `Text` stays as is, we add a virtual `FormattedText` (get-only).
- In practice, `Text` is only set internally, so we could keep the setter private and/or use a field. Then `Text` becomes get-only (basically acting as the `FormattedText` mentioned before).

Eideren: I didn't really like this change either, but wasn't comfortable changing public apis.
Removed the setter on `Text` - `FormattedText` has larger implications both on changing the 63 loc which reads from `Text` as well as what should or should not be formatted within

xen2: Sounds good. Not a bad thing to make it readonly anyway.

PR #3209, opened 2026-06-07
Fire tracking remove events after mutation

### LINKED ISSUE #3171: TrackingCollection.CollectionChanged fired too early when removing items

**Version**: 4.3.0.2507 and prior

**Platform(s)**: All

**Describe the bug**
`TrackingCollection.Remove`/`RemoveAt` fires the `CollectionChanged` event before it changes the collection.
This means it is not safe to query the collection in the event handler.

**To Reproduce**
Rough code:
```
var coll = new TrackingCollection<int>();
coll.Add(1);
coll.CollectionChanged += (sender, e) => { Debug.Assert(coll.Count == 0); };
coll.RemoveAt(0);  // Will fail the assert
```

**Expected behavior**
Item should already be removed when the event is fired.

**Workaround**
Delay any code that requires iterating or calling the collection directly when the event fires.

**Related discussion**
1. Not sure if `TrackingCollection` is meant to be used by external users. Having a `CollectionChanging` would ensure the correct event is fired for the correct timing (assuming you want to see the changing event instead of always changed event).
2. The set item method triggers 2 events `NotifyCollectionChangedAction.Remove` -> `NotifyCollectionChangedAction.Add`, instead of a singular `NotifyCollectionChangedAction.Replace` which is a bit clunky but would be considered another breaking change to fix.
3. `FastTrackingCollection` has the same issue.
4. `TrackingDictionary` has the same issue.

marko1olo:
Closes #3171.

`TrackingCollection`, `FastTrackingCollection`, and `TrackingDictionary` all exposed remove notifications before the backing collection had actually changed. That made `CollectionChanged` handlers observe stale state during `Remove`, `RemoveAt`, and clear-driven remove notifications.

This updates remove notifications so handlers observe post-removal state while keeping the existing payloads and remove event ordering:

- `TrackingCollection.RemoveItem` stores the removed item, removes it, then fires `Remove`.
- `FastTrackingCollection.RemoveItem` does the same for fast event args.
- `TrackingCollection.ClearItems` and `FastTrackingCollection.ClearItems` snapshot remove payloads, clear the collection, then raise remove events in the existing reverse order.
- `TrackingDictionary.Remove(TKey)` removes before firing.
- `TrackingDictionary.Remove(KeyValuePair<TKey,TValue>)` now also raises the remove event when it removes an item.

`SetItem` behavior is intentionally unchanged because its existing remove/add notifications are marked `CollectionChanged=false`, and the issue calls replace semantics out as a separate breaking-change question.

Local validation:

- `dotnet restore sources/core/Stride.Core.Tests/Stride.Core.Tests.csproj`
- `dotnet test sources/core/Stride.Core.Tests/Stride.Core.Tests.csproj --filter "FullyQualifiedName~TrackingCollectionTests|FullyQualifiedName~FastTrackingCollectionTests|FullyQualifiedName~TrackingDictionaryTests" --no-restore --nologo` -> 27 passed
- `git diff --cached --check`

Notes:

- The local checkout was sparse/blobless, so I added `sources/tests/xunit.runner.stride` to the sparse checkout to make the test project compile locally.
- Push used `GIT_LFS_SKIP_PUSH=1` because this commit does not modify LFS files, and the local partial clone reported an unrelated missing promisor object for an existing FFmpeg LFS blob during the default LFS pre-push scan.

### DIFF
--- a/sources/core/Stride.Core.Tests/Collections/FastTrackingCollectionTests.cs
+++ b/sources/core/Stride.Core.Tests/Collections/FastTrackingCollectionTests.cs
@@ -44,12 +44,14 @@ public class FastTrackingCollectionTests
     {
         var collection = new FastTrackingCollection<string> { "test" };
         var eventRaised = false;
+        var itemRemovedBeforeEvent = false;
         FastTrackingCollectionChangedEventArgs capturedArgs = default;
 
         FastTrackingCollection<string>.FastEventHandler<FastTrackingCollectionChangedEventArgs> handler =
             (object sender, ref FastTrackingCollectionChangedEventArgs e) =>
         {
             eventRaised = true;
+            itemRemovedBeforeEvent = !collection.Contains("test");
             capturedArgs = e;
         };
         collection.CollectionChanged += handler;
@@ -60,6 +62,7 @@ public class FastTrackingCollectionTests
         Assert.Equal(NotifyCollectionChangedAction.Remove, capturedArgs.Action);
         Assert.Equal("test", capturedArgs.Item);
         Assert.Equal(0, capturedArgs.Index);
+        Assert.True(itemRemovedBeforeEvent);
     }
 
     [Fact]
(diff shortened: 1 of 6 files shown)


Kryptos-FR:
I think this has already been discussed and it might be on purpose. The event could maybe be called `Changing` instead of `Changed` or we could expose both. However, making the current change will impact the behavior or the calling sites and might not be desirable.

In addition, please follow the template when creating PR or issues.

marko1olo:
Thanks for the review. I agree this is a behavior-compatibility risk: even though #3171 describes the current timing as surprising, changing the timing of the existing event can break callers that depend on the pre-mutation notification.

I am going to close this PR rather than push the change as-is. A safer direction would need an explicit API decision first, for example whether Stride wants separate changing/changed notifications or only documentation/naming clarification around the existing timing.

PR #3212, opened 2026-06-08
fix: Finalize array support in Quantum

### LINKED ISSUE #2386: Stride Prefab Selecting from array not working

**Version**: 4.2.0.2188

**Platform(s)**: Windows

**Describe the bug**
When I create an array of prefabs in my code I can't select it in the editor 

**To Reproduce**
1. create a script with a list of prefabs.
[https://youtu.be/7-qpzwTlSW8](url)

Eideren:
# PR Details
Arrays previously had limited support in the gamestudio, asset types could not be assigned to one of its slots for example. This PR fixes that.

Note that this PR does not affect collection id support for arrays; derived arrays still won't inherit appropriately as it requires rewriting some of the internal logic called through `AssetPropertyGraph.ReconcileWithBaseNode` that work with the expectation that one can remove and add to the collection.

Also note that some of these changes refactor some of the reflection logic around collection types to move type-specific logic to their own implementation. Let me know if this is a pain on your side @Kryptos-FR and we'll see what I can do to help you out.

## Related Issue
Fixes #2386

## Types of changes

## Checklist

### DIFF
--- a/sources/assets/Stride.Core.Assets/CollectionIdGenerator.cs
+++ b/sources/assets/Stride.Core.Assets/CollectionIdGenerator.cs
@@ -110,6 +110,12 @@ public class CollectionIdGenerator : DataVisitorBase
 
     private bool ShouldGenerateItemIdCollection(object collection)
     {
+        // Arrays do not support Collection Ids right now;
+        // a bunch of blockers in AssetPropertyGraph.ReconcileWithBaseNode around the expectations
+        // that add/remove work on this type of collection
+        if (collection is Array)
+            return false;
+
         // Do not generate for value types (collections that are struct) or null
         if (collection?.GetType().IsValueType != false)
             return false;
--- a/sources/core/Stride.Core.Reflection/TypeDescriptors/CollectionBaseDescriptor.cs
+++ b/sources/core/Stride.Core.Reflection/TypeDescriptors/CollectionBaseDescriptor.cs
@@ -34,7 +34,7 @@ public abstract class CollectionBaseDescriptor : ObjectDescriptor
     /// Validates whether the type of this key would be valid
     /// </summary>
     /// <remarks>
-    /// For ints it also checks if its value is above 0
+    /// For ints it also checks if its value is greater than or equal to 0
     /// </remarks>
     public abstract bool IsKeyValid(object? key);
 
--- a/sources/core/Stride.Core.Reflection/TypeDescriptors/CollectionBaseDescriptor.cs
+++ b/sources/core/Stride.Core.Reflection/TypeDescriptors/CollectionBaseDescriptor.cs
@@ -1,0 +1,45 @@
+// Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net)
+// Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using Stride.Core.Yaml.Serialization;
+
+namespace Stride.Core.Reflection;
+
+public abstract class CollectionBaseDescriptor : ObjectDescriptor
+{
+    protected CollectionBaseDescriptor(ITypeDescriptorFactory factory, Type type, bool emitDefaultValues, IMemberNamingConvention namingConvention) : base(factory, type, emitDefaultValues, namingConvention)
+    {
+    }
+
+    /// <summary>
+    /// Returns the value at the key specified in the collection.
+    /// </summary>
+    /// <remarks> The key would be the integer index for list-like collections </remarks>
+    public abstract object? GetValue(object collection, object key);
+
+    /// <summary>
+    /// Sets the value at key to value in the collection provided.
+    /// </summary>
+    /// <remarks> The key would be the integer index for list-like collections </remarks>
+    public abstract void SetValue(object collection, object key, object? value);
+
+    /// <summary>
+    /// The type of value returned through <see cref="GetValue"/>
+    /// </summary>
+    public abstract Type ValueType { get; }
+
+    public abstract IEnumerable<object> EnumerateKeys(object collection);
+
+    /// <summary>
+    /// Validates whether the type of this key would be valid
+    /// </summary>
+    /// <remarks>
+    /// For ints it also checks if its value is above 0
+    /// </remarks>
+    public abstract bool IsKeyValid(object? key);
+
+    /// <summary>
+    /// Checks whether there is a value for this key
+    /// </summary>
+    public abstract bool ContainsKey(object collection, object? key);
+}
--- a/sources/core/Stride.Core.Reflection/TypeDescriptors/DictionaryDescriptor.cs
+++ b/sources/core/Stride.Core.Reflection/TypeDescriptors/DictionaryDescriptor.cs
@@ -10,7 +10,7 @@ namespace Stride.Core.Reflection;
 /// <summary>
 /// Provides a descriptor for a <see cref="IDictionary"/>.
 /// </summary>
-public class DictionaryDescriptor : ObjectDescriptor
+public class DictionaryDescriptor : CollectionBaseDescriptor
 {
     private static readonly List<string> ListOfMembersToRemove = ["Comparer", "Keys", "Values", "Capacity"];
 
@@ -86,7 +86,7 @@ public class DictionaryDescriptor : ObjectDescriptor
     /// Gets the type of the value.
     /// </summary>
     /// <value>The type of the value.</value>
-    public Type ValueType { get; }
+    public override Type ValueType { get; }
 
     /// <summary>
     /// Gets or sets a value indicating whether this instance is pure dictionary.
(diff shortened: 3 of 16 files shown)


sources/assets/Stride.Core.Assets/CollectionIdGenerator.cs
```diff
@@ -110,6 +110,12 @@ public override void VisitSet(IEnumerable set, SetDescriptor descriptor)
 
     private bool ShouldGenerateItemIdCollection(object collection)
     {
+        // Arrays do not support Collection Ids right now;
```
Ethereal77, line 113: As it is a "right now" and talks about a known shortcoming of the current solution, I'd put a "TODO" there so we won't forget

Eideren: Created an issue (#3213) instead, TODOs feel a bit too easy to ignore, let me know if you feel like I should include it with the comment anyway

Ethereal77: No, you are right. An issue is better

Kryptos-FR:
I'll try to review it on the weekend.

sources/core/Stride.Core.Reflection/TypeDescriptors/SetDescriptor.cs
```diff
@@ -212,4 +214,24 @@ protected override bool PrepareMember(MemberDescriptorBase member, MemberInfo me
 
         return base.PrepareMember(member, metadataClassMemberInfo);
     }
+
+    public override IEnumerable<object> EnumerateKeys(object coll)
+    {
+        var enumerable = coll as System.Collections.IEnumerable;
+        var enumerator = enumerable.GetEnumerator();
```
Kryptos-FR: `using` to dispose the enumerator

sources/core/Stride.Core.Reflection/TypeDescriptors/SetDescriptor.cs
```diff
@@ -212,4 +214,24 @@ protected override bool PrepareMember(MemberDescriptorBase member, MemberInfo me
 
         return base.PrepareMember(member, metadataClassMemberInfo);
     }
+
+    public override IEnumerable<object> EnumerateKeys(object coll)
+    {
+        var enumerable = coll as System.Collections.IEnumerable;
```
Kryptos-FR: This will cast as `IEnumerable?` which could cause a `NullReferenceException` below. Either direct cast to `(IEnumerable)` or add a null-check.

sources/core/Stride.Core.Reflection/TypeDescriptors/SetDescriptor.cs
```diff
@@ -212,4 +214,24 @@ protected override bool PrepareMember(MemberDescriptorBase member, MemberInfo me
...
+        var enumerator = enumerable.GetEnumerator();
+        while (enumerator.MoveNext())
```
Kryptos-FR: Any reason why a simple `foreach` wasn't used here (like in DictionaryDescriptor.cs:246)?

sources/core/Stride.Core.Reflection/TypeDescriptors/DictionaryDescriptor.cs
```diff
@@ -158,7 +158,7 @@ public void Remove(object dictionary, object key)
     /// </summary>
     /// <param name="dictionary">The dictionary.</param>
     /// <param name="key">The key.</param>
-    public bool ContainsKey(object dictionary, object key)
+    public override bool ContainsKey(object dictionary, object key)
```
Kryptos-FR: Inconsistent signature. Should be `bool ContainsKey(object dictionary, object key)` but with a `ArgumentNullException.ThrowIfNull(key)`.

Either that, or the base signature in `CollectionBaseDescriptor` shoudln't allow a null key.

sources/presentation/Stride.Core.Quantum/DynamicNode.cs
```diff
@@ -182,32 +182,10 @@ protected static bool IsIndexExisting(IGraphNode node, NodeIndex index)
...
-                        && index.Int >= 0
-                        && index.Int < collectionDescriptor.GetCollectionCount(value))
-                    {
-                        return true;
-                    }
-                }
-            }
-            else if (descriptor is DictionaryDescriptor dictionaryDescriptor)
-            {
-                if (dictionaryDescriptor.KeyType.IsInstanceOfType(index.Value)
```
Kryptos-FR, line 206: We probably still need to check the type and short-circuit, otherwise it will throw in dictionaryDescriptor.

Either that, or opposite to the comment in `DictionaryDescriptior` return false instead of throwing if key is null.

Whatever the choice is, we need new tests to cover these cases.

sources/presentation/Stride.Core.Quantum/DynamicNode.cs
```diff
@@ -221,21 +199,11 @@ protected static bool IsIndexValid(IGraphNode node, NodeIndex index)
...
-            }
-            else
-            {
-                return index.IsInt && index.Int >= 0;
-            }
-        }
-        else if (node.Descriptor is DictionaryDescriptor dictionaryDescriptor)
-        {
-            return dictionaryDescriptor.KeyType.IsInstanceOfType(index.Value);
+            return collectionBaseDescriptor.IsKeyValid(index.Value);
```
Kryptos-FR, line 204: Notice how here we do check if the key is valid.

sources/core/Stride.Core.Reflection/TypeDescriptors/CollectionBaseDescriptor.cs
```diff
@@ -0,0 +1,45 @@
...
+    /// </summary>
+    public abstract Type ValueType { get; }
+
+    public abstract IEnumerable<object> EnumerateKeys(object collection);
+
+    /// <summary>
+    /// Validates whether the type of this key would be valid
+    /// </summary>
+    /// <remarks>
+    /// For ints it also checks if its value is above 0
```
Kryptos-FR: 0 is allowed

Kryptos-FR (changes requested):
Overall good, a few minor issues, and we should add more tests to cover the specifics.

Possible area:
- reflection: DescriptorTests for each descriptor type, test each methods from the base class and the correct behavior for it
- quantum: TestCollections add TestArrays
- asset: in Stride.Core.Quantum.test, serialize/deserialize a component with an array of content references, do a round trip
- archetype regression (will be useful for #3213): a base + derived asset using arrays. Confirm no exception during `ReconcileWithBaseNode` and no `~Id` emitted.

sources/core/Stride.Core.Reflection/TypeDescriptors/SetDescriptor.cs
Eideren: The three issues you raised here can be explained by the fact that it is a straight port of the one it replaces in [GraphNodeBase.cs](https://github.com/stride3d/stride/pull/3212/changes#diff-c4d3d0c581adba8bb2bdeaa9c33bd9f52310b70017d874370aa7a20563c2b6be). 
Still, I agree it could be improved. Will replace the whole thing by a direct cast and a foreach.

sources/core/Stride.Core.Reflection/TypeDescriptors/DictionaryDescriptor.cs
Eideren: Hashsets do support null keys, so changing the base signature doesn't sound like the right call.
And throwing when providing a null key doesn't sound like the right one either given that the base signature explicitly allows null keys, so I think the sanest solution to this is to return false when the key is null. 
This does break the expectation one might have that this API matches the `IDictionary` one 1:1 of throwing on null keys side effect for this function signature, but to me this seems fairly benign compared to the others ?
Let me know what you think.

sources/presentation/Stride.Core.Quantum/DynamicNode.cs
Eideren: See other comment, will produce tests once we resolve this.

sources/presentation/Stride.Core.Quantum/DynamicNode.cs
Eideren: I think your comment here is just a hint for the one above to let me know how I could use this as part of the changes for `IsIndexExisting()`. I got it, let's continue in the two other comments. If you meant something else by it let me know.

Eideren:
Kindly pinging @Kryptos-FR if he can reply to the discussion above - no rush, just in case you forgot ;)

sources/core/Stride.Core.Reflection/TypeDescriptors/DictionaryDescriptor.cs
Kryptos-FR: Returning false when null seems best for now.

Kryptos-FR:
@Eideren I did forget ;). Looks ok to me now.

Kryptos-FR (approved):
With a few more tests it should be good.

Eideren:
@Kryptos-FR Looks like the `DynamicNode` and all derived types are unused in the source - the only other spot where they are being used is in a test suite which is entirely skipped, and does throw when forced.

I don't think it makes sense to maintain these given that they are not really user-facing, nor particularly useful ?

I added tests to cover the other changes.

Kryptos-FR:
@Eideren yeah `DynamicNode` looks like an unfinished attempt. I don't see any use case for it today so I think it's fine to remove it.

Eideren:
@Kryptos-FR done in #3265 given that it's unrelated to these changes

PR #3228, opened 2026-06-20
fix: Incorrect bounding box for sprites with non-default Center

### LINKED ISSUE #3227: Sprite disappears with non standard Center
**Release Type**: GitHub 

**Version**: master 

**Platform(s)**: Reproduced on Windows; not platform-specific

**Describe the bug**
`SpriteComponent`'s bounding box, used for frustum culling, does not account for a non-default `Sprite.Center` (pivot offset). When a sprite's pivot is moved away from the texture's geometric center — for example to align the sprite's bottom edge with the entity's position so it appears to stand on the ground — the bounding box used for culling stays centered on the entity's world position instead of following the visual offset of the sprite.
As a result, the sprite is culled (disappears) as soon as the **entity's position**, not the sprite's actual visible edge, crosses the frustum boundary. This causes the sprite to vanish from view while a significant portion of it (up to roughly half, depending on the pivot offset) is still visually on screen.As a result, the sprite is culled (disappears) as soon as the **entity's position**, not the sprite's actual visible edge, crosses the frustum boundary. This causes the sprite to vanish from view while a significant portion of it (up to roughly half, depending on the pivot offset) is still visually on screen.

**To Reproduce**

1. Import a texture, e.g. 1920x1080 pixels.
2. Add a `SpriteComponent` to an entity using this texture, with `CenterFromMiddle` enabled.
3. Set `Center` to `(0, 540)` — this offsets the pivot down by half the texture height, moving the sprite's pivot to its bottom edge.
4. Position the entity in the scene so the sprite sits at the desired location (e.g. on the ground, `Position.Y` set so the bottom of the sprite touches the floor).
5. In running game, move the camera or the entity so the sprite moves toward the bottom edge of the screen.
6. Observe that the sprite disappears entirely as soon as the entity's position (i.e. the midpoint of the sprite, before the Center offset) crosses the bottom of the screen — while the upper half of the sprite is still visually within the camera's view.

**Expected behavior**

The sprite should remain visible and only be culled once it is actually fully outside the camera's frustum.

**Screenshots**

https://github.com/user-attachments/assets/ac098014-0385-4350-bd03-8c91481ce247

w0wca7a:
# PR Details

When a sprite's pivot is offset from the geometric center of the texture (e.g. anchoring the sprite to its bottom edge so it sits on the ground), the sprite is rendered visually shifted relative to the entity's position — but the bounding box used for frustum culling stays centered on the entity itself.

## Related Issue

fixes: #3227
## Types of changes

## Checklist


xen2:
👋 About this PR.

PR #3230, opened 2026-06-22
Fix minimized window raw client size handling

Redwarx008:
# PR Details

Fixes minimized-window resize handling so minimized or zero-sized native client areas are not treated as valid swapchain/presenter sizes.

## Problem observed

We hit this in a Stride game using the WinForms game window:

1. Start the game in windowed mode.
2. Minimize the game window.
3. Restore it from the taskbar.

After restore, the window could come back as a tiny title-bar-only window instead of its previous client size. In the rendering path this also allowed a minimized/tiny client size to propagate as a valid render size, eventually producing a `1x1` viewport/backbuffer state. With post-processing enabled, that can fail in the luminance/downscale chain with:

```text
System.InvalidOperationException: Input and output texture cannot have same size [(1,1,1)]
```

The issue is not specific to a render feature or post-processing effect. The root problem is that the window/device resize path was using `ClientBounds` for resize decisions. `ClientBounds` intentionally clamps dimensions to at least `1x1` to keep rendering code away from zero-sized viewports, but that makes it unsafe as the source of truth for deciding whether a native window size should rebuild the swapchain or presenter.

During minimize/restore, WinForms can report a zero client area, and in some cases a tiny non-zero minimized client area. If that value is treated as a valid user resize, it can update cached/preferred window size and resize the presenter/backbuffer to an invalid tiny size. Once that happens, later rendering code sees a legal-looking `1x1` size rather than an invalid minimized window state.

This PR separates those concerns:

- `ClientBounds` can continue to provide rendering-safe clamped bounds.
- `RawClientSize` is used by window/device resize decisions so minimized or zero-sized native client areas can be rejected before they affect swapchain/presenter sizing.

This PR:

- Adds `GameWindow.RawClientSize` for resize/device decisions that need native client size rather than clamped `ClientBounds`.
- Skips zero-sized WinForms client-size notifications before they update cached/preferred window sizes.
- Treats minimized WinForms windows as raw size `(0, 0)`, even if the host reports a tiny non-zero client area.
- Uses raw client size in `GraphicsDeviceManager` and `GameWindowRenderer` before resizing the backbuffer or presenter.
- Adds an SDL override so SDL resize paths do not inherit the clamped `ClientBounds` fallback.

## Related Issue

No existing issue found. This was reproduced locally from a game window minimize/restore path that produced an invalid tiny render size during restore.

## Types of changes

## Checklist

Validation performed:

- `dotnet build sources/engine/Stride.Games/Stride.Games.csproj -c Debug`
- `dotnet build sources/editor/Stride.GameStudio/Stride.GameStudio.csproj -c Debug`
- Launched `Stride.GameStudio.exe`; process stayed alive and responsive, with the main window reaching `Opening session...`.
- Result: 0 errors. Existing warnings remain.


sources/engine/Stride.Games/GameWindowRenderer.cs
```diff
@@ -202,19 +202,19 @@ protected override void Destroy()
         ///   otherwise, it defaults to <see cref="PixelFormat.R8G8B8A8_UNorm"/>.
         /// </param>
         /// <returns>
-        ///   An <see cref="Int2"/> structure representing the width and height of the requested Back-Buffer size.
-        ///   If the preferred Back-Buffer dimensions are not set or the window has been resized by the user,
-        ///   the current window dimensions are used.
+        ///   <see langword="true"/> when a valid requested size could be determined; otherwise, <see langword="false"/>.
         /// </returns>
-        private Int2 GetRequestedSize(out PixelFormat format)
```
Ethereal77: If you are changing the return type and transforming it into an `out` parameter, reuse the previous `<returns>` documentation for the new `out Int2 size` parameter instead of removing it.

sources/engine/Stride.Games/GameWindowRenderer.cs
```diff
@@ -225,11 +225,15 @@ private Int2 GetRequestedSize(out PixelFormat format)
         ///   using the requested size and format. It configures the presentation parameters,
         ///   including Depth-Stencil format and presentation interval.
         /// </remarks>
-        protected virtual void CreateOrUpdatePresenter()
```
Ethereal77: The new `bool` return type needs a `<returns>` documentation.

sources/engine/Stride.Games/SDL/GameWindowSDL.cs
```diff
@@ -347,6 +347,16 @@ public override Rectangle ClientBounds
             }
         }
 
+        internal override Int2 RawClientSize
```
Ethereal77, line 351: `/// <inheritdoc/>` as other inherited properties.

sources/engine/Stride.Games/Desktop/GameWindowWinforms.cs
```diff
@@ -376,6 +381,19 @@ public override Stride.Core.Mathematics.Rectangle ClientBounds
             }
         }
 
+        internal override Int2 RawClientSize
```
Ethereal77, line 385: `/// <inheritdoc/>` as other inherited properties.

Ethereal77 (changes requested):
Overall looks good.

I'm not entirely convinced another property is needed for this, but as it is an `internal` detail, if it works that's ok for me.

I've left some comments regarding docs.

Ethereal77:
Also, not sure if possible, but it would be nice to have some tests to check this keeps working in the future (simply checking the logic around `ClientBounds`, `RawClientSize`, and `IsMinimized`). It would also serve to check the behavior is preserved if/when we introduce other windowing platforms.

xen2:
👋 About this PR.

PR #3231, opened 2026-06-23
Build tooling & sample versioning

xen2:
# PR Details

# Build tooling & sample versioning

Follow-up to the build/release modernization: refines the existing per-checkout dev-versioning, modernizes the asset compiler, and puts in-repo samples on a clean release-vs-dev versioning scheme. This is primarily build, tooling, and asset-pipeline work — no engine/runtime behavior change beyond one serialization default.

## Highlights

- **Asset compiler modernization.** `Stride.Core.Assets.CompilerApp` → `Stride.AssetCompiler`, with an action-verb CLI (`build`/`pack`/`upgrade`/`generate-code`) and a positional input.
  Upgrades reconcile asset bases by default (like a GameStudio open + save), and NuGet restore is solution-scoped.
  Engine asset-graph definitions move out of the WPF editor into `Stride.Assets` so headless tools get them without referencing the editor.
  The verb shape is a stepping stone toward a future unified `stride` CLI (`new`, `upgrade`, `build`, …).

- **Sample & template versioning.** In-repo samples are committed at the clean release version and flipped to/from the local dev build on demand (real csproj edits); the version bump runs the real package upgraders over the whole solution in one process. Sample and starter template packages are content-versioned independently of the engine.
  Samples migrated to 4.4.0.

- **Dev-versioning refinements.** Iterates on the existing per-checkout `-devN` ledger: the first checkout now also gets a suffix (`-dev`) so no local build ever carries a clean release version and dev + release versions are unified behind one generator driven by explicit `MajorMinor`/`MinPatch` inputs.

- **Docs.** New build versioning & release guide.


Ethereal77:
Would it be possible to somehow specify a custom local version suffix? I mean, if I have several parallel worktrees for specific features, instead of them being `-dev`, `-dev2`, etc. (which is a nice default in case you do not care), to have a way to specify, for example, `-custom`, `-bug33`, or whatever, so generated packages are easily distinguishable.

sources/assets/Stride.AssetCompiler/Stride.AssetCompiler.csproj
```diff
@@ -23,6 +23,7 @@
     <PackageReference Include="Microsoft.NET.StringTools" ExcludeAssets="runtime" PrivateAssets="all" />
     <PackageReference Include="Microsoft.Build.Framework" ExcludeAssets="runtime" PrivateAssets="all" />
     <PackageReference Include="Microsoft.Build.Utilities.Core" ExcludeAssets="runtime" PrivateAssets="all" />
+    <ProjectReference Include="..\Stride.Core.Assets.Quantum\Stride.Core.Assets.Quantum.csproj" />
     <ProjectReference Include="..\..\engine\Stride.Assets.Models\Stride.Assets.Models.csproj" />
```
Kryptos-FR, line 27: Why do we need to reference explicitly Stride runtime libraries? How would this work for third-party engine plugins.

I think originally the name "Core" in the app was because it was supposed to be asset-agnostic (only deal with Core asset types and Core quantum).

Obviously that is not the case anymore, but maybe we need a second PR to be able to register (pae configuration or command line argument) additional assembly references.

xen2:
> Would it be possible to somehow specify a custom local version suffix? I mean, if I have several parallel worktrees for specific features, instead of them being `-dev`, `-dev2`, etc. (which is a nice default in case you do not care), to have a way to specify, for example, `-custom`, `-bug33`, or whatever, so generated packages are easily distinguishable.

Sure!
Edit `%LocalAppData%\Stride\worktree-ids.txt`, it should pick it up next build.

Issue #3238, opened 2026-06-26
[VISIX] - Stride extensions does not work if solution is in `.slnx` format


antokhio:
**Release Type**: Official Release

**Version**: 4.3.0

**Platform(s)**: Windows

**Describe the bug**
VISIX Extension stops to work if solution converted to `slnx` from `sln`.

**To Reproduce**
1.  Create project check that VISIX works
2. Modernize -> "Convert to slnx" 
3. Notice extension disappears.

**Expected behavior**
VISIX works whatever solution format.

**Screenshots**
`slnx`
<img width="768" height="290" alt="Image" src="https://github.com/user-attachments/assets/5dd36a17-58d7-424d-87bd-99b801387244" />

`sln`
<img width="701" height="200" alt="Image" src="https://github.com/user-attachments/assets/dc6664c1-5b8a-4851-b609-ad315fc3b876" />

**Log and callstacks**
```
The command "dotnet "C:\...\.nuget\packages\stride.core.assets.compilerapp\4.2.1.2487\buildTransitive\..\lib\net8.0-windows7.0\Stride.Core.Assets.CompilerApp.dll"  --disable-auto-compile --project-configuration "Debug" --platform=Windows --project-configuration=Debug --compile-property:StrideGraphicsApi=Direct3D11 --output-path="D:\...\Tester\Tester\Bin\Windows\Debug\win-x64\data" --build-path="D:\...\Tester\Tester\Tester.Windows\obj\stride\assetbuild\data" --package-file="D:\...\Tester\Tester\Tester.Windows\Tester.Windows.csproj" --msbuild-uptodatecheck-filebase="D:\...\Tester\Tester\Tester.Windows\obj\Debug\win-x64\stride\assetcompiler-uptodatecheck"" exited with code 1.
```

antokhio:
Should be fixed by:

https://github.com/stride3d/stride/commit/d8fd605b5de295860974a3e24165a7b8dc48fb76
https://github.com/stride3d/stride/commit/9d27d76305a60460655e9245ef1200d653b83642

antokhio:
Verified and `visix` works in `slnx` after those 2 commits

PR #3239, opened 2026-06-26
Added .slnx parsing for visix extension

antokhio:
# PR Details

I had to compile stride shaders as separate package rescently and noticed that visix extensions does not work. Upon digging on source of evil I found that it only supports `.sln` solutions.

Key edits:
* Adds SolutionXReader based on https://github.com/microsoft/vs-solutionpersistence/blob/main/src/Microsoft.VisualStudio.SolutionPersistence/Serializer/Xml/Slnx.xsd
* Adds check if we are trying to write `.slnx` will throw. (adds todo the way that slnx's are written needs to be somehow considered)
* Checks if we are reading `.slnx` parsing it to solution correctly

## Related Issue

https://github.com/stride3d/stride/issues/3238

## Types of changes

## Checklist

I'm not sure how to run test's here, tested that compiled extension works on test `slnx` and `sln` project:

image

### DIFF
--- a/sources/core/Stride.Core.Design/Solutions/Solution.cs
+++ b/sources/core/Stride.Core.Design/Solutions/Solution.cs
@@ -11,10 +11,10 @@
 // documentation files (the "Software"), to deal in the Software without restriction, including without limitation
 // the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and
 // to permit persons to whom the Software is furnished to do so, subject to the following conditions:
-// 
+//
 // The above copyright notice and this permission notice shall be included in all copies or substantial portions
 // of the Software.
-// 
+//
 // THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED
 // TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
 // THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF
@@ -46,9 +46,14 @@ public class Solution
     }
 
     private Solution(Solution original)
-        : this(original.FullPath, original.Headers, original.Projects, original.GlobalSections, original.Properties)
-    {
-    }
+        : this(
+            original.FullPath,
+            original.Headers,
+            original.Projects,
+            original.GlobalSections,
+            original.Properties
+        )
+    { }
 
     /// <summary>
     /// Initializes a new instance of the <see cref="Solution" /> class.
@@ -58,7 +63,13 @@ public class Solution
     /// <param name="projects">The projects.</param>
     /// <param name="globalSections">The global sections.</param>
     /// <param name="properties">The properties.</param>
-    public Solution(string fullpath, IEnumerable<string> headers, IEnumerable<Project> projects, IEnumerable<Section> globalSections, IEnumerable<PropertyItem> properties)
+    public Solution(
+        string fullpath,
+        IEnumerable<string> headers,
+        IEnumerable<Project> projects,
+        IEnumerable<Section> globalSections,
+        IEnumerable<PropertyItem> properties
+    )
     {
         FullPath = fullpath;
         this.Headers = new List<string>(headers);
@@ -79,10 +90,7 @@ public class Solution
     /// <value>The children.</value>
     public IEnumerable<Project> Children
     {
-        get
-        {
-            return Projects.Where(project => project.GetParentProject(this) == null);
-        }
+        get { return Projects.Where(project => project.GetParentProject(this) == null); }
     }
 
     /// <summary>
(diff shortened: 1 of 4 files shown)


antokhio:
@dotnet-policy-service agree

sources/core/Stride.Core.Design/Solutions/Solution.cs
```diff
@@ -46,9 +46,14 @@ public Solution()
     }
 
     private Solution(Solution original)
-        : this(original.FullPath, original.Headers, original.Projects, original.GlobalSections, original.Properties)
-    {
-    }
+        : this(
```
Kryptos-FR: Avoid cosmetic change like this one that aren't related to the issue at hand. It doesn't follow the general style of this codebase.

sources/core/Stride.Core.Design/Solutions/Solution.cs
```diff
@@ -11,10 +11,10 @@
 // documentation files (the "Software"), to deal in the Software without restriction, including without limitation
 // the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and
 // to permit persons to whom the Software is furnished to do so, subject to the following conditions:
-// 
+//
```
Kryptos-FR: Unnecessary change.

sources/core/Stride.Core.Design/Solutions/Solution.cs
```diff
@@ -11,10 +11,10 @@
...
 // The above copyright notice and this permission notice shall be included in all copies or substantial portions
 // of the Software.
-// 
+//
```
Kryptos-FR: Unnecessary change.

sources/core/Stride.Core.Design/Solutions/Solution.cs
```diff
@@ -58,7 +63,13 @@ private Solution(Solution original)
     /// <param name="projects">The projects.</param>
     /// <param name="globalSections">The global sections.</param>
     /// <param name="properties">The properties.</param>
-    public Solution(string fullpath, IEnumerable<string> headers, IEnumerable<Project> projects, IEnumerable<Section> globalSections, IEnumerable<PropertyItem> properties)
+    public Solution(
+        string fullpath,
```
Kryptos-FR: Avoid cosmetic change like this one that aren't related to the issue at hand. It doesn't follow the general style of this codebase.

sources/core/Stride.Core.Design/Solutions/Solution.cs
```diff
@@ -132,6 +140,16 @@ public void Save()
     /// <param name="solutionPath">The solution path.</param>
     public void SaveAs(string solutionPath)
     {
+        // TODO:
+        // Writing only supports the legacy text based .sln format. Routing a .slnx through the text
+        // writer would corrupt the file, so fail explicitly instead.
```
Kryptos-FR, line 137: It's problematic. We do call `Save()` from `PackageSession.Save()`, which itself is called from GameStudio's `SessionViewModel.SaveSession()`. It would be better to also support saving the slnx format.

Kryptos-FR (changes requested):
Thanks for the contribution.

A few remarks, code style should follow as much as possible the existing codebase, though it isn't always consistent.

Saving the format should also be supported or it will cause issue in the GameStudio.

sources/core/Stride.Core.Design/Solutions/Solution.cs
antokhio: I'm using https://github.com/belav/csharpier for code formatting so it just formats it all...
Gonna try to redo without it I guess, but I would suggest some standardized approach...

sources/core/Stride.Core.Design/Solutions/Solution.cs
antokhio: This is a question we need to decide somehow, I can implement that blindfold-ly,  so it would alternate to `slnx` saving, but:

* How to test?

Only test I got is by migrating to `slnx` ready made solution, meaning we need to open it in Game Studio, does game studio would support opening `.slnx`? 

So it feels like I can add this but, I'm not sure about side effects, how to test and would that be usable at all?

antokhio: Anyways the scope of the PR a `visix` fix, not the `slnx` support for stride in general, so i suggest we create issue from that, and i will fix it in next pr, since otherewise we will block the fix.... @Kryptos-FR ...

antokhio:
I'm also thinking might we consider adding this https://github.com/microsoft/vs-solutionpersistence

Ethereal77:
Iirc Microsoft points to [VS-SolutionPersistence](https://github.com/microsoft/vs-solutionpersistence) as the recommended way whenever there is need to read/write solution files.
Why keep the current outdated and hardcoded way of processing solution files, mixing it with parts from that library **instead** of using that library directly (and strip out our own SLN-reading logic)?

xen2:
Sorry, I worked on exactly the same thing the last few days. Check those commits:
d8fd605b
9d27d763
It's already in master. Is that enough for your use case?

antokhio:
@xen2 yes, seems it should fix the VISIX Package [issue I had](https://github.com/stride3d/stride/issues/3238) so looks like it safe to close issue with those commits.

antokhio:
Just verified visix works in slnx

PR #3240, opened 2026-06-27
stride CLI tool

xen2:
# PR Details

A cross-platform `dotnet tool` for managing Stride versions and creating/building projects from the command line.
No editor required!

```bash
dotnet tool install -g Stride.Cli
```

## Commands

```bash
# Engine versions
stride install                 # install the latest Stride engine
stride install 4.4             # install the newest 4.4.x
stride list                    # list installed versions
stride update                  # roll installed lines to their newest patch
stride uninstall 4.4.0

# Projects
stride new                     # list the templates the installed version offers
stride new fps -n MyShooter    # create from a template: game, fps, platformer2d, topdownrpg, vrsandbox, ...
stride new game -n MyGame -o ./out   # -n name, -o output dir; extra --key value pairs forward to the template
stride new fps --help          # show that template's parameters
cd MyShooter && dotnet run --project MyShooter.Windows

# Tools (resolved for the project in the current directory)
stride studio                  # open Game Studio
stride asset build MyGame.sln  # run the Asset Compiler (all args forwarded)

# The CLI itself
stride version                 # CLI version + resolved Stride version
stride self update             # update the tool via `dotnet tool update`
```

The `stride-` template prefix is optional (`stride new game` == `stride new stride-game`), and `--version` overrides the Stride version resolved from the current directory.

## Also in this PR

- **Release pipeline** — `release-cli.yml` builds, code-signs (Azure KeyVault, reusing the engine signing), and publishes `Stride.Cli` to NuGet.org, tagging `cli/<version>` + a GitHub Release. Versioned independently via SemVer in `Stride.Cli.csproj`; prereleases via the `version-suffix` dispatch input (`1.0.0-beta1`).
- **Docs** — CLI section in the root README, `sources/launcher/README.md`, and a "Stride CLI" section in `docs/build/versioning.md`.


Kryptos-FR:
The `dotnet` cli uses `new list` to give the list of available template, instead of just `new`. Calling just `dotnet new` without any other arguments prints some help and instructions. I kind of like it and think we should follow it closely so that it is more familiar to users.

On a side note, it has additional interesting commands such as `dotnet new search web` which queries nuget for all available compatible templates including third-party's.

.github/workflows/release-cli.yml
```diff
@@ -0,0 +1,214 @@
...
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          lfs: true
+          fetch-depth: 0  # full history + tags for the re-publish guard
+          fetch-tags: true
+
+      - uses: actions/setup-dotnet@v4
+        with:
+          # 6.0.x required by deps/Gettext.Net pulled in transitively (matches release.yml).
```
Kryptos-FR: We have released a .NET 10 version of Gettext.Net (see https://www.nuget.org/packages/Stride.GNU.Gettext/3.0.0) we shouldn't need .NET 6.0 SDK anymore.

build/Stride.Launcher.sln
```diff
@@ -31,90 +31,250 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Stride.Core.Translation.Pre
...
 EndProject
 Global
-	GlobalSection(SharedMSBuildProjectFiles) = preSolution
-		..\sources\assets\Stride.Core.Assets.Yaml\Stride.Core.Assets.Yaml.projitems*{0f8be30e-c41f-4747-b52b-d2d4e13ec6a2}*SharedItemsImports = 5
-		..\sources\editor\Stride.Core.MostRecentlyUsedFiles\Stride.Core.MostRecentlyUsedFiles.projitems*{0f8be30e-c41f-4747-b52b-d2d4e13ec6a2}*SharedItemsImports = 5
-		..\sources\editor\Stride.PrivacyPolicy\Stride.PrivacyPolicy.projitems*{0f8be30e-c41f-4747-b52b-d2d4e13ec6a2}*SharedItemsImports = 5
-	EndGlobalSection
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
+		Debug|x64 = Debug|x64
```
Kryptos-FR: Why add those targets? Code should be portable with `Any CPU`.

sources/launcher/README.md
```diff
@@ -1,9 +1,33 @@
...
+Stride Launcher & CLI
+=====================
 
-Source for Stride launcher and installer.
+User-facing entry points to a Stride install:
 
-# Build instructions
+- **Stride.Cli** — the cross-platform `stride` command-line tool (a `dotnet tool`): install Stride versions and create, build, and manage projects.
+- **Stride.Launcher** — the WPF launcher/installer application (Windows).
+- **Stride.Launcher.Core** — shared logic used by both (version discovery/management, NuGet store access).
```
Kryptos-FR: I don't see any changes in the Launcher.csproj that would reference the core library.

sources/assets/Stride.Core.Packages/NugetStore.cs
```diff
@@ -279,6 +279,59 @@ public IEnumerable<NugetLocalPackage> GetLocalPackages(string packageId)
...
+        {
+            if (installPath == null)
+                continue;
+            try
+            {
+                foreach (var info in new FindLocalPackagesResourceV3(installPath).GetPackages(NativeLogger, CancellationToken.None))
+                    TryAdd(info);
+            }
+            catch
+            {
```
Kryptos-FR, line 328: Avoid catch all, better catch the right exception types.

sources/assets/Stride.Core.Packages/NugetStore.cs
```diff
@@ -279,6 +279,59 @@ public IEnumerable<NugetLocalPackage> GetLocalPackages(string packageId)
...
+        {
+            if (!source.IsLocal)
+                continue;
+            try
+            {
+                foreach (var info in LocalFolderUtility.GetPackagesV2(source.Source, NativeLogger))
+                    TryAdd(info);
+            }
+            catch
+            {
```
Kryptos-FR, line 345: Avoid catch all, better catch the right exception types.

sources/core/Stride.Core.Design/Solutions/KnownProjectTypeGuid.cs
```diff
@@ -28,12 +28,14 @@ namespace Stride.Core.Solutions;
 public static class KnownProjectTypeGuid
 {
     public static readonly Guid VisualBasic = new("F184B08F-C81C-45F6-A57F-5ABD9991F28F");
-    public static readonly Guid CSharp = new ("FAE04EC0-301F-11D3-BF4B-00C04F79EFBC");
+    // SDK-style C# project: the guid used for projects created today.
+    public static readonly Guid CSharp = new("9A19103F-16F7-4668-BE54-9A1E7A4F7556");
```
Kryptos-FR, line 34: That's actually not entirely true. In latest versions of MSBuild and Visual Studio, all projects use `FAE04EC0-301F-11D3-BF4B-00C04F79EFBC` whether they are legacy or sdk style. The build engine determines which kind it is based on the fact that it has a "sdk" declaration at the root.

In our cases, since we use a customized sdk without the sdk declaration, then we do have to use `9A19103F-16F7-4668-BE54-9A1E7A4F7556` which is a way to force the build engine to recognize it as SDK-style.

You can easily test this by creating a new solution using the CLI (with `--format sln`), then create a project and add it to the solution with `dotnet sln add <projectPath>`. It will use `FAE04EC0-301F-11D3-BF4B-00C04F79EFBC`.

xen2: Still generating new GUID but adjusted comment

sources/launcher/README.md
xen2: Good catch, fixed it!
(I thought I would share it at first; now folded into Stride.Cli, I will split a lib later if necessary)

build/Stride.Launcher.sln
xen2: Thanks visual studio, fixed!

xen2:
> The `dotnet` cli uses `new list` to give the list of available template, instead of just `new`. Calling just `dotnet new` without any other arguments prints some help and instructions. I kind of like it and think we should follow it closely so that it is more familiar to users.
> 
> On a side note, it has additional interesting commands such as `dotnet new search web` which queries nuget for all available compatible templates including third-party's.

I swapped it, `stride new` now do what `stride new --help` used to do, and I've added `stride new list` (I did it after this PR, it's on master)

xen2:
> The `dotnet` cli uses `new list` to give the list of available template, instead of just `new`. Calling just `dotnet new` without any other arguments prints some help and instructions. I kind of like it and think we should follow it closely so that it is more familiar to users.
> 
> On a side note, it has additional interesting commands such as `dotnet new search web` which queries nuget for all available compatible templates including third-party's.

BTW, except for that `list` (now fixed), I already took special care to stay close to `dotnet` as much as I could.
i.e., `stride new` doesn't try to find a unused folder by suffixing a number like GameStudio does but stick closer to `dotnet new`: error if default folder name exists already, and with a `--name` and `--force` flag. In practice it's actually much better for predictability when using it in scripts.

PR #3242, opened 2026-06-28
fix: Platform detection logic for Linux

Jklawreszuk:
# PR Details

`StridePlatform` property used by i.e. AssetCompiler, is not set correctly on Linux. Platform detection works only if the user set RuntimeIdentifier to linux-*.

User case: 
1. As an user, I create new console project.
2. I add Stride libraries, including AssetCompiler
3. I add some assets thus involving asset compiling process
4. I build an app using `dotnet build` - asset compiler executes
5. Asset compiler sets --plaform to `windows`

Parameters in this case: 
- TargetFramework=net10.0, 
- RuntimeIdentifier="" **(empty)**, 

thus according to this logic:
```xml
    <StridePlatform Condition="'$(StridePlatform)' == '' And $(RuntimeIdentifier.StartsWith('linux'))">Linux</StridePlatform>
    <StridePlatform Condition="'$(StridePlatform)' == ''">Windows</StridePlatform>
```

it sets to "default" option - Windows

**Edit**

My solution requires testing, and it might be a good idea to simplify the platform detection logic as a whole. 

## Types of changes

## Checklist


xen2:
👋 About this PR.

PR #3195, opened 2026-05-28
feat: Rework CharacterController based on Bepu's

Eideren:
# PR Details
The implementation for bepu's character controller was a bit wonky;
- The gravity property was mutated internally to prevent the body from sliding down slopes
- Moving surfaces would not carry the body along with them
- Moving past a slope would cause the body to fly off instead of continuing horizontally

## Related Issue
None logged

## Types of changes

- [x] Docs change / refactoring / dependency upgrade
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist
- [ ] My change requires a change to the documentation.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.
- [x] **I have built and run the editor to try this change out.**

N.B.: `CharacterControllers.cs`, `CharacterMotionConstraint.cs` and `CharacterMotionConstraint.tt` are effectively copies from bepu's demo. Mostly left as is to make porting new features and fixes introduced over there easier.

Issue #3249, opened 2026-06-30
Game Studio crashes when opening/upgrading the project using 4.4.0-beta1


dawnmichal:
**Release Type**: Official Release

**Version**: 4.4.0-beta1

**Platform(s)**: Windows 11 Pro

**Describe the bug**
Game Studio crashes when opening/upgrading the project from version 4.3.0.2507 to 4.4.0-beta1.

**Log and callstacks**
Crash report attached.

[Report.txt](https://github.com/user-attachments/files/29516392/Report.txt)

dawnmichal:
The crash occurs only when opening the project for the first time and attempting to upgrade it. When I then start Game Studio again and try to open the project once more, Studio no longer crashes and the project finishes loading, but errors are listed in the log. I’m therefore attaching the log from the second opening.

[Opening log.txt](https://github.com/user-attachments/files/29516672/Opening.log.txt)

dawnmichal:
The errors reported when opening the project were caused by missing namespaces and changes to the names of some properties (for example, the IndexBufferBinding class has count renamed to indexCount). Opening the project now works, and it can also be built.

I have run into a few other minor bugs, so I will continue testing. However, I still don’t have an explanation for the crash during the first upgrade attempt, so that probably remains relevant.

At the same time, I’m not sure whether I should consider the upgrade completed, or whether the project is not fully and correctly upgraded even though it can now be opened, since the upgrade did not complete properly on the first attempt.

xen2:
First of all, GameStudio should not crash even if upgrade unsuccessful, so we've got to fix at least that one.

Second, we've added a new code upgrader system to make this kind of changes (IndexBufferBinding class has count renamed to indexCount) automatic.

Do you happen to have an easy repro project before it's ugpraded?

dawnmichal:
Ok, here we go...
I managed to trim my project down to a reasonable minimum, removing as many irrelevant parts, models, etc. as possible. What remains is a relatively simple base for demonstrating my issues with migrating the project from Stride version 4.3.0.2507 to version 4.4.0-beta1.

1. After extracting the archive, the project should open in Game Studio version 4.3.0.2507 without any issues. It should also build and run without problems. In the main menu, only the buttons for starting and quitting the game are functional. Everything should work correctly, including controlling the player character with the mouse, the WASD keys, and the spacebar for jumping.

[...]

PR #3253, opened 2026-07-02
fix: CharacterComponent handedness

Eideren:
# PR Details
Fixes the handedness of the character component, should fix most discrepancies from the previous breaking (#3195) change when running the upcoming stride release out of the box.

This also changes the basis used for the non-deprecated path with lateral movement to avoid the changes in basis between character and transforms.

## Related Issue
None reported here, one surfaced through a [discord discussion](https://discord.com/channels/500285081265635328/1173235382356299897/1521591273939800136)

## Types of changes

## Checklist


dawnmichal:
As part of another issue (https://github.com/stride3d/stride/issues/3249#issuecomment-4860038512), I prepared a test project where this problem also occurred, so it can be used to test the fix if needed.

xen2 (approved):

PR #3255, opened 2026-07-04
Modernize the Visual Studio extension (out-of-process) + SDSL syntax highlighting

xen2:
Replaces the legacy net472 in-process VSPackage with a .NET 8 out-of-process `VisualStudio.Extensibility` extension, versioned independently from the engine,
and adds `.sdsl`/`.sdfx` syntax coloring.

## Extension (`Stride.VisualStudio.Package`, net8, out-of-process)

- New out-of-process extension: Stride version resolver + **Open in Game Studio**.
- **Update Stride Version** and **Regenerate Stride Shaders** commands.
- Loads only for Stride solutions.
- Bundles the net10 `stride` CLI under `tools\stride\` so commands shell out to it.
- Removes the old in-process package and its dead platform-switching (startup-project/config sync) and launchSettings devenv workaround.

## Independent versioning

- The package now versions itself (starting at **4.4.0**), decoupled from the engine's `SharedAssemblyInfo` — same model as the Stride CLI.
- `release-vspackage.yml` gains an optional prerelease `version-suffix`.
- Fixes the uninstall to target the current package id (was the stale `.2022`).

## SDSL/SDFX syntax highlighting (`Stride.VisualStudio.Package.Shaders`)

- Companion classic (in-process) content-only VSIX that registers a TextMate grammar for instant `.sdsl`/`.sdfx` coloring — the out-of-process extension can't register grammars itself, so it ships alongside as a second VSIX inside the same NuGet package.
- The grammar + companion `.tmTheme` are generated by a `dotnet run` tool that reflects Stride's own reserved keyword/type vocabulary, so coloring can't drift from the language.

## CLI

- `stride sdk available` + auto-install target in `upgrade`.
- `stride generate-legacy-shader-code` for 4.0–4.3 projects.

## Build / CI

- VSIX packaging builds and packs both vsixes via `dotnet build ... -t:VSIXPlugin`.
- Sign the bundled `packageinstall.exe`/`dll` (VS package **and** GameStudio).
- Renamed `build-vs-package.yml` → `build-vspackage.yml` to match `release-vspackage.yml`.


mihai-ene-public:
In addition to highlighting it would be nice to have in the future some form of intellisense

xen2:
> In addition to highlighting it would be nice to have in the future some form of intellisense

Agree. I have already investigated a bit and what we want to do is a LSP (language server) so that it works for both VS and VSCode.
This is a huge undertaking so it will come later.

Issue #630, opened 2020-03-19
[Vulkan] WriteDiscard support


lee19811:
When I try to use config "win-vulkan" in project file compile Starbreach successful，but run with error: System.InvalidOperationException: Can't use WriteDiscard on Graphics API that doesn't support renaming
   Xenko.Graphics.CommandList.MapSubresource(GraphicsResource resource, Int32 subResourceIndex, MapMode mapMode, Boolean doNotWait, Int32 offsetInBytes, Int32 lengthInBytes)
   Xenko.Graphics.FastTextRenderer.End(GraphicsContext graphicsContext)
   Xenko.Profiling.GameProfilingSystem.Draw(GameTime gameTime)
   Xenko.Games.GameSystemCollection.Draw(GameTime gameTime)
   Xenko.Games.GameBase.Draw(GameTime gameTime)
   Starbreach.StarbreachGame.Draw(GameTime gameTime)

Jklawreszuk:
I've managed to reproduce it too. 

runtime : Fedora Linux

1. Create new project.
2. Edit any script and try to call DebugText.Print() inside Update() method
3. Exception : Can't use WriteDiscard on Graphics API that doesn't support renaming

xen2:
This was fixed in a8eb1556a
We might still have issues with dynamic textures but the dynamic buffer case is solved.

edit: I will keep open for buffers, but `FastTextRenderer` should be OK already.

PR #3261, opened 2026-07-06
fix: GameProfilingSystem crashes on Vulkan (FastTextRenderer null command buffer)

### LINKED ISSUE #1020: Game Profiler does not work with Vulkan
**Release Type**: Official 4.0

**Version**: 4.0.1.1400

**Platform(s)**: Windows

**Describe the bug**
Game Profiler does not work with Vulkan. I remove the Api reference in .csproj and it works no problem. Looks like it's a command list thing, see error below.

**To Reproduce**
1. Attempt to attach game profiler to game running with Vulkan.

**Expected behavior**
Expect GameProfiler to work with all renderer API.

[Game]: Error: Unexpected exception. System.InvalidOperationException: Can't use WriteDiscard on Graphics API that doesn't support renaming
   at Stride.Graphics.CommandList.MapSubresource(GraphicsResource resource, Int32 subResourceIndex, MapMode mapMode, Boolean doNotWait, Int32 offsetInBytes, Int32 lengthInBytes) in C:\BuildAgent\work\b5f46e3c4829a09e\sources\engine\Stride.Graphics\Vulkan\CommandList.Vulkan.cs:line 1280
   at Stride.Graphics.FastTextRenderer.End(GraphicsContext graphicsContext) in C:\BuildAgent\work\b5f46e3c4829a09e\sources\engine\Stride.Graphics\FastTextRenderer.cs:line 196
   at Stride.Profiling.GameProfilingSystem.Draw(GameTime gameTime) in C:\BuildAgent\work\b5f46e3c4829a09e\sources\engine\Stride.Engine\Profiling\GameProfilingSystem.cs:line 414
   at Stride.Games.GameSystemCollection.Draw(GameTime gameTime) in C:\BuildAgent\work\b5f46e3c4829a09e\sources\engine\Stride.Games\GameSystemCollection.cs:line 125
   at Stride.Games.GameBase.Draw(GameTime gameTime) in C:\BuildAgent\work\b5f46e3c4829a09e\sources\engine\Stride.Games\GameBase.cs:line 792
   at Stride.Games.GameBase.RawTick(TimeSpan elapsedTimePerUpdate, Int32 updateCount, Single drawInterpolationFactor, Boolean drawFrame) in C:\BuildAgent\work\b5f46e3c4829a09e\sources\engine\Stride.Games\GameBase.cs:line 639
   at Stride.Games.GameBase.RawTickProducer() in C:\BuildAgent\work\b5f46e3c4829a09e\sources\engine\Stride.Games\GameBase.cs:line 590

sasvdw:
# PR Details

Fixes a native crash (`0xC0000005`) on the **Vulkan** backend: enabling the on-screen game profiler (`GameProfilingSystem`, toggled with **F7**) crashes within a couple of frames inside `FastTextRenderer.End`.

```
Fatal error. 0xC0000005
   at Vortice.Vulkan.Vulkan.vkCmdBindVertexBuffers(...)   // on master; 4.3.x faults one call earlier at vkCmdBindPipeline
   at Stride.Graphics.CommandList.SetVertexBuffer(...)
   at Stride.Graphics.FastTextRenderer.End(...)
   at Stride.Profiling.GameProfilingSystem.Draw(...)
```

### Root cause

`GameProfilingSystem.Draw` rendered `FastTextRenderer` through `RenderContext.GetShared(Services).GetThreadContext()`. On Vulkan that returns a **thread-local pipeline-worker** command list, which is `Close()`d during the compositor's parallel scene rendering (`RenderSystem.Draw`) and **never re-begun** afterwards — so its native `VkCommandBuffer` is null, and recording draw commands into it in `FastTextRenderer.End` dereferences a null handle → access violation.

It only worked on Direct3D11 because D3D11 records serially (`IsDeferred == false`) and its thread contexts alias the single main command list via `GraphicsDevice.InternalMainCommandList` — which Vulkan intentionally does **not** set, because it does genuine parallel command-buffer recording. So the profiler was the one `GameSystem.Draw`-level caller reaching into a pipeline-internal context that isn't valid outside the render pipeline.

### Fix

Render through **`Game.GraphicsContext`** — the main, per-frame command list that `GameBase.BeginDraw` resets and begins every frame — exactly as **`DebugTextSystem`** (the other `FastTextRenderer` consumer) already does. Behaviour-preserving on Direct3D11 (that was already the main command list); fixes the crash on Vulkan. Single method changed; no `FastTextRenderer` change needed.

### Validation

Built the engine at a single consistent version and A/B-tested a minimal real game: **without** the change it crashes 3/3, **with** it runs cleanly and the profiler overlay renders. Minimal self-contained reproduction (clone + `dotnet run` on a Vulkan GPU): **https://github.com/sasvdw/stride-1020-vulkan-profiler-repro**

### Note / follow-up

This is the targeted caller fix. The deeper issue — `GetThreadContext()` silently handing a `GameSystem.Draw` an invalid (closed) command list on Vulkan, so *any* future out-of-pipeline drawing system is a latent crash — is a framework concern (`DebugTextSystem` even carries a `// TODO GRAPHICS REFACTOR where to get command list from?`). I've kept this PR minimal and low-risk and will raise that separately as a proposal.

## Related Issue

Closes #1020. Related: #630 (`[Vulkan] WriteDiscard support` — same `FastTextRenderer`-on-Vulkan area).

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Engine/Profiling/GameProfilingSystem.cs
+++ b/sources/engine/Stride.Engine/Profiling/GameProfilingSystem.cs
@@ -346,50 +346,51 @@ namespace Stride.Profiling
                 renderTargetSize = new Size2(renderTarget.Width, renderTarget.Height);
             }
 
-            var renderContext = RenderContext.GetShared(Services);
-            var renderDrawContext = renderContext.GetThreadContext();
-
             if (fastTextRenderer == null)
             {
-                fastTextRenderer = new FastTextRenderer(renderDrawContext.GraphicsContext)
+                fastTextRenderer = new FastTextRenderer(Game.GraphicsContext)
                 {
                     DebugSpriteFont = Content.Load<Texture>("StrideDebugSpriteFont"),
                     TextColor = TextColor,
                 };
             }
 
-            using (renderDrawContext.PushRenderTargetsAndRestore())
+            // Render through the game's main GraphicsContext (as DebugTextSystem does). The per-thread
+            // RenderContext command list used previously is Close()d after the compositor renders and is never
+            // re-begun, so on Vulkan its native command buffer is null and FastTextRenderer.End() crashes when
+            // recording draw commands into it (0xC0000005 — issue #1020). The main command list is reset and
+            // begun every frame in GameBase.BeginDraw, so it is always valid here. On Direct3D11 this is
+            // equivalent (the thread context already shared the main command list).
+            var graphicsContext = Game.GraphicsContext;
+            graphicsContext.CommandList.SetRenderTargetAndViewport(null, renderTarget);
+            viewportHeight = graphicsContext.CommandList.Viewport.Height;
+            fastTextRenderer.Begin(graphicsContext);
+            lock (stringLock)
             {
-                renderDrawContext.CommandList.SetRenderTargetAndViewport(null, renderTarget);
-                viewportHeight = renderDrawContext.CommandList.Viewport.Height;
-                fastTextRenderer.Begin(renderDrawContext.GraphicsContext);
-                lock (stringLock)
+                var currentHeight = textDrawStartOffset.Y;
+                fastTextRenderer.DrawString(graphicsContext, fpsStatString, textDrawStartOffset.X, currentHeight);
+                currentHeight += TopRowHeight;
+
+                if (FilteringMode == GameProfilingResults.CpuEvents)
                 {
-                    var currentHeight = textDrawStartOffset.Y;
-                    fastTextRenderer.DrawString(renderDrawContext.GraphicsContext, fpsStatString, textDrawStartOffset.X, currentHeight);
+                    fastTextRenderer.DrawString(graphicsContext, gcMemoryString, textDrawStartOffset.X, currentHeight);
                     currentHeight += TopRowHeight;
-
-                    if (FilteringMode == GameProfilingResults.CpuEvents)
-                    {
-                        fastTextRenderer.DrawString(renderDrawContext.GraphicsContext, gcMemoryString, textDrawStartOffset.X, currentHeight);
-                        currentHeight += TopRowHeight;
-                        fastTextRenderer.DrawString(renderDrawContext.GraphicsContext, gcCollectionsString, textDrawStartOffset.X, currentHeight);
-                        currentHeight += TopRowHeight;
-                    }
-                    else if (FilteringMode == GameProfilingResults.GpuEvents)
-                    {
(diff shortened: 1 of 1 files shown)


sources/engine/Stride.Engine/Profiling/GameProfilingSystem.cs
```diff
@@ -346,50 +346,51 @@ public override void Draw(GameTime gameTime)
...
                 };
             }
 
-            using (renderDrawContext.PushRenderTargetsAndRestore())
+            // Render through the game's main GraphicsContext (as DebugTextSystem does). The per-thread
+            // RenderContext command list used previously is Close()d after the compositor renders and is never
+            // re-begun, so on Vulkan its native command buffer is null and FastTextRenderer.End() crashes when
+            // recording draw commands into it (0xC0000005 — issue #1020). The main command list is reset and
+            // begun every frame in GameBase.BeginDraw, so it is always valid here. On Direct3D11 this is
+            // equivalent (the thread context already shared the main command list).
```
xen2: Let's remove that whole comment.
It's overly specific, and no need to mention/explain about code/issue that is now gone.
New approach is proper graphics API usage so nothing specific to comment for just assigning a variable.

sasvdw: Will do!

sasvdw: Alrighty. This should now be fixed.

sources/engine/Stride.Engine/Profiling/GameProfilingSystem.cs
```diff
@@ -346,50 +346,45 @@ public override void Draw(GameTime gameTime)
...
-                fastTextRenderer = new FastTextRenderer(renderDrawContext.GraphicsContext)
+                fastTextRenderer = new FastTextRenderer(Game.GraphicsContext)
                 {
                     DebugSpriteFont = Content.Load<Texture>("StrideDebugSpriteFont"),
                     TextColor = TextColor,
                 };
             }
 
-            using (renderDrawContext.PushRenderTargetsAndRestore())
+            var graphicsContext = Game.GraphicsContext;
```
Ethereal77: `fastTextRenderer ??= new ...` (the modern null-conditional assignment).

Also, you can move the `var graphicsContext = ...` to the top and reuse it instead of `new fastTextRenderer(Game.GraphicsContext)`.

sasvdw: These changes are sensible yes. Thanks @Ethereal77

PR #3272, opened 2026-07-17
fix: Game settings config overrides not working properly

ferafiks:
# PR Details

Currently there is an issue with using [config overrides](https://doc.stride3d.net/latest/en/manual/game-studio/game-settings.html#overrides), where the `Specific Filter` field doesn't actually do anything. This is because `PlatformConfiguration` has two static fields that are never set which are used to determine the user's graphics card and cpu model.

This PR fixes this and changes some of the logic that is used when choosing the correct configuration:
* Specific filter can now be used to filter the gpu, but not the cpu because I'm not sure if Stride even stores that property anywhere.
* Configuration overrides are now prioritized from last to first, so that the behavior is more intuitive.
* Overrides with specific filters remain prioritized over those without one.

Sidenote, this feature is extremely useless. Aside from setting a different graphics level, there is almost never a need to use a different config based on the user's platform or gpu. I could alternatively close this PR and open a different one to remove the feature altogether.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Engine/Engine/Game.cs
+++ b/sources/engine/Stride.Engine/Engine/Game.cs
@@ -263,7 +263,7 @@ namespace Stride.Engine
             {
                 PlatformConfigurations.RendererName = GraphicsAdapterFactory.DefaultAdapter?.Description ?? string.Empty;
 
-                databaseFileProvider = InitializeAssetDatabase();
+                databaseFileProvider = InitializeAssetDatabase(AssetBundleName);
                 ((DatabaseFileProviderService)Services.GetService<IDatabaseFileProviderService>()).FileProvider = databaseFileProvider;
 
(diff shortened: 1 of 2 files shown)


xen2:
> Sidenote, this feature is extremely useless. Aside from setting a different graphics level, there is almost never a need to use a different config based on the user's platform or gpu. I could alternatively close this PR and open a different one to remove the feature altogether.

I think this is what we should do.
Changing graphics level isn't going to work well if assets are not adjusted anyway.

One idea for a more generic approach: add a filter/configuration to the new "asset replace" feature (not trivial if we want proper selection support at runtime, but at least it's a decent starting point UX-wise: user overrides an asset and specialize/adjust it on given conditions).

ferafiks:
In that case, I'm gonna close this one and hopefully have the other ready in a few weeks

PR #3274, opened 2026-07-17
fix: Update required components to build the engine

Eideren:
# PR Details
MSVC Build Tools for ARM64/ARM64EC is required as of latest changes, tested out on a clean VS. Build probes and fails to find MSVC.dll in VS ARM64 directory.

## Related Issue
None logged

## Types of changes

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ Our [Roadmap](https://doc.stride3d.net/latest/en/contributors/roadmap.html) comm
    - **In the 'Individual Components' Tab**
      - **MSVC Build Tools for ARM64/ARM64EC (Latest)** *(currently v14.5x in VS 2026)*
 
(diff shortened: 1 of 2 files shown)


xen2:
Sorry, I didn't get to it because I wanted to double check if our system to skip ARM64 when not existing was broken.

But after a quick look, it's still there:
https://github.com/stride3d/stride/blob/460ac7e487ae99d78c35a0f8fc565d3cd2256562/sources/native/Stride.Native.targets#L43-L63

Do you happen to remember how it was failing in your case?

Eideren:
@xen2 Not really, but from what I wrote above the build fails to find that `MSVC.dll` in the arm directory. You should be able to reproduce it by uninstalling VS, installing the required components, and doing a clean build of the engine.

xen2:
The ARM64 link failure was a bug in toolset check, not a missing prerequisite: it accepted ARM64 libs from any installed VS toolset while the link uses only one.
Fixed on master (ARM64 optional as it should have).

PR #3280, opened 2026-07-19
GameStudio: edit System.Guid properties in the property grid

Nicogo1705:
### PR Details

`Guid` members on components/assets previously fell through to the read-only
`ObjectPropertyTemplateProvider` fallback and rendered as a non-editable "Guid" row in the
property grid — there was no way to set them from Game Studio at all.

This PR adds a proper Guid editor, plus a few usability layers on top:

**Editing** — a `GuidPropertyTemplateProvider` (`TypeMatchTemplateProvider`, so it also covers
`Nullable<Guid>`) mirroring the string editor: a `TextBox` bound through a new `GuidToString`
value converter. No registration code needed: providers declared in
`DefaultPropertyTemplateProviders.xaml` are auto-registered, and the default `OverrideRule.Some`
takes precedence over the read-only fallback.

**Commit-time validation** — invalid text is passed through by `ConvertBack`, so the
strongly-typed node setter throws, `TextBoxBase` raises `TextToSourceValueConversionFailed`, and
the attached `TextBoxPropertyValueValidationBehavior` (same pattern as the UFile editor) plays
the red error adorner while the text reverts to the source value. The stored value can never be
corrupted.

**Ghost template** — while the editor has focus, the typed prefix is completed inline with the
remaining `0000…-0000` pattern in faint text (via a new `GuidTemplateRemainder` converter), so
the missing length and dash positions are visible at a glance.

**Input mask** (`GuidInputMaskBehavior`) —
- hex input is regrouped live as 8-4-4-4-12, dashes inserted automatically (the caret hops past
  a freshly inserted dash);
- deletions are not fought: backspacing a dash does not re-append it;
- input is capped at the full 36-char template (`MaxLength`), replacing a selection still works;
- pasted `{guid}` / `(guid)` forms are unwrapped;
- non-hex content turns the text red immediately (purely visual, nothing is blocked or
  rewritten — commit-time validation does the hard reject).

**Before:** a `[DataMember] public Guid MyId` property shows a non-editable row displaying "Guid".
**After:** the value is editable as text with format guidance, auto-formatting and live feedback.

### Related Issue

No linked issue — quality-of-life editor feature; `Guid` properties on user components were
previously impossible to set from Game Studio.

### Types of changes

### Checklist

### DIFF
--- a/sources/editor/Stride.Core.Assets.Editor/View/Behaviors/GuidInputMaskBehavior.cs
+++ b/sources/editor/Stride.Core.Assets.Editor/View/Behaviors/GuidInputMaskBehavior.cs
@@ -50,7 +50,9 @@ namespace Stride.Core.Assets.Editor.View.Behaviors
             // Also tolerate the brace/paren Guid forms on paste ("{...}", "(...)"): the wrapper
             // characters are dropped so the content still fits the mask.
             var raw = text.Replace("-", "").Trim('{', '}', '(', ')', ' ');
-            if (raw.Length > HexDigits || !IsHex(raw))
+            var valid = raw.Length <= HexDigits && IsHex(raw);
+            SetInvalidHint(!valid);
+            if (!valid)
                 return;
 
             var formatted = Format(raw, appendTrailingDash: grew);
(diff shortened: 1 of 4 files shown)


Nicogo1705:
image

PR #3285, opened 2026-07-23
fix: Member whose pre-existing value's type does not match the serialized value's type

Eideren:
# PR Details
```cs
public interface ITest;
[DataContract] public class A : ITest { public string S; }
[DataContract] public class B : ITest { public float F; }
public class MyScript : StartupScript
{
    // Changing this value to 'B' in the editor, then saving and re-opening the project throws;
    // The generic serializer gets 'Test' value ('new A()') and passes it onto 'BSerializer' 
    // for the serialized data to be deserialized into it.
    // BSerializer, expecting a null or a value of type 'B', casts this object which is an 'A' to 'B' -> throwing
    // If Test is null, it would simply instantiate an 'B'
    public ITest Test = new A();
}
```

## Related Issue
None.

## Types of changes

## Checklist

### DIFF
--- a/sources/core/Stride.Core/Serialization/MemberSerializerCore.ttinclude
+++ b/sources/core/Stride.Core/Serialization/MemberSerializerCore.ttinclude
@@ -28,6 +28,9 @@
                 // If we still have no serializer, throw an exception
                 if (dataSerializer is null)
                     throw new ArgumentException("No serializer available for type " + <#= typeExpr #>.FullName);
+
+                if (obj is not null && obj.GetType().IsAssignableTo(dataSerializer.SerializationType) == false)
+                    obj = default; // pre-existing obj is not compatible with the serialized type; they are two distinct types deriving from the same T
             }
 <# } // supportsNullDataSerializer #>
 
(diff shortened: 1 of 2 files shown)


xen2:
I have looked into it and I think this should be a better fix:
- where cast error happens
- no perf cost (instead of hot path)
- handle some weird case a bit better (i.e. PreSerialize override without calling base)

https://github.com/stride3d/stride/compare/master...xen2:stride:fix/member-deserialize-type-mismatch

Eideren:
@xen2 This was the first solution I came up with, I didn't end up implementing it that way as it makes the assumption that no other classes would override that method, or that if they do, they would be aware of and handle this unexpected input.
You could seal the overridden method to alleviate this concern.
It's not making the serializer any easier to maintain either way.

xen2:
I see your point.
The fix in MemberSerializer also makes lot of other assumptions.
But your idea to seal is good, let me see if it's possible.

xen2:
It seems to work well with `seal`. Also I can move the fix only in `PreSerialize` (I also had it in  `Serialize` before, for the reason you mentioned: if `PreSerialize` override doesn't call `base.PreSerialize`, it wouldn't work, as was the case in `VertexDeclaration.Serializer`).
I will run tests on it to make sure nothing is broken, and it should be fine.

xen2:
Superseded by #3305

PR #3293, opened 2026-07-25
fix: SDSL multidimensional array dimensions emitted in reverse order

### LINKED ISSUE #3292: SDSL: multidimensional array dimensions emitted in reverse order (breaks compute LambertianPrefilteringSH)
**Release Type**: GitHub / Beta channel

**Version**: 4.4.0-beta4 (also current `master` — code unchanged)

**Platform(s)**: All — the reversal is in the SDSL→SPIR‑V compiler (platform‑independent); the downstream `fxc` failure surfaces on Direct3D.

**Describe the bug**

The SDSL shader compiler emits **multidimensional array dimensions in reverse order**. `float4 Data[2][3][5]` is emitted (in the generated SPIR‑V, and the HLSL SPIRV‑Cross produces from it) as `[5][3][2]` — reversed, not rotated. Rank‑1 arrays are unaffected, so it's specific to rank ≥ 2.

This is the production 4.4 pipeline: `EffectCompiler` → `ShaderMixer.MergeSDSL` → `SpirvTranslator.Translate(Hlsl)` → fxc/dxc (used by both the runtime `EffectCompilerFactory` and asset‑build `EffectCompileCommand`).

**To Reproduce**

Minimal standalone repro (no Stride checkout, no GPU — just `dotnet run`): https://github.com/sasvdw/stride-sdsl-array-repro

It compiles three shaders via `MergeSDSL` + `SpirvTranslator` and compares emitted vs declared dimensions:

```
=== LambertianPrefilteringSHPass1<4,3> ===   (Stride's own shipped shader)
    declared: PartialSHCoeffs[4][4][9]
    emitted:  PartialSHCoeffs[9][4][4]   MISMATCH
=== ArrayDimsRank3 ===
    declared: Data[2][3][5]
    emitted:  Data[5][3][2]              MISMATCH
=== ArrayDimsRank1 (control) ===
    declared: Data[7]   emitted: Data[7]   OK
```

**Expected behavior**

Dimensions preserved in declaration order: `Data[2][3][5]` → `[2][3][5]`.

**Real‑world impact**

The only stock `.sdsl` with a multidimensional array is `LambertianPrefilteringSHPass1` (the **compute** SH/IBL prefilter): `groupshared float4 PartialSHCoeffs[TBlockSize][TBlockSize][CoefficientsCount]` (`[4][4][9]`). Emitted as `[9][4][4]` but indexed as `[4][4][9]`, so the third index runs 0..8 against a dimension of 4 and fxc rejects the generated HLSL:

```
error X3504: literal loop terminated early due to out of bounds array access
error X3696: infinite loop detected - loop never exits
```

The compute `LambertianPrefilteringSH` therefore can't compile. It has gone unnoticed because both stock consumers (`SkyboxGenerator`, `LightProbeGenerator`) use the pixel‑shader `LambertianPrefilteringSHNoCompute` variant (no multidimensional array), and the compute path in `TestLambertPrefilteringSH` is gated behind a manual keypress the automated test never sends.

**Root cause**

`sources/shaders/Stride.Shaders.Parsers/Parsing/SDSL/AST/Literals.cs` `GenerateArrayType` iterates the declared dimensions left‑to‑right and wraps each around the accumulated type, so the **last** bracket becomes the **outermost** array. `arraySizes` is in declaration order and `RegisterArrayType` faithfully emits the nesting it's handed — the reversal originates solely in that loop. Fix: wrap from the last dimension inward.

**Additional context**

Fix + regression test in the linked PR (closes this issue). `Stride.Shaders.Tests` isn't currently run by any CI solution filter, which is why the compiler had no automated guard for this — worth wiring in (it has some pre‑existing, unrelated failures to triage first).

sasvdw:
# PR Details

The SDSL compiler emitted **multidimensional array dimensions in reverse order** — `float4 Data[2][3][5]` came out as `[5][3][2]` (reversed, not rotated; rank‑1 unaffected). This is the production 4.4 path (`EffectCompiler` → `MergeSDSL` → `SpirvTranslator` → fxc/dxc).

### Root cause

`GenerateArrayType` (`sources/shaders/Stride.Shaders.Parsers/Parsing/SDSL/AST/Literals.cs`) iterated the declared dimensions left‑to‑right and wrapped each around the accumulated type, so the last bracket became the outermost array. `arraySizes` is in declaration order and `RegisterArrayType` faithfully emits the nesting it's handed, so the reversal originated solely in that loop.

### Fix

Wrap from the last dimension inward, so the first‑declared dimension is outermost. One‑line change; rank‑1 behaviour unchanged. This unbreaks the stock compute shader `LambertianPrefilteringSHPass1` (`PartialSHCoeffs[4][4][9]`, previously emitted `[9][4][4]` → indexed out of bounds → fxc `X3504/X3696`).

### Test

`MultidimensionalArrayDimensionsKeepDeclarationOrder` (`Stride.Shaders.Tests`) compiles a small compute shader with `Data3D[2][3][5]` (+ a rank‑1 control) via `MergeSDSL` + `SpirvTranslator` and asserts the emitted HLSL keeps `[2][3][5]`. Compiler‑only (no GPU), so it's CI‑safe. Red before the fix (`Actual "[5][3][2]"`), green after; the rest of the suite is unchanged.

The shader asset lives in a new `assets/SDSL/CompilerTests/` folder rather than `ComputeTests/`, deliberately: `ComputeTest1` runs every `ComputeTests` shader through a D3D11 dispatch, which wouldn't be CI‑safe.

### Note

`Stride.Shaders.Tests` isn't currently referenced by any CI `.slnf`, so these shader unit tests are built but never run in CI — worth wiring in (the suite has a handful of pre‑existing, unrelated `ErrorUnsupportedSpirv` failures to triage first).

### Validation

Reproduced and verified with a standalone repro (`dotnet run`, no GPU): https://github.com/sasvdw/stride-sdsl-array-repro — all three cases correct after the fix, including the real `LambertianPrefilteringSHPass1`.

## Related Issue

Closes #3292.

## Types of changes

## Checklist

### DIFF
--- a/sources/shaders/Stride.Shaders.Parsers/Parsing/SDSL/AST/Literals.cs
+++ b/sources/shaders/Stride.Shaders.Parsers/Parsing/SDSL/AST/Literals.cs
@@ -787,8 +787,11 @@ public partial class TypeName(string name, TextLocation info) : Literal(info)
 
     public static SymbolType GenerateArrayType(SymbolTable table, SpirvContext context, SymbolType arraySymbolType, List<Expression> arraySizes)
     {
-        foreach (var arraySize in arraySizes)
+        // Wrap from the last declared dimension inward so the first dimension ends up outermost:
+        // for `T[a][b][c]` (arraySizes = [a, b, c]) this builds Array(a, Array(b, Array(c, T))).
+        for (var index = arraySizes.Count - 1; index >= 0; index--)
         {
+            var arraySize = arraySizes[index];
             if (!table.ResolveArraySizes || arraySize is EmptyExpression)
                 arraySymbolType = new ArrayType(arraySymbolType, -1);
             else
--- a/sources/shaders/Stride.Shaders.Tests/Stride.Shaders.Tests.csproj
+++ b/sources/shaders/Stride.Shaders.Tests/Stride.Shaders.Tests.csproj
@@ -41,9 +41,11 @@
 
   <ItemGroup>
     <Using Include="Xunit" />
+    <StrideShaderFiles Include="..\assets\SDSL\CompilerTests\**" LinkBase="assets\SDSL\CompilerTests\" />
     <StrideShaderFiles Include="..\assets\SDSL\RenderTests\**" LinkBase="assets\SDSL\RenderTests\" />
     <StrideShaderFiles Include="..\assets\SDSL\ComputeTests\**" LinkBase="assets\SDSL\ComputeTests\" />
     <StrideShaderFiles Include="..\assets\SDSL\StreamOutTests\**" LinkBase="assets\SDSL\StreamOutTests\" />
+    <Content Include="..\assets\SDSL\CompilerTests\**" CopyToOutputDirectory="PreserveNewest" LinkBase="assets\SDSL\CompilerTests\" />
     <Content Include="..\assets\SDSL\RenderTests\**" CopyToOutputDirectory="PreserveNewest" LinkBase="assets\SDSL\RenderTests\" />
     <Content Include="..\assets\SDSL\ComputeTests\**" CopyToOutputDirectory="PreserveNewest" LinkBase="assets\SDSL\ComputeTests\" />
     <Content Include="..\assets\SDSL\StreamOutTests\**" CopyToOutputDirectory="PreserveNewest" LinkBase="assets\SDSL\StreamOutTests\" />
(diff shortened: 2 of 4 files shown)


xen2:
LGTM

I have found a related bug while reviewing: cbuffer reflection only keeps the outermost array dimension in `Elements`, so `float4 Data[2][3]` reports 2 elements instead of 6.
Could you cherry-pick it into this PR?
(I can't seem to be able to push it myself to your fork even though "Maintainers are allowed to edit this pull request." is green; it might be due because your fork is owned by an org? but not sure)

```
    git fetch https://github.com/xen2/stride.git fix/sdsl-multidim-array-reflection
    git cherry-pick 2dbe810834
```

xen2:
Never mind about previous message, I simply merged your branch and added my changes on top directly.
Thanks again

Issue #3288, opened 2026-07-24
4.4.0.0-beta4 Window minimize/restore zero size or crash


Basewq:
**Release Type**: Beta Release

**Version**: 4.4.0.0-beta4

**Platform(s)**: Windows

**Describe the bug**
Minimizing then restoring the game window results in a zero sized window (technically the window menu options are still visible) or crash due to ImageMultiScaler post processing.

**To Reproduce**
1. Create new game template (all default settings)
2. Run the game
3. Minimize window -> Restore window

**Expected behavior**
Restore window back to original size.

**Log and callstacks**
Depending on lighting/post processing setup, the game will either be zero sized window (default settings the window cannot be resized back to the original size), or crash, due to the zero sized window, with the following:

```
[Game]: Error: Unexpected exception. System.InvalidOperationException: Input and output texture cannot have same size [(1,1,1)]
   at Stride.Rendering.Images.ImageMultiScaler.PrepareScaling(Texture inputTexture, Int32& scalingDirection, PixelFormat& outputPixelFormat) in /_/sources/engine/Stride.Rendering/Rendering/Images/ImageScaler/ImageMultiScaler.cs:line 139
   at Stride.Rendering.Images.ImageMultiScaler.DrawCore(RenderDrawContext context) in /_/sources/engine/Stride.Rendering/Rendering/Images/ImageScaler/ImageMultiScaler.cs:line 55
   at Stride.Rendering.RendererBase.Draw(RenderDrawContext context) in /_/sources/engine/Stride.Rendering/Rendering/RendererBase.cs:line 51
   at Stride.Rendering.Images.LuminanceEffect.DrawCore(RenderDrawContext context) in /_/sources/engine/Stride.Rendering/Rendering/Images/LuminanceEffect/LuminanceEffect.cs:line 155
   at Stride.Rendering.RendererBase.Draw(RenderDrawContext context) in /_/sources/engine/Stride.Rendering/Rendering/RendererBase.cs:line 51
   at Stride.Rendering.Images.PostProcessingEffects.DrawCore(RenderDrawContext context) in /_/sources/engine/Stride.Rendering/Rendering/Images/PostProcessingEffects.cs:line 449
   at Stride.Rendering.RendererBase.Draw(RenderDrawContext context) in /_/sources/engine/Stride.Rendering/Rendering/RendererBase.cs:line 51
[...]

PR #3297, opened 2026-07-27
Fix minimized window raw client size handling

xen2:
# PR Details

Revived PR #3230 from @Redwarx008 
Kudos to his work on this issue!

Please check that PR details for the actual description.
I have adjusted a few things (extra commits) and added a few unit tests as requested in [this comment](https://github.com/stride3d/stride/pull/3230#issuecomment-4827449484).

As an alternative approach, I was hesitating to go further and mark the old `ClientBounds` obsolete (nobody needs it as a `Rectangle`, that's XNA leftover when it was bounds on screen, we always pretend it's at corner 0,0).
Instead, we could simply add a proper `public Int2 ClientSize` (name TBD, or maybe even use nullable when undefined) without the Max(1) and then have call site dealing with rendering & expecting size 1 do the max themselves (or even skip rendering).

I am happy to do it if people agree.
What do you think?

## Related Issue

#3230
#3288 

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Engine.NoAssets.Tests/GameWindowMinimizeTest.cs
+++ b/sources/engine/Stride.Engine.NoAssets.Tests/GameWindowMinimizeTest.cs
@@ -1,0 +1,270 @@
+// Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net)
+// Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using System;
+using System.Diagnostics;
+using System.Runtime.ExceptionServices;
+using System.Runtime.InteropServices;
+using System.Threading;
+using System.Threading.Tasks;
+
+using Xunit;
+
+using Stride.Core;
+using Stride.Core.Mathematics;
+using Stride.Games;
+using Stride.Graphics;
+using Stride.Graphics.Regression;
+
+namespace Stride.Engine.Tests;
+
+/// <summary>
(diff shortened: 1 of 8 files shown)


Ethereal77:
I have always (even on XNA) seen `ClientBounds` as a convenience property to know in one go both the size and the position of the window. In fullscreen apps, or in platforms where you own the screen, the position will always be (0,0), but on desktop platforms it can change.
If we can get/set the position and we can get/set the size via other means, maybe `ClientBounds` can stay as just a getter-only property?

Or are you talking about something else?

xen2:
In Stride every implementation returns new Rectangle(0, 0, w, h).
X and Y are hardcoded 0, and the actual window position is a separate `Position` property.
`ClientBounds` is used as a replacement for Width/Height so far.

It's a leftover from XNA API that we use this specific `Rectangle` property rather than `Int2`.

We could add `Position` into it but:
- doing the Max(1) on the size part is quite weird as it doesn't correspond to the real `ClientBounds` (which are undefined or 0).
- weird coupling, might need an extra heavy SDL/WinAPI call to check position on top of size.
- usually we care only about size, rarely about position (so better separate)

Ethereal77:
I mean, we continue using just size internally for our computations and preparing the presenters, etc.
Just that we continue to expose the bounds as a rectangle with (position, size), a getter-only property for convenience. That is what I mean.

xen2:
Closed in favor of #3300

PR #3298, opened 2026-07-27
fix(StackPanel): Correct gap spacing when children are collapsed

bai-chi-xing:
# PR Details

- Introduce GetSafeChildAndGapSize method to replace GetSafeChildSize for proper gap handling
- Modify measure logic to correctly compute gaps when children are collapsed
- Update arrange logic to properly allocate gaps in the presence of collapsed elements
- Fix gap handling in scrolling positioning calculations
- Add test cases to verify support for collapsed child elements

My English is not very good, so this description was translated with AI — I hope it conveys my intention clearly.

This PR fixes a bug I discovered: when a `StackPanel` with a non-zero `Gap` is placed inside a `ScrollViewer`, dragging the scrollbar becomes janky and difficult to control.

While investigating, I noticed that gap handling was problematic in several places throughout `StackPanel`. When collapsed (non-occupying) children were present, gaps were still being included in both measure and arrange calculations — even the final `DesiredSize` was incorrect.

To fix this, I consolidated gap sizes into the preceding sibling element of each visible child (regardless of whether that sibling is collapsed or not). This ensures smooth scrolling behavior. However, this changed the meaning of `ScrollPosition`, which previously did not account for gaps at all. I've updated the XML documentation comments to clarify the new behavior and warn about potential pitfalls.

I've tested the changes locally and confirmed that the scrolling issue is resolved. Since these changes only affect gap handling, I extended the existing tests rather than adding new test files.

This is my first time submitting a PR, so please let me know if I've missed any guidelines or conventions. I'm happy to revise.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.UI.Tests/Layering/StackPanelTests.cs
+++ b/sources/engine/Stride.UI.Tests/Layering/StackPanelTests.cs
@@ -83,6 +83,15 @@ namespace Stride.UI.Tests.Layering
             var expectedSize = new Vector3(60, 85, 0); // width is max, height is sum + gaps
             Assert.Equal(expectedSize, stackPanel.DesiredSizeWithMargins);
 
+            // Test measure with collapsed child
+            child3.Visibility = Visibility.Collapsed;
+            stackPanel.Measure(measureSize);
+            expectedSize = new Vector3(50, 55, 0);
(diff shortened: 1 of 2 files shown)


bai-chi-xing:
@dotnet-policy-service agree

PR #3302, opened 2026-07-28
fix: map GraphicsProfile.Level_11_2 to a valid Direct3D feature level

### LINKED ISSUE #3301: GraphicsProfile.Level_11_2 is unusable on Direct3D (invalid feature-level mapping)

**Version**: `master` (reproduced on `03edac194`); the mapping has been wrong since the initial
open-source commit (2018), so this affects all versions.

**Platform(s)**: Windows — Direct3D11 **and** Direct3D12 (Vulkan is unaffected).

**Describe the bug**

Setting **Game Settings → Rendering → Default Graphics Profile = Level 11.2** crashes the game at
startup on the Direct3D backends. It is not hardware-dependent: it reproduces on any GPU, and even on
the WARP software rasterizer.

`GraphicsProfile.Level_11_2 = 0xB200` has no corresponding native Direct3D feature level — real D3D
feature levels jump from `11_1 (0xB100)` straight to `12_0 (0xC000)` (Direct3D 11.2 was an API
revision that runs on FL 11_1 hardware, not a new feature level). `GraphicsProfileHelper.ToFeatureLevel`
raw-casts the profile to `(D3DFeatureLevel) 0xB200`, a value the API never returns, so device creation
can never satisfy it.

**To Reproduce**

1. Create a game project on the **Direct3D11** backend.
2. Open the `GameSettings` asset → **Rendering** → set **Default Graphics Profile = Level 11.2**.
3. Run the game → it crashes at startup with the exception below.
4. (Optional) Switch the backend to **Direct3D12** → also crashes (at device creation). **Vulkan** →
   runs.

Minimal repro repository (code-only game + WARP software-renderer proof):
https://github.com/sasvdw/stride-graphicsprofile-level112-repro

**Expected behavior**

`Level_11_2` should behave like every other profile — the game should launch (running at its real
capability tier, FL 11_1, the same as `Level_11_1` and what its own tooltip claims: "~ Direct3D 11.2
/ Vulkan 1.1"), consistent with Vulkan. Only profiles a device genuinely cannot provide (e.g.
`Level_11_x` on FL10-only hardware) should be rejected.

**Screenshots**

N/A — the crash occurs during device creation, before anything renders.

**Log and callstacks**

```
System.InvalidOperationException: None of the graphics profiles [Level_11_2] are supported by the Graphics Device. The highest available profile is [Level_11_1].
   at Stride.Games.GraphicsDeviceManager.FindBestDevice(Boolean anySuitableDevice)
   at Stride.Games.GraphicsDeviceManager.<ChangeOrCreateDevice>g__ChangeOrCreateDevice|128_0(<>c__DisplayClass128_0&)
   at Stride.Games.GraphicsDeviceManager.ChangeOrCreateDevice(Boolean forceCreate)
   at Stride.Games.GraphicsDeviceManager.Stride.Games.IGraphicsDeviceManager.CreateDevice()
   at Stride.Games.GameBase.InitializeBeforeRun()
   at Stride.Games.GamePlatform.OnInitCallback()
   at Stride.Games.GameWindowWinforms.Run()
   at Stride.Games.GamePlatform.Run(GameContext gameContext)
   at Stride.Games.GameBase.Run(GameContext gameContext)
   at Program.<Main>$(String[] args)
```

(That is the Direct3D11 message. On Direct3D12 the availability check passes and it instead fails
later inside `CreateDevice`.)

**Additional context**

Per-backend behaviour with Level 11.2:

| Backend | Behaviour |
|---|---|
| **Direct3D11** | `GraphicsAdapter.IsProfileSupported` requests `CreateDevice` at `0xB200` and requires an exact match; it never matches, so the profile is reported unsupported and `GraphicsDeviceManager` throws (above). |
| **Direct3D12** | `IsProfileSupported` returns `true` unconditionally, so the availability check passes — then `CreateDevice(0xB200)` fails (the existing guard only clamps the *lower* bound `< 11_0`). |
| **Vulkan** | Ignores the profile entirely and runs. |

That Vulkan runs while both Direct3D backends fail is the tell: the fault is the Direct3D
profile→feature-level mapping, not the enum value.

*Proof it is not a hardware/driver capability query.* A standalone program mirroring Stride's exact
`D3D11CreateDevice` call against the **WARP software rasterizer** (source in the repro) produces:

```
WARP max feature level : 0xC100   [0x00000000]                    <- WARP supports FL 12_1
Request 0xB200 (11_2)  : FAILED   [0x80070057]  matched=0x0000    <- E_INVALIDARG
Request 0xB100 (11_1)  : SUCCEEDED   [0x00000000]  matched=0xB100
```

WARP — a *software* rasterizer reporting FL 12_1 — rejects `0xB200` with `E_INVALIDARG` while
accepting `0xB100`, confirming the failure is the invalid requested value.

*History.* Present since the initial open-source commit (2018). The June-2025 "Documentation pass for
GraphicsDevice" (`beb66ce9b`) added the `[Display("Level 11.2 ~ like Direct3D 11.2 / Vulkan 1.1")]`
label, which makes an unusable option look legitimate in the settings dropdown.

*Suggested fix.* Map `Level_11_2` to its real capability tier (FL 11_1) in the Direct3D
profile→feature-level mapping, so all backends accept it and run at 11_1 while genuine capability gaps
still fail. A fix PR follows.

sasvdw:
# PR Details

**Summary** — `GraphicsProfile.Level_11_2` does not work on the Direct3D backends. This PR maps it to
the feature level it represents (FL 11_1), so it behaves like every other profile.

**This PR unlocks no Direct3D 11.2 capabilities.** There is no feature level 11_2. Direct3D 11.2 was a
runtime revision for Windows 8.1, and the level enum goes from `11_1 (0xB100)` to `12_0 (0xC000)`. Its
headline feature, tiled resources, is an optional capability. The application queries it after device
creation with `CheckFeatureSupport(D3D11_FEATURE_D3D11_OPTIONS1)` → `TiledResourcesTier`, and Tier 1
needs FL 11_0 hardware plus a WDDM 1.3 driver. This PR only stops an enum value that we already ship
from failing.

**Description** — `GraphicsProfile.Level_11_2 = 0xB200` has no native Direct3D feature level.
`GraphicsProfileHelper.ToFeatureLevel` cast the profile to the non-existent `0xB200`. Direct3D11 then
rejected **Level 11.2** in `GraphicsAdapter.IsProfileSupported`. Direct3D12 passed its unconditional
`IsProfileSupported` and then failed at `CreateDevice(0xB200)`. Vulkan ignores the profile, so Vulkan
ran.

`Level_11_2` reaches those paths from disk, not only from code. `GraphicsProfile` is a `[DataContract]`
and `Level_11_2` has a `[Display]` name. It is therefore a selectable value for
`RenderingSettings.DefaultGraphicsProfile` in Game Settings, and `Game.cs` feeds that value into both
`deviceManager.ShaderProfile` and `PreferredGraphicsProfile`.

Five Direct3D files change. The `GraphicsProfile` enum stays as it is, so this PR breaks nothing and
changes no serialized data.

- **`GraphicsProfileHelper.cs`** — `ToFeatureLevel` maps `Level_11_2` to `11_1` with an explicit
  table, instead of a cast that depends on the two enums sharing values. `ToFeatureLevels` maps a
  sequence one element at a time. It replaces the array overload, whose span reinterpretation skipped
  the mapping.
- **`GraphicsAdapter.Direct3D.cs`** (Direct3D11 `IsProfileSupported`) resolves the feature level with
  `ToFeatureLevel()` instead of a cast, so the exact-match check tests the real FL 11_1.
- **`GraphicsOutput.Direct3D11.cs`** and **`GraphicsOutput.Direct3D12.cs`**
  (`FindClosestMatchingDisplayMode`) map the caller's profile span instead of reinterpreting it. See
  the next section.
- **`ShaderCompiler.cs`** (`ShaderProfileFromGraphicsProfile`) lists `Level_11_2` with `11_0` and
  `11_1` in the shader-model `5_0` case, so it no longer throws.

## The display-mode path was a second crash

`FindClosestMatchingDisplayMode` exists once per backend. Both copies reinterpreted the whole profile
span as feature levels. `GamePlatform` passes the game's own `PreferredGraphicsProfile` to that method
when `IsFullScreen` is set. With 11.2 selected, the fullscreen path carried `0xB200` into device
creation.

- *Direct3D12* tried each level in turn and failed every attempt. It then threw from
  `ThrowNoCompatibleProfile`. Fullscreen therefore still rejected the profile, even with the mapping
  fixed elsewhere.
- *Direct3D11* discarded the `CreateDevice` HRESULT and then released the device and the context
  unconditionally. On failure both pointers are still `null`, so it faulted instead of reporting an
  unsupported profile. **This fault is not specific to 11.2.** Any unsupported profile in fullscreen
  reached the same null release, for example `Level_11_1` on FL10 hardware. Direct3D11 now checks the
  HRESULT before it uses the device, which is what Direct3D12 already did.

Direct3D11 and Direct3D12 now accept `Level_11_2` and run at FL 11_1, windowed and fullscreen, the
same as Vulkan. **The code still rejects what it must reject.** A profile that a device cannot provide
fails the capability check as before. Only the invalid-value failure goes away.

## Consequences to know about

- **The round trip loses information, by design.** `FromFeatureLevel` does not change, so a game that
  requests 11.2 reports `Features.CurrentProfile == Level_11_1`. `RequestedProfile` keeps the request.
  The `case Level_11_2:` arm in `GraphicsDeviceFeatures.Direct3D11.cs` is therefore unreachable. It is
  harmless, because it shares a row with `11_1` and `11_0`.
- **`IsProfileSupported(Level_11_2)` now succeeds on FL 11_1 hardware.**
  `IsPreferredProfileAvailable` walks every enum value, so the message *"the highest available profile
  is [...]"* can now name `Level_11_2`. This follows from treating 11.2 as the 11_1 tier. It is not a
  regression.
- **Shader macros still carry `0xB200`.** `EffectCompiler` emits `STRIDE_GRAPHICS_PROFILE` from the
  requested profile, plus one `GRAPHICS_PROFILE_LEVEL_*` macro per enum value. The ordered comparisons
  in `.sdsl` stay correct, because `0xB200` is above every other level. `Stride.Shaders.Tests` already
  pins those values. Shader-model resolution is the part that needed the fix.
- **The Null backend is the concrete reason the `ShaderCompiler` arm is necessary.**
  `Null/GraphicsDeviceFeatures.Null.cs` sets `RequestedProfile = CurrentProfile = Level_11_2`, and
  `EffectSystem` feeds `Features.RequestedProfile` into the compiler.
- **The test project takes its own `Silk.NET.Direct3D11` reference.** `Stride.Graphics` keeps Silk.NET
  at `PrivateAssets="compile"`, so `D3DFeatureLevel` does not flow to consumers. The reference is
  conditioned to the desktop frameworks, the same way `Stride.Graphics` conditions its own.

**Motivation and context** — see #3301. The failure does not depend on hardware. A WARP
software-renderer proof in the repro shows WARP reporting FL 12_1, rejecting `0xB200` with
`E_INVALIDARG`, and accepting `0xB100`. The mapping has been wrong since the 2018 open-source commit.
A June-2025 documentation pass added the `[Display("Level 11.2 ~ …")]` tooltip, which makes the broken
option look legitimate in Game Settings.

## Related Issue

Fixes #3301. Repro (minimal code-only game and WARP proof):
https://github.com/sasvdw/stride-graphicsprofile-level112-repro

## Types of changes

## Checklist

**Validation status** (kept honest — check the boxes above as each one is satisfied):

- *Docs*: no documentation change is necessary. The profile works now, and the existing tooltip
  becomes accurate.
- *Tests*: `sources/engine/Stride.Graphics.Tests/TestGraphicsProfileHelper.cs` covers both mappings.
  `Level_11_2` resolves to FL 11_1 and not to `0xB200`. Every profile maps to a defined
  `D3DFeatureLevel`. A sequence maps one element at a time, so no entry can leak `0xB200`. Verified
  red and green: the sequence tests fail against a cast and pass with the mapping.
- *Build*: `Stride.Graphics` builds clean under `StrideGraphicsApi=Direct3D11` and under
  `=Direct3D12`. The two `GraphicsOutput` files sit behind mutually exclusive `#if` guards, so each
  backend needs its own build for full coverage. `Stride.Shaders.Compilers` builds clean. CI runs the
  full test suite.
- *Editor*: **outstanding**. Build and run the editor with a project set to Level 11.2, windowed and
  fullscreen, to confirm that it starts. This is the mandatory personal-testing step.

### DIFF
--- a/sources/engine/Stride.Graphics.Tests/Stride.Graphics.Tests.csproj
+++ b/sources/engine/Stride.Graphics.Tests/Stride.Graphics.Tests.csproj
@@ -15,6 +15,9 @@
     <Compile Remove="Compiler\CubemapEffect.sdfx.cs" />
   </ItemGroup>
   <ItemGroup>
+    <PackageReference Include="Silk.NET.Direct3D11" />
+  </ItemGroup>
+  <ItemGroup>
     <ProjectReference Include="..\..\tests\xunit.runner.stride\xunit.runner.stride.csproj" />
     <ProjectReference Include="..\Stride.Engine\Stride.Engine.csproj" />
     <ProjectReference Include="..\Stride.Graphics.Regression\Stride.Graphics.Regression.csproj" />
--- a/sources/engine/Stride.Graphics.Tests/Stride.Graphics.Tests.csproj
+++ b/sources/engine/Stride.Graphics.Tests/Stride.Graphics.Tests.csproj
@@ -15,7 +15,10 @@
     <Compile Remove="Compiler\CubemapEffect.sdfx.cs" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Silk.NET.Direct3D11" />
+    <!-- Stride.Graphics references Silk.NET with PrivateAssets="compile", so D3DFeatureLevel does not
+         flow to consumers; TestGraphicsProfileHelper needs its own reference to name it. Desktop-only,
+         matching the Direct3D references in Stride.Graphics.csproj. -->
+    <PackageReference Include="Silk.NET.Direct3D11" Condition="'$(StrideTargetFramework)' == '$(StrideFramework)' Or '$(StrideTargetFramework)' == '$(StrideFrameworkWindows)'" />
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\..\tests\xunit.runner.stride\xunit.runner.stride.csproj" />
--- a/sources/engine/Stride.Graphics.Tests/TestGraphicsProfileHelper.cs
+++ b/sources/engine/Stride.Graphics.Tests/TestGraphicsProfileHelper.cs
@@ -1,0 +1,42 @@
+// Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net)
+// Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
+
+#if STRIDE_GRAPHICS_API_DIRECT3D
+
+using System;
+
+using Xunit;
+
+using Silk.NET.Core.Native;
+
+namespace Stride.Graphics.Tests;
+
+/// <summary>
+/// Regression tests for the GraphicsProfile -> D3DFeatureLevel mapping in GraphicsProfileHelper.
+/// Direct3D-only: the helper and D3DFeatureLevel exist only under STRIDE_GRAPHICS_API_DIRECT3D.
+/// </summary>
+public class TestGraphicsProfileHelper
+{
+    [Fact]
+    public void Level_11_2_MapsTo_Level_11_1()
+    {
+        // 0xB200 has no native D3D feature level; it must resolve to the real 11_1 (0xB100),
+        // matching Vulkan, instead of the raw-cast 0xB200 that fails CreateDevice on D3D11/D3D12.
+        Assert.Equal(GraphicsProfile.Level_11_1.ToFeatureLevel(), GraphicsProfile.Level_11_2.ToFeatureLevel());
+        Assert.NotEqual((D3DFeatureLevel)0xB200, GraphicsProfile.Level_11_2.ToFeatureLevel());
+    }
+
+    [Fact]
+    public void EveryProfile_MapsToADefinedFeatureLevel()
+    {
+        // Guards against any profile (now or future) mapping to a value with no native feature level.
+        foreach (GraphicsProfile profile in Enum.GetValues<GraphicsProfile>())
+        {
+            var featureLevel = profile.ToFeatureLevel();
+            Assert.True(Enum.IsDefined(featureLevel),
+                $"{profile} (0x{(int)profile:X4}) maps to undefined D3DFeatureLevel 0x{(int)featureLevel:X4}");
+        }
+    }
+}
+
+#endif
--- a/sources/engine/Stride.Graphics.Tests/TestGraphicsProfileHelper.cs
+++ b/sources/engine/Stride.Graphics.Tests/TestGraphicsProfileHelper.cs
@@ -48,6 +48,41 @@ public class TestGraphicsProfileHelper
     {
         Assert.Throws<ArgumentOutOfRangeException>(() => ((GraphicsProfile)0x1234).ToFeatureLevel());
     }
+
+    [Fact]
+    public void Level_11_1_And_11_2_ShareAFeatureLevel()
+    {
+        // 11.2 is an API revision on 11_1 hardware, so the two profiles are the same capability tier.
+        // Anything that probes a device with them (GraphicsAdapter.IsProfileSupported) answers alike.
+        Assert.Equal(GraphicsProfile.Level_11_1.ToFeatureLevel(), GraphicsProfile.Level_11_2.ToFeatureLevel());
+    }
+
+    [Fact]
+    public void Profiles_MapElementWise_NotByReinterpretingTheSpan()
+    {
+        GraphicsProfile[] profiles = [GraphicsProfile.Level_11_2, GraphicsProfile.Level_11_1, GraphicsProfile.Level_10_0];
+
+        var featureLevels = GraphicsProfileHelper.ToFeatureLevels(profiles);
+
+        D3DFeatureLevel[] expected = [D3DFeatureLevel.Level111, D3DFeatureLevel.Level111, D3DFeatureLevel.Level100];
+        Assert.Equal(expected, featureLevels);
+        // A bulk reinterpretation of the span would leak 0xB200 through as a feature level.
+        Assert.All(featureLevels, featureLevel => Assert.True(Enum.IsDefined(featureLevel)));
+    }
+
+    [Fact]
+    public void NoProfiles_MapToNoFeatureLevels()
+    {
+        Assert.Empty(GraphicsProfileHelper.ToFeatureLevels([]));
+    }
+
+    [Fact]
+    public void UnknownProfileInASequence_Throws()
+    {
+        GraphicsProfile[] profiles = [GraphicsProfile.Level_11_0, (GraphicsProfile)0x1234];
+
+        Assert.Throws<ArgumentOutOfRangeException>(() => GraphicsProfileHelper.ToFeatureLevels(profiles));
+    }
 }
 
 #endif
--- a/sources/engine/Stride.Graphics.Tests/TestGraphicsProfileHelper.cs
+++ b/sources/engine/Stride.Graphics.Tests/TestGraphicsProfileHelper.cs
@@ -17,25 +17,36 @@ namespace Stride.Graphics.Tests;
 /// </summary>
 public class TestGraphicsProfileHelper
 {
-    [Fact]
-    public void Level_11_2_MapsTo_Level_11_1()
+    [Theory]
+    [InlineData(GraphicsProfile.Level_9_1, D3DFeatureLevel.Level91)]
+    [InlineData(GraphicsProfile.Level_9_2, D3DFeatureLevel.Level92)]
+    [InlineData(GraphicsProfile.Level_9_3, D3DFeatureLevel.Level93)]
+    [InlineData(GraphicsProfile.Level_10_0, D3DFeatureLevel.Level100)]
+    [InlineData(GraphicsProfile.Level_10_1, D3DFeatureLevel.Level101)]
+    [InlineData(GraphicsProfile.Level_11_0, D3DFeatureLevel.Level110)]
+    [InlineData(GraphicsProfile.Level_11_1, D3DFeatureLevel.Level111)]
+    [InlineData(GraphicsProfile.Level_11_2, D3DFeatureLevel.Level111)]
+    public void Profile_MapsToExpectedFeatureLevel(GraphicsProfile profile, D3DFeatureLevel expected)
     {
-        // 0xB200 has no native D3D feature level; it must resolve to the real 11_1 (0xB100),
-        // matching Vulkan, instead of the raw-cast 0xB200 that fails CreateDevice on D3D11/D3D12.
-        Assert.Equal(GraphicsProfile.Level_11_1.ToFeatureLevel(), GraphicsProfile.Level_11_2.ToFeatureLevel());
-        Assert.NotEqual((D3DFeatureLevel)0xB200, GraphicsProfile.Level_11_2.ToFeatureLevel());
+        Assert.Equal(expected, profile.ToFeatureLevel());
     }
 
     [Fact]
     public void EveryProfile_MapsToADefinedFeatureLevel()
     {
-        // Guards against any profile (now or future) mapping to a value with no native feature level.
+        // Guards against a profile added later without a matching entry in the mapping.
         foreach (GraphicsProfile profile in Enum.GetValues<GraphicsProfile>())
         {
             var featureLevel = profile.ToFeatureLevel();
             Assert.True(Enum.IsDefined(featureLevel),
                 $"{profile} (0x{(int)profile:X4}) maps to undefined D3DFeatureLevel 0x{(int)featureLevel:X4}");
         }
+    }
+
+    [Fact]
+    public void UnknownProfile_Throws()
+    {
+        Assert.Throws<ArgumentOutOfRangeException>(() => ((GraphicsProfile)0x1234).ToFeatureLevel());
     }
 }
 
--- a/sources/engine/Stride.Graphics/Direct3D/GraphicsAdapter.Direct3D.cs
+++ b/sources/engine/Stride.Graphics/Direct3D/GraphicsAdapter.Direct3D.cs
@@ -269,7 +269,11 @@ namespace Stride.Graphics
             ID3D11DeviceContext* deviceContext = null;
 
             D3DFeatureLevel matchedFeatureLevel = 0;
-            var featureLevel = (D3DFeatureLevel) graphicsProfile;
+            // Use ToFeatureLevel (not a raw cast) so Level_11_2 resolves to the real FL 11_1 it maps
+            // to; a raw cast would ask CreateDevice for the non-existent 0xB200, which can never match
+            // and would wrongly report the profile as unsupported on every adapter. The exact-match
+            // check below still legitimately rejects real capability gaps (e.g. FL10-only hardware).
+            var featureLevel = graphicsProfile.ToFeatureLevel();
             var featureLevels = stackalloc D3DFeatureLevel[] { featureLevel };
 
             HResult result = d3d11.CreateDevice(pAdapter: null, D3DDriverType.Hardware, Software: IntPtr.Zero,
--- a/sources/engine/Stride.Graphics/Direct3D/GraphicsAdapter.Direct3D.cs
+++ b/sources/engine/Stride.Graphics/Direct3D/GraphicsAdapter.Direct3D.cs
@@ -269,10 +269,6 @@ namespace Stride.Graphics
             ID3D11DeviceContext* deviceContext = null;
 
             D3DFeatureLevel matchedFeatureLevel = 0;
-            // Use ToFeatureLevel (not a raw cast) so Level_11_2 resolves to the real FL 11_1 it maps
-            // to; a raw cast would ask CreateDevice for the non-existent 0xB200, which can never match
-            // and would wrongly report the profile as unsupported on every adapter. The exact-match
-            // check below still legitimately rejects real capability gaps (e.g. FL10-only hardware).
             var featureLevel = graphicsProfile.ToFeatureLevel();
             var featureLevels = stackalloc D3DFeatureLevel[] { featureLevel };
 
--- a/sources/engine/Stride.Graphics/Direct3D/GraphicsProfileHelper.cs
+++ b/sources/engine/Stride.Graphics/Direct3D/GraphicsProfileHelper.cs
@@ -38,6 +38,28 @@ internal static class GraphicsProfileHelper
     };
 
     /// <summary>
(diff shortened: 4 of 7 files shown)


Ethereal77:
LGTM. Good catch!

A little nitpick: The comments look a bit too verbose for my taste (even though I tend to write big explanatory comments 😁)

sasvdw:
@Ethereal77, happy to take suggestions... Was a really tough one from a comments point of view. There's a genuine footgun highlighted in the comments that is bound to cause issues if someone just looks at the changes at surface level.

Ethereal77:
I agree with you. Better documentation is always preferable, specially around conflicting code. But, I think that if you have already created and used a helper `ToFeatureLevel()`, and that helper already documents why the mapping `11_2` is special, then the call site of the helper does not need the same comment, as the potential bug is not there anymore.

In the shader compiler, as it does not use the helper, the comment makes sense.

sasvdw:
@Ethereal77, after further consideration, I think I've landed on something more concise while also adding tests to ensure everything stays up to date.

I also removed the comment from the shader compiler because it's an entirely different case and should be straightforward knowledge for anyone with an understanding of this domain.

Ethereal77:
👍 Looks nice

xen2:
If you don't mind, I would prefer to hold on for this one and take time to properly review the feature level system (inherited from D3D11 when D3D11 was our "main" API model), as this doesn't translate well cross-API. Also this level seems to translate weirdly as it seems to be a D3D software-side change, not HW support.

I think later we might need to go more toward Vulkan model (capabilities, which D3D11/12 also had anyway on top of feature level), and maybe a shader model choice.

Now, my question regarding what to do with this PR is:
Is there any actual need for 11.2 in the meantime? If yes, we can merge this as a stopgap, otherwise let's not add unnecessary complexity until something better is in place.

Ethereal77:
> I think later we might need to go more toward Vulkan model (capabilities, which D3D11/12 also had anyway on top of feature level), and maybe a shader model choice.

That would be really nice. If we design it well it would simplify a lot of logic. I hope however we won't go overboard and end up with the caps inferno of D3D9 times 😁

> Is there any actual need for 11.2 in the meantime?

I'd say not really. Direct3D Feature Level 11.2 included:
* Tiled resources: Would be nice, but there is currently no support for that in `Stride.Graphics` anyway. It would need to be implemented.
* Improvements for WARP devices (support for FL 11.1, double in shaders, 16x MSAA): Would be nice, but mostly for tests, so not really useful for everyday use.
* Annotate graphics commands: Again, nice for better debugging, but not critical.
* HLSL shader linking: Not really useful for us I believe, as that role is already covered by the current shader system, no?

sasvdw:
I've discovered some more silent casts, and I've made those changes. The PR description has also been updated to reflect this.

The gist is that this change isn't so much introducing DirectX3D 11.2, but rather making the config parameter we're exposing "safe" for consumers in Game Studio. There's a larger scope of changes that can express hardware by capability in the long term.

sasvdw:
@xen2, to answer your question directly: nothing needs the 11.2 *features*, and @Ethereal77's list is
right. I went digging because I assumed that a request for 11.2 would at least give us tiled
resources. That assumption is wrong at the API level.

- There is no `D3D_FEATURE_LEVEL` 11_2. Direct3D 11.2 is a runtime revision for Windows 8.1. The
  level enum goes from `11_1 = 0xB100` to `12_0 = 0xC000`.
- Tiled resources are an optional capability. The application queries them after device creation with
  `CheckFeatureSupport(D3D11_FEATURE_D3D11_OPTIONS1)` → `TiledResourcesTier`. Tier 1 needs *feature
  level 11_0 hardware* plus a WDDM 1.3 driver
  ([docs](https://learn.microsoft.com/en-us/windows/win32/direct3d11/tiled-resources-features-tiers)).
- D3D12 has the same shape. `D3D12_FEATURE_D3D12_OPTIONS.TiledResourcesTier` reports it, and adapters
  that support feature level 12_0 all support TIER_2 or above.

So this PR enables nothing, and I will not argue for 11.2 on capability grounds.

What still needs a decision is that we already ship `Level_11_2` and already offer it. It reaches the
engine from disk, not only from code. `GraphicsProfile` is a `[DataContract]`, and `Level_11_2`
carries `[Display("Level 11.2 ~ like Direct3D 11.2 / Vulkan 1.1")]`. It therefore appears in the Game
Settings dropdown for `RenderingSettings.DefaultGraphicsProfile`, and `Game.cs` feeds that value into
both `deviceManager.ShaderProfile` and `PreferredGraphicsProfile`.

Reviewing the PR against the tree turned up more than I first reported. A project that selects 11.2
fails in four places:

1. `GraphicsDevice.Direct3D11.cs` passes `0xB200` to `D3D11CreateDevice` and gets `E_INVALIDARG`. No
   candidate device survives, so the user sees *"None of the graphics profiles [Level_11_2] are
   supported by the Graphics Device. The highest available profile is [Level_11_1]"*. The hardware
   runs everything 11.2 exposes.
2. `ShaderProfileFromGraphicsProfile` in the D3D11 shader compiler has no `Level_11_2` arm and throws
   `Graphics Profile not supported.`
3. The Null backend sets `RequestedProfile = CurrentProfile = Level_11_2`, and `EffectSystem` feeds
   `Features.RequestedProfile` into the compiler.
4. `FindClosestMatchingDisplayMode` exists once per Direct3D backend, and both copies reinterpreted
   the caller's whole profile span as feature levels. `GamePlatform` passes the game's own
   `PreferredGraphicsProfile` there when `IsFullScreen` is set, so `0xB200` reached device creation.
   D3D12 threw from `ThrowNoCompatibleProfile`. D3D11 discarded the `CreateDevice` HRESULT and then
   released the device and the context unconditionally, so it faulted on two null pointers.

Item 4 is worth a look on its own merits. That D3D11 fault never depended on 11.2. Any unsupported
profile in fullscreen reached the same null release, `Level_11_1` on FL10 hardware included. The PR
now maps the span element-wise on both backends and checks the HRESULT before it uses the device.

That leaves two honest options, and I am happy with either:

- **this PR** — keep the value and map it to its real capability tier, so a choice of 11.2 behaves
  like a choice of 11.1 instead of failing.
- **remove or `[Obsolete]` `Level_11_2`** and stop offering it. I will write that PR instead if you
  prefer it.

The state worth avoiding is the current one, where the value sits in the UI and always fails. I have
no urgency about which way it goes.

On the capability model, I agree, and it is closer than it looks. `GraphicsDeviceFeatures` is already
that struct. D3D11 fills four of its flags from real `CheckFeatureSupport` calls. D3D12 already reads
`D3D12_FEATURE_D3D12_OPTIONS`, which is the struct that carries `TiledResourcesTier`. Even
`ComputeShaderFormatSupport.Tiled`, that is `D3D11_FORMAT_SUPPORT2_TILED`, is already populated per
format on both D3D backends, and nothing reads it. The real gap is Vulkan, where `GraphicsProfileHelper`
is commented out and `GraphicsDeviceFeatures.Vulkan` hard-codes every flag with `CurrentProfile` set
to the request.

One design point decides how far capabilities can go, so it is worth settling early. The profile does
two jobs that capabilities cannot merge. Runtime queries can replace the `CurrentProfile >=
Level_11_0` gates. The asset pipeline cannot use them, because it has no device. `TextureHelper`
derives the maximum texture size from the profile, and `EffectCompilerParameters.Profile` picks the
shader model. Both run on a build machine that may have no GPU at all. So the end state is probably
this. The profile, or a renamed successor, stays as the content target. Shader model becomes an
explicit choice. `GraphicsDeviceFeatures` becomes the only runtime authority.

I have built that first step and opened it as #3368. It adds three questions the engine currently
answers with one number: what a backend implements, what a device provides, and both together. Seven
of the twenty-five profile comparisons become device queries, and the other eighteen are grouped by
why they stay — nine have no device to ask, seven are device questions with no capability term yet,
and two read a profile that is not the right source.

It also fixes five bugs the audit turned up, including one that matters here: Vulkan rejects 32-bit
index buffers whenever a project targets 9.x, because `CurrentProfile` is a device fact only on
Direct3D 11 and Vulkan echoes the request. That is the same confusion this PR is about.

The end-user sample suite passes on all four legs, and the branch is rebased on current master.

Ethereal77:
This is an interesting topic. When you have to compile content for a "backend" but does not have a device, how to determine the capabilities level to compile for?

While we decide the direction to go, I think this is a good opportunity to remove all the checks that are old enough to not make sense. D3D9 and OpenGL are already removed, and almost everyone have a D3D12-capable GPU these days (even iGPUs from 2018+). So those checks that limit things if below FL10 or even FL11 are never going to be met.

sasvdw:
@Ethereal77 that's perhaps a batch of tech debt that could land as 4.4?

> This is an interesting topic. When you have to compile content for a "backend" but does not have a device, how to determine the capabilities level to compile for?

"Backend" is the wrong word for it (as you mentioned in my proposal PR). Backend refers to whether there is Stride code that exposes the capability for said API (D3D11/D3D12/Vulkan). There's no device needed to compile for said backend. It's only once running that the game will need to check if the underlying device is actually providing the capabilities the game needs, with certain capabilities degrading to known fallbacks if unsupported.

xen2:
> So this PR enables nothing, and I will not argue for 11.2 on capability grounds.
> 
> What still needs a decision is that we already ship `Level_11_2` and already offer it. It reaches the engine from disk, not only from code. `GraphicsProfile` is a `[DataContract]`, and `Level_11_2` carries `[Display("Level 11.2 ~ like Direct3D 11.2 / Vulkan 1.1")]`. It therefore appears in the Game Settings dropdown for `RenderingSettings.DefaultGraphicsProfile`, and `Game.cs` feeds that value into both `deviceManager.ShaderProfile` and `PreferredGraphicsProfile`.

I think in practice, it's just an enum that was copied from D3D list or added for future purpose (assuming it would work like other feature levels). So it's never been supported and nobody is using it in practice (it's broken anyway if I understood correctly).
Since we will probably switch to a more fine-grained feature/capability approach (i.e. #3368), I think we can close this PR?

sasvdw:
The only reason I'd advocate merging this is to prevent anyone from setting the language level in their game settings and then not understanding why it isn't launching. #3368 isn't going to make the 4.4 release; at least this will remove the footgun.

xen2:
If I understood correctly, right now changing it crashes, right? (either right away or after a restart)
Those previous 4.3 users who might have changed it  (if not crashing right away and saved) probably fixed it back already.

For new 4.4 users, we could keep it simple by simply removing the enum entry for 11.2 -- footgun gone.

xen2:
Note: when getting back to the graphics device capabilities (#3368) we might still want to reuse some part of this PR depending on what we do exactly with our API and how we map it to D3D11/12.

sasvdw:
> If I understood correctly, right now changing it crashes, right? (either right away or after a restart) Those previous 4.3 users who might have changed it (if not crashing right away and saved) probably fixed it back already.
> 
> For new 4.4 users, we could keep it simple by simply removing the enum entry for 11.2 -- footgun gone.

Changing it will lead to your game crashing on startup with the error below. Removing the enum value is also an option, but in all honesty it looks like it'll have the same amount of code changes needed than just ensuring the enum is interpreted correctly.

Your call in the end, the test harness here is still a valuable insight for running certain checks that might be relevant to #3368.

```
[StrideExperimentsGame]: Error: Unexpected exception. System.InvalidOperationException: None of the graphics profiles [Level_11_2] are supported by the Graphics Device. The highest available profile is [Level_11_1].
   at Stride.Games.GraphicsDeviceManager.FindBestDevice(Boolean anySuitableDevice)
   at Stride.Games.GraphicsDeviceManager.<ChangeOrCreateDevice>g__ChangeOrCreateDevice|128_0(<>c__DisplayClass128_0&)
   at Stride.Games.GraphicsDeviceManager.ChangeOrCreateDevice(Boolean forceCreate)
   at Stride.Games.GraphicsDeviceManager.Stride.Games.IGraphicsDeviceManager.CreateDevice()
   at Stride.Games.GameBase.InitializeBeforeRun()
System.InvalidOperationException: None of the graphics profiles [Level_11_2] are supported by the Graphics Device. The highest available profile is [Level_11_1].
   at Stride.Games.GraphicsDeviceManager.FindBestDevice(Boolean anySuitableDevice)
   at Stride.Games.GraphicsDeviceManager.<ChangeOrCreateDevice>g__ChangeOrCreateDevice|128_0(<>c__DisplayClass128_0&)
   at Stride.Games.GraphicsDeviceManager.ChangeOrCreateDevice(Boolean forceCreate)
   at Stride.Games.GraphicsDeviceManager.Stride.Games.IGraphicsDeviceManager.CreateDevice()
   at Stride.Games.GameBase.InitializeBeforeRun()
Unhandled exception. System.InvalidOperationException: None of the graphics profiles [Level_11_2] are supported by the Graphics Device. The highest available profile is [Level_11_1].
   at Stride.Games.GraphicsDeviceManager.FindBestDevice(Boolean anySuitableDevice)
   at Stride.Games.GraphicsDeviceManager.<ChangeOrCreateDevice>g__ChangeOrCreateDevice|128_0(<>c__DisplayClass128_0&)
   at Stride.Games.GraphicsDeviceManager.ChangeOrCreateDevice(Boolean forceCreate)
   at Stride.Games.GraphicsDeviceManager.Stride.Games.IGraphicsDeviceManager.CreateDevice()
   at Stride.Games.GameBase.InitializeBeforeRun()
   at Stride.Games.GamePlatform.OnInitCallback()
   at Stride.Games.GameWindowWinforms.Run()
   at Stride.Games.GamePlatform.Run(GameContext gameContext)
   at Stride.Games.GameBase.Run(GameContext gameContext)
   at Program.<Main>$(String[] args) in C:\Development\lazy-stride-experiments\app\LazyStrideExperiments.Windows\StrideExperimentsApp.cs:line 54
   ```

PR #3304, opened 2026-07-29
fix: invalid SPIR-V that fails HLSL legalization ([numthreads] on non-entry methods, implicit-LOD sampling outside pixel/compute)

### LINKED ISSUE #3303: [numthreads] on a non-entry compute method emits invalid SPIR-V (D3D11 compile fails with a bare InternalError)

**Version**: `master` (reproduced against `4.4.0-beta4`). New in 4.4 — the D3D11 path now round-trips
SPIR-V → HLSL via SPIRV-Cross; 4.3 went straight to HLSL and never hit this.

**Platform(s)**: Windows — Direct3D11 only. (Vulkan/D3D12 never call `LegalizeForHlsl`, so they don't
hit the failure, though the module is still invalid.)

**Describe the bug**

A compute shader with `[numthreads(...)]` on a **non-entry** method — e.g. `override void Compute()`
inheriting `ComputeShaderBase` — fails to compile on Direct3D11. The compiler emits the `[numthreads]`
`OpExecutionMode LocalSize` on the attribute-bearing method (`%..._Compute`), but the compute entry
point is a generated `CSMain_Wrapper`, so the mode is stranded on a non-entry function (invalid
SPIR-V). `EffectCompiler` then hands the invalid module to `LegalizeForHlsl`, which fails with a bare
`spvOptimizerRun InternalError` — nothing validates the module first, so the real cause is hidden. (The
declared thread group size is also silently dropped.)

**To Reproduce**

Minimal repro: **https://github.com/sasvdw/stride-spirv-legalize-repro** (`dotnet run` — no Stride checkout, no GPU).

**Expected behavior**

HLSL/FXC/DXC treat `[numthreads]` on a non-entry function as legal and silently ignored (size comes
only from the entry point). Stride should match that — compile and ignore it, ideally with a warning
pointing to `ThreadNumberX/Y/Z`.

**Screenshots**

N/A.

**Log and callstacks**

spirv-val on the emitted module:

```
OpExecutionMode Entry Point <id> '...[%<Shader>_Compute]' is not the Entry Point operand of an OpEntryPoint.
```

What the engine reports today (`LegalizeForHlsl`):

```
InvalidOperationException: spvOptimizerRun failed: InternalError
```

**Additional context**

Root cause: `MethodOrMember.Compile` emits `OpExecutionMode(function.Id, LocalSize)` on the
attribute-bearing method; `InterfaceProcessor`'s wrapper relocation only moves the entry method's
(`CSMain`'s) modes. Not a regression — the classic ≤4.3 path also dropped a `[numthreads]` on a
non-entry method (fxc honors it only on the entry). Fix PR follows.

### LINKED ISSUE #3313: Texture.Sample() in a vertex-stage material surface emits invalid SPIR-V (D3D11 compile fails with a bare InternalError)

**Version**: `master` (reproduced against a 4.4.1 fork). New in 4.4 — the D3D11 path now round-trips SPIR-V → HLSL via SPIRV-Cross; 4.3 went straight to HLSL and never hit this.

**Platform(s)**: Windows — Direct3D11 only. (Vulkan/D3D12 never call `LegalizeForHlsl`, so they don't hit the failure, though the module is still invalid.)

**Describe the bug**

A `Texture.Sample(...)` call in a **vertex-stage** material surface fails to compile on Direct3D11. The SDSL→SPIR-V frontend emits `OpImageSampleImplicitLod` for `.Sample()` regardless of stage, but SPIR-V forbids implicit-LOD instructions outside the Fragment / GLCompute / Mesh / Task execution models (no screen-space derivatives exist in the vertex stage). `EffectCompiler` then hands the invalid module to `LegalizeForHlsl`, which fails with a bare `spvOptimizerRun InternalError` — nothing validates the module first, so the real cause is hidden.

The built-in `MaterialDisplacementMapFeature` does **not** hit this: it sets `context.IsNotPixelStage = true`, which routes its compute-color texture graph through the `ComputeColorTextureLod*` shaders (`SampleLevel`). That workaround only covers the compute-color path (`ComputeColors/ComputeTextureBase.cs`); a custom SDSL shader that calls `.Sample()` directly is lowered by `TextureMethodsImplementations.CompileSample`, which is stage-unaware and always emits implicit LOD.

**To Reproduce**

Minimal SDSL surface that triggers it (the texture fetch is legal HLSL in a vertex shader):

```csl
shader VertexTextureFetchSurface : IMaterialSurface, PositionStream, Texturing
{
    override void Compute()
    {
        float h = Texture0.Sample(PointSampler, float2(0.5, 0.5)).r;
        streams.Position = float4(streams.Position.x, h, streams.Position.z, streams.Position.w);
    }
};
```

…injected into a material's vertex stage via a `MaterialFeature` using the stock `SetStreamFinalModifier` API (the same hook `MaterialDisplacementMapFeature` uses):

```csharp
context.SetStreamFinalModifier<MyFeature>(MaterialShaderStage.Vertex, new ShaderClassSource("VertexTextureFetchSurface"));
```

Apply that material to any mesh and run. No texture needs to be bound — the crash is in shader compilation, not rendering. *(Full GameStudio repro: shader + feature + steps available if useful.)*

Repo with reproduce project: https://github.com/delustra/stride-vtf-spirv-repro

**Expected behavior**

HLSL/FXC/DXC treat `Texture.Sample(...)` in a vertex shader as legal and lower it to LOD 0 (no gradients available). Stride should match that — compile and emit `OpImageSampleExplicitLod` with `Lod = 0.0` for implicit-LOD samples reachable from a Vertex / Hull / Domain / Geometry entry point (Fragment / Compute / Mesh / Task left untouched).

**Screenshots**

N/A.

**Log and callstacks**

`spirv-val` on the emitted module (dumped pre-legalization):

```
OpEntryPoint Entry Point <id> '...VSMain_Wrapper's callgraph contains function <id> '...VertexTextureFetchSurface_Compute', which cannot be used with the current execution model:
ImplicitLod instructions require Fragment, GLCompute, MeshEXT or TaskEXT execution model: ImageSampleImplicitLod
```

What the engine reports today (`LegalizeForHlsl`):

```
InvalidOperationException: spvOptimizerRun failed: InternalError
   at Stride.Shaders.Spirv.Tools.SpirvTools.RunAndCopy(...)
   at Stride.Shades.Spirv.Tools.SpirvTools.LegalizeForHlsl(...)
   at Stride.Shaders.Compiler.EffectCompiler.Compile(...)
   at ...RootEffectRenderFeature.PrepareEffectPermutations
```

**Additional context**

Root cause: `TextureMethodsImplementations.CompileSample` (`sources/shaders/Stride.Shaders.Parsers/Parsing/SDSL/AST/TextureMethodsImplementations.cs`) unconditionally emits `OpImageSampleImplicitLod`; there is no stage awareness on that path. Compare `CompileSampleLevel` (~line 139), which correctly emits `OpImageSampleExplicitLod`.

Sibling of #3303 — different invalid instruction (`OpImageSampleImplicitLod` vs `OpExecutionMode LocalSize`), different stage, different root cause, but the same failure mode: the 4.4 SDSL→SPIR-V frontend emits spec-invalid SPIR-V that surfaces as a bare `InternalError` because the module isn't validated before `spirv-opt`. #3304 fixes the `[numthreads]` variant; this one needs its own fix.

Suggested fix: a post-merge pass that walks the call-graph from Vertex / Hull / Domain / Geometry entry points and rewrites any reachable `OpImageSampleImplicitLod` (+ Dref/Proj variants) to the matching `…ExplicitLod` opcode with `Lod = 0.0` — mirroring the HLSL lowering, and orthogonal to the existing `IsNotPixelStage` compute-color workaround. (A broader hardening worth considering for this whole family: run `spirv-val` and surface its message before `spirv-opt`, so variants like #3303 and this one self-diagnose instead of throwing a bare `InternalError`.)

sasvdw:
# PR Details

**Summary**

Two unrelated SDSL constructs emit invalid SPIR-V. On Direct3D11 both fail the same way.
`LegalizeForHlsl` throws a bare `spvOptimizerRun InternalError`, because nothing validates the module
first. This PR fixes both causes and reports the shared failure clearly.

## 1. `[numthreads]` on a non-entry compute method (#3303)

The compiler emitted `OpExecutionMode LocalSize` on the method that carries the attribute. The compute
entry point is a generated `CSMain_Wrapper`, so the mode landed on a non-entry function. The result was
invalid SPIR-V.

fxc and dxc accept `[numthreads]` on a non-entry function and then ignore it. The thread group size
comes only from the entry point. Stride now does the same.

- `MethodOrMember.cs`: emit `LocalSize` only when the attribute is on the compute entry
  (`EntryPoint == ComputeShader`). Otherwise skip it and log a warning that points to
  `ThreadNumberX/Y/Z`.
- `ComputeEffectShader.cs`: clarify the XML documentation. `ThreadNumbers` sets the thread group size of
  the compute shader, which is its `[numthreads]`. `ThreadGroupCounts` is the dispatch count.

This is not a regression. The classic 4.3 path also dropped a misplaced `[numthreads]`.

## 2. Implicit-LOD texture sampling outside the pixel and compute stages (#3313)

A `Texture.Sample(...)` call reachable from a vertex, hull, domain or geometry entry point emitted
`OpImageSampleImplicitLod`. SPIR-V allows that instruction only in `Fragment`, `GLCompute`, `MeshEXT`
and `TaskEXT`. Implicit LOD needs screen-space derivatives, and those stages do not have them.

**This change reports an error. It does not lower the call to `SampleLevel(..., 0)`, because the shader
was never valid.** I checked both Microsoft compilers directly:

| | `Texture0.Sample(s, uv)` in a vertex shader |
|---|---|
| `fxc /T vs_5_0` | `error X4532: cannot map expression to vs_5_0 instruction set` |
| `dxc -T vs_6_0` | `error: Opcode Sample not valid in shader model vs_6_0` → *"Function uses features incompatible with the shader stage (vs)"* |
| `dxc -T vs_6_0 -spirv` | `error: sampling with implicit lod is only allowed in fragment and compute shaders` |
| SPIR-V spec | `ImplicitLod instructions require Fragment, GLCompute, MeshEXT or TaskEXT execution model` |

The Microsoft documentation for `Sample` agrees: *"This method can be invoked within a pixel shader, but
it is not supported in a vertex shader or a geometry shader."* The "Vertex-shader stage" page does list
`Sample` as legal in a vertex shader. That entry is a long-standing documentation error and should read
`SampleLevel`. `SampleGrad` and `SampleCmpLevelZero` do compile for `vs_5_0`.

So this is not a functional regression, because such a shader never compiled. It is a diagnostics
regression. Version 4.3 reported `X4532`. Version 4.4 reported `spvOptimizerRun InternalError`.

The engine has always treated this as illegal. That is why `MaterialDisplacementMapFeature` sets
`IsNotPixelStage` and routes texture reads through `ComputeColorTextureLod`, which uses `SampleLevel`.

- `ShaderMixer.Validation.cs`: after interface processing, walk the call graph from each non-fragment entry point.
  Report every implicit-LOD sample that the walk reaches, with its source location and the explicit-LOD
  alternatives. The same walk covers `ddx`, `ddy`, `fwidth` and `CalculateLevelOfDetail`, which need the
  same derivatives:

  ```
  VSTextureSample.sdsl(14,25): A texture sample with an implicit level of detail needs a pixel or
  compute shader. A Vertex shader reaches this code. Use SampleLevel, SampleGrad or SampleCmpLevelZero
  to give an explicit level of detail.
  ```

A silent lowering to LOD 0 would make Stride accept a shader that fxc, dxc and the SPIR-V validator all
reject. It would also change the sampling behavior without telling the author. I am happy to change this
if you prefer the permissive behavior.

## 3. Shared diagnostics

- `EffectCompiler.cs`: when `LegalizeForHlsl` throws, validate the module and report the instruction
  that caused the failure, instead of a bare `InternalError`. Issue #3313 asks for this. It also covers
  future problems of the same family.

## Related Issue

Fixes #3303.
Fixes #3313.

Repro for #3303: https://github.com/sasvdw/stride-spirv-legalize-repro
Repro for #3313: https://github.com/delustra/stride-vtf-spirv-repro

## Types of changes

## Checklist

**Validation status**

- *Tests*: `StrideShaderTests` gains `TextureSampleOutsideFragmentStageIsReported`, which uses
  `CompilerTests/VSTextureSample.sdsl`. It also gains `TextureSampleInFragmentStageCompiles`, which uses
  `CompilerTests/PSTextureSample.sdsl`. The second test is the control. It proves that pixel-stage
  sampling still compiles to valid SPIR-V. `NumThreadsOnNonEntryMethodIsIgnoredWithWarning` still passes.
  I checked each test red before the fix and green after it. The test for #3313 is a separate commit that
  comes before its fix, so the failure is on record.
- *Suite*: 518 tests pass. 10 failures are pre-existing on this branch and unrelated. The test harness
  parses floats from `.sdsl` test headers with the current culture, so they fail on any comma-decimal
  locale. The same tests fail when I revert these changes. #3325 fixes them separately.
- *Editor*: outstanding. This is the mandatory personal-testing step.
- *Docs*: the XML documentation for `ComputeEffectShader.ThreadNumbers` and `ThreadGroupCounts` is now
  clearer, and it flows to the API reference. The manual has no compute-shader guide. I track that as a
  separate `stride-docs` follow-up.


sources/engine/Stride.Rendering/Rendering/ComputeEffect/ComputeEffectShader.cs
```diff
@@ -43,12 +43,14 @@ public ComputeEffectShader(RenderContext context)
...
-        /// Gets or sets the number of group counts the shader should be dispatched to.
+        /// Gets or sets the number of thread groups to dispatch; total threads = <see cref="ThreadGroupCounts"/> × <see cref="ThreadNumbers"/>.
         /// </summary>
         public Int3 ThreadGroupCounts { get; set; }
 
         /// <summary>
-        /// Gets or sets the number of threads desired by thread group.
+        /// Gets or sets the compute shader's thread group size — the shader's <c>[numthreads(x, y, z)]</c>, applied via
+        /// the <c>ThreadNumberX/Y/Z</c> macros on <c>CSMain</c>. This is the supported way to set it; a literal
+        /// <c>[numthreads]</c> in the shader body is ignored. Keep consistent with <see cref="ThreadGroupCounts"/>.
```
xen2, line 53: New description is too complex/long.
Maybe such details fit better in <remarks> tag?

sources/shaders/Stride.Shaders.Compilers/SDSL/ShaderMixer.cs
```diff
@@ -94,6 +97,9 @@ public bool MergeSDSL(ShaderSource shaderSource, Options options, ILogger log, o
         };
         (entryPoints, globalContext.Reflection.InputAttributes) = interfaceProcessor.Process(table, temp, context);
 
+        if (!ValidateImplicitLodSampling(context, temp, entryPoints, log))
```
xen2, line 100: I am hesitating if this validation is better separate/standalone (as is) or as part of `InterfaceProcessor` (which already compute caller and method list per entry point).

On one hand, it's nice to have it standalone.
On the other hand, it's reprocessing a lot of info we already have at hand during `InterfaceProcessor`.

It might be worth checking which implementation makes more sense.)

If `InterfaceProcessor` context doesn't help much, I am fine to keep in ShaderMixer.cs (but at least in a separate ShaderMixer.Validation.cs partial file).

xen2: Thinking more about it, maybe better not to mix it with InterfaceProcessor.

Another option is to consider that we could later easily factor/reuse the analysis that compute caller/callee chains from a given entry point, and reuse it in this validation and any other future validations as it is probably a common use-case (and maybe even in InterfaceProcessor, but not sure it's possible).

But if it's the only place need such caller/callee analysis from an entrypoint so far, it might be fine to keep it as is and do that split later. (only thing to do would be to move it to a ShaderMixer.Validation.cs partial file).

sources/engine/Stride.Rendering/Rendering/ComputeEffect/ComputeEffectShader.cs
sasvdw: Done in fcec6f0a2. `<summary>` is one line now; the `ThreadNumberX/Y/Z` macro mechanism and the "a literal `[numthreads]` in the shader body is ignored" caveat moved to `<remarks>`.

sources/shaders/Stride.Shaders.Compilers/SDSL/ShaderMixer.cs
sasvdw: Moved to `ShaderMixer.Validation.cs` in 23146b733. Pure move, no behaviour change.

I checked the `InterfaceProcessor` option before doing that, and it does not buy much:

- The per-entry-point method set lives in `LiveAnalysis` / `MethodInfo`, both `internal` to `Stride.Shaders.Parsers`, and there is no `InternalsVisibleTo` for `Stride.Shaders.Compilers`. `InterfaceProcessor.Result` only exposes `EntryPoints` and `InputAttributes`.
- That state is transient by design. `ResetUsedThisStage` clears it between stages and methods are cloned per stage (`OriginalMethodCode` / `ThisStageMethodId`), so nothing survives `Process` for a validator to read.
- `ReadWriteAnalyzer.AnalyzeStreamReadWrites` does recurse through `OpFunctionCall`, but the recursion is tied to stream read/write bookkeeping and it drops `OpLine`, which this diagnostic needs for the source location in the message.

So I agree with your second reading: keep it standalone for now. If a second validation needs the same caller/callee walk from an entry point, that is the moment to factor out a shared call-graph analysis, and this method is a straightforward first caller for it.

sources/shaders/Stride.Shaders.Compilers/EffectCompiler.cs
```diff
@@ -233,7 +233,21 @@ public override TaskOrResult<EffectBytecodeCompilerResult> Compile(ShaderMixinSo
...
+                    {
+                        legalizedSpirv = SpirvTools.LegalizeForHlsl(MemoryMarshal.Cast<byte, uint>(spirvBytecode));
+                    }
+                    catch (Exception legalizeException)
+                    {
+                        var validationResult = Spirv.Tools.Spv.ValidateBinary(spirvBytecode, targetVulkan: false);
+                        if (!validationResult.IsValid)
+                        {
+                            log.Error($"SPIR-V for effect {fullEffectName} (id: {mixinObjectId}) is invalid and cannot be legalized for HLSL: {validationResult.Output}", legalizeException);
+                            return new EffectBytecodeCompilerResult(null, log);
```
xen2, line 247: Outer catch won't work anymore (the one doing `WriteSpvDebugFiles(effectDir, mixinObjectId.ToString(), spirvBytecodeForDebug);`)

sources/shaders/Stride.Shaders.Compilers/SDSL/ShaderMixer.Validation.cs
```diff
@@ -0,0 +1,108 @@
...
+                        : null;
+                    break;
+
+                case Op.OpFunctionCall:
+                    if (!callees.TryGetValue(currentFunction, out var calls))
+                        callees.Add(currentFunction, calls = []);
+                    calls.Add(((OpFunctionCall)i).Function);
+                    break;
+
+                case Op.OpImageSampleImplicitLod:
```
xen2, line 57: Can you check other opcodes that might need this?
OpDPdx, OpDPdy, OpFwidth, OpImageQueryLod

Otherwise I think ddx() and so on will still have same error.

sources/shaders/Stride.Shaders.Compilers/SDSL/ShaderMixer.Validation.cs
```diff
@@ -0,0 +1,108 @@
...
+
+                case Op.OpLine:
+                    OpLine line = i;
+                    currentLocation = sourceFiles.TryGetValue(line.File, out var fileName)
+                        ? $"{fileName}({line.Line},{line.Column})"
+                        : null;
+                    break;
+
+                case Op.OpFunctionCall:
+                    if (!callees.TryGetValue(currentFunction, out var calls))
```
xen2, line 52: FunctionCall is not enough. Hull PatchConstant method is linked using PatchConstantFuncSDSL decoration.

sources/shaders/Stride.Shaders.Compilers/SDSL/ShaderMixer.cs
```diff
@@ -79,6 +79,9 @@ public bool MergeSDSL(ShaderSource shaderSource, Options options, ILogger log, o
             return false;
         }
 
+        foreach (var warning in table.Warnings)
```
xen2, line 82: SDSLC.cs line 98 already drains warnings (and info).
So I wonder why is this needed?

sources/shaders/Stride.Shaders.Parsers/Parsing/SDSL/AST/ShaderElements.MethodOrMember.cs
```diff
@@ -520,17 +520,25 @@ public void Compile(SymbolTable table, ShaderClass shader, CompilerUnit compiler
                 {
                     if (anyAttribute.Name == "numthreads")
                     {
-                        var parameters = attrParamBuffer[..anyAttribute.Parameters.Count];
-                        for (var index = 0; index < anyAttribute.Parameters.Count; index++)
+                        if (EntryPoint != EntryPoint.ComputeShader)
```
xen2, line 523: Won't this trigger on each `CSMain` override? Right now we don't override it, but it could happen.
We need a way to select which one (either most overriden, or the first one declaring it starting from most derived? -- maybe you could check what we do for other HSMain/DSMain attributes, possibly it's not handled either)

sources/engine/Stride.Rendering/Rendering/ComputeEffect/ComputeEffectShader.cs
```diff
@@ -43,13 +43,18 @@ public ComputeEffectShader(RenderContext context)
...
         public Int3 ThreadGroupCounts { get; set; }
 
         /// <summary>
-        /// Gets or sets the number of threads desired by thread group.
+        /// Gets or sets the compute shader's thread group size.
         /// </summary>
+        /// <remarks>
+        /// This is the shader's <c>[numthreads(x, y, z)]</c>, applied through the <c>ThreadNumberX/Y/Z</c> macros
+        /// on <c>CSMain</c>. It is the supported way to set the thread group size; a literal <c>[numthreads]</c>
+        /// in the shader body is ignored. Keep consistent with <see cref="ThreadGroupCounts"/>.
```
xen2, line 56: Slightly misleading: You can still add numthread on your own CSMain override (well, depending what is done in `ShaderElements.MethodOrMembers.cs`. What is not authorized is to add it to `Compute()` but this should never have worked and is not part of HLSL spec either.
So maybe this part can be removed?

xen2:
As a side note, one issue I noticed is that once the shader come from cache, it doesn't have the warning that happened during initial compilation.
Just like we do for the asset compiler cache, we probably want to add warning/diagnostic in the cached SPV.
Something like `OpDiagnosticSDSL(severity, fileId, line, column, message)`

Not needed for this PR, but recording this as memory for a future follow-up

sources/shaders/Stride.Shaders.Compilers/EffectCompiler.cs
sasvdw: Fixed. The early return skipped the outer catch, so the `.spv` file never got written. It now writes before returning.

sources/shaders/Stride.Shaders.Compilers/SDSL/ShaderMixer.Validation.cs
sasvdw: Added `OpImageQueryLod` and the derivative opcodes (`OpDPdx`, `OpDPdy`, `OpFwidth`, plus Fine and Coarse). `ddx()` in a vertex shader now reports. I also split the message and renamed the method to `ValidateDerivativeUsage`, since the old wording did not fit a bare `ddx()`. Test: `DerivativeOutsideFragmentStageIsReported`.

sources/shaders/Stride.Shaders.Compilers/SDSL/ShaderMixer.Validation.cs
sasvdw: This one already works. Validation runs after `InterfaceProcessor.Process`, and the generated hull wrapper calls the patch constant function with an `OpFunctionCall`, so the walk finds it. I tried a hull shader sampling in `HSConstantMain` and it reports. Added that as a test. No code change.

sources/shaders/Stride.Shaders.Compilers/SDSL/ShaderMixer.cs
sasvdw: Different tables. `MergeSDSL` makes its own `SymbolTable` at `ShaderMixer.cs:50`, so it is not a duplicate of `SDSLC.cs:98`. I checked: mixing `CSNumThreadsOnNonEntry` cold gives 1 warning here, mixing it again gives 0, because it comes from the cache and `Compile` does not re-run. So it is needed, and the cached case is the `OpDiagnosticSDSL` follow-up you mentioned. Left alone.

sources/shaders/Stride.Shaders.Parsers/Parsing/SDSL/AST/ShaderElements.MethodOrMember.cs
sasvdw: You are right, and it is already broken on master. `EntryPoint` comes from the method name (`MethodOrMember.cs:378`), so the base and the override both count as the compute entry point. With an override, the base keeps its `OpExecutionMode LocalSize` on a function that is no longer an entry point. `MergeSDSL` returns true but validation fails with `OpExecutionMode Entry Point <id> ... is not the Entry Point operand of an OpEntryPoint`.

This PR wraps the old body in `if`/`else` and the override path takes the `else` branch unchanged, so it behaves the same as master. I think the fix belongs where the wrapper is generated, dropping execution modes whose target is no longer an entry point. Want me to open an issue? I have not looked at `HSMain`/`DSMain`.

sources/engine/Stride.Rendering/Rendering/ComputeEffect/ComputeEffectShader.cs
sasvdw: Removed.

sources/shaders/Stride.Shaders.Parsers/Parsing/SDSL/AST/ShaderElements.MethodOrMember.cs
xen2: Yes, let's do a separate issue.
We'll need to have a proper policy on what to do when overriding an entrypoint: transfer their attribute (+ override/merge if same attribute on overriding method) or simply drop them, even if attribute not specified again.

sources/shaders/Stride.Shaders.Compilers/SDSL/ShaderMixer.cs
xen2: The "different tables" part is right, but that is exactly why the loop cannot fire: nothing ever adds a warning to the mixer's table.

I checked it on 0880ddb9b5. Deleting the two lines leaves your own test passing, along with all five new tests in this PR. With your `AddWarning` in place I also printed `table.Warnings.Count` where you drain, and it is 0, so the drain body never runs.

`AddWarning` is only reachable from `MethodOrMember.Compile`, which only ever runs with the table SDSLC creates at `SDSLC.cs:51`. `LoadFromCode` takes no table and `Warnings` is a per-instance list, so the mixer's `new SymbolTable(...)` at `ShaderMixer.cs:50` never receives it. The warning you see on a cold compile comes from the existing drain at `SDSLC.cs:98`, which logs into the same `ILogger` because `MergeSDSL` sets `shaderLoaderBase.Log = log` at line 46. That is also why it goes away on a cache hit.

Could you remove the two lines, and the matching bullet in the PR description? One detail in passing: SDSLC prints `warning.ToString()`, which carries the location and the source line, while this loop prints `.Message`, which carries neither.

sources/shaders/Stride.Shaders.Parsers/Parsing/SDSL/AST/ShaderElements.MethodOrMember.cs
xen2: (I let you open the issue since you offered, thanks)

sources/shaders/Stride.Shaders.Compilers/SDSL/ShaderMixer.cs
sasvdw: You're right, thanks for checking. Removed the two lines and the bullet.

sources/shaders/Stride.Shaders.Parsers/Parsing/SDSL/AST/ShaderElements.MethodOrMember.cs
sasvdw: Opened #3430.

sasvdw: Fix is up in #3431.

PR #3324, opened 2026-08-04
fix: synchronise Vulkan attachment reads, buffer barriers and swapchain acquisition

sasvdw:
# PR Details

> Rebased onto `master` after #3337 (Vulkan 1.3, dynamic rendering) and #3338 (explicit transitions) merged. Two things changed in the rebase: the swapchain block section 3 deletes is now the `vkQueueSubmit2` form of the same code, and the `VkValidationFeaturesEXT` struct is chained ahead of the MoltenVK `VkLayerSettingsCreateInfoEXT` that #3337 added, so both reach `vkCreateInstance` on Apple platforms.

Vulkan barriers left three classes of access unsynchronized. This PR fixes all three, and adds the switch that reports them.

### 1. Attachment barriers covered writes but not reads

`BarrierMapping.ToVkAccessFlags` gave `RenderTarget` only `ColorAttachmentWrite`, and `DepthStencilWrite` only `DepthStencilAttachmentWrite`. A color attachment is also read, by blending and by a load operation that preserves the existing contents. A writable depth attachment is read by the depth and stencil tests. Every barrier the engine issued for a render target or a depth buffer therefore left the render pass read unsynchronized.

`Texture.Vulkan.cs` applied the same write-only masks as a texture's initial access state. A fix to the mapping alone was not enough.

Direct3D12 already implements this model. One `RENDER_TARGET` access bit covers both directions, and `DEPTH_STENCIL_WRITE` permits the test reads. Vulkan keeps its access flags strictly separate, so the same intent has to be spelled out.

`BarrierLayout` never documented which accesses each member covers. `UnorderedAccess` is the one member whose documentation said "reading and writing", and it is also the one member whose mapping was already correct. The two members that were wrong are the two that documented nothing. Each member now states what it covers, so a backend maps against a contract instead of a name.

Under dynamic rendering the hazard is reported at `vkCmdBeginRendering` for every attachment with `LOAD_OP_LOAD`: colour, depth and stencil. Synchronization validation names the missing `COLOR_ATTACHMENT_READ` and `DEPTH_STENCIL_ATTACHMENT_READ` bits directly.

### 2. `ResourceBarrierTransition` threw for buffers

The method handled `Texture` and threw `NotImplementedException` for everything else. A compute shader could not hand a structured buffer write to a later reader.

Buffers have no layout in Vulkan, so the new branch derives access and stage flags from `BarrierLayout` and never calls `ToVkImageLayout`. It records the last access per command buffer, so a later transition names an accurate source instead of the buffer's static usage superset. On first touch it reads `NativeAccessMask` and `NativePipelineStageMask` as a conservative source, but it never writes them. The copy and upload paths need those to stay a superset of every legal usage.

### 3. Swapchain images were transitioned before acquisition

`CreateBackBuffers` transitioned every swapchain image to Present and submitted that command buffer before the first acquisition. Vulkan permits use of a presentable image only between `vkAcquireNextImageKHR` and `vkQueuePresentKHR`, and a layout transition is such a use. Validation reports the violation once per image per swapchain creation, including after every resize.

The transition existed to give the tracked layout a valid source. `AcquireNextImage` now sets that state to `Undefined`, which is correct because the contents of a newly acquired image are undefined. This also removes a `vkQueueWaitIdle` from every swapchain creation and resize.

This defect needs a real swapchain, so no headless test reaches it.

## Opt-in synchronization validation

`STRIDE_VULKAN_SYNC_VALIDATION=1` enables `VK_EXT_validation_features` with `SynchronizationValidation`. Core validation does not report a missing barrier. This does, and it is what found the first defect above.

The switch is opt-in for a reason. Synchronization validation also reports hazards that predate any given change, which would fail unrelated tests. Its extension comes from the validation layer rather than the ICD, so an instance extension query made with no layer name never lists it.

## Tests

`TestBufferBarrier` is new. A compute shader writes a structured buffer, and a second dispatch reads that buffer as a shader resource.

The consumer is a dispatch rather than a readback on purpose. A readback copy emits its own barrier from the buffer's access and stage masks. That barrier would synchronize the write even with no transition present, and the test could never fail.

The honest limit of this test: it proves the barrier is emitted and legal, and that the round trip produces the expected values. It does not prove that removing the barrier fails deterministically, because the hazards synchronization validation catches are reported at `vkQueueSubmit` time and across command buffers. Run the suite with `STRIDE_VULKAN_SYNC_VALIDATION=1` to have Vulkan report the hazard directly.

## Verification

Debug build, so the validation layers load. `Stride.Graphics.Tests.11_0` and `Stride.Graphics.Tests.10_0` pass on Vulkan (Lavapipe), Direct3D11 (WARP) and Direct3D12 (WARP). Both suites also pass on Vulkan with `STRIDE_VULKAN_SYNC_VALIDATION=1` and report no hazard. The default path, with the switch unset, is unchanged.

Each defect was checked A/B on the rebased code:

- **Defect 1.** With the access-mask fix reverted and synchronization validation on, `TestHammersley` and `TestRadiancePrefilteringGgxFaceContinuity` fail on `READ_AFTER_WRITE` hazards at `vkCmdBeginRendering`. With the fix, no hazard.
- **Defect 2.** `TestBufferBarrier` passes on all three backends.
- **Defect 3.** A windowed game with two mid-run resizes, so three swapchain creations. With `master`'s presenter, validation reports `performs a layout transition on presentable VkImage … but the image has not been acquired` on `vkQueueSubmit2` for every image of every swapchain. With this PR, no error and no warning, with and without synchronization validation.

Direct3D is unaffected by design. Every behavioral change is inside the Vulkan backend. `BarrierLayout.cs` gains documentation only.

## Types of changes

## Checklist

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01VHYKv129fEUp6sxMbybZWf

### DIFF
--- a/sources/engine/Stride.Graphics.Tests.11_0/TestBufferBarrier.cs
+++ b/sources/engine/Stride.Graphics.Tests.11_0/TestBufferBarrier.cs
@@ -1,0 +1,94 @@
+// Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net)
+// Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using System.Threading.Tasks;
+
+using Xunit;
+
+using Stride.Core;
+using Stride.Core.Mathematics;
+using Stride.Rendering;
+using Stride.Rendering.ComputeEffect;
+
+namespace Stride.Graphics.Tests;
+
+/// <summary>
+/// Covers a compute write handed to a second dispatch that reads it as a shader resource.
+/// </summary>
+/// <remarks>
+/// The consumer is a dispatch rather than a readback: a readback copy emits its own barrier from
+/// the buffer's access and stage masks, which would synchronise the write even with no transition
+/// and leave the test unable to fail. Run with STRIDE_VULKAN_SYNC_VALIDATION=1 to have Vulkan
+/// report the hazard when the transition is missing.
+/// </remarks>
+public class TestBufferBarrier : GraphicTestGameBase
+{
+    private const int ElementCount = 64;
+
+    private ComputeEffectShader writeEffect;
+    private ComputeEffectShader readEffect;
+    private Buffer sharedBuffer;
+    private Buffer resultBuffer;
+
+    protected override void RegisterTests()
+    {
+        base.RegisterTests();
+
+        FrameGameSystem.Draw(ComputeWriteIsVisibleToASubsequentDispatch);
+    }
+
+    protected override async Task LoadContent()
+    {
+        await base.LoadContent();
+
+        sharedBuffer = Buffer.Structured.New<uint>(GraphicsDevice, ElementCount, unorderedAccess: true).DisposeBy(this);
+        resultBuffer = Buffer.Structured.New<uint>(GraphicsDevice, ElementCount, unorderedAccess: true).DisposeBy(this);
+
+        var renderContext = RenderContext.GetShared(Services);
+        writeEffect = new ComputeEffectShader(renderContext)
+        {
+            ShaderSourceName = "BufferBarrierTestShader",
+            ThreadNumbers = new Int3(ElementCount, 1, 1),
+            ThreadGroupCounts = new Int3(1, 1, 1),
+        };
+        writeEffect.DisposeBy(this);
+
+        readEffect = new ComputeEffectShader(renderContext)
+        {
+            ShaderSourceName = "BufferBarrierReadTestShader",
+            ThreadNumbers = new Int3(ElementCount, 1, 1),
+            ThreadGroupCounts = new Int3(1, 1, 1),
+        };
+        readEffect.DisposeBy(this);
+    }
+
+    private void ComputeWriteIsVisibleToASubsequentDispatch()
+    {
+        var commandList = GraphicsContext.CommandList;
+        var renderDrawContext = new RenderDrawContext(Services, RenderContext.GetShared(Services), GraphicsContext);
+
+        commandList.ResourceBarrierTransition(sharedBuffer, BarrierLayout.UnorderedAccess);
+        writeEffect.Parameters.Set(BufferBarrierTestShaderKeys.Output, sharedBuffer);
+        ((RendererBase)writeEffect).Draw(renderDrawContext);
+
+        commandList.ResourceBarrierTransition(sharedBuffer, BarrierLayout.ShaderResource);
+        commandList.ResourceBarrierTransition(resultBuffer, BarrierLayout.UnorderedAccess);
+        readEffect.Parameters.Set(BufferBarrierReadTestShaderKeys.Input, sharedBuffer);
+        readEffect.Parameters.Set(BufferBarrierReadTestShaderKeys.Result, resultBuffer);
+        ((RendererBase)readEffect).Draw(renderDrawContext);
+
+        var values = resultBuffer.GetData<uint>(commandList);
+
+        Assert.Equal(ElementCount, values.Length);
+        for (uint i = 0; i < ElementCount; i++)
+        {
+            Assert.Equal((i * 3) + 1, values[i]);
+        }
+    }
+
+    [SkippableFact]
+    public void ComputeWriteIsVisibleToASubsequentRead()
+    {
+        RunGameTest(new TestBufferBarrier());
+    }
+}
--- a/sources/engine/Stride.Graphics/Vulkan/CommandList.Vulkan.cs
+++ b/sources/engine/Stride.Graphics/Vulkan/CommandList.Vulkan.cs
@@ -36,6 +36,10 @@ namespace Stride.Graphics
         // that's a known limitation to be addressed by adding a "last-submitted layout" tracker.
         private readonly Dictionary<Texture, BarrierLayout> currentCbLayouts = new();
 
+        // Buffers have no Vulkan layout: this records the last access THIS CB synchronised against, so
+        // a later transition can name an accurate source instead of the buffer's static usage superset.
+        private readonly Dictionary<Buffer, BarrierLayout> currentCbBufferLayouts = new();
+
         private readonly Dictionary<FramebufferKey, VkFramebuffer> framebuffers = new();
         private readonly VkImageView[] framebufferAttachments = new VkImageView[9];
         private int framebufferAttachmentCount;
@@ -87,6 +91,7 @@ namespace Stride.Graphics
             CleanupRenderPass();
             boundDescriptorSets.Clear();
             currentCbLayouts.Clear();
+            currentCbBufferLayouts.Clear();
 
             framebuffers.Clear();
             framebufferDirty = true;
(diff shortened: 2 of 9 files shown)


xen2:
~~Thanks, the Voxels rewrite is right. I pushed two commits on top, please pull before continuing.~~

~~- **Mixer:** a `stage compose` declared only by a nested shader could not be supplied at the root either. Now a stage slot always takes its value from the root. Two tests added.~~
~~- **Voxels:** marchers bind their own samplers (`ApplyAttributeSamplers`), so `LightVoxelRenderer` no longer hard-codes the `Marcher` path.~~

~~Shader tests pass (849). I did not run the voxel GI demo, could you check it once?~~

xen2:
Sorry, wrong PR for previous message!

sources/engine/Stride.Graphics/Vulkan/SwapChainGraphicsPresenter.Vulkan.cs
```diff
@@ -223,6 +231,11 @@ private unsafe void AcquireNextImage()
             // Flip render targets
             backBuffer.SetNativeHandles(swapchainImages[currentBufferIndex].NativeImage, swapchainImages[currentBufferIndex].NativeColorAttachmentView);
 
+            backBuffer.NativeLayout = VkImageLayout.Undefined;
```
xen2, line 234: This is the right model, but it exposes a latent problem in the copy paths. `Copy`, `CopyRegion` and `UpdateSubresource` in `CommandList.Vulkan.cs` transition the image to a transfer layout and then restore it with `newLayout = texture.NativeLayout` (for example `imageBarriers[...].newLayout = sourceParent.NativeLayout` in `Copy`).

Before this change a fresh back buffer had `NativeLayout = PresentSrcKHR`, which is a legal restore target. Now it is `Undefined`. If the first use of the back buffer in a frame is a copy (for example `commandList.Copy(texture, Presenter.BackBuffer)` before any `ResourceBarrierTransition`), the restore barrier has `newLayout = VK_IMAGE_LAYOUT_UNDEFINED`, which the spec forbids (VUID-VkImageMemoryBarrier-newLayout-01198).

I found no engine code that does this today, but user code can. Two possible fixes:
- in the copy paths, when the tracked layout is `Undefined`, skip the restore and record the transfer layout as the new tracked state, or
- restore to a valid default such as `ColorAttachmentOptimal` and update the tracked state to match.

If you prefer to keep it out of this PR, please at least document it as a known limitation here.

sources/engine/Stride.Graphics/Vulkan/CommandList.Vulkan.cs
```diff
@@ -543,6 +554,42 @@ public unsafe void ResourceBarrierTransition(GraphicsResource resource, BarrierL
...
+                    sourceStages = BarrierMapping.ToVkPipelineStageFlags(fromLayout);
+                }
+                else
+                {
+                    oldAccessMask = buffer.NativeAccessMask;
+                    sourceStages = buffer.NativePipelineStageMask;
+                }
+
+                var newAccessMask = BarrierMapping.ToVkAccessFlags(newLayout);
+                var newStages = BarrierMapping.ToVkPipelineStageFlags(newLayout);
```
xen2, line 576: `ToVkPipelineStageFlags(BarrierLayout.ShaderResource)` returns `FragmentShader | ComputeShader`, with no `VertexShader`. For buffers this matters more than for textures: a compute-written structured buffer read in a vertex shader (GPU skinning, instancing, particles) is a common pattern, and this barrier would not name that stage.

This PR already fixes the same gap for textures in `Texture.Vulkan.cs` (`VertexInput` -> `VertexShader`), and `Buffer.Vulkan.cs` uses `VertexShader | FragmentShader | ComputeShader` for its static mask. I suggest adding `VertexShader` to the `ShaderResource` entry in `BarrierMapping.ToVkPipelineStageFlags` so all three agree.

sources/engine/Stride.Graphics/Vulkan/GraphicsAdapterFactory.Vulkan.cs
```diff
@@ -207,9 +215,10 @@ public unsafe GraphicsAdapterFactoryInstance(bool enableValidation)
                 VK_KHR_XCB_SURFACE_EXTENSION_NAME,
                 VK_EXT_METAL_SURFACE_EXTENSION_NAME,
                 VK_KHR_PORTABILITY_ENUMERATION_EXTENSION_NAME,
-                VK_EXT_DEBUG_UTILS_EXTENSION_NAME
+                VK_EXT_DEBUG_UTILS_EXTENSION_NAME,
+                VK_EXT_VALIDATION_FEATURES_EXTENSION_NAME
```
xen2, line 219: This entry looks unused, please double-check.
`availableExtensionNames` is never checked for it, and as the description says, the query without a layer name never lists this extension anyway.
The extension is added to `desiredExtensionNames` directly further down.

If so, please revert this entry (and adjust count from 9 to 8 if not using Span<>, but shouldn't be necessary)

sources/engine/Stride.Graphics.Tests.11_0/TestBufferBarrier.cs
```diff
@@ -0,0 +1,94 @@
...
+
+        commandList.ResourceBarrierTransition(sharedBuffer, BarrierLayout.ShaderResource);
+        commandList.ResourceBarrierTransition(resultBuffer, BarrierLayout.UnorderedAccess);
+        readEffect.Parameters.Set(BufferBarrierReadTestShaderKeys.Input, sharedBuffer);
+        readEffect.Parameters.Set(BufferBarrierReadTestShaderKeys.Result, resultBuffer);
+        ((RendererBase)readEffect).Draw(renderDrawContext);
+
+        var values = resultBuffer.GetData<uint>(commandList);
+
+        Assert.Equal(ElementCount, values.Length);
```
xen2, line 82: As the remarks say, this cannot fail when the barrier is missing, so in practice it guards against the old `NotImplementedException`. That is still useful. Could CI run this test (or the whole Vulkan suite) once with `STRIDE_VULKAN_SYNC_VALIDATION=1`, so a missing barrier fails the test? It would need the debug callback to turn sync hazards into a test failure, so maybe a follow-up.

sources/engine/Stride.Graphics/Vulkan/GraphicsAdapterFactory.Vulkan.cs
```diff
@@ -207,9 +215,10 @@ public unsafe GraphicsAdapterFactoryInstance(bool enableValidation)
...
             };
-            var supportedExtensions = new Span<VkUtf8String>(supportedExtensionNames, 8);
+            var supportedExtensions = new Span<VkUtf8String>(supportedExtensionNames, 9);
```
xen2, line 221: Let's use `Span<VkUtf8String> supportedExtensionNames ` instead of `var supportedExtensionNames` so that this line is not necessary and no hardcoded 8/9.

sources/engine/Stride.Graphics/Vulkan/SwapChainGraphicsPresenter.Vulkan.cs
sasvdw: Fixed in 58fd2e6be, option (a). `AdoptTransferLayoutIfUndefined` runs before the restore barrier in `Copy`, `CopyRegion` and `UpdateSubResource`: when the tracked layout is `Undefined`, the transfer layout becomes the tracked state (`NativeLayout`, access and stage masks, `LayoutTracker`, the per-CB map), so the restore is a same-layout transfer barrier and the next `ResourceBarrierTransition` names the transfer layout as its source.

Two findings while reproducing it:

- Through the game loop the case is unreachable: `GraphicsPresenter.BeginDraw` (66b29bd43) transitions the back buffer to `RenderTarget` before user `Draw` runs, so a `Copy` into the back buffer as the frame's first command already sees `ColorAttachmentOptimal`. It needs a caller that drives the presenter without `BeginDraw`.
- Forcing a texture's tracked layout to `Undefined` and issuing `Copy` (as source and as destination), `CopyRegion` and `SetData` reproduced `newLayout is VK_IMAGE_LAYOUT_UNDEFINED` in all three paths on the previous head. With the commit, the same run is clean, with and without synchronization validation.

Only the swapchain image can be `Undefined` here; every other texture starts from `GetInitialBarrierLayout()`, which never returns it.

sources/engine/Stride.Graphics/Vulkan/CommandList.Vulkan.cs
sasvdw: Fixed in 4dcaf2e5b: `ShaderResource` now maps to `VertexShader | FragmentShader | ComputeShader`, matching the static masks in `Buffer.Vulkan.cs` and `Texture.Vulkan.cs`. It applies to textures too, since `ResourceBarrierTransition` stores the destination stages in `NativePipelineStageMask`.

sources/engine/Stride.Graphics/Vulkan/GraphicsAdapterFactory.Vulkan.cs
sasvdw: Removed in 58097fe84. It was unused: nothing consulted `availableExtensionNames` for it.

sources/engine/Stride.Graphics/Vulkan/GraphicsAdapterFactory.Vulkan.cs
sasvdw: Done in 58097fe84: `Span<VkUtf8String> supportedExtensionNames = stackalloc …`, and the `new Span<>(…, 9)` line is gone.

sources/engine/Stride.Graphics.Tests.11_0/TestBufferBarrier.cs
sasvdw: The callback side already exists: `GameTestBase.RunGameTest` collects every `DebugLogModule` error during the run and fails the test with them, and synchronization hazards arrive at error severity. With the access-mask fix reverted and `STRIDE_VULKAN_SYNC_VALIDATION=1`, `TestHammersley` and `TestRadiancePrefilteringGgxFaceContinuity` fail on `READ_AFTER_WRITE` at `vkCmdBeginRendering`; with the fix, both graphics suites pass under the switch with no hazard.

What CI lacks is only the environment variable. I will open a follow-up PR adding a Vulkan leg with it to `test-windows-game.yml`, after running the full GPU solution filter under the switch so the leg is scoped to what is actually clean.

sasvdw: Opened #3457: sets `STRIDE_VULKAN_SYNC_VALIDATION=1` on the existing Vulkan leg of `test-windows-game.yml`, no new matrix leg. All six suites in the GPU solution filter pass under it on this branch.

PR #3325, opened 2026-08-04
fix: parse and format shader numbers with the invariant culture

sasvdw:
# PR Details

**Summary**

The shader tooling read and wrote numbers with the culture of the machine. Every render test that carries
a float parameter failed on a comma-decimal locale. The SDSL front end also parsed generic arguments with
the current culture, although every producer writes them invariantly.

**Test harness**

A test header is authored in the shader source, so it always uses a period as the decimal separator:

```
// PSMain(ExpectedResult=#7F7F7F7F, stream.EXTRA_COLOR=(0.498 0.498 0.498 0.498))
```

`FrameRenderer` read those values with `float.Parse(value)`, which uses the current culture. On an en-ZA
machine the decimal separator is a comma, so that call throws `FormatException: The input string '0.498'
was not in a correct format`. This failed 9 render tests that pass on CI.

- `FrameRenderer.cs` and `FrameRenderer.D3D11.cs`: parse with `CultureInfo.InvariantCulture`, as the
  neighboring vector case already did. Also pass `NumberStyles.Float` or `NumberStyles.Integer`.

The styles matter. The default set allows a group separator, so `float.Parse("0,498", invariant)` returns
**498** instead of a `FormatException`. A group separator has no meaning here. Rejecting it turns a silent
wrong value into a clear error.

**Compiler**

@Ethereal77 asked for this in review. Parsing a programming language must not depend on the culture of
the machine.

- `Builder.Class.cs:272` and `:275`: `GenericResolverFromValues.TryResolveGenericValue` read generic
  arguments with the current culture. Every producer writes them invariantly. See
  `ShaderClassSource.ConvertGenericArgToString`, which documents the string as the SDSL-parseable
  representation. A float generic argument therefore failed to parse on a comma-decimal locale. The float
  also takes `NumberStyles.Float`, for the reason above.
- `Builder.Class.cs:584`: the resolved generic value becomes macro text that the compiler substitutes into
  shader source. It now goes through `ConvertGenericArgToString` instead of `ToString`, so the text stays
  SDSL-parseable.
- `NumberParsers.cs:37` and `:43`, `SemanticAnalyzer.cs:19`, `BuiltinProcessor.cs:30`: the integer literal
  parser, the semantic index parser and the location decoration parser. A sweep for the same mistake found
  these. They read digits only, so they could not misbehave in practice. They are aligned so the front end
  is consistent.
- `Context.Constants.cs`: name float constants with the invariant culture. `$"{lit.Type}_{lit.Value}"`
  wrote `float_0,5` into `OpName` instead of `float_0.5`, so one shader source compiled to different
  SPIR-V bytes on different machines.

  This last change makes the output deterministic. It does not fix a defect. Nothing reads the name of a
  constant. SPIRV-Cross inlines constant values into the generated HLSL and formats them itself, so the
  name never becomes an identifier. I checked this on an en-ZA machine. I compiled a shader with float
  literals and read the generated HLSL. It contains `float a = 0.5f;` and no `float_` name. The only
  effect is that compiled effects are not byte-identical between machines, so their caches do not transfer.

**Motivation and context** — I found the test failures while I ran `Stride.Shaders.Tests` locally on an
en-ZA machine.

## Related Issue

None. I found this in a local test run.

## Types of changes

## Checklist

**Validation status**

- *Tests*: `Stride.Shaders.Tests` on an en-ZA machine gives **10 failures before and 1 after**, with 524
  passed. The 9 fixed tests are `RenderTest1` for `StreamVSToPS`, `StreamPassthroughPS`, `StreamParameter`,
  `GenericsSemantic`, `GenericsExternal`, `CompositionExternalCBuffer` and
  `CompositionInheritanceStageIndirectAndDirect`, plus `StreamOutTest1` for `StreamGS` and
  `StreamTessellation`. The compiler changes leave that result unchanged at 524 passed and 1 failed.
- *Not covered*: `RenderTest1(GenericsEffect)` still fails with a `NullReferenceException` in
  `GenericIdentifier.ResolveExternalShader` (`Literals.cs:640`). It fails the same way on a clean `master`
  checkout, so it is a separate bug that already exists and has nothing to do with the culture. I can file
  it separately.
- *No new tests*: the existing render tests prove the harness fix, because they pass on a comma-decimal
  locale only with this change. The compiler paths are either unreachable today or read digits only, so a
  test cannot assert anything that the suite does not already cover. A regression test for the harness
  would need to set the thread culture. Tell me if you want one.
- *Editor*: not exercised. This changes the shader test harness, the SDSL front end and the name of a
  constant.

### DIFF
--- a/sources/shaders/Stride.Shaders.Tests/FrameRenderer.D3D11.cs
+++ b/sources/shaders/Stride.Shaders.Tests/FrameRenderer.D3D11.cs
@@ -397,7 +397,7 @@ float4 main(vs_out input) : SV_TARGET {
                     });
 
                     // Also create the vertex and bind it right away
-                    var floatValues = parameter.Value.TrimStart('(').TrimEnd(')').Split(' ', StringSplitOptions.TrimEntries).Select(x => float.Parse(x)).ToArray();
+                    var floatValues = parameter.Value.TrimStart('(').TrimEnd(')').Split(' ', StringSplitOptions.TrimEntries).Select(x => float.Parse(x, NumberStyles.Float, CultureInfo.InvariantCulture)).ToArray();
                     bufferDesc = new BufferDesc
                     {
                         ByteWidth = (uint)(sizeof(float) * floatValues.Length), // up to 4 floats
@@ -948,17 +948,17 @@ float4 main(vs_out input) : SV_TARGET {
                 foreach (var comp in TestHeaderParser.SplitArgs(value))
                 {
                     if (type.Type == EffectParameterType.Float)
-                        *((float*)&cbufferDataPtr[offset + compIndex * sizeof(float)]) = float.Parse(comp, CultureInfo.InvariantCulture);
+                        *((float*)&cbufferDataPtr[offset + compIndex * sizeof(float)]) = float.Parse(comp, NumberStyles.Float, CultureInfo.InvariantCulture);
                     else if (type.Type == EffectParameterType.Int)
-                        *((int*)&cbufferDataPtr[offset + compIndex * sizeof(int)]) = int.Parse(comp, CultureInfo.InvariantCulture);
+                        *((int*)&cbufferDataPtr[offset + compIndex * sizeof(int)]) = int.Parse(comp, NumberStyles.Integer, CultureInfo.InvariantCulture);
                     compIndex++;
                 }
                 break;
             case { Type: EffectParameterType.Int }:
-                *((int*)&cbufferDataPtr[offset]) = int.Parse(value);
+                *((int*)&cbufferDataPtr[offset]) = int.Parse(value, NumberStyles.Integer, CultureInfo.InvariantCulture);
                 break;
             case { Type: EffectParameterType.Float }:
-                *((float*)&cbufferDataPtr[offset]) = float.Parse(value);
+                *((float*)&cbufferDataPtr[offset]) = float.Parse(value, NumberStyles.Float, CultureInfo.InvariantCulture);
                 break;
             default:
                 throw new NotImplementedException();
--- a/sources/shaders/Stride.Shaders.Tests/FrameRenderer.cs
+++ b/sources/shaders/Stride.Shaders.Tests/FrameRenderer.cs
@@ -33,10 +33,10 @@ public abstract class FrameRenderer(uint width = 800, uint height = 600, byte[]?
                 }
                 break;
             case { Type: EffectParameterType.Int }:
-                *((int*)&cbufferDataPtr[offset]) = int.Parse(value);
+                *((int*)&cbufferDataPtr[offset]) = int.Parse(value, NumberStyles.Integer, CultureInfo.InvariantCulture);
                 break;
             case { Type: EffectParameterType.Float }:
-                *((float*)&cbufferDataPtr[offset]) = float.Parse(value);
+                *((float*)&cbufferDataPtr[offset]) = float.Parse(value, NumberStyles.Float, CultureInfo.InvariantCulture);
                 break;
             default:
                 throw new NotImplementedException();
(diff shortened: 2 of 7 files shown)


Ethereal77:
Good catch! Parsing a programming language must always be culture-invariant.

> GenericResolverFromValues.TryResolveGenericValue
(Builder.Class.cs:272 and :275) parses generic values with int.Parse and float.Parse under the current culture

Can you also fix that, please? No reason to leave that mistake there knowing it is wrongly parsing values, even if currently not used. Else, we could forget, and then it becomes a time bomb for when this path is exercised in the future.

sources/shaders/Stride.Shaders.Tests/FrameRenderer.cs
```diff
@@ -33,10 +33,10 @@ protected static unsafe void FillCBufferData(string value, EffectTypeDescription
                 }
                 break;
             case { Type: EffectParameterType.Int }:
-                *((int*)&cbufferDataPtr[offset]) = int.Parse(value);
+                *((int*)&cbufferDataPtr[offset]) = int.Parse(value, NumberStyles.Integer, CultureInfo.InvariantCulture);
```
Kryptos-FR, line 36: Why does it need both? Wouldn't `CultureInfo.InvariantCulture` be enough? `NumberStyles.Integer` should be implicit.

sources/shaders/Stride.Shaders.Tests/FrameRenderer.D3D11.cs
```diff
@@ -948,17 +948,17 @@ private static unsafe void FillData(string value, EffectTypeDescription type, in
                 foreach (var comp in TestHeaderParser.SplitArgs(value))
                 {
                     if (type.Type == EffectParameterType.Float)
-                        *((float*)&cbufferDataPtr[offset + compIndex * sizeof(float)]) = float.Parse(comp, CultureInfo.InvariantCulture);
+                        *((float*)&cbufferDataPtr[offset + compIndex * sizeof(float)]) = float.Parse(comp, NumberStyles.Float, CultureInfo.InvariantCulture);
```
Kryptos-FR, line 951: Same here. What does it add? The number styles is implicit given the fact it's using `float.Parse`.

sources/shaders/Stride.Shaders.Tests/FrameRenderer.cs
sasvdw: InvariantCulture uses `,` as a thousands separator. I've gone through existing shaders and hadn't seen any cases where commas are used. By including the `NumberStyles.Integer/Float` it ensures that a `FormatException` is raised if thousands separators are included.

There's a rather substantial piece of MSDN documentation on the `NumberStyles` enum:
https://learn.microsoft.com/en-us/dotnet/api/system.globalization.numberstyles?view=net-10.0

The alternative would be that we build our own helper that provides valid numbstyles for shader parsing.

Kryptos-FR: The call to `int.Parse` without the `styles` parameter does call the overload with that same `NumberStyles.Integer`. So it's a no-op and redundant to set it.

see https://github.com/dotnet/dotnet/blob/090e98ab4beb4ba2eb972031ff1e4e59f2057287/src/runtime/src/libraries/System.Private.CoreLib/src/System/Int32.cs#L149

My remark was that it was unnecessarily noisy, making it look like it was a special case when it's actually not.

sources/shaders/Stride.Shaders.Tests/FrameRenderer.D3D11.cs
Kryptos-FR: This seems to make sense here since by default float do parse thousand separators.

sources/shaders/Stride.Shaders.Tests/FrameRenderer.cs
xen2: Sorry I thought it was resolved with previous comment and I merged.
I can still take further changes in another PR if necessary (but if just an extra unused parameter, maybe not a big deal).

Kryptos-FR: No it's fine. I also wrongly thought that it was unnecessary for `float.Parse` (because `NumberStyles.Float` should have been the default), but turns out the overload adds the thousand separator (probably for legacy reasons).

see https://github.com/dotnet/dotnet/blob/090e98ab4beb4ba2eb972031ff1e4e59f2057287/src/runtime/src/libraries/System.Private.CoreLib/src/System/Single.cs#L395

PR #3327, opened 2026-08-05
ContentManager.Load: better message when asset is not found (and throw exception in all cases)

xen2:
# PR Details

As discussed on Discord, I tried to make a better error message when an asset is not found.

Case 1: root-differing candidates ("did you mean"):
```
The asset 'StrideDefaultFont' could not be found. Did you mean:
  '/MyLib/StrideDefaultFont'
  '/Stride.Engine/StrideDefaultFont'
```

Case 2: same name in other folders (capped at 5):
```
The asset 'WrongFolder/Ground' could not be found. Assets with the same name exist at:
  '/A/Ground'
  '/B/Ground'
  '/C/Ground'
  '/D/Sub/Ground'
  '/E/Ground'
  (and 2 more)
```

I also made it a proper exception (previous code was a bit weird: log error and caught exception if debugger attached)
It didn't feel right that `ContentManager.Load` could return null with a simple log error (it was highly unlikely user was checking for null).
We could later add a `ContentManager.TryLoad` if really necessary, but `ContentManager.Exists` should be enough for now.

## Types of changes

## Checklist

- [x] My change requires a change to the documentation (done in XML doc, need a comment in release notes)

### DIFF
--- a/sources/core/Stride.Core.Serialization/Serialization/Contents/ContentManager.cs
+++ b/sources/core/Stride.Core.Serialization/Serialization/Contents/ContentManager.cs
@@ -1,7 +1,6 @@
 // Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net) and Silicon Studio Corp. (https://www.siliconstudio.co.jp)
 // Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
 
-using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Reflection;
 using Stride.Core.Diagnostics;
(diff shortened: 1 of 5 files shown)


IXLLEGACYIXL:
"(and 2 more)"
why not list all?

xen2:
> "(and 2 more)" why not list all?

I was afraid if loading a generic name like `Texture` or `Scene` and it matches hundreds of them in subfolder (i.e. `Scene1\Scene` `Scene2\Scene` and so on), it wouldn't be a great experience.
So it is limited to 5 (in practice I think 95% of cases there will be only 1 match).
For the context, it's just meant as a hint that needs developer intervention anyway. If this happens, something is seriously broken, and usually the right asset to load is obvious just looking at the assets.

PR #3336, opened 2026-08-08
Vulkan: switch to Vulkan 1.3 and use dynamic rendering

xen2:
## PR Details

Moves the Vulkan backend to Vulkan 1.3 core:

- **Require Vulkan 1.3**: clear error at device creation otherwise. Excludes almost nothing in practice: timeline semaphores already made 1.2 the effective floor, and Android drivers jumped from 1.1 straight to 1.3.
  Note that we could have emulated timeline semaphore on 1.1 using https://github.com/KhronosGroup/Vulkan-ExtensionLayer but we're talking about pre-2022 devices. Let's drop support and if some game/user really need such old devices for a large scale game, we could reconsider.
- **Dynamic rendering**: `vkCmdBeginRendering` replaces render pass objects, deleting the VkRenderPass/VkFramebuffer machinery. Pipelines only declare attachment formats, and the render pass instance now stays open across pipeline changes. Also unblocks the upcoming explicit resource transition work.
- **Aggregate feature structs** (`VkPhysicalDeviceVulkan11/12/13Features`) instead of many individual structs.
- **`vkQueueSubmit2`**: simpler submits, and fixes wait stage masks that were too weak.

Net ~200 lines removed.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Graphics.Tests/TestFramebufferAttachmentGuard.cs
+++ b/sources/engine/Stride.Graphics.Tests/TestFramebufferAttachmentGuard.cs
@@ -19,12 +19,12 @@ public class TestFramebufferAttachmentGuard : GraphicTestGameBase
     }
 
     /// <summary>
-    /// The Vulkan backend's debug guard turns a framebuffer/render-pass attachment-count mismatch into a clear
-    /// exception instead of a device loss. Verify it fires when more render targets are bound than the active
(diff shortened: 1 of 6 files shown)


xen2:
Recreated as #3337 (to try stacked PR, needs to be on origin rather than fork)

PR #3340, opened 2026-08-08
fix(msbuild): Stride.AssetCompiler.targets - use MSBuild::NormalizePath to normalize paths

VaclavElias:
#### PR Classification
Bug fix to ensure correct path handling in asset build processes.

#### PR Summary
Replaced string concatenation with `System.IO.Path.Combine` for `StrideCompileAssetBuildPath` and `StrideCompileAssetUpToDateCheckFileBase` in `Stride.AssetCompiler.targets`, ensuring intermediate output paths are handled correctly for both normal projects and file-based apps (`dotnet run app.cs`), whose `obj`/`bin` paths are already rooted under the SDK's temp cache.
- `Stride.AssetCompiler.targets`: use `Path.Combine` instead of concatenation for `StrideCompileAssetBuildPath` and `StrideCompileAssetUpToDateCheckFileBase`; added explanatory comments

#### Testing notes

This PR fixes `StrideCompileAssetBuildPath` and `StrideCompileAssetUpToDateCheckFileBase` to use
`Path.Combine` instead of string concatenation, so absolute intermediate paths (e.g. file-based
apps via `dotnet run app.cs`, whose obj/bin live under the SDK's temp cache) are handled correctly
instead of producing a malformed path.

Verified on Windows. `Path.Combine`/`Path.IsPathRooted` rootedness semantics can differ slightly
across platforms, so it would be good to double check normal + file-based app builds on
Linux/macOS, if a maintainer has the setup handy.


VaclavElias:
References: https://learn.microsoft.com/en-us/dotnet/core/sdk/file-based-apps

Additional context, this change will allow to run Stride (code-only) from any C# file e.g. `Program.cs` without a project file `csproj` like the example below.

Without this PR, we have to add also these two lines

```
#:property BaseIntermediateOutputPath=obj\
#:property OutputPath=bin\
``` 

```csharp
#:package Stride.CommunityToolkit.Bepu@1.0.0-dev
#:package Stride.CommunityToolkit.Skyboxes@1.0.0-dev
#:package Stride.CommunityToolkit.Windows@1.0.0-dev

using Stride.CommunityToolkit.Bepu;
using Stride.CommunityToolkit.Engine;
using Stride.CommunityToolkit.Rendering.ProceduralModels;
using Stride.CommunityToolkit.Skyboxes;
using Stride.Core.Mathematics;
using Stride.Engine;

using var game = new Game();

game.Run(start: Start);

void Start(Scene rootScene)
{
    game.SetupBase3DScene();
    game.AddSkybox();

    var entity = game.Create3DPrimitive(PrimitiveModelType.Capsule);
    entity.Transform.Position = new Vector3(0, 8, 0);
    entity.Scene = rootScene;
}
```

sources/assets/Stride.AssetCompiler/build/Stride.AssetCompiler.targets
```diff
@@ -65,12 +65,17 @@
...
     <StrideCompileAssetBuildPath Condition="'$(StrideCompileAssetBuildRedirected)' == 'true'">$([System.IO.Path]::GetFullPath($([System.IO.Path]::Combine($([System.IO.Path]::GetDirectoryName('$(StrideCurrentPackagePath)')), 'obj/stride/assetbuild/data'))))</StrideCompileAssetBuildPath>
-    <StrideCompileAssetBuildPath Condition="'$(StrideCompileAssetBuildPath)' == ''">$(ProjectDir)$(BaseIntermediateOutputPath)stride\assetbuild\data</StrideCompileAssetBuildPath>
+    <!-- Combine rather than concatenate: BaseIntermediateOutputPath/IntermediateOutputPath are
+         relative to the project for a normal project, but absolute for a file-based app
+         (dotnet run app.cs), whose obj/bin live under the SDK's temp cache. Path.Combine returns
+         the second path unchanged when it is already rooted, so the relative case is unaffected. -->
+    <StrideCompileAssetBuildPath Condition="'$(StrideCompileAssetBuildPath)' == ''">$([System.IO.Path]::Combine('$(ProjectDir)', '$(BaseIntermediateOutputPath)stride\assetbuild\data'))</StrideCompileAssetBuildPath>
```
xen2: Let's use this instead:
`$([MSBuild]::NormalizePath('$(ProjectDir)', '$(BaseIntermediateOutputPath)', 'stride\assetbuild\data'))`
It would match other use (i.e. line 65) and it applies GetFullPath which resolves `..\` from the resulting path and OK with missing trailing slash).

xen2 (changes requested):

sources/assets/Stride.AssetCompiler/build/Stride.AssetCompiler.targets
VaclavElias: Thanks, I will test it locally and report back.

VaclavElias: Done! Tested! It works as well in my case and I was able to launch Game Studio.

Kryptos-FR: Fix the PR description that still mentions `System.IO.Path.Combine`

VaclavElias: Thanks! Also done!

xen2 (approved):

PR #3341, opened 2026-08-08
fix: read by direction when resampling the prefiltered radiance highest level

### LINKED ISSUE #3334: Prefiltered radiance mip 0 rescales each cube face on its own, so face edges lose their neighbors

**Version**: `master` (`81b17f0e2`)

**Platform(s)**: All. This is backend independent.

**Describe the bug**

`RadiancePrefilteringGGXNoCompute` fills mip 0 of the prefiltered cubemap from the source when
`DoNotFilterHighestLevel` is set. A straight copy serves that only while the sizes, the mip count and the
format all agree (`RadiancePrefilteringGGXNoCompute.cs:103`). Everything else falls to `ImageScaler` over a
single face view (`:110-117`). That rescale clamps at the face border, so the border texels never take the
neighboring face into account and mip 0 stops being continuous around every face edge.

**How it is reached**

`SkyboxGenerator` builds the output at `SpecularCubeMapSize`, which defaults to 64
(`SkyboxGenerator.cs:144`), and fixes the filtering format at `:149`:

```csharp
var filteringTextureFormat = skyboxTexture.Format.IsHDR ? skyboxTexture.Format : PixelFormat.R8G8B8A8_UNorm;
```

A low dynamic range skybox therefore compiles to a block format while the filtering format stays
`R8G8B8A8_UNorm`. The two never match, so the rescale runs. `TextureAsset.IsCompressed` defaults to `true`
(`TextureAsset.cs:83`), which makes this the ordinary case rather than an unusual one.

High dynamic range sources avoid it. The texture compiler never compresses them, because
`TextureHelper.cs:295-299` keeps the input format and carries a TODO about BC6H. So an HDR skybox stays at
`R16G16B16A16_Float`, matches the filtering format and takes the copy.

Two other ways in, both ordinary. A skybox texture with `GenerateMipmaps: false` and a source larger than
`SpecularCubeMapSize` fails the mip test. A source smaller than `SpecularCubeMapSize` fails the width test.

Every skybox texture in this repository sets `IsCompressed: false` by hand, against the default. Three out
of three, counting the two templates and the one in `Stride.Graphics.Tests.10_0`. That reads like people
have already stepped around this.

**Measured**

I ran the filter over the `GraceCathedral` cubemap into a 64 pixel output, which is the
`SpecularCubeMapSize` default, and changed only the output format, since that is what selects the branch.
For each result I measured the mean absolute luminance step across the 12 shared face edges, against the
step between neighboring texels inside a face.

| output format | mip | branch | seam | within-face | ratio |
|---|---|---|---|---|---|
| `R16G16B16A16_Float` | 0 | `CopyRegion` | 0.77 | 123.66 | 0.01 |
| `R8G8B8A8_UNorm` | 0 | **`ImageScaler`** | 3.46 | 4.03 | **0.86** |
| `R8G8B8A8_UNorm` | 1 | GGX filter | 1.44 | 4.53 | 0.32 |

Read the last two rows together. They share a texture, a format and a content path. The level the rescale
produces steps by 0.86, and the level the filter produces directly below it steps by 0.32.

The effect is real but moderate. At 0.86 the step across a face edge is about the size of an ordinary step
between neighboring texels, so this softens the boundary rather than drawing a hard seam.

**Expected behavior**

Mip 0 stays continuous around face edges, like every other level.

**Note for anybody fixing this**

The obvious fix does not work, and it fails quietly. The filter cannot serve mip 0. At roughness 0 the GGX
distribution returns a zero probability density, so the solid angle goes to infinity and the sample mip
clamps to `MipmapCount` (`RadiancePrefilteringGGXNoComputeShader.sdsl:41-44`). Every texel then reads the
smallest mipmap and the whole level becomes one flat color. I measured a within-face variation of 0.00
that way. A continuity check passes on that output, because a flat level has no seam to find.

What works is to read the source by direction rather than through a single face view, so a border texel
crosses the edge the way the hardware samples a cubemap.

**A second, opposite inconsistency**

The compute path checks the size but not the format before `CopyRegion`
(`RadiancePrefilteringGGX.cs:92`). The no-compute path checks the format
(`RadiancePrefilteringGGXNoCompute.cs:103`). One of the two is wrong. The compute path also falls into the
filter when its check fails, which walks into the roughness 0 problem above.

**Relationship to #3328**

This is not the cause of the residual seam in #3328. Two measurements say so. The residual appears in the
compute path as well, which has no per-face structure. `TestRadiancePrefilteringGgx` never reaches this
branch, and never displays mip 0.

sasvdw:
## PR Details

**Summary**

Mip 0 of the prefiltered cubemap lost continuity around every face edge whenever the copy shortcut did not
apply, because the fallback rescaled each face on its own.

**Description**

`RadiancePrefilteringGGXNoCompute` fills mip 0 from the source when `DoNotFilterHighestLevel` is set. The
copy at `:103` serves that only while the sizes, the mip count and the format agree. Everything else went
through `ImageScaler` over a single face view, which clamps at the face border, so the border texels never
saw the neighboring face.

This replaces that fallback with `CubemapFaceResampleShader`, which reads the source by direction. A
border texel then crosses the face edge the way the hardware samples a cubemap.

Falling through to the existing filter would look like the tidier fix, and it is what the compute path
appears to do, but it does not work. At roughness 0 the GGX distribution returns a zero probability
density, the solid angle goes to infinity, and the sample mip clamps to `MipmapCount`
(`RadiancePrefilteringGGXNoComputeShader.sdsl:41-44`). Every texel reads the smallest mipmap and the level
becomes one flat color. That is why mip 0 carries a special case at all.

Measured over the 12 shared face edges, against the step between neighboring texels inside a face:

| | within-face | seam ratio |
|---|---|---|
| before, `ImageScaler` | 4.03 | 0.86 |
| falling through to the filter | 0.00 | 0.00 |
| after, resample by direction | 5.16 | 0.11 |

The middle row is the trap. A flat level reports no seam, so a continuity check alone would call it fixed.
The test asserts the level still carries an image for that reason.

**Motivation and context** — see #3334.

## Related Issue

Fixes #3334.

## Types of changes

## Checklist

**Validation status**

- *Tests*: `TestRadiancePrefilteringGgxFaceContinuity` drives the filter with an output format that differs
  from the source, which is the case `SkyboxGenerator` creates for a low dynamic range skybox. It measures
  the step across the 12 shared face edges and asserts the level still carries an image. Red at 0.86 before
  the change, green at 0.11 after, verified both ways.
- *No gold image moves.* Every cubemap in the tree takes the copy path, so nothing that is captured today
  reaches the changed branch. `Stride.Graphics.Tests.11_0` passes 5 of 5, including
  `TestRadiancePrefilteringGgx`. The skybox cases in `Stride.Graphics.Tests.10_0` pass 3 of 3.
- *Not covered*: the same roughness 0 problem sits in the compute path if its size check ever fails, and
  that path also skips the format check the no-compute path makes. Both are described in the issue and left
  alone here.
- *Editor*: not exercised.

### DIFF
--- a/sources/engine/Stride.Graphics.Tests.11_0/TestRadiancePrefilteringGgxFaceContinuity.cs
+++ b/sources/engine/Stride.Graphics.Tests.11_0/TestRadiancePrefilteringGgxFaceContinuity.cs
@@ -1,0 +1,137 @@
+// Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net)
+// Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using System;
+
+using Xunit;
+
+using Stride.Core.Mathematics;
+using Stride.Rendering;
+using Stride.Rendering.ComputeEffect.GGXPrefiltering;
+
+namespace Stride.Graphics.Tests;
+
+public class TestRadiancePrefilteringGgxFaceContinuity : GraphicTestGameBase
+{
+    // The 12 edges cube faces share, in Direct3D face order, with the orientation each pair meets at.
+    private static readonly (int FaceA, Edge EdgeA, int FaceB, Edge EdgeB, bool Reversed)[] SharedEdges =
+    {
+        (0, Edge.Bottom, 3, Edge.Right,  false), (0, Edge.Left,  4, Edge.Right,  false),
+        (0, Edge.Right,  5, Edge.Left,   false), (0, Edge.Top,   2, Edge.Right,  true),
+        (1, Edge.Bottom, 3, Edge.Left,   true),  (1, Edge.Left,  5, Edge.Right,  false),
+        (1, Edge.Right,  4, Edge.Left,   false), (1, Edge.Top,   2, Edge.Left,   false),
+        (2, Edge.Bottom, 4, Edge.Top,    false), (2, Edge.Top,   5, Edge.Top,    true),
+        (3, Edge.Bottom, 5, Edge.Bottom, true),  (3, Edge.Top,   4, Edge.Bottom, false),
+    };
+
+    private enum Edge { Top, Bottom, Left, Right }
+
+    private const int OutputSize = 64;
+
+    private double seamRatio;
+
+    protected override void RegisterTests()
+    {
+        base.RegisterTests();
+
+        FrameGameSystem.Draw(MeasureFaceContinuity);
+    }
+
+    /// <summary>
+    /// A cubemap discretizes a function over a sphere, so neighboring texels on either side of a shared
+    /// face edge must stay as close as neighboring texels within a face. The output format differs from
+    /// the source here, which is what a low dynamic range skybox produces in <c>SkyboxGenerator</c>.
+    /// </summary>
+    private void MeasureFaceContinuity()
+    {
+        var commandList = GraphicsContext.CommandList;
+        var input = Content.Load<Texture>("CubeMap");
+
+        using var output = Texture.New2D(GraphicsDevice, OutputSize, OutputSize, MathUtil.Log2(OutputSize),
+            PixelFormat.R8G8B8A8_UNorm, TextureFlags.ShaderResource | TextureFlags.RenderTarget, 6);
+
+        var filter = new RadiancePrefilteringGGXNoCompute(RenderContext.GetShared(Services))
+        {
+            RadianceMap = input,
+            PrefilteredRadiance = output,
+            MipmapGenerationCount = MathUtil.Log2(OutputSize),
+        };
+        filter.Draw(new RenderDrawContext(Services, RenderContext.GetShared(Services), GraphicsContext));
+
+        var faces = new double[6][,];
+        for (var face = 0; face < faces.Length; face++)
+            faces[face] = ReadLuminance(output, commandList, face);
+
+        var seam = 0.0;
+        foreach (var (faceA, edgeA, faceB, edgeB, reversed) in SharedEdges)
+        {
+            var a = ReadEdge(faces[faceA], edgeA);
+            var b = ReadEdge(faces[faceB], edgeB);
+            if (reversed)
+                Array.Reverse(b);
+
+            var total = 0.0;
+            for (var i = 0; i < OutputSize; i++)
+                total += Math.Abs(a[i] - b[i]);
+            seam += total / OutputSize;
+        }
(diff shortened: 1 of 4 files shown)


sources/engine/Stride.Graphics.Tests.11_0/TestRadiancePrefilteringGgxFaceContinuity.cs
```diff
@@ -0,0 +1,144 @@
...
+                _ => face[i, OutputSize - 1],
+            };
+        }
+        return result;
+    }
+
+    [SkippableFact]
+    public void HighestLevelIsContinuousAcrossFaces()
+    {
+        SkipTestForGraphicPlatform(GraphicsPlatform.Vulkan);
```
Ethereal77: Why is this skipped for Vulkan?

sasvdw: It was a bad copy on my end. Thought I was still working with the code from another branch where I had the fixes living, for skipping some related tests and ended up duplicating the test suite without a proper look.

xen2:
Thanks, the change and the test look good.

One follow-up: this resample path is mostly reached by sources without mipmaps (a skybox made from a 2D panorama always has a single mip level).

Missing mipmaps cause two problems:

- Mip 0 is resampled from a much larger image with one bilinear tap, so it aliases.
- The GGX filter clamps its sampling level to `MipLevelCount - 1`, so every roughness level reads the full resolution texture and gets noisy.

I made a small commit that builds the mip chain in `SkyboxGenerator` when the source has none. Then your resample reads the matching level 1:1, and the GGX mip selection works as designed.

Could you cherry-pick it into this PR?

```
git fetch https://github.com/xen2/stride fix/ggx-prefilter-mip0-per-face-rescale
git cherry-pick 88b974b85c84794064a0cf1f6705125261269f00
```

xen2:
@sasvdw can you include the previously mentioned commit?

Or another option (maybe better long term): can you give me permission on your fork?
Even though " Maintainers are allowed to edit this pull request."  is checked, it seems it doesn't work because it's not a fork on your account but on an org account.

sasvdw:
I somehow missed your comment. I'll get it cherry picked and added.

sasvdw:
I've also added you as a contributor to my fork. I'll try and submit future PRs from a different repo.

Issue #3175, opened 2026-05-15
BepuConfiguration.NewInstance - inconsistent behaviour


ferafiks:
**Release Type**: Release

**Version**: 4.3.0.2507

**Platform(s)**: All

**Describe the bug**
If BepuConfiguration isn't added to game settings, it will be created at runtime. However, this behavior can differ:
* If the game settings asset exists -> configuration has no simulations, so nothing works
* If the game settings asset does not exist (code-only) -> configuration has a single simulation with default values

Having the latter always be the case would reduce a lot of confusion for users that didn't read the manual.

**To Reproduce**
1. Create a new project
2. Add some physics objects
3. Observe that physics don't work and there's a warning that there are no simulations
4. Delete the game settings asset
5. Set the graphics compositor and scene manually
6. Physics works

**Expected behavior**
A single simulation is added no matter if the game settings asset exists or not

PR #3343, opened 2026-08-09
Removed game settings overrides

ferafiks:
# PR Details

As discussed in https://github.com/stride3d/stride/pull/3272, removes the partially not working overrides from the game settings asset.

Breaking changes:
* `Game.Settings.Configurations.Get<>()` is now `Game.Settings.GetConfiguration<>()`

## Types of changes

## Checklist


Ethereal77:
I've never used this feature. Do we have alternative for this?
I mean (correct me if I'm wrong) I've always imagined this was for scenarios like: I have a game with a specific configuration and graphics compositor, but for iOS and Android I want a specific configuration with smaller textures and a special compositor that disables shadows and post-FX.

If it is like that, it is an useful feature to have for multi-platform builds.

ferafiks:
@Ethereal77 What you are describing wasn't possible previously, as the graphics compositor was not in a configuration, so it couldn't be overridden. The things you could override were mostly not related to hardware or platform (e.g. physics) except for graphics level, which from what xen mentioned in the linked PR, it probably wouldn't have worked. Also specific filters didn't work at all.

An alternative right now is to override the game class and add your own logic there (which is probably what you would have to do for most graphics-related settings). In the future, there could be a more proper way, like a replacement asset, but for now I'm just trying to remove this mostly broken feature

sources/engine/Stride.Engine/Engine/Design/GameSettings.cs
```diff
@@ -54,6 +56,22 @@ public GameSettings()
...
         /// Gets or sets configuration for the actual running platform as compiled during build
         /// </summary>
-        public PlatformConfigurations Configurations { get; set; }
+        public List<Configuration> Configurations { get; set; }
+
+        /// <summary>Retrieves a configuration of a given type.</summary>
+        /// <typeparam name="T">The type of the configuration to retrieve.</typeparam>
+        /// <returns>Returns the configuration.</returns>
+        /// <remarks>If <see cref="Configurations"/> doesn't contain the configuration, the method will return a newly created instance and add it to the list, so that future calls will keep returning the same instance.</remarks>
+        public T GetConfiguration<T>() where T : Configuration, new()
```
xen2: Let's rename it `GetOrCreateConfiguration()` so that it's more clear what it does.

xen2 (commented):
1. Found plenty of `Overrides:/PlatformFilters:` in the repo (i.e. new game and other templates). Scan for any *.sdgamesettings in the repo and clean them up.
2. RecastMeshSystem.cs:55 doesn't compile anymore
3. BepuConfiguration.cs:25-27 logic is now wrong I think (old Get<T>() returned `new T()` without saving it so it used to work). New code needs to run the presence check before calling GetConfiguration/GetOrCreateConfiguration (also removes the duplicate OfType scan).

ferafiks:
1. Done
2. I don't know why, but the `Stride.BepuPhysics.Navigation` project isn't in `Stride.slnx`, which is probably how I missed it.
3. I addressed https://github.com/stride3d/stride/issues/3175 while I was at it. Now, when you have no `BepuConfiguration` in game settings, the physics will work just like you would expect them to, unlike previously where there would be no simulations and everything would remain static. This is the same behavior that is present in code-only (no game settings).

sources/engine/Stride.BepuPhysics/Stride.BepuPhysics.Navigation/Processors/RecastMeshSystem.cs
```diff
@@ -52,7 +52,7 @@ public RecastMeshSystem(IServiceRegistry registry) : base(registry)
         _shapeCache = registry.GetSafeServiceAs<ShapeCacheSystem>();
 
         var gameSettings = registry.GetSafeServiceAs<IGameSettingsService>();
-        _navSettings = gameSettings.Settings.Configurations.Get<RecastNavigationConfiguration>() ?? new();
+        _navSettings = gameSettings.Settings.GetOrCreateConfiguration<RecastNavigationConfiguration>() ?? new();
```
xen2: `?? new()` is not necessary anymore.

sources/editor/Stride.Assets.Presentation/Templates/DotNetNewTemplateParametersWindow.xaml.cs
```diff
@@ -13,6 +13,7 @@
 using Stride.Core.Presentation.Controls;
 using Stride.Data;
```
xen2: This using can be removed

sources/engine/Stride.Assets/GameSettingsAssetCompiler.cs
```diff
@@ -80,29 +77,15 @@ protected override Task<ResultStatus> DoCommandOverride(ICommandContext commandC
                     DoubleViewSplashScreen = Parameters.DoubleViewSplashScreen,
                     EffectCompilation = entryPackage.UserSettings.GetValue(GameUserSettings.Effect.EffectCompilation),
                     RecordUsedEffects = entryPackage.UserSettings.GetValue(GameUserSettings.Effect.RecordUsedEffects),
-                    Configurations = new PlatformConfigurations(),
+                    Configurations = [],
```
xen2: `Configurations = Parameters.Defaults.Where(x => !x.OfflineOnly).ToList()`
then you can remove the loop doing `result.Configurations.Add()`

sources/engine/Stride.Engine/Engine/Design/GameSettings.cs
```diff
@@ -54,6 +56,22 @@ public GameSettings()
         /// <summary>
         /// Gets or sets configuration for the actual running platform as compiled during build
         /// </summary>
-        public PlatformConfigurations Configurations { get; set; }
+        public List<Configuration> Configurations { get; set; }
```
xen2: Maybe we should add a  `= []; ` initializer so that it doesn't fail if created programatically?

ferafiks: done

Issue #2696, opened 2025-04-02
Target .NET STS Versions


ds5678:
**Is your feature request related to a problem? Please describe.**

I'm always frustrated when I can't use the latest API offerings in the BCL.

In addition, bumping the .NET version generally gives many free performance improvements.

**Describe the solution you'd like**

Stride should bump its .NET version to the latest STS version every November instead of only bumping for LTS versions (even-numbered years).

"Customers can choose Long Term Support (LTS) releases or Standard Term Support (STS) releases. The quality of all releases is the same. The only difference is the length of support. LTS releases get free support and patches for three years. STS releases get free support and patches for two years."

This means that .NET 10 and 11 will both reach end of support on the same day (November 14, 2028).

https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core

**Describe alternatives you've considered**

* Multi-targeting the engine
  * This is likely to be a lot of extra maintenance.
* Compile my game using the newer .NET version
  * I don't think the game studio supports this.
  * Even with a code-only approach, I would still be exposed to issues like #3351.

Doprez:
Some added context https://github.com/stride3d/stride/discussions/2687#discussioncomment-12640834

There is no current plan to constantly update to the STS versions of .NET but I think there is a way to enable the builder to try and use the latest even without the libraries using STS. I do agree it s a bit annoying **especially** when VS auto installs STS versions by default.

Doprez:
You can keep this issue open. I think its still relevant for the second portion of your enhancement request. I just need to confirm that the current builder for projects is able to use the latest without issue. Once I have time to confirm that then I can update this issue.

VaclavElias:
Maybe one day, when we have enough resources, we could provide a dedicated branch or a NuGet STS (Standard Term Support) version for .NET STS as well. Unfortunately, at the moment, we don't have the resources or an available maintainer to actively drive this effort.

ds5678:
@VaclavElias @sasvdw @xen2 I have updated the issue description with my thoughts from #3351.

PR #3351, opened 2026-08-11
fix: Prepare ThreadPool for .Net11+

### LINKED ISSUE #3326: regression: LowLevelLifoSemaphore change in signature in dotnet11
**Release Type**: Future under .net11+

**Describe the bug**
`LowLevelLifoSemaphore.Release(int)` will be replaced by [Signal()](https://github.com/dotnet/runtime/commit/c14473701d9106c7be977b25b90211b99231b54b#diff-89db4da8c13ce558b4e96fe84ed99903ed03a2dbd30ce602fdaf69dacb71b52b) in .Net11. This internal type is used in our [ThreadPool](https://github.com/stride3d/stride/blob/ed6f739f25c92a9a6b9c4454f694b4d345b52667/sources/core/Stride.Core/Threading/ThreadPool.cs#L65) implementation.
Thankfully we have a fallback for this specific case, but we'll need to account for this change to avoid the performance overhead.

Eideren:
# PR Details
See #3326
Very minor changes to the dispatcher as well to make the logic a bit more readable and reduce the cruft.

## Related Issue
fixes #3326

## Types of changes

## Checklist

### DIFF
--- a/sources/core/Stride.Core/Threading/ThreadPool.cs
+++ b/sources/core/Stride.Core/Threading/ThreadPool.cs
@@ -22,11 +22,10 @@ public sealed partial class ThreadPool : IDisposable
     /// </summary>
     public static ThreadPool Instance = new();
 
-    private static readonly bool SingleCore;
     [ThreadStatic]
-    private static bool isWorkedThread;
+    private static bool isWorkerThread;
     /// <summary> Is the thread reading this property a worker thread </summary>
-    public static bool IsWorkedThread => isWorkedThread;
+    public static bool IsWorkedThread => isWorkerThread;
 
     private static readonly ProfilingKey ProcessWorkItemKey = new($"{nameof(ThreadPool)}.ProcessWorkItem");
 
@@ -62,13 +61,14 @@ public sealed partial class ThreadPool : IDisposable
         }
         try
         {
-            semaphore = new DotnetLifoSemaphore(spinCount);
+            if (Environment.Version.Major >= 11)
+                semaphore = new Dotnet11OrAboveLifoSemaphore();
+            else
+                semaphore = new Dotnet10OrUnderLifoSemaphore(spinCount);
         }
         catch (Exception e)
         {
-            // For net6+ this should not happen, logging instead of throwing as this is just a performance regression
-            if (Environment.Version.Major >= 6)
-                Logger.Warning($"Could not bind to dotnet's Lifo Semaphore, falling back to suboptimal semaphore:\n{e}");
+            Logger.Warning($"Could not bind to dotnet's Lifo Semaphore, falling back to suboptimal semaphore:\n{e}");
 
             semaphore = new SemaphoreW(spinCountParam: 70);
         }
@@ -81,39 +81,22 @@ public sealed partial class ThreadPool : IDisposable
         }
     }
 
-    static ThreadPool()
-    {
-        SingleCore = Environment.ProcessorCount < 2;
-    }
-
     /// <summary>
     /// Queue an action to run on one of the available threads,
     /// it is strongly recommended that the action takes less than a millisecond.
     /// </summary>
     public unsafe void QueueWorkItem([Pooled] Action workItem, int amount = 1)
     {
-        // Throw right here to help debugging
-        if (workItem == null)
-        {
-            throw new NullReferenceException(nameof(workItem));
-        }
-
-#if NET8_0_OR_GREATER
+        ArgumentNullException.ThrowIfNull(workItem);
         ArgumentOutOfRangeException.ThrowIfLessThan(amount, 1);
         ObjectDisposedException.ThrowIf(disposing > 0, this);
-#else
-        if (amount < 1) throw new ArgumentOutOfRangeException(nameof(amount));
-        if (disposing > 0) throw new ObjectDisposedException(ToString());
-#endif // NET8_0_OR_GREATER
 
-        Interlocked.Add(ref workScheduled, amount);
         var work = new Work { WorkHandler = &ActionHandler, Data = workItem };
+
         for (int i = 0; i < amount; i++)
-        {
             PooledDelegateHelper.AddReference(workItem);
-            workItems.Enqueue(work);
-        }
-        semaphore.Release(amount);
+
+        QueueWorkItemInner(amount, work);
     }
 
     static void ActionHandler(object param)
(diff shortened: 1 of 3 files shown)


sources/core/Stride.Core/Threading/ThreadPool.cs
```diff
@@ -281,4 +260,31 @@ public DotnetLifoSemaphore(int spinCount)
...
+    {
+        private readonly object semaphore;
+        private readonly Func<int, bool> wait;
+        private readonly Action release;
+
+        public Dotnet11OrAboveLifoSemaphore()
+        {
+            // The semaphore Dotnet uses for its own threadpool is more efficient than what's publicly available,
+            // but sadly it is internal - we'll hijack it through reflection
+            Type lifoType = Type.GetType("System.Threading.LowLevelLifoSemaphore")!;
```
Kryptos-FR, line 274: Can we copy/paste the code instead?
Reflection doesn't necessarily work on all platforms and it certainly won't if we ever support Native AOT.

Ethereal77: If we end up copying it, I'd wait for the final release of .NET 11, or else we are obligued to watch if they ever change it, correct bugs, redesign it, etc. (which may very well happen since it's still in preview)

Edit: ...or at least the RC

xen2: > Can we copy/paste the code instead? Reflection doesn't necessarily work on all platforms and it certainly won't if we ever support Native AOT.

FYI, we do support Native AOT already since PR #3204 (not only for iOS but also desktop).
It's even [tested regularly](https://github.com/xen2/stride/blob/e95f9c6d73daf1cc5839c1143f2a191a716c68f7/tests/enduser/Stride.Samples.Tests/SampleScreenshotTests.cs#L58).

Kryptos-FR: How did the tests no fail in such case? Or is it not part of tests triggered by PR changes?

Eideren: AOT does support reflection, it doesn't support generating new IL on the fly, so no generics like `Type.MakeGenericType()` basically

Eideren: I wasn't aware some platform don't support reflection, is that documented somewhere ?
Afair, the fallback semaphore is platform agnostic, so even if reflection failed it won't bring the whole engine down with it.
As for whether to copy, I do agree with Ethereal, in that case I would wait for the release. There's a reason I wasn't doing it previously, the implementation is platform specific and calls into a lot of internal logic, some of which is not in c# land. I've not looked deep enough to know whether there would be any blockers.

ds5678: UnsafeAccessor seems ideal for this situation.

https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.unsafeaccessorattribute?view=net-10.0

Eideren: From what I remember, you need to be able to access the type of the class you want to access through that API, here the type is internal

ds5678: > From what I remember, you need to be able to access the type of the class you want to access through that API, here the type is internal

That's only true on .NET 9 and lower.

https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.unsafeaccessortypeattribute?view=net-10.0

Kryptos-FR: > AOT does support reflection, it doesn't support generating new IL on the fly, so no generics like `Type.MakeGenericType()` basically

Yes. I think in my mind I had a simpler rule about reflection because in some cases, the related types could have been trimmed. However, in the present case, it's part of the framework so unlikely to be trimmed.

Maybe there is no issue after all.

Eideren: @ds5678 cool stuff, thanks! Do you happen to know whether there is a way to validate unsafe accessor mappings without running them - we need to know whether those method exist for the fallback, I could call them with a null instance and filter based on if I get a nullref exception but that's... kind of awful

ds5678: If the target member cannot be found, I think the runtime implements it as throwing `MissingMethodException` (or similar).

I don't know of any way to check if a method is available other than with a try catch block and caching the result.

xen2: Just checking, should I wait for .NET 11 then? Or ready for merge?

VaclavElias: Are we actually planning to upgrade to .NET 11? We used to only to LTS.

xen2: No, but you could make your game use (multitarget?) .NET 11 and it would use the new code.

ds5678: > you could make your game use (multitarget?) .NET 11 and it would use the new code.

Unless users are compiling their own engine packages, this wouldn't work because `Stride.Core` solo-targets .NET 10. Bumping their game to .NET 11 would still reference .NET 10 engine code (and therefore still hit the error that this pull request is trying to fix).

> Are we actually planning to upgrade to .NET 11? We used to only to LTS.

I think we should. .NET 10 and 11 both reach end of support at the same time (November 14, 2028).

"Customers can choose Long Term Support (LTS) releases or Standard Term Support (STS) releases. The quality of all releases is the same. The only difference is the length of support. LTS releases get free support and patches for three years. STS releases get free support and patches for two years."

https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core

sasvdw: > No, but you could make your game use (multitarget?) .NET 11 and it would use the new code.

That does mean that the engine core would need to support multi-targeting.

I wouldn't jump to .NET 11 in general, though. Those who want .NET 11 features can target .NET 11, and we can use .NET 11 code branches as a test bed for engine development that adopts new features that would be guaranteed LTS support in the next release. Microsoft guarantees LTS stability for even-numbered releases; Stride should match this cadence, but could trail certain features in a .NET 11-specific build. There's like a lot of build machinery behind the scenes that needs to be fixed to support this, but I do see a path forward.

Eideren: @xen2 feel free to merge. @ds5678's suggestion would have been great in a context where we don't have to maintain a fallback, better to leave it as is. The rest of the discussion is out of scope, @sasvdw @ds5678 @VaclavElias feel free to continue in a dedicated discussion.

ds5678: > The rest of the discussion is out of scope, feel free to continue in a dedicated discussion.

I've updated #2696 with my recent thoughts.

> @ds5678's suggestion would have been great in a context where we don't have to maintain a fallback

If #2696 is accepted, there wouldn't need to be a fallback. Stride would adopt .NET 11 in November.

Eideren: @ds5678 The fallback will stay regardless, it is setup to avoid breaking the engine if we don't keep up with the internals of dotnet, or if dotnet does not support that particular synchronization mechanism on other platforms.

PR #3355, opened 2026-08-13
Update templates and sample code to better support pausing

ds5678:
# PR Details

A couple places incorrectly used `Game.DrawTime` in an `Update` method. I replaced them with with `Game.UpdateTime`.

https://github.com/stride3d/stride/discussions/2999 says that the proper way to do game pausing for animations, particles, and physics is to set `Game.UpdateTime.Factor` to zero. However, most of the sample code used `Game.UpdateTime.Elapsed` instead of `Game.UpdateTime.WarpElapsed`. `Game.UpdateTime.WarpElapsed` is equal to `Game.UpdateTime.Factor * Game.UpdateTime.Elapsed`. Both are automatically calculated every frame, so there is no performance difference for using one over the other.

## Types of changes

## Checklist

### DIFF
--- a/samples/Input/TouchInputs/TouchInputs.Game/TouchInputsScript.cs
+++ b/samples/Input/TouchInputs/TouchInputs.Game/TouchInputsScript.cs
@@ -81,7 +81,7 @@ namespace TouchInputs
 
         public override void Update()
         {
-            var currentTime = Game.DrawTime.Total;
+            var currentTime = Game.UpdateTime.Total;
 
             keyDown = "";
             keyEvents = "";
--- a/sources/editor/Stride.Assets.Presentation/Templates/Assets/Scripts/Camera/BasicCameraController.cs
+++ b/sources/editor/Stride.Assets.Presentation/Templates/Assets/Scripts/Camera/BasicCameraController.cs
@@ -59,7 +59,7 @@ public class ##Scriptname## : SyncScript
 
     private void ProcessInput()
     {
-        float deltaTime = (float)Game.UpdateTime.Elapsed.TotalSeconds;
+        float deltaTime = (float)Game.UpdateTime.WarpElapsed.TotalSeconds;
         translation = Vector3.Zero;
         yaw = 0f;
         pitch = 0f;
--- a/samples/Games/JumpyJet/JumpyJet.Game/CharacterScript.cs
+++ b/samples/Games/JumpyJet/JumpyJet.Game/CharacterScript.cs
@@ -124,7 +124,7 @@ namespace JumpyJet
                 if (!isRunning)
                     continue;
 
-                var elapsedTime = (float)Game.UpdateTime.Elapsed.TotalSeconds;
+                var elapsedTime = (float)Game.UpdateTime.WarpElapsed.TotalSeconds;
 
                 // apply impulse on the touch/space
                 if (Input.IsKeyPressed(Keys.Space) || UserTappedScreen())
--- a/samples/Games/JumpyJet/JumpyJet.Game/PipesScript.cs
+++ b/samples/Games/JumpyJet/JumpyJet.Game/PipesScript.cs
@@ -66,7 +66,7 @@ namespace JumpyJet
             if (!isScrolling)
                 return;
 
-            var elapsedTime = (float)Game.UpdateTime.Elapsed.TotalSeconds;
+            var elapsedTime = (float)Game.UpdateTime.WarpElapsed.TotalSeconds;
 
             for (int i = 0; i < pipeSets.Count; i++)
             {
--- a/samples/Games/SpaceEscape/SpaceEscape.Game/Background/BackgroundScript.cs
+++ b/samples/Games/SpaceEscape/SpaceEscape.Game/Background/BackgroundScript.cs
@@ -78,7 +78,7 @@ namespace SpaceEscape.Background
             if (!isScrolling)
                 return;
 
-            var elapsedTime = (float)Game.UpdateTime.Elapsed.TotalSeconds;
+            var elapsedTime = (float)Game.UpdateTime.WarpElapsed.TotalSeconds;
 
             // Check if needed to remove the first block
             var firstBlock = levelBlocks[0];
--- a/samples/Graphics/AnimatedModel/AnimatedModel.Game/RotateEntity.cs
+++ b/samples/Graphics/AnimatedModel/AnimatedModel.Game/RotateEntity.cs
@@ -24,7 +24,7 @@ namespace AnimatedModel
                     rotationSpeed = 200f * Input.PointerEvents.Sum(x => x.DeltaPosition.X);
 
                 rotationSpeed *= 0.93f;
-                var elapsedTime = (float) Game.UpdateTime.Elapsed.TotalSeconds;
+                var elapsedTime = (float) Game.UpdateTime.WarpElapsed.TotalSeconds;
                 Entity.Transform.Rotation *= Quaternion.RotationY(rotationSpeed * elapsedTime);
             }
         }
--- a/samples/Graphics/SpriteStudioDemo/SpriteStudioDemo.Game/PlayerScript.cs
+++ b/samples/Graphics/SpriteStudioDemo/SpriteStudioDemo.Game/PlayerScript.cs
@@ -103,7 +103,7 @@ namespace SpriteStudioDemo
                 if (inputState == InputState.RunLeft || inputState == InputState.RunRight)
                 {
                     // Update Agent's position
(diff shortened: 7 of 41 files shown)


ds5678:
@xen2 When you have time, can you review this?

Kryptos-FR:
For better illustration purpose I'd rather we reserve using `WarpElapsed` in samples that do demonstrate a pausing mechanism.

Having both properly conveys an intent:
- either that there is no pausing mechanism in the game, or that it ignores pausing (e.g. to play an idle animation on a model/particle or to animate a UI
- or that it is a pausing mechanism and it is affected by it

ds5678:
> For better illustration purpose I'd rather we reserve using `WarpElapsed` in samples that do demonstrate a pausing mechanism.

I have the complete opposite view that things should be pause-friendly by default since pause menus are so ubiquitous in games. `Elapsed` is the one that should require justification for use.

xen2:
> @xen2 When you have time, can you review this?

Yes sorry, I have a long backlog and currently prioritize SDSL regressions but I'll get to it!

samples/Input/TouchInputs/TouchInputs.Game/TouchInputsScript.cs
xen2: This will end up comparing DrawTime with `UpdateTime` captured earlier.
Since no interpolation is involved, we should use `UpdateTime.Total`?

xen2 (commented):
Two spots outside the diff look like they were missed in the same sweep, plus one question about the camera controllers.

**1. The script templates still use `UpdateTime.Elapsed`,** although the samples with the identical code were converted:

- `sources/editor/Stride.Assets.Presentation/Templates/Assets/Scripts/Prefabs/PrefabInstance.cs:118` — same `// Countdown` block as `samples/Particles/ParticlesSample/ParticlesSample.Game/PrefabInstance.cs:118`, which this PR does change.
- `sources/editor/Stride.Assets.Presentation/Templates/Assets/Scripts/Prefabs/PrefabInstance.cs:72` — `timeIntervalCountdown` in `IsTriggered()`.
- `sources/editor/Stride.Assets.Presentation/Templates/Assets/Scripts/Events/EventBroadcaster.cs:47` — same `timeIntervalCountdown` pattern in `Update()`.

These are what Game Studio inserts on "Add script", so they should follow the same rule as the samples.

**2. `samples/NewGame/NewGame/MyTemplate.Game/BasicCameraController.cs:62`.** That folder is the input for the "New Game" project template (`Stride.Templates.Games.csproj` -> `SamplePath`), so every new project ships this file. It is a separate committed copy, not generated, so it does not follow the change you made to `Templates/Assets/Scripts/Camera/BasicCameraController.cs`. The two now disagree.

**3. The converted camera controllers become internally inconsistent,** because only some of their input paths go through `deltaTime`. Taking `Templates/Assets/Scripts/Camera/BasicCameraController.cs` as the example, at `Factor = 0`:

| input path | line | scaled? | result |
|---|---|---|---|
| WASD translation | 78 | yes | frozen |
| arrow-key look | 196 | yes | frozen |
| mouse look | 218-219 | no | still works |
| touch drag | 237-238 | no | still works |

So the keyboard half of the camera pauses and the mouse half does not. `FreeCamera`, `FpsCamera` and `Physics/PlayerController` have the same split.

Note: I would keep `Elapsed` (instead of `WarpElapsed`) at least on the free/debug cameras, since not being able to fly around while paused is a step back.

samples/Input/TouchInputs/TouchInputs.Game/TouchInputsScript.cs
ds5678: I verified that the change to `UpdateTime.Total` functions as intended.

ds5678:
Note: I deleted `BasicCameraController` from the `ThirdPersonPlatformer` project because it's not used for anything. `ThirdPersonPlatformer` has its own camera implementation.

samples/NewGame/NewGame/MyTemplate.Game/BasicCameraController.cs
```diff
@@ -173,15 +173,17 @@ private void ProcessInput()
 
         // Mouse movement and gestures
         {
+            float speed = (float)Game.UpdateTime.Factor
```
xen2: Missing semicolon (also in 2 other places)

sources/editor/Stride.Assets.Presentation/Templates/Assets/Scripts/Camera/BasicCameraController.cs
```diff
@@ -215,8 +217,8 @@ private void ProcessInput()
                     Input.LockMousePosition();
                     Game.IsMouseVisible = false;
 
-                    yaw -= Input.MouseDelta.X * MouseRotationSpeed.X;
-                    pitch -= Input.MouseDelta.Y * MouseRotationSpeed.Y;
+                    yaw -= Input.MouseDelta.X * MouseRotationSpeed.X * speed;
```
xen2: Doesn't this make the mouse half sensitive if Factor 0.5?

samples/NewGame/NewGame/MyTemplate.Game/BasicCameraController.cs
ds5678: Sorry about that. It's fixed now

sources/editor/Stride.Assets.Presentation/Templates/Assets/Scripts/Camera/BasicCameraController.cs
ds5678: Yes, `WarpElapsed = Factor * Elapsed`

xen2: I'd revert the three `BasicCameraController` copies to `Elapsed` and drop the `speed = Factor` on mouse/touch (but keep it for keyboard), same as you did for FreeCamera and FpsCamera. The rest of the PR is good.

Two reasons. `MouseDelta`, gesture deltas and the wheel are already amounts, not rates, so multiplying them by anything only changes sensitivity: in slow motion the camera lags the hand. And stopping the camera at `Factor = 0` is a pause policy, better left to the game than baked into a template.

Sorry if my earlier comment pointed you the wrong way. The rule isn't mouse vs keyboard, it's that a rate needs a duration and an amount does not.

ds5678: I reverted them

xen2:
Two camera scripts kept `WarpElapsed`: `CameraOrbitScript.cs:36` and `FindAndAttachCameraComponent.cs:19`. Same reasoning as BasicCameraController, the orbit one especially since it's touch-driven (same reasoning as mouse delta applies).

xen2:
> For better illustration purpose I'd rather we reserve using `WarpElapsed` in samples that do demonstrate a pausing mechanism.
> 
> Having both properly conveys an intent:
> 
> * either that there is no pausing mechanism in the game, or that it ignores pausing (e.g. to play an idle animation on a model/particle or to animate a UI
> * or that it is a pausing mechanism and it is affected by it

@Kryptos-FR I get your point, and I think the real problem is the naming: `Elapsed` is short and looks like the default, `WarpElapsed` looks like an opt-in feature, so people reach for the one that quietly breaks pause. Unity and Unreal do the opposite, the scaled one is the short name. Also there is no warped `Total`.

Maybe we can rethink this in future version (something like `ScaledElapsed`/`UnscaledElapsed` + `ScaledTotal`, with the bare ones obsoleted) than settle it through sample style here. More discussion/thinking needed, also I would argue it's not high priority.

In the meantime what we can do is: match the clock of what you change. World state (speeds, cooldowns, timers, animation) scaled, camera/UI/debug unscaled. That's what this PR does now, so I'm fine with it.

ds5678:
> Two camera scripts kept `WarpElapsed`: `CameraOrbitScript.cs:36` and `FindAndAttachCameraComponent.cs:19`. Same reasoning as BasicCameraController, the orbit one especially since it's touch-driven (same reasoning as mouse delta applies).

I fixed them. Let me know if there's anything else you want changed.

xen2:
Two more and I think we're done:

`TutorialUI.cs:127` is named UI but it's a camera transition (it lerps `Camera.Transform`), so same as the ones you just reverted.

`GameMenu/MainScript.cs:201` (`FillLifeBar`): that one is in a menu, so it shouldn't pause, right? A menu is exactly what's on screen while the game is paused (double check behavior, typing from phone so I didn't run it)

ds5678:
I reverted those two.

I agree with your statements about the main menu.

PR #3365, opened 2026-08-22
feat: GameStudio: show where a dragged entity will land in the entity tree

### LINKED ISSUE #3364: GameStudio: the entity tree gives no indication of where a dragged entity will land
## Description

Reordering and reparenting entities in the Scene editor's hierarchy tree is hard to predict. While dragging, the only feedback is the message window that follows the cursor. The on-screen indicator is a single flat grey line that looks the same at every nesting level, and some gestures are offered that the drop cannot honour.

Five separate problems, all present in the current editor:

### 1. The insert line carries no depth information

`InsertAdorner` draws one full-width 3 px `#ADADAD` line across the whole row. "Insert after Skybox" (a deep last child) and "Insert before ShootTheCubes" (the next top-level item) mean different parents and different indices, but they render identically, and their hot zones are 4 px apart.

Below, the same line for two different drops:

Image

### 2. "Insert before a folder" is offered, but cannot be honoured

Hovering the top band of a folder row shows "Insert before &lt;folder&gt;". Dropping there leaves the entity below the folder.

The cause is in the asset model rather than the view. `EntityDesign.Folder` is a path string with no index, and `EntityHierarchyItemViewModel` keeps `Folders` in an alphabetically sorted collection which it adds before the entities, so a folder has no position to insert around. `EntityFolderViewModel` does not override `GetInsertionIndex`, so it inherits the base implementation that returns `0` and ignores `InsertPosition` entirely.

Image

### 3. Harmless gestures are shown as errors

Dropping an entity on itself, or on the parent it already has, shows a red cross with "Can't drop an entity on itself or its children" or "Entity already has this parent". Nothing invalid was attempted — the gesture simply changes nothing.

### 4. Some positions cannot be reached at all

An entity cannot be placed after an expanded item that is the last child at its level. The only anchor would be the next sibling's "insert before" band, and there is no next sibling. At root level a drop in the empty area hides this; nested, nothing does.

### 5. A reparent has no row indication

The middle band of a row reparents into that entity, but nothing marks the row, so "into" and "between" look the same.

## Steps to reproduce

1. Open a sample with a nested scene, for example `samples/Templates/FirstPersonShooter`.
2. Expand a folder so that a deep entity sits directly above a top-level row.
3. Drag an entity across the boundary between those two rows. The indicator is the same line in the same place for both targets, although they mean different parents.
4. Hover the top band of a folder row. It offers "Insert before &lt;folder&gt;". Drop there, and the entity does not move above the folder.

## Proposal

Make the indicator state where the items land:

- indent the insert line to the target's depth and precede it with a marker, so the nesting level is readable
- highlight the row when the drop reparents into it
- make the insert bands a fraction of the row height instead of a fixed 4 px, so they scale with DPI and font size
- stop offering insert positions that cannot be honoured
- tell a harmless no-op apart from an error
- let the pointer's horizontal position choose the ancestor level in the gap below a subtree, which makes the unreachable position in (4) reachable

I have this working against `master` and will open a pull request shortly.

sasvdw:
# PR Details

The entity tree gave no reliable indication of where a dragged entity would land. The insert line was a flat grey line drawn across the whole row at the same place for every nesting level, so two drops that mean different parents looked identical, and some gestures were offered that the drop could not honour.

This makes the indicator state the outcome before you commit to it.

## What changed

**The insert line carries depth.** It starts at the indentation of its target, is preceded by a marker, and uses the accent colour of the theme instead of a fixed grey.

A1_before_after

**The drop target is highlighted**, in three states: green to accept, grey when nothing would change, red to refuse.

B4_three_states

**The insert bands scale with the row** — 30% of the row height, clamped to [3, 10] px and then to 40% of the row so a drop-into zone always survives — instead of a fixed 4 px. `DropZoneCalculator` holds this geometry and is unit tested.

A3_hit_zones

**A position that could not be reached is now reachable.** An item could not be placed after an expanded item that is the last child at its level. In the gap below a subtree, the horizontal position of the pointer now selects the ancestor level, and the indentation of the line shows which one.

C2_ancestor_picking

**An item can be moved to the end of its parent**, by dropping it in the empty area below the tree or on the row of its parent. `MoveChildren` already corrected the insertion index for an item it removes from the same parent, so the move itself was supported; only the check stopped it.

C3_empty_area_states

## Behaviour changes worth reviewing

Three things change for existing users. None of them is silent — each is stated in the drag message before the drop happens.

**1. Folder rows no longer offer insert bands.** The gesture existed, but it could not honour the position it promised. `EntityDesign.Folder` is a path string with no index, and folders are sorted by name and always come before the entities, so a folder has no position to insert around. `EntityFolderViewModel` did not override `GetInsertionIndex`, so it inherited the base implementation that returns `0` and ignores `InsertPosition`. The whole folder row now adds the items as its children, which works.

C1_folder_before_after

**2. Dropping an item on the parent it already has now moves it to the end** of that parent, instead of being refused. It stays a no-op only when the item is already the last child.

**3. A drop that changes nothing is neutral, not an error.** Dropping an item on itself used to show a red cross. `DropAcceptance` adds a third result between accepted and refused. The interfaces give it a default implementation, therefore the other view models keep their behaviour.

## Implementation notes

- `TreeViewItem` gains `HeaderHeight` and `HeaderElement`. The control template puts the header and the items presenter in one stack panel, so `ActualHeight` includes all children of an expanded item and cannot be used for row geometry. Adorning the header rather than the item is also what keeps the highlight on the row instead of the whole subtree.
- `DropAcceptance` reaches the behaviour through default interface members on `IAddChildViewModel` and `IInsertChildViewModel`, so none of the existing implementers changed. Only `EntityHierarchyItemViewModel` overrides them.
- The boolean members map `== Accepted`, so `NoOp` is a subset of the previous `false` and every existing caller behaves exactly as before.
- One pre-existing leak is fixed along the way: the `finally` in `DoDragDrop` never cleared the insert adorner, so an aborted drag left the line on screen.

## Testing

29 unit tests covering the new pure pieces: the band geometry in `DropZoneCalculator`, the tri-state on `DragContainer`, and proof that the default interface members keep the old boolean behaviour for view models that implement only the old members.

**These tests do not run in CI on this PR.** The `runtime` filter in `changes.yml` does not list `sources/editor/**` or `sources/presentation/**`, so only `Windows-Full` (which builds with `-p:StrideSkipUnitTests=true`) and `VS-Package` are triggered. `Windows-Tests-Simple`, the only lane that runs `Stride.Core.Assets.Editor.Tests`, is gated on `runtime` and stays skipped, and `Gate` still passes. To run them:

```
dotnet test build\Stride.Tests.Simple.slnf --filter "FullyQualifiedName~Stride.Core.Assets.Editor.Tests"
```

I did not want to add an unrelated file just to trip the path filter, but I am happy to if you would prefer the lane to run. The change is visible in GameStudio, so the `ci-editor` label may also be worth applying.

Verified by hand in a Debug GameStudio on `samples/Templates/FirstPersonShooter`, checking the resulting order of the tree after each drop rather than only the indicator: folder rows, the three drop states, ancestor selection in the gap below a subtree, move to the end from the empty area and from the row of the parent, and that a no-op writes nothing (a single undo after a no-op reverts the move that preceded it).

## Known limitations

- The ancestor zones are narrow. The level changes at the indentation of the target, which is about 10 px per level at the left edge of the tree.
- In the bottom band of an item whose parent is a folder, moving the pointer left does not step up a level. The walk reaches the folder, folders refuse insertion, and it falls through to "add as a child" on the hovered row. It is safe and the message states it, but it changes the operation instead of the level. There is no valid shallower target to offer in that case.
- The classification in the view model has no automated coverage, because there is no test project for `Stride.Assets.Presentation`. The geometry and the tri-state plumbing are covered by tests; the acceptance decisions are covered by hand.
- Themes other than the default, and virtualized scrolling, were not exercised.

## Related Issue

Fixes #3364

## Types of changes

## Checklist

### DIFF
--- a/sources/editor/Stride.Core.Assets.Editor.Tests/TestDragContainer.cs
+++ b/sources/editor/Stride.Core.Assets.Editor.Tests/TestDragContainer.cs
@@ -1,0 +1,59 @@
+// Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net) and Silicon Studio Corp. (https://www.siliconstudio.co.jp)
+// Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
+using System.Collections.Generic;
+using System.ComponentModel;
+using Xunit;
+using Stride.Core.Assets.Editor.View.Behaviors;
+using Stride.Core.Assets.Editor.ViewModel;
+
+namespace Stride.Core.Assets.Editor.Tests
+{
+    public class TestDragContainer
+    {
+        [Fact]
+        public void TestAcceptedShowsAcceptedOnly()
+        {
+            var container = new DragContainer(new object[] { new object() }) { Acceptance = DropAcceptance.Accepted };
+            Assert.True(container.IsAccepted);
+            Assert.False(container.IsRejected);
+        }
+
+        [Fact]
+        public void TestRejectedShowsRejectedOnly()
+        {
+            var container = new DragContainer(new object[] { new object() }) { Acceptance = DropAcceptance.Rejected };
+            Assert.False(container.IsAccepted);
+            Assert.True(container.IsRejected);
+        }
+
+        [Fact]
+        public void TestNoOpShowsNeither()
+        {
+            // The drag visual binds one image to each of these. A no-op must show no image at all.
+            var container = new DragContainer(new object[] { new object() }) { Acceptance = DropAcceptance.NoOp };
+            Assert.False(container.IsAccepted);
+            Assert.False(container.IsRejected);
+        }
+
+        [Fact]
+        public void TestDefaultIsRejected()
+        {
+            var container = new DragContainer(new object[] { new object() });
+            Assert.Equal(DropAcceptance.Rejected, container.Acceptance);
+        }
+
+        [Fact]
+        public void TestAcceptanceNotifiesTheDerivedProperties()
+        {
+            var container = new DragContainer(new object[] { new object() });
+            var changed = new List<string>();
+            ((INotifyPropertyChanged)container).PropertyChanged += (s, e) => changed.Add(e.PropertyName);
+
+            container.Acceptance = DropAcceptance.Accepted;
+
+            Assert.Contains(nameof(DragContainer.Acceptance), changed);
+            Assert.Contains(nameof(DragContainer.IsAccepted), changed);
+            Assert.Contains(nameof(DragContainer.IsRejected), changed);
+        }
+    }
+}
--- a/sources/editor/Stride.Core.Assets.Editor.Tests/TestDragContainer.cs
+++ b/sources/editor/Stride.Core.Assets.Editor.Tests/TestDragContainer.cs
@@ -1,4 +1,4 @@
-// Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net) and Silicon Studio Corp. (https://www.siliconstudio.co.jp)
+// Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net)
 // Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
 using System.Collections.Generic;
 using System.ComponentModel;
--- a/sources/editor/Stride.Core.Assets.Editor.Tests/TestDropAcceptanceDefaults.cs
+++ b/sources/editor/Stride.Core.Assets.Editor.Tests/TestDropAcceptanceDefaults.cs
@@ -1,0 +1,73 @@
+// Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net) and Silicon Studio Corp. (https://www.siliconstudio.co.jp)
+// Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
+using System.Collections.Generic;
+using Xunit;
+using Stride.Core.Assets.Editor.View.Behaviors;
+using Stride.Core.Assets.Editor.ViewModel;
+
+namespace Stride.Core.Assets.Editor.Tests
+{
+    /// <summary>
+    /// Verifies that a view model which implements only the boolean members keeps its behaviour.
+    /// </summary>
+    public class TestDropAcceptanceDefaults
+    {
+        private const string ExpectedMessage = "message from the view model";
+
+        private class LegacyAddChild : IAddChildViewModel
+        {
+            private readonly bool canAdd;
+
+            public LegacyAddChild(bool canAdd) { this.canAdd = canAdd; }
+
+            public bool CanAddChildren(IReadOnlyCollection<object> children, AddChildModifiers modifiers, out string message)
+            {
+                message = ExpectedMessage;
+                return canAdd;
+            }
+
+            public void AddChildren(IReadOnlyCollection<object> children, AddChildModifiers modifiers) { }
+        }
+
+        private class LegacyInsertChild : IInsertChildViewModel
+        {
+            private readonly bool canInsert;
+
+            public LegacyInsertChild(bool canInsert) { this.canInsert = canInsert; }
+
+            public bool CanInsertChildren(IReadOnlyCollection<object> children, InsertPosition position, AddChildModifiers modifiers, out string message)
+            {
+                message = ExpectedMessage;
+                return canInsert;
+            }
+
+            public void InsertChildren(IReadOnlyCollection<object> children, InsertPosition position, AddChildModifiers modifiers) { }
+        }
+
+        [Theory]
+        [InlineData(true, DropAcceptance.Accepted)]
+        [InlineData(false, DropAcceptance.Rejected)]
+        public void TestAddChildrenDefaultFollowsTheBoolean(bool canAdd, DropAcceptance expected)
+        {
+            IAddChildViewModel viewModel = new LegacyAddChild(canAdd);
+
+            var acceptance = viewModel.GetAddChildrenAcceptance(new object[0], AddChildModifiers.None, out var message);
+
+            Assert.Equal(expected, acceptance);
+            Assert.Equal(ExpectedMessage, message);
+        }
+
+        [Theory]
+        [InlineData(true, DropAcceptance.Accepted)]
+        [InlineData(false, DropAcceptance.Rejected)]
+        public void TestInsertChildrenDefaultFollowsTheBoolean(bool canInsert, DropAcceptance expected)
+        {
+            IInsertChildViewModel viewModel = new LegacyInsertChild(canInsert);
+
+            var acceptance = viewModel.GetInsertChildrenAcceptance(new object[0], InsertPosition.Before, AddChildModifiers.None, out var message);
+
+            Assert.Equal(expected, acceptance);
+            Assert.Equal(ExpectedMessage, message);
+        }
+    }
+}
--- a/sources/editor/Stride.Core.Assets.Editor.Tests/TestDropAcceptanceDefaults.cs
+++ b/sources/editor/Stride.Core.Assets.Editor.Tests/TestDropAcceptanceDefaults.cs
@@ -1,4 +1,4 @@
-// Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net) and Silicon Studio Corp. (https://www.siliconstudio.co.jp)
+// Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net)
 // Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
 using System.Collections.Generic;
 using Xunit;
--- a/sources/editor/Stride.Core.Assets.Editor.Tests/TestDropZoneCalculator.cs
+++ b/sources/editor/Stride.Core.Assets.Editor.Tests/TestDropZoneCalculator.cs
@@ -1,0 +1,96 @@
+// Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net) and Silicon Studio Corp. (https://www.siliconstudio.co.jp)
+// Distributed under the MIT license. See the LICENSE.md file in the project root for more information.
+using Xunit;
+using Stride.Core.Assets.Editor.View.Behaviors;
+
+namespace Stride.Core.Assets.Editor.Tests
+{
+    public class TestDropZoneCalculator
+    {
+        private const double RowHeight = 24.0;
+
+        [Theory]
+        [InlineData(0.0)]
+        [InlineData(3.0)]
+        [InlineData(7.0)]
+        public void TestTopBandInsertsBefore(double positionY)
+        {
+            Assert.Equal(InsertPosition.Before, DropZoneCalculator.GetInsertPosition(positionY, RowHeight, true));
+        }
+
+        [Theory]
+        [InlineData(20.0)]
+        [InlineData(24.0)]
+        public void TestBottomBandInsertsAfter(double positionY)
+        {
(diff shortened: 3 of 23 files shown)


sources/editor/Stride.Core.Assets.Editor.Tests/TestDragContainer.cs
```diff
@@ -0,0 +1,59 @@
+// Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net) and Silicon Studio Corp. (https://www.siliconstudio.co.jp)
```
Ethereal77: New files don't need the "Silicon Studio Corp." copyright notice nor its URL

sources/editor/Stride.Core.Assets.Editor.Tests/TestDropAcceptanceDefaults.cs
```diff
@@ -0,0 +1,73 @@
+// Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net) and Silicon Studio Corp. (https://www.siliconstudio.co.jp)
```
Ethereal77: New files don't need the "Silicon Studio Corp." copyright notice nor its URL

sources/editor/Stride.Core.Assets.Editor.Tests/TestDropZoneCalculator.cs
```diff
@@ -0,0 +1,96 @@
+// Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net) and Silicon Studio Corp. (https://www.siliconstudio.co.jp)
```
Ethereal77: New files don't need the "Silicon Studio Corp." copyright notice nor its URL

sources/editor/Stride.Core.Assets.Editor/View/Behaviors/DragDrop/DropZoneCalculator.cs
```diff
@@ -0,0 +1,71 @@
+// Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net) and Silicon Studio Corp. (https://www.siliconstudio.co.jp)
```
Ethereal77: New files don't need the "Silicon Studio Corp." copyright notice nor its URL

sources/editor/Stride.Core.Assets.Editor/View/Behaviors/DragDrop/TreeViewDragDropBehavior.cs
```diff
@@ -53,21 +54,99 @@ protected override IEnumerable<object> GetItemsToDrag(FrameworkElement container
...
+        /// the item". To go after an expanded item, use the gap below its last descendant and move the pointer to the
+        /// left. See <see cref="ResolveInsertAfterTarget"/>.
+        /// The empty area below the last item has no row to measure. Therefore it keeps bands of a fixed size.
+        /// </remarks>
         protected override IInsertChildViewModel GetInsertTargetItem(FrameworkElement container, Point mousePosition, out InsertPosition insertPosition)
         {
             insertPosition = InsertPosition.Before;
 
-            if (mousePosition.Y >= 0 && mousePosition.Y <= InsertThreshold)
+            if (!(container is TreeViewItem item))
```
Ethereal77: `if (container is not TreeViewItem item)`

sources/editor/Stride.Core.Assets.Editor/ViewModel/DropAcceptance.cs
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net) and Silicon Studio Corp. (https://www.siliconstudio.co.jp)
```
Ethereal77: New files don't need the "Silicon Studio Corp." copyright notice nor its URL

sources/editor/Stride.Core.Assets.Editor/View/Behaviors/DragDrop/DragDropAdornerManager.cs
```diff
@@ -86,9 +96,100 @@ internal static void ClearInsertAdorner()
...
+            dropTargetLayer = null;
         }
 
-        internal static void UpdateInsertAdorner([NotNull] UIElement container, InsertPosition position)
+        [NotNull]
+        private static Brush CreateFallbackNeutralBrush()
+        {
+            var brush = new SolidColorBrush(Color.FromArgb(255, 148, 148, 148));
+            brush.Freeze();
+            return brush;
```
Ethereal77: I'll leave this to someone more knowledgeable about the XAML part of GameStudio, but...

Shouldn't all the style-related things (colors, brushes, etc.) be queried from the theme instead of hardcoded?

sources/editor/Stride.Core.Assets.Editor/View/Behaviors/DragDrop/InsertAdorner.cs
```diff
@@ -1,42 +1,87 @@
...
+    /// If the adorned element is a <see cref="TreeViewItem"/>, the line starts at the indentation of that item and a
+    /// marker precedes it. This shows the nesting level at which the items land. The line follows the header of the
+    /// item, because the height of an expanded item includes its child items.
+    /// </remarks>
     public class InsertAdorner : Adorner
     {
-        private readonly Pen renderPen;
+        private const string LineBrushKey = "SelectedBackgroundBrush";
+        private const double LineThickness = 2.0;
+        private const double MarkerRadius = 3.0;
```
Ethereal77: Same concern about hardcoded styles:
Would this not be better defined by the theme styles?

sources/editor/Stride.Core.Assets.Editor/View/Behaviors/DragDrop/InsertAdorner.cs
```diff
@@ -1,42 +1,87 @@
...
+
             base.OnRender(drawingContext);
         }
+
+        [NotNull]
+        private static Brush CreateFallbackBrush()
+        {
+            var brush = new SolidColorBrush(Color.FromArgb(255, 173, 173, 173));
+            brush.Freeze();
+            return brush;
```
Ethereal77: Again, hardcoded styles

Ethereal77 (commented):
Looks good. I'll leave a thorough review to someone that knows better about the XAML part.

I've left some comments regarding headers and styles.

sources/editor/Stride.Core.Assets.Editor.Tests/TestDragContainer.cs
sasvdw: Fixed in 2afdbe8b3 — thanks. It was on all five new files; I had copied the header from an existing file without noticing the clause only belongs to the ones inherited from the original code base.

sources/editor/Stride.Core.Assets.Editor.Tests/TestDropAcceptanceDefaults.cs
sasvdw: Fixed in 2afdbe8b3 — thanks. It was on all five new files; I had copied the header from an existing file without noticing the clause only belongs to the ones inherited from the original code base.

sources/editor/Stride.Core.Assets.Editor.Tests/TestDropZoneCalculator.cs
sasvdw: Fixed in 2afdbe8b3 — thanks. It was on all five new files; I had copied the header from an existing file without noticing the clause only belongs to the ones inherited from the original code base.

sources/editor/Stride.Core.Assets.Editor/View/Behaviors/DragDrop/DropZoneCalculator.cs
sasvdw: Fixed in 2afdbe8b3 — thanks. It was on all five new files; I had copied the header from an existing file without noticing the clause only belongs to the ones inherited from the original code base.

sources/editor/Stride.Core.Assets.Editor/ViewModel/DropAcceptance.cs
sasvdw: Fixed in 2afdbe8b3 — thanks. It was on all five new files; I had copied the header from an existing file without noticing the clause only belongs to the ones inherited from the original code base.

sources/editor/Stride.Core.Assets.Editor/View/Behaviors/DragDrop/TreeViewDragDropBehavior.cs
sasvdw: Fixed in a1657789e.

sources/editor/Stride.Core.Assets.Editor/View/Behaviors/DragDrop/DragDropAdornerManager.cs
sasvdw: Good catch, and yes. Addressed in 0a32da956.

To be precise about what was and was not theme-driven before: the two colours were already queried (`SelectedBackgroundBrush` for the insert line, `NormalBorderBrush` here), but the fallbacks and the geometry were not, and the accept/refuse colours came from `HighlightBorderAdorner`'s own defaults. So the concern was justified.

There are now six keys. The four theme files give the colours, since a colour is a decision of the theme, and `generic.xaml` gives the two sizes, since they do not vary by theme:

    DropInsertLineBrush, DropTargetNeutralBrush, DropTargetAcceptBrush, DropTargetRefuseBrush
    DropInsertLineThickness, DropInsertMarkerRadius

`DropFeedbackResources` reads them, so one class holds the key names and the search order (element first, then the application, then a default). No colour literal is left in this file — the hardcoded grey is gone.

The accept/refuse colours are applied to this feature's adorner only. `HighlightBorderAdorner` itself is untouched, so the text-box validation behaviours, `ContainTextAdornerBehavior` and the panel-wide drop adorner keep their existing appearance.

sources/editor/Stride.Core.Assets.Editor/View/Behaviors/DragDrop/InsertAdorner.cs
sasvdw: Addressed in 0a32da956. The constants are gone: the brush, the line thickness and the marker radius are now theme keys read through `DropFeedbackResources`. See the reply on `DragDropAdornerManager` for the full key list.

sources/editor/Stride.Core.Assets.Editor/View/Behaviors/DragDrop/InsertAdorner.cs
sasvdw: Addressed in 0a32da956. This fallback is gone. The insert line keeps one non-colour default — `SystemColors.HighlightBrush` — only for the case where no theme is loaded at all, such as a unit test or the designer; the shipped editor always resolves the theme key.

Ethereal77:
LGTM. Thanks for your contribution!

@Kryptos-FR can you check this? This is more your area of expertise

Kryptos-FR (approved):
Looks good to me. Thanks a lot.

@xen2 the gap with editor tests not triggered by CI should be fixed in another PR I guess.

xen2:
> @xen2 the gap with editor tests not triggered by CI should be fixed in another PR I guess.

Editor tests are optional (since they take 25 min). Same goes for enduser tests (checking samples).
They can be added while reviewing by using proper label on PR, `ci-editor`
(sorry I merged thinking I already did that earlier, I will launch them manually)

I will think a bit more if we can improve this workflow or set them up by default (i.e. on some specific changes).

Ideally I would like to separate iteration vs final check (so that it doesn't trigger every single push).

PR #3368, opened 2026-08-23
feat: ask the graphics device what it provides, instead of comparing profiles

sasvdw:
# PR Details

**Summary** — Adds a capability vocabulary to `Stride.Graphics`, so a renderer asks what the device
provides instead of comparing graphics profiles.

## What it adds

Three questions that Stride currently answers with one number:

```csharp
GraphicsBackend.Implements(kind)      // static, per backend, needs no device
features.Provides(capability)         // the hardware and the driver
features.Supports(capability)         // both, three-valued
```

`GraphicsCapabilitySupport` is `Available`, `NotProvidedByDevice` or `NotImplementedByBackend`. That
third state is the point. The two failures have different remedies — another device fixes one, and
only a change to Stride fixes the other — and there was no way to say so. The Vulkan backend has never
implemented multisampling, and today reports it as hardware that cannot do it.

A `GraphicsCapability` owns its identity and how to read it from `GraphicsDeviceFeatures`, so a
renderer states what it needs rather than whether it has it. Two kinds so far: a device feature flag,
and `Multisampling(format, count)`, which is per pixel format and graded.

## Why

A feature-level ladder assumes capabilities are ordered and cumulative. Raytracing, mesh shaders and
variable-rate shading are optional and orthogonal — a card can be feature level 12_1 with no
raytracing. A profile comparison cannot ask about those, and it cannot tell hardware that will not do
something from a backend that does not do it yet.

## One behaviour change

`VoxelRenderer` now asks for compute shaders instead of comparing profiles. `HasComputeShaders` is
queried only on Direct3D 11 and hardcoded `true` on the other backends, so voxelization now runs where
the profile comparison refused it. The capability is also coarser than voxelization needs, because the
set cannot yet say shader model 5. This one needs a review.

## Call sites

The engine compares a profile against a level in twenty-five places. Seven now ask the device. The
eighteen that stay are three different things, and only one of them is "not a device question".

**Nine have no device to ask.** The asset compilers and
`MaterialSpecularMicrofacetEnvironmentGGXLUT.Generate` run at build time, and two backend sites pick a
profile before `GraphicsDeviceFeatures` exists. No query can replace these.

**Seven are device questions with no capability yet.** `SamplerState.Direct3D11.cs:87` caps anisotropy
on 9.1, `Texture.Direct3D11.cs:1015` asks for the standard multisample quality pattern, and the three
`CheckMipLevels` workarounds ask about a Direct3D 9 driver limit. These want the query. They keep the
comparison because the capability set has no term for what they ask.

**Two read a profile that is not the right source**, `GraphicsCompositorHelper.cs:60` and
`LightShaderGroupDynamic.cs:94`. The second is below.

Two further sites read a profile without comparing it: `EffectSystem.cs:145` *is* the content target,
and `GraphicsDeviceManager.cs:624` compares one request to another. Both are right as they are.

## Fixed

- **Vulkan rejected 32-bit index buffers** when a project targeted 9.x, though `VK_INDEX_TYPE_UINT32`
  is core Vulkan. `CurrentProfile` is a device fact only on Direct3D 11; Vulkan echoes the request.
- **`LightShafts` dereferenced a null `ShadowMapRenderer`** at draw time.
  `ForwardLightingRenderFeature` treats it as optional, and `GraphicsCompositorHelper` builds the
  feature without one below profile 10.0. It now throws where it initialises.
- **`CheckMipLevels` was inverted on Direct3D 11.** The workaround targets block-compressed textures,
  which are never depth-stencil, so the condition never fired. Direct3D 12 and Vulkan have the `!`.
- **Vulkan reported no format support for anything.** `FormatSupport` now comes from
  `vkGetPhysicalDeviceFormatProperties`. `MultisampleCountMax` deliberately stays `None`, because
  `ForwardRenderer` and `Texture.InitializeFrom` read it raw and it is what keeps the unimplemented
  multisampling path unreachable.
- **Nothing logged the graphics device.** `GraphicsDeviceFeatures.ToString()` had no caller anywhere.
  `GraphicsDevice.Recreate` now logs it, on both the create and the reset path, and `ToString()` gained
  the flags it omitted — every one that makes a renderer quietly do something else.

## Found, not fixed

- The GGX LUT is keyed on a profile when the question is texture format support.
- `Texture.Vulkan.cs:670` truncates compressed mip chains on Vulkan when a project targets 9.x.
- `LightShaderGroupDynamic.cs:39` ignores `ShaderProfile`, so its permutation count can disagree with
  what `EffectSystem` compiles.

## What it does not do

Nothing touches `GraphicsProfile`, the asset pipeline or serialized settings. The profile stays the
build-time content target and the device-creation target, which no capability query can replace. Only
one new capability is added. The seven sites that want the query each need a different one, and each
of those deserves a considered design rather than a term bolted on here.

## Related Issue

Follows the discussion on #3302 about the feature level system and moving toward capabilities.

## Types of changes

## Checklist

**Validation status**

- *Build*: clean under `StrideGraphicsApi=Direct3D11`, `=Direct3D12` and `=Vulkan`. **Not `=Null`** —
  that backend does not compile on `master` today, with partial-method signature drift in
  `Null/CommandList.Null.cs` and `Null/Texture.Null.cs`. The `GraphicsBackend.Null.cs` partial this PR
  adds is therefore unverified, and an earlier claim here that Null built clean was an incremental
  no-op on my part.
- *Tests*: three facts in `TestGraphicsCapability`, all passing, no device needed. They guard how the
  set grows: a `GraphicsCapabilityKind` that nothing answers for, two capabilities on one kind, and a
  kind with no switch arm, which throws today because the switch has a default arm the compiler
  accepts. They do not validate this change.
- *Run*: the end-user sample suite passes on every leg — Direct3D11, Direct3D12 and Vulkan on Windows,
  Vulkan on Linux — at **17 of 17** each, plus both packaging jobs. No frame drifted past its LPIPS
  baseline and none deferred to the vision gate.
- *Editor*: built and run on Direct3D11. GameStudio opens a project, and a game launched from it runs
  and renders.

The device report makes it clear what that run did and did not cover. Every sample logs:

```
Graphics device: Microsoft Basic Render Driver (Direct3D11)
Requested: Level_9_3, Current: Level_9_3, HasComputeShaders: False, ... HasDepthAsSRV: False
```

The samples render at profile 9_3, so they do exercise the paths this PR changes most: `CheckMipLevels`
now fires where the missing `!` kept it unreachable, the depth-SRV flags read `False`, and
`HasIndex32Bits` is `False` exactly as the old `CurrentProfile <= Level_9_3` was. Gold images still
match, so none of those changed what is drawn.

Two things the suite does not reach: the `VoxelRenderer` behaviour change, because no fixture uses
voxel GI, and multisampling, because every fixture renders with `MSAALevel: None`.

A game run from the editor on real hardware covers the other side of every one of those flags:

```
Graphics device: Intel(R) Arc(TM) 140V GPU (16GB) (Direct3D11)
Requested: Level_10_0, Current: Level_10_0, HasComputeShaders: True, ... HasDepthAsSRV: True,
HasDepthAsReadOnlyRT: False, ... HasIndex32Bits: True
```


sources/engine/Stride.Graphics/Direct3D11/GraphicsDeviceFeatures.Direct3D11.cs
```diff
@@ -66,6 +66,7 @@ internal GraphicsDeviceFeatures(GraphicsDevice device)
             HasDepthAsSRV = CurrentProfile >= Level_10_0;
             HasDepthAsReadOnlyRT = CurrentProfile >= Level_11_0;
             HasMultiSampleDepthAsSRV = CurrentProfile >= Level_11_0;
+            HasIndex32Bits = CurrentProfile > Level_9_3;
```
Ethereal77: Not important, as DX9 is already removed as well as OpenGL, but...
shouldn't this be a `>=`? I remember using 32-bit indices on Direct3D 9.0c

sources/engine/Stride.Graphics/Vulkan/VulkanConvertExtensions.cs
```diff
@@ -281,6 +281,30 @@ public static VkFormat ConvertPixelFormat(PixelFormat inputFormat)
...
+            {
+                ConvertPixelFormat(inputFormat, out format, out _, out _);
+                return true;
+            }
+            catch (InvalidOperationException)
+            {
+                format = VkFormat.Undefined;
+                return false;
+            }
+        }
```
Ethereal77, line 297: I would do this the other way around, i.e., having the `TryConvertPixelFormat` as the "cheap" method that determines the conversion and then doing `ConvertPixelFormat` as the stub that calls `TryConvertPixelFormat` and throwns an exception when unsuccesful.

That way the `Try...` variant is the cheap one. And I also would document the exception.

sources/engine/Stride.Graphics/GraphicsCapabilityKind.cs
```diff
@@ -0,0 +1,54 @@
...
+
+namespace Stride.Graphics;
+
+/// <summary>
+///   The kinds of capability a renderer can declare it needs.
+/// </summary>
+/// <remarks>
+///   A backend implements a kind or it does not. The device answers per capability instance, because
+///   multisampling support depends on the pixel format and the sample count.
+/// </remarks>
```
Ethereal77: The remark about "multisampling support" is too specific for an `enum` `GraphicsCapabilityKind` that describes capabilities in the general sense.

I mean, explain that "backends" declare they support specific capabilities, but the user device can or cannot support some of them.

Ethereal77:
As I already said, I'm not opposed to moving from a feature level to a capability-based system if done well. I think however this may be a big breaking change. Right now we are at `-beta5`, I don't know if this is "beta" in the classic sense (bug fixing, polishing, etc.) and this would make more sense in an "alpha" to release on the next version of Stride.

Anyway, I have a few worries:

* Up until now, "backend" have not been an _"official" term_ in Stride. We referred to graphics APIs (even though we meant the backend that used that API). If it reads easier for users, having a `GraphicsBackend` seems to be a good idea. But better decide what to do, as having several names for the same thing may confuse people.

* Right now, `GraphicsPlatform` is the way of identifying the platform that is being run. It can be queried in the `GraphicsDevice`'s static `Platform` property. It's an `enum`.

* Also, if `GraphicsBackend` were a static type that represents the currently executing graphics backend, I'd centralize some of the information that is now in other parts. For example, it could have a name, a `GraphicsPlatform`, etc.

* What is the benefit of it being a full type instead of an `enum` (as it is right now) and having backend-dependent extensions on it declaring what is supported?

* Another concern is that it seems to be a lot of overhead having capabilities being classes. It seems it inverts the control for no good reason. Let me explain:
  - I thought (and tried myself to keep) `Features` was as light as possible. A structure populated once that could be queried. It could be made lazy even to make it even lighter. But having the capabilities being classes that accept a `GraphicsDeviceFeatures` just to query it and answer indirectly seems off, and now it is not a light struct, but heap allocated classes.
  - This can become problematic specially for formats. For example, the multisample support capabilities seem to be thought for having one per format, and there could be hundreds of formats + multi-sample counts combinations.
  - For backend capabilities, the backend could answer the "supports this" questions on its own, without needing another type.
  
I like where this goes, but I feel it a bit over-abstracted for not the correct reasons. A direct way to query the backend, and a direct way for the device to answer the "supports" questions (that would also ask the backend) could be lighter while offering the same functionality. That's my honest opinion.

@xen2 What's your take on this?

sources/engine/Stride.Graphics/Vulkan/VulkanConvertExtensions.cs
sasvdw: I agree with this. Been looking at it all for too long to realise.

xen2:
Yes, this new capability system will be for 4.5. I would like to take some time to make sure it's designed properly before moving forward.

Also I agree with several points from @Ethereal77 
I think we will need a few iterations on the design.

I will focus on 4.4 release first, then I will come back on this topic, so please wait a few days/weeks.

sources/engine/Stride.Graphics/Vulkan/VulkanConvertExtensions.cs
sasvdw: Done in f442c3760.

`TryConvertPixelFormat` now holds the switch and returns `false` for a format with no Vulkan equivalent; `ConvertPixelFormat` calls it and throws, with an `<exception>` tag on it as you suggested.

Worth noting the cost was not only stylistic: `GraphicsDeviceFeatures.Vulkan` walks all 256 `PixelFormat` values at device creation, and most have no Vulkan equivalent, so the old wrapper threw and caught an `InvalidOperationException` for nearly every one of them.

All four existing signatures are unchanged, so every call site compiles as before — the only addition is a four-argument `TryConvertPixelFormat` overload that carries the conversion, with the two-argument one delegating to it. Verified by building the Vulkan variant specifically (`-p:StrideGraphicsApis=Vulkan`), since the file is behind `STRIDE_GRAPHICS_API_VULKAN` and a default Direct3D11 build does not compile it.

Still looking at your other two comments.

sources/engine/Stride.Graphics/Direct3D11/GraphicsDeviceFeatures.Direct3D11.cs
sasvdw: You are correct. The reality is that the enum value still exists, so a decision needs to be made on whether to remove it entirely.

In all honesty, the `GraphicsProfile` enum feels more and more like a feature gate for different versions of SDSL, but there is an argument to be made that capability-based SDSL versioning might be the better way. I don't know; more thinking is still needed.

sasvdw:
> What is the benefit of it being a full type instead of an enum (as it is right now) and having backend-dependent extensions on it declaring what is supported?

The back-end full type is mostly due to us not having discriminated unions available to us yet. Some capabilities carry metadata, so a pure enum doesn't answer the full question. The Depth Buffer multisample count is an example here.

> Another concern is that it seems to be a lot of overhead having capabilities being classes. It seems it inverts the control for no good reason. Let me explain:
> 
> * I thought (and tried myself to keep) Features was as light as possible. A structure populated once that could be queried. It could be made lazy even to make it even lighter. But having the capabilities being classes that accept a GraphicsDeviceFeatures just to query it and answer indirectly seems off, and now it is not a light struct, but heap allocated classes.
> * This can become problematic specially for formats. For example, the multisample support capabilities seem to be thought for having one per format, and there could be hundreds of formats + multi-sample counts combinations.
> * For backend capabilities, the backend could answer the "supports this" questions on its own, without needing another type.

There's plenty of room for optimisation. I'd thought I'd try to first get something that expresses the right intent before going and optimising it. I'll start taking a deeper dive into this as the inversion is likely a bit overkill.

sources/engine/Stride.Graphics/Direct3D11/GraphicsDeviceFeatures.Direct3D11.cs
sasvdw: Changed to `>=`. You're right — 9_3 is the first feature level that carries 32-bit index buffers, 9_1 and 9_2 are the 16-bit ones.

The `>` came from `PrefabModelAssetCompiler`, which uses the same test to decide mesh splitting. I've left that one alone: it runs at content-compile time with no device to ask, and changing it would change compiled output. Worth a separate look.

On removing the now-meaningless enum values entirely — agreed, but that's a bigger change than this PR should carry.

sources/engine/Stride.Graphics/GraphicsCapabilityKind.cs
sasvdw: Fair point — leading with multisampling made the enum read as if it were about multisampling.

The remarks now state the general contract: a backend declares the kinds it implements, and that holds for every device it runs on; whether the user's device provides one is a separate question, asked of a `GraphicsCapability` rather than of a kind. The per-format-and-sample-count detail moved onto the `Multisampling` member.

PR #3370, opened 2026-08-23
fix: load the assets a run-time material references

### LINKED ISSUE #3369: Material.New binds an empty texture for the default specular model
## Release Type

- [x] Bug

## Version

`master` at `2f6f79b4b`. Observed on Windows, Direct3D11, Intel Arc 140V, at profile `Level_10_0`.

## Description

A material created in code with `Material.New` binds an empty `Texture` for the environment DFG
lookup table. The default specular model uses that lookup table, so this applies to most materials
created at runtime.

**Observed at run time.** A probe script in a game built from the First Person Shooter template reports
the same asset resolving two different ways in the same frame, on the same device:

```
[built by Material.New] url=/Stride.Engine/StrideEnvironmentLightingDFGLUT16 IsProxy=True  size=0x0   format=None
[built by Material.New] CONFIRMED: the lookup table is an unresolved proxy with no pixels.
[loaded from an asset]  url=/Stride.Engine/StrideEnvironmentLightingDFGLUT16 IsProxy=False size=32x64 format=R16G16_Float
[loaded from an asset]  the lookup table is a real texture.
```

The material loaded from an asset holds a real 32x64 `R16G16_Float` texture. The material built by
`Material.New` holds a 0x0 texture with format `None` that is still marked `IsProxy`.

A third material that selects `MaterialSpecularMicrofacetEnvironmentGGXPolynomial` binds no texture at
all, which agrees with the cause below: only the lookup table version is affected.

## Cause

`MaterialSpecularMicrofacetEnvironmentGGXLUT.Generate` binds the lookup table as a proxy:

https://github.com/stride3d/stride/blob/2f6f79b4b/sources/engine/Stride.Rendering/Rendering/Materials/MaterialSpecularMicrofacetEnvironmentGGXLUT.cs#L27-L29

`AttachedReferenceManager.CreateProxyObject<T>` is `new T()` plus reference metadata. Its own
documentation says it lets you "construct and save object references without actually loading them":

https://github.com/stride3d/stride/blob/2f6f79b4b/sources/core/Stride.Core.Serialization/Serialization/AttachedReferenceManager.cs#L78-L89

A proxy becomes a loaded object in one place only: `ReferenceSerializer<T>.Serialize`, and only when a
`ContentSerializerContext` is present — that is, during a `ContentManager` load.

`Material.New` runs `MaterialGenerator.Generate` outside that path:

https://github.com/stride3d/stride/blob/2f6f79b4b/sources/engine/Stride.Rendering/Rendering/Material.cs#L49-L68

Nothing resolves the proxy afterward. The parameter collection keeps a `Texture` with no GPU resource.

A material loaded from an asset is not affected. The generator runs at build time, and the resulting
material goes through the content pipeline, which resolves the reference.

## Affected defaults

Two features default `Environment` to the lookup table version:

- `Materials/MaterialSpecularMicroFacetModelFeature.cs:44`
- `Materials/Hair/MaterialSpecularHairModelFeature.cs:258`

`MaterialSpecularMicrofacetEnvironmentGGXPolynomial` computes the term in the shader and needs no
texture, so a material that selects it is not affected.

## Steps to reproduce

1. Make a `MaterialDescriptor` with a `MaterialSpecularMicrofacetModelFeature` and keep the default
   `Environment`.
2. Call `Material.New(GraphicsDevice, descriptor)`.
3. Read the lookup table back out of the material and look at it:

```csharp
var texture = material.Passes[0].Parameters
    .Get(MaterialSpecularMicrofacetEnvironmentGGXLUTKeys.EnvironmentLightingDFG_LUT);
var reference = AttachedReferenceManager.GetAttachedReference(texture);
// reference.IsProxy is true, and texture.Width and texture.Height are 0.
```

To see both sides at once, put a material asset that uses the same environment function beside it and
read the same key from that one. It reports `IsProxy=False` and a 32x64 texture.

Expected: both give a real texture, because both name the same asset.

## Possible directions

`MaterialGeneratorContext` holds a `GraphicsDevice` but no `ContentManager`, so `Generate` has no way
to load the asset. Options, roughly in order of size:

- Give `Material.New` an optional `ContentManager` and resolve attached references after generation.
- Let `Generate` load through a content manager on the context.
- Document the limit, and have `Material.New` log a warning when generation leaves an unresolved proxy.

## Related

Found while auditing `GraphicsProfile` comparisons for a capability model. The same `Generate` method
also selects between the 16-bit and the 8-bit lookup table on a profile comparison, when the real
question is texture format support. That part is a separate concern.

sasvdw:
# PR Details

**Summary** — A material built by `Material.New` bound an empty texture for the environment DFG lookup
table. Fixes #3369.

## The two halves

The reported bug is that `Material.New` never resolves the references its features attach:

```
[built by Material.New] url=/Stride.Engine/StrideEnvironmentLightingDFGLUT16 IsProxy=True  size=0x0   format=None
[loaded from an asset]  url=/Stride.Engine/StrideEnvironmentLightingDFGLUT16 IsProxy=False size=32x64 format=R16G16_Float
```

Fixing that alone is not enough, and writing the test is what showed why. The first red run failed with
`The asset ... could not be found`, because **the table may not be in the build at all**. So there are
two problems, and a game whose materials are all built in code needs both fixed.

**The reference was never resolved.** `MaterialSpecularMicrofacetEnvironmentGGXLUT.Generate` attaches
the table with `AttachedReferenceManager.CreateProxyObject`, which is `new T()` plus reference metadata.
A proxy becomes the real asset in one place only, `ReferenceSerializer`, and only during a content load.
`Material.New` runs the generator outside that path.

`Material.New` now takes an optional `ContentManager` and, after generation, walks each pass's
parameters and loads anything still marked `IsProxy`. It resolves generically rather than special-casing
the lookup table, so `Generate` stays untouched and the asset compiler keeps serializing proxies exactly
as before.

**The asset was never guaranteed to be there.** The asset compiler starts from roots and follows
references. `Package.RootAssets` is the declared escape hatch for assets nothing references — *"needs to
be compiled even if not directly or indirectly referenced (useful for explicit code references)"*.
`Stride.Engine` already declares the default font, the splash screen and the debug sprite font that way;
the splash screen is the identical pattern, named from code in `GameSettingsFactory.cs:23`. The two
lookup tables were never declared, so they reach a build only when some material *asset* happens to use
that environment function and drags them in.

Both are now declared. They ship in every bundle as a result, including games that never use that
environment function — 32x64 `R16G16_Float` plus the 8-bit variant, the same unconditional treatment the
splash screen already has.

## When it still cannot resolve

Without a `ContentManager`, or when the asset is genuinely absent, the material keeps the empty object
and reports it, naming the parameter and the URL. It does not throw. Turning a subtly wrong render into
a crash would be the worse trade, and the roughly twelve existing callers — editor gizmos, previews,
debug shapes, fallbacks — pass no content manager and none of them uses a specular model.

## Tests

- `TestMaterialProxyResolution` in `Stride.Graphics.Tests`: the table resolves to a real texture, and the
  no-content-manager path reports rather than stays silent. That project holds **no** material asset
  referencing the tables, so it passes only if both halves hold. Reverting just the `.sdpkg` change puts
  it back to red with `ContentManagerException`, so it measures what it claims to.
- `TestGeneratorLeavesTextureReferencesAsProxies` in `Stride.Assets.Tests`: pins that
  `MaterialGenerator.Generate` on its own still yields a proxy, which is what the asset compiler
  serializes. Four tests there already built this feature and asserted only on shader source.

## Related Issue

Fixes #3369.

## Types of changes

## Checklist

**Validation status**

- *Build*: `Stride.Rendering` clean under `StrideGraphicsApi=Direct3D11`, `=Direct3D12` and `=Vulkan`,
  with `-t:Rebuild` rather than an incremental build. Not `=Null`: that backend does not compile on
  `master` today, with partial-method signature drift in `Null/CommandList.Null.cs` and
  `Null/Texture.Null.cs`. Unrelated to this change and worth its own issue.
- *Tests*: 2 passing in `Stride.Graphics.Tests`, 7 in `TestMaterialGenerator` — the new contract test and
  the six that were already there, so the asset compiler's path is unchanged.
- *Gold images*: none should move. The only rendering test that builds this material in code,
  `MaterialLayerABBWithAPI`, is `[Fact(Skip = ...)]` and never runs, so no baseline encodes the broken
  appearance. Left skipped; the stated reason is debugging convenience and I have no evidence it relates
  to this bug.
- *Editor*: outstanding.

## Not fixed here

`Generate` chooses between the 16-bit and 8-bit tables on a graphics profile when the real question is
texture format support. That is a separate concern, already listed under "found, not fixed" in #3368.

### DIFF
--- a/sources/buildengine/Stride.Core.BuildEngine.Common/BuildTransaction.cs
+++ b/sources/buildengine/Stride.Core.BuildEngine.Common/BuildTransaction.cs
@@ -31,6 +31,41 @@ internal class BuildTransaction
         }
     }
 
+    public IEnumerable<KeyValuePair<string, ObjectId>> GetMergedIdMap()
+    {
+        var result = new Dictionary<string, ObjectId>();
+
+        lock (transactionOutputObjects)
+        {
+            foreach (var outputObject in transactionOutputObjects)
+            {
+                if (outputObject.Key.Type == UrlType.Content)
+                    result[outputObject.Key.Path] = outputObject.Value;
+            }
+
+            foreach (var outputObjects in outputObjectsGroups)
+            {
+                // Lock underlying EnumerableBuildStep.OutputObjects
+                lock (outputObjects)
+                {
+                    foreach (var outputObject in outputObjects)
+                    {
+                        if (outputObject.Key.Type == UrlType.Content)
+                            result.TryAdd(outputObject.Key.Path, outputObject.Value.ObjectId);
+                    }
+                }
+            }
+
+            if (contentIndexMap != null)
+            {
+                foreach (var entry in contentIndexMap.GetMergedIdMap())
+                    result.TryAdd(entry.Key, entry.Value);
+            }
+        }
+
+        return result;
+    }
+
     public bool TryGetValue(string url, out ObjectId objectId)
     {
         var objUrl = new ObjectUrl(UrlType.Content, url);
@@ -114,8 +149,7 @@ internal class BuildTransaction
 
         public IEnumerable<KeyValuePair<string, ObjectId>> GetMergedIdMap()
         {
-            // Shouldn't be used
-            throw new NotImplementedException();
+            return buildTransaction.GetMergedIdMap();
         }
 
         public void Dispose()
--- a/sources/core/Stride.Core.Serialization/Serialization/Contents/ContentManager.cs
+++ b/sources/core/Stride.Core.Serialization/Serialization/Contents/ContentManager.cs
@@ -670,46 +670,53 @@ public sealed partial class ContentManager : IContentManager
     {
         var errorMessage = $"The asset '{url}' could not be found. Asset path should be 'MyFolder/MyAssetName', or '/PackageName/MyFolder/MyAssetName' for an asset from a namespaced package. Check that the path is correct and that the asset has been included into the build.";
 
-        var contentIndexMap = FileProvider?.ContentIndexMap;
-        if (contentIndexMap != null)
+        try
         {
-            var indexMap = contentIndexMap.GetMergedIdMap();
+            var contentIndexMap = FileProvider?.ContentIndexMap;
+            if (contentIndexMap != null)
+            {
+                var indexMap = contentIndexMap.GetMergedIdMap();
 
-            // First suggest assets differing from the requested URL only by the '/PackageName' root
-            var isRooted = url.StartsWith('/');
-            var candidates = new List<string>();
-            foreach (var entry in indexMap)
-            {
-                if (isRooted ? DiffersByRoot(url, entry.Key) : DiffersByRoot(entry.Key, url))
-                    candidates.Add(entry.Key);
-            }
-            if (candidates.Count > 0)
-            {
-                candidates.Sort();
-                errorMessage = $"The asset '{url}' could not be found. Did you mean:"
-                    + string.Concat(candidates.Select(c => $"{Environment.NewLine}  '{c}'"))
-                    + $"{Environment.NewLine}An asset from a namespaced package is addressed by a rooted URL: '/PackageName/MyFolder/MyAssetName'.";
-            }
-            else
-            {
-                // Otherwise fall back to assets with the same name in other folders
-                var assetName = url.Substring(url.LastIndexOf('/') + 1);
-                var sameNameCandidates = new List<string>();
+                // First suggest assets differing from the requested URL only by the '/PackageName' root
+                var isRooted = url.StartsWith('/');
+                var candidates = new List<string>();
                 foreach (var entry in indexMap)
                 {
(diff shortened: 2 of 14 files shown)


xen2 (approved):
Thanks, good fix and the tests pin exactly the right behavior.

I pushed a few commits on top:

1. Restricted the resolve loop to object parameter keys (replacing a permutation value would not bump the permutation counter) and moved it to a shared `ParameterCollectionExtensions.ResolveAttachedReferences`.
2. `ParameterCollection.UpdateLayout` now warns (once per URL) when an unresolved reference is about to be bound for rendering, so this bug class fails loudly instead of rendering black.
3. Fixed the same bug in the editor's `SkyboxPreview`, which also needed the LUT compiled into the preview database. Verified in GameStudio.
4. Fixed a crash found while testing: the asset-not-found suggestions were not implemented for the editor's build-transaction database.
5. Small test and UI fixes.

PR #3360, opened 2026-08-16
Update sample project assets

ds5678:
# PR Details

This pull request updates the sample projects with reimported assets.

## Related Issue

I was preparing to do other pull requests related to the sample projects, and the Game Studio prompted me to reimport a bunch of assets. Apparently, their data format has changed a bit since creation.

## Types of changes

- [x] Docs change / refactoring / dependency upgrade

## Checklist

- [x] All new and existing tests passed.
- [x] **I have built and run the editor to try this change out.**

PR #3373, opened 2026-08-24
Bug fix: Remove `redMaterial` field from AsyncCollisionTriggerDemo

ds5678:
## Related Issue

It was always null since the field is never assigned. This caused an exception while I was testing #3360.

## Types of changes

## Checklist


xen2:
Good find.
It's a side effect of https://github.com/stride3d/stride/pull/3327

PR #3376, opened 2026-08-25
fix: Avoid reading unmapped memory in FastTextRenderer

### LINKED ISSUE #3375: FastTextRenderer reads a Vulkan mapping after it has been unmapped
**Release Type**: GitHub source

**Version**: `4.4.0-dev` at commit `2f6f79b`

**Platform(s)**: Linux/Vulkan

**Describe the bug**

On the first frame that draws debug text, Linux/Vulkan terminates with an `AccessViolationException`.
When `FastTextRenderer` initializes, it creates the index data for the character quads in a mapped temporary buffer and saves a pointer to that data.
It then unmaps the buffer before passing the pointer to `Buffer.Index.New`, which reads from it to initialize the final index buffer.
The pointer is no longer valid after the Vulkan buffer has been unmapped.

**To Reproduce**

1. Build a code-only game against Stride source packages from commit `2f6f79b`.
2. Select Vulkan.
3. Queue one debug message with `game.DebugTextSystem.Print(...)`.
4. Start the game.
5. Observe a fatal access violation while constructing the index buffer.

**Expected behavior**

The debug text should render without accessing unmapped memory.

**Screenshots**

Not applicable because the process terminates.

**Log and callstacks**

```csharp
Fatal error.
System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
   at System.SpanHelpers.Memmove(Byte ByRef, Byte ByRef, UIntPtr)
   at System.Buffer.MemoryCopy(Void*, Void*, Int64, Int64)
   at Stride.Core.MemoryUtilities.CopyWithAlignmentFallback(Void*, Void*, UInt32)
   at Stride.Graphics.Buffer.Recreate(IntPtr)
   at Stride.Graphics.Buffer.InitializeFromImpl(Stride.Graphics.BufferDescription ByRef, Stride.Graphics.BufferFlags, Stride.Graphics.PixelFormat, IntPtr)
   at Stride.Graphics.Buffer.InitializeFrom(System.ReadOnlySpan`1<Byte>, Int32, Stride.Graphics.BufferFlags, Stride.Graphics.PixelFormat, Stride.Graphics.GraphicsResourceUsage)
   at Stride.Graphics.Buffer.New(Stride.Graphics.GraphicsDevice, System.ReadOnlySpan`1<Byte>, Int32, Stride.Graphics.BufferFlags, Stride.Graphics.PixelFormat, Stride.Graphics.GraphicsResourceUsage)
   at Stride.Graphics.Buffer+Index.New(Stride.Graphics.GraphicsDevice, System.ReadOnlySpan`1<Byte>, Stride.Graphics.GraphicsResourceUsage)
   at Stride.Graphics.FastTextRenderer.Initialize(Stride.Graphics.GraphicsContext, Int32)
   at Stride.Graphics.FastTextRenderer..ctor(Stride.Graphics.GraphicsContext, Int32)
   at Stride.Profiling.DebugTextSystem.Draw(Stride.Games.GameTime)
   at Stride.Games.GameSystemCollection.Draw(Stride.Games.GameTime)
   at Stride.Games.GameBase.Draw(Stride.Games.GameTime)
   at Stride.Games.GameBase.RawTick(System.TimeSpan, Int32, Single, Boolean)
   at Stride.Games.GameBase.RawTickProducer()
   at Stride.Games.GameBase.Tick()
   at Stride.Games.GamePlatform.Tick()
   at Stride.Games.GamePlatform.OnRunCallback()
   at Stride.Games.GameWindowSDL+<>c__DisplayClass17_0.<Run>b__0()
   at Stride.Games.SDLMessageLoop.Run(Stride.Graphics.SDL.Window, RenderCallback)
   at Stride.Games.GameWindowSDL.Run()
   at Stride.Games.GamePlatform.Run(Stride.Games.GameContext)
   at Stride.Games.GameBase.Run(Stride.Games.GameContext)
   at Stride.CommunityToolkit.Engine.GameExtensions.Run(Stride.Engine.Game, Stride.Games.GameContext, System.Action`1<Stride.Engine.Scene>, System.Action`2<Stride.Engine.Scene,Stride.Games.GameTime>)
   at Program.<Main>$(System.String[])
```

**Additional context**

I fixed this locally by building the indices in a managed `int[]` and passing that array directly to `Buffer.Index.New`.
This removes the temporary buffer and keeps the index data alive for the entire call.
The method no longer uses pointers after this change, so I also removed its `unsafe` modifier.
I rebuilt the local Stride packages and confirmed that the game now renders successfully on Linux/Vulkan.

BenjaBobs:
# PR Details

`FastTextRenderer.Initialize` retained a pointer to mapped index-buffer memory and passed it to `Buffer.Index.New` after unmapping the buffer.
On Linux/Vulkan the pointer was no longer valid, which caused an `AccessViolationException` when `Buffer.Index.New` tried to copy the index data.
This change builds the indices in a managed `int[]` and passes that array directly to `Buffer.Index.New`.
This also removes the temporary graphics buffer and the now-unnecessary `unsafe` modifier.
I rebuilt the local Stride packages and confirmed that a code-only game using Vulkan can render debug text without crashing.

## Related Issue

Fixes #3375

## Types of changes

## Checklist

I tested the change in a code-only game on Linux/Vulkan.
I did not run Game Studio because it is not currently available on Linux.

### DIFF
--- a/sources/engine/Stride.Graphics/FastTextRenderer.cs
+++ b/sources/engine/Stride.Graphics/FastTextRenderer.cs
@@ -86,33 +86,25 @@ namespace Stride.Graphics
         /// Initializes a FastTextRendering instance (create and build required ressources, ...).
         /// </summary>
         /// <param name="graphicsContext">The current GraphicsContext.</param>
-        private unsafe void Initialize(GraphicsContext graphicsContext, int maxCharacters)
+        private void Initialize(GraphicsContext graphicsContext, int maxCharacters)
         {
             maxCharacterCount = maxCharacters;
-            var indexBufferSize = maxCharacters * 6 * sizeof(int);
-            var indexBufferLength = indexBufferSize / IndexStride;
-
-            // Map and build the indice buffer
-            indexBuffer = graphicsContext.Allocator.GetTemporaryBuffer(new BufferDescription(indexBufferSize, BufferFlags.IndexBuffer, GraphicsResourceUsage.Dynamic));
-
-            var mappedIndices = graphicsContext.CommandList.MapSubResource(indexBuffer, 0, MapMode.WriteNoOverwrite, false, 0, indexBufferSize);
-            var indexPointer = mappedIndices.DataBox.DataPointer;
+            var indices = GC.AllocateUninitializedArray<int>(maxCharacters * 6);
 
             var i = 0;
             for (var c = 0; c < maxCharacters; c++)
             {
-                *(int*)(indexPointer + IndexStride * i++) = c * 4 + 0;
-                *(int*)(indexPointer + IndexStride * i++) = c * 4 + 1;
-                *(int*)(indexPointer + IndexStride * i++) = c * 4 + 2;
+                indices[i++] = c * 4 + 0;
+                indices[i++] = c * 4 + 1;
+                indices[i++] = c * 4 + 2;
 
-                *(int*)(indexPointer + IndexStride * i++) = c * 4 + 1;
-                *(int*)(indexPointer + IndexStride * i++) = c * 4 + 3;
-                *(int*)(indexPointer + IndexStride * i++) = c * 4 + 2;
+                indices[i++] = c * 4 + 1;
+                indices[i++] = c * 4 + 3;
+                indices[i++] = c * 4 + 2;
             }
 
-            graphicsContext.CommandList.UnmapSubResource(mappedIndices);
-
-            indexBufferBinding = new IndexBufferBinding(Buffer.Index.New(graphicsContext.CommandList.GraphicsDevice, new ReadOnlySpan<byte>((void*)indexPointer, indexBufferSize)), true, indexBufferLength);
+            indexBuffer = Buffer.Index.New(graphicsContext.CommandList.GraphicsDevice, indices);
+            indexBufferBinding = new IndexBufferBinding(indexBuffer, true, indices.Length);
 
             // Create vertex buffers
             vertexBuffers = new Buffer[VertexBufferCount];
--- a/sources/engine/Stride.Graphics/FastTextRenderer.cs
+++ b/sources/engine/Stride.Graphics/FastTextRenderer.cs
@@ -19,8 +19,6 @@ namespace Stride.Graphics
     {
         private const int VertexBufferCount = 2;
 
-        private const int IndexStride = sizeof(int);
-
         private Buffer[] vertexBuffers;
         private int activeVertexBufferIndex;
         private VertexBufferBinding[] vertexBuffersBinding;
@@ -54,25 +52,10 @@ namespace Stride.Graphics
 
         protected override void Destroy()
         {
-            for (int i = 0; i < VertexBufferCount; i++)
-                vertexBuffers[i].Dispose();
+            ReleaseResources();
 
             activeVertexBufferIndex = -1;
-
-            if (indexBuffer != null)
-            {
-                indexBuffer.Dispose();
-                indexBuffer = null;
-            }
-
-            indexBufferBinding = null;
             pipelineState = null;
-
-            if (simpleEffect != null)
-            {
-                simpleEffect.Dispose();
-                simpleEffect = null;
-            }
 
             for (int i = 0; i < VertexBufferCount; i++)
                 inputElementDescriptions[i] = null;
@@ -80,6 +63,25 @@ namespace Stride.Graphics
             charsToRenderCount = -1;
 
             base.Destroy();
+        }
+
+        /// <summary>
+        /// Releases the graphics resources created by <see cref="Initialize"/>.
+        /// </summary>
+        private void ReleaseResources()
+        {
+            for (int i = 0; i < VertexBufferCount; i++)
+                vertexBuffers[i].Dispose();
+
+            vertexBuffers = null;
+            vertexBuffersBinding = null;
+
+            indexBuffer?.Dispose();
+            indexBuffer = null;
+            indexBufferBinding = null;
+
+            simpleEffect?.Dispose();
+            simpleEffect = null;
         }
 
         /// <summary>
@@ -174,6 +176,7 @@ namespace Stride.Graphics
             if (charsToRenderCount > maxCharacterCount)
             {
                 maxCharacterCount = (int)(1.5f * charsToRenderCount);
+                ReleaseResources();
                 Initialize(graphicsContext, maxCharacterCount);
             }
 


BenjaBobs:
@dotnet-policy-service agree

sources/engine/Stride.Graphics/FastTextRenderer.cs
```diff
@@ -86,33 +86,25 @@ protected override void Destroy()
...
-                *(int*)(indexPointer + IndexStride * i++) = c * 4 + 3;
-                *(int*)(indexPointer + IndexStride * i++) = c * 4 + 2;
+                indices[i++] = c * 4 + 1;
+                indices[i++] = c * 4 + 3;
+                indices[i++] = c * 4 + 2;
             }
 
-            graphicsContext.CommandList.UnmapSubResource(mappedIndices);
-
-            indexBufferBinding = new IndexBufferBinding(Buffer.Index.New(graphicsContext.CommandList.GraphicsDevice, new ReadOnlySpan<byte>((void*)indexPointer, indexBufferSize)), true, indexBufferLength);
```
Ethereal77, line 115: I think the problem with this code is not using the now unmapped buffer to construct the index buffer, but constructing a new unnecessary index buffer using the pointer. Let me explain:
* The index buffer is created above in line 96 as a dynamic buffer. It is then mapped and populated, then unmapped.
* Then, when creating the `IndexBufferBinding` instead of using our already populated `indexBuffer` we do another `Buffer.Index.New` passing the now invalid pointer to the data to populate it. But that is exactly what we already did above!

This seems like a mistake from when this was ported from C++ to C#, and not a problem per-se of persisting the pointer incorrectly.

BenjaBobs: From my understanding, the way it was before two buffers were used because the final one should be immutable for read optimization, and so a dynamic one was created to fill in the actual data afterwards.

`create mutable buffer` > `fill with data` > `copy to immutable buffer` > `cleanup mutable buffer`.

My change would then change this to
`build data in int[]` > `copy to immutable buffer`

So I guess you could say the double buffering from before was unnecessary, but from as a consumer the fatal issue was what is effectively a "read-after-free".

What do you think if the solution?
I also considered using `NativeMemory` to avoid GC but that would also mean keeping the `unsafe` and doing a `try`/`finally`.

Ethereal77: Sorry for the late reply. I was busy.

Looking into this with a fresh mindset, I think we both are correct and both are wrong at the same time.

I initially misread your post and thought you were referring to the vertex buffer. That one must be dynamic because it is rewritten every frame. The index buffer, however, is fixed, immutable and uploaded only once.

But note that when using `Buffer.New()` for the index buffer and passing initialization data, Stride is already uploading the data to GPU memory through a staging buffer when needed, so I think you are doing here double work. I can only talk about D3D implementations here, which are what I worked on, but I think Vulkan must do the same.

Because of that, manually mapping a temporary buffer and then creating a second buffer from that pointer ends up duplicating work, and in this case it also caused the read-after-unmap issue.

sources/engine/Stride.Graphics/FastTextRenderer.cs
```diff
@@ -86,33 +86,25 @@ protected override void Destroy()
...
             maxCharacterCount = maxCharacters;
-            var indexBufferSize = maxCharacters * 6 * sizeof(int);
-            var indexBufferLength = indexBufferSize / IndexStride;
-
-            // Map and build the indice buffer
-            indexBuffer = graphicsContext.Allocator.GetTemporaryBuffer(new BufferDescription(indexBufferSize, BufferFlags.IndexBuffer, GraphicsResourceUsage.Dynamic));
-
-            var mappedIndices = graphicsContext.CommandList.MapSubResource(indexBuffer, 0, MapMode.WriteNoOverwrite, false, 0, indexBufferSize);
-            var indexPointer = mappedIndices.DataBox.DataPointer;
+            var indices = new int[maxCharacters * 6];
```
ds5678:
```suggestion
            var indices = GC.AllocateUninitializedArray<int>(maxCharacters * 6);
```

If we're using an array, this is faster, since the default value of `maxCharacters` (7000) is larger than `2048 / sizeof(int)`.

https://github.com/dotnet/dotnet/blob/15bf8906e62eda052a8d07729d3adddea4e1ea12/src/runtime/src/coreclr/nativeaot/System.Private.CoreLib/src/System/GC.NativeAot.cs#L822-L825

ds5678 (changes requested):

sources/engine/Stride.Graphics/FastTextRenderer.cs
BenjaBobs: I did not know of this, thanks! I added the change.

sources/engine/Stride.Graphics/FastTextRenderer.cs
BenjaBobs: Np, we're all busy at times :)
So do we agree on the current fix then?

xen2: LGTM, resolving

sources/engine/Stride.Graphics/FastTextRenderer.cs
xen2: This is now unused

sources/engine/Stride.Graphics/FastTextRenderer.cs
```diff
@@ -86,33 +86,25 @@ protected override void Destroy()
...
-                *(int*)(indexPointer + IndexStride * i++) = c * 4 + 2;
+                indices[i++] = c * 4 + 1;
+                indices[i++] = c * 4 + 3;
+                indices[i++] = c * 4 + 2;
             }
 
-            graphicsContext.CommandList.UnmapSubResource(mappedIndices);
-
-            indexBufferBinding = new IndexBufferBinding(Buffer.Index.New(graphicsContext.CommandList.GraphicsDevice, new ReadOnlySpan<byte>((void*)indexPointer, indexBufferSize)), true, indexBufferLength);
+            indexBuffer = Buffer.Index.New(graphicsContext.CommandList.GraphicsDevice, indices);
```
xen2, line 108: `Initialize` is also called from `End()` when `charsToRenderCount > maxCharacterCount`, and it overwrites `indexBuffer` here (and `vertexBuffers` a few lines below) without disposing the previous ones, so the old GPU buffers leak until the process exits. Sane goes for `simpleEffect` I think.

Not introduced by this PR, but this is the line that makes `indexBuffer` a real device
buffer owned by the class. Something like this at top of `Initialize`:
```csharp
// Release the resources of a previous Initialize (End() re-initializes on character count growth)
indexBuffer?.Dispose();
if (vertexBuffers != null)
{
    for (int j = 0; j < VertexBufferCount; j++)
        vertexBuffers[j]?.Dispose();
}
simpleEffect?.Dispose();
```

xen2 (changes requested):

sources/engine/Stride.Graphics/FastTextRenderer.cs
Ethereal77: If the index buffer is created, initialized and disposed (i.e., not immutable) then possibly the use of `Allocator.GetTemporaryBuffer` did make sense, no? It is acting like a discardable buffer after all.

I ask this because while having the index buffer belong to this class a nice "encapsulation" of its resource, `Allocator` may have in the future (if it does not already) optimizations specific for these use once, upload, then discard buffers.
For example, Unreal Engine has a big temporary buffer that it suballocates in an arena-like fashion for these kind of workloads specifically (text rendering, debug lines, etc.)

xen2: I want to do a pass in the future to review all GPU memory alloc.
A lot of pattern we use were more tailored for D3D11 (cycling temp buffer, map-discard)
D3D12/Vulkan allow more stuff (draw while mapped, etc.) but our current API doesn't allow it.
Also, our emulated D3D11-style map-discard is not great with D3D12/Vulkan

I would rather reconsider that when we switch to a proper D3D12/Vulkan-first approach (with maybe D3D11 in back-compat mode).
I am far from certain `Allocator.GetTemporaryBuffer` will be what we want to use then.

I would say, let's fix the `AccessViolationException` the simple way for now, and reconsider all those memory alloc, ring-buffer etc. separately.

xen2: Done with 3a0d448f09

sources/engine/Stride.Graphics/FastTextRenderer.cs
xen2: Done with 3a0d448f09

PR #3379, opened 2026-08-25
Simplify PR template (free-text sections instead of checkboxes)

xen2:
Simplifies the PR template so pull requests no longer show an incomplete task counter ("0 of 8 tasks") in the PR list.
Every `- [ ]` in a description counts toward that badge, and since almost nobody ticks all the boxes, every PR looked unfinished forever.

Changes:
- Replaced the "Types of changes" and "Checklist" checkbox lists with free-text sections.
- Removed "All new and existing tests passed", tests now run on every PR in CI.
- Kept the important signals as prompts instead: issue link (with a `Fixes #1234` hint), how the change was tested (editor changes should still be tried in the editor, CI does not cover it fully), breaking changes and documentation impact.

## How did you test this?

No code change, github template only.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,33 +1,19 @@
-# PR Details
-
-<!--- Provide a general summary of your changes in the Title of this PR -->
 <!--- Describe your changes in detail here -->
 <!--- Visit https://doc.stride3d.net/latest/en/contributors/contribution-workflow/github-pull-request-guidelines.html for more -->
 
 ## Related Issue
 
-<!--- If suggesting a new feature or change, please discuss it in an issue first -->
-<!--- If fixing a bug, there should be an issue describing it with steps to reproduce -->
-<!--- Please link to the issue here: -->
+<!--- New features and changes should be discussed in an issue first. -->
+<!--- Bug fixes should link an issue with steps to reproduce. -->
+<!--- Link the issue here (use "Fixes #1234" to auto-close it on merge): -->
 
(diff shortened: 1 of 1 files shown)


Kryptos-FR:
> tests now run on every PR in CI

Only if the workflows are approved, which is not by default for non-members. So if we remove that, there should be a check that is enforced and blocks the PR until the workflows run.

xen2:
> > tests now run on every PR in CI
> 
> Only if the workflows are approved, which is not by default for non-members. So if we remove that, there should be a check that is enforced and blocks the PR until the workflows run.

It's already enforced, with the "CI / Gate" job. Example:
image

Kryptos-FR (approved):

.github/pull_request_template.md
```diff
@@ -1,33 +1,19 @@
...
+<!--- Bug fixes should link an issue with steps to reproduce. -->
+<!--- Link the issue here (use "Fixes #1234" to auto-close it on merge): -->
 
-- [ ] Docs change / refactoring / dependency upgrade
-- [ ] Bug fix (non-breaking change which fixes an issue)
-- [ ] New feature (non-breaking change which adds functionality)
-- [ ] Breaking change (fix or feature that would cause existing functionality to change)
+## How did you test this?
 
-## Checklist
```
MsEpsilon, line 22: I don't think this has to be removed (##Checklist), especially for bigger PRs, but that's maybe just bikeshedding on my part.

Issue #3374, opened 2026-08-25
Voxel cone tracing doesn't work on 4.4.0-beta 5


meriaizen86:
**Release Type**: Beta

**Version**: 4.4.0-beta 5 (works on 4.3.0.2507 and previous versions)

**Platform(s)**:  Windows 11 25H2 26200.9168

**Describe the bug**
The scene does't render, instead it pops up the following error message:

```
Error: One or more errors occurred. (One or more errors occurred. (Error: The method or operation is not implemented.. System.NotImplementedException: The method or operation is not implemented.
   at Stride.Shaders.Parsing.SDSL.AST.AccessorChainExpression.CompileHelper(SymbolTable table, CompilerUnit compiler) in /_/sources/shaders/Stride.Shaders.Parsers/Parsing/SDSL/AST/Expression.cs:line 1286
   at Stride.Shaders.Parsing.SDSL.AST.AccessorChainExpression.ProcessSymbol(SymbolTable table, SymbolType expectedType) in /_/sources/shaders/Stride.Shaders.Parsers/Parsing/SDSL/AST/Expression.cs:line 780
   at Stride.Shaders.Parsing.SDSL.AST.AssignExpression.ProcessSymbol(SymbolTable table, SymbolType expectedType) in /_/sources/shaders/Stride.Shaders.Parsers/Parsing/SDSL/AST/Expression.cs:line 90
   at Stride.Shaders.Parsing.SDSL.AST.ExpressionStatement.ProcessSymbol(SymbolTable table) in /_/sources/shaders/Stride.Shaders.Parsers/Parsing/SDSL/AST/Statements.cs:line 36
   at Stride.Shaders.Parsing.SDSL.AST.BlockStatement.ProcessSymbol(SymbolTable table) in /_/sources/shaders/Stride.Shaders.Parsers/Parsing/SDSL/AST/Statements.cs:line 225
   at Stride.Shaders.Parsing.SDSL.AST.ShaderMethod.ProcessSymbolBody(SymbolTable table, SpirvContext context) in /_/sources/shaders/Stride.Shaders.Parsers/Parsing/SDSL/AST/ShaderElements.MethodOrMember.cs:line 506
   at Stride.Shaders.Parsing.SDSL.AST.ShaderClass.Compile(SymbolTable table, CompilerUnit compiler) in /_/sources/shaders/Stride.Shaders.Parsers/Parsing/SDSL/AST/Shader.cs:line 616
[...]

PR #3382, opened 2026-08-29
Fix partially voxel GI on 4.4: SDSL mixer, D3D12 root signature, Vulkan MSAA

Nicogo1705:
# PR Details

`Stride.Voxels` renders correctly on 4.3 and crashes on 4.4: its shaders do not get through the new SDSL compiler. Past the crashes they compile and dispatch, but global illumination contributes exactly zero, with nothing logged. On Direct3D12 and Vulkan both backends throw on the first frame.

Every fix here is a general engine issue that `Stride.Voxels` happens to uncover, not a workaround inside it; the two backend fixes affect any scene. Verified on Direct3D11, Direct3D12 and Vulkan.

**Reproduction:** https://github.com/Nicogo1705/StrideVoxelGI, a Cornell box using `Stride.Voxels` (`dotnet run --project Demo`). It targets `4.4.0-beta5`, so it reproduces every failure below as published; pointed at this branch it renders. `G` toggles the indirect light, `V` cycles the voxel debug views.

## Commits

| Commit | Fix |
|---|---|
| `76c532e` | ignore `using` namespace declarations instead of erroring |
| `80eff8d` | emit the access chain for literal-indexed shader compositions |
| `eafee2e` | name the missing composition instead of `KeyNotFoundException` |
| `692a965` | index `perMapOffsetScale` with an int, not a float |
| `d4288e7` | atomics on typed buffers via `OpImageTexelPointer` |
| `a1b6f29` | drop geometry stream parameters from every method, not just the entry point |
| `6c44263` | typed buffers are opaque resources too; say why the optimizer refused a module |
| `a2791d4` | number pixel shader UAVs after the render targets; drop execution modes with their function |
| `0ace208` | widen image sampled types: an image cannot deliver 16 bits |
| `9e2681b` | clear the voxel buffer with a 2D dispatch (65535 groups per dimension on every API) |
| `ad961c9` | a composition cannot supply the shader's entry point |
| `ba272d0` | `streams = input[i]` keeps the members the stage input does not carry |
| `2f8e603` | **D3D12**: size the static sampler buffer from the samplers, not the root parameters |
| `26a8ed2`, `f741c7a` | **Vulkan**: report the multisample counts the device supports, per format |
| `978bfdd` | drop every geometry stream parameter of a method, not only the last one |
| `fff9a24` | tests pinning down what namespaces and `using` do in SDSL today |

Two of these made GI render zero rather than crash, each by silently deleting code at mix time:

- `ad961c9`: a composition sharing its base with the shader it is composed into won the `CSMain` group, so the mipmap shader compiled to a bare `ret` with no resources.
- `ba272d0`: `streams = input[i]` zeroed everything the input did not carry, wiping the dominant axis each vertex; only surfaces already facing that axis voxelized.

The two backend fixes are not specific to voxels:

- `2f8e603`: the immutable sampler buffer was sized from `rootSignatureParameters.Count` instead of `immutableSamplers.Count`; an effect with more samplers than root parameters overran it (`Destination is too short`).
- `26a8ed2`, `f741c7a`: `GraphicsDeviceFeatures` never queried multisample support on Vulkan, so every format reported `MultisampleCount.None` and `Texture.New` rejected every multisampled render target. The count now comes from `vkGetPhysicalDeviceImageFormatProperties`, per format, with the usage `CreateImage` gives an attachment; formats the device cannot use that way stay at `None`. Note that `CreateImage` still hardcodes `samples = Count1`, so Vulkan never actually creates a multisampled image; that is a separate fix.

## Related Issue

https://github.com/stride3d/stride/issues/3374

## Types of changes

## Checklist

Every shader fix has a regression test, verified failing with the fix reverted. The two backend fixes need a device and were verified by running the demo on the backend concerned. Game Studio was not run against these changes.

**Removed from this PR:** the two commits that hoisted a `stage compose` value supplied by a nested effect (`4df2f3d`, `6a90900`), as discussed in review. Without them `Stride.Voxels` does not get through the mixer yet: `AttributeSamplers` is supplied under `environmentLights[i]`. The mixer rule is discussed in #3423; `Stride.Voxels` will be adapted once it is settled.

Known limitation left as is: on D3D12, mesa's `spirv_to_dxil` reports `SampledBuffer` / `ImageBuffer` (`Buffer<T>` / `RWBuffer<T>`) as unsupported; the scene renders, but that path is a backend limitation none of this touches.

---
<sub>⚠️ Written with AI assistance (Claude Code), driven and verified by me; needs a real review by someone who knows the SDSL mixer and the SPIR-V interface processor. 🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

### DIFF
--- a/sources/engine/Stride.Graphics/Vulkan/GraphicsDeviceFeatures.Vulkan.cs
+++ b/sources/engine/Stride.Graphics/Vulkan/GraphicsDeviceFeatures.Vulkan.cs
@@ -3,6 +3,8 @@
 
 #if STRIDE_GRAPHICS_API_VULKAN
 
+using Vortice.Vulkan;
+
 namespace Stride.Graphics
 {
     /// <summary>
@@ -40,9 +42,39 @@ namespace Stride.Graphics
 
             HasResourceRenaming = false;
 
-            // TODO D3D12
+            // Multisample support was never queried here: every format was reported as
+            // MultisampleCount.None, so Texture.New threw "the maximum supported level is None" for
+            // any multisampled render target - MSAA was simply unusable on Vulkan. Stride.Voxels'
+            // dominant-axis voxelization allocates an 8x target and died on it.
+            //
+            // Only the framebuffer masks are read, not per-format image properties: this runs for
+            // all 256 PixelFormat values at device creation, and the framebuffer limits are what
+            // actually bound a render target's sample count.
+            var physicalDevice = deviceRoot.NativePhysicalDevice;
+            deviceRoot.NativeInstanceApi.vkGetPhysicalDeviceProperties(physicalDevice, out var physicalDeviceProperties);
+            var colorSampleCounts = physicalDeviceProperties.limits.framebufferColorSampleCounts;
+            var depthSampleCounts = physicalDeviceProperties.limits.framebufferDepthSampleCounts;
+
+            static MultisampleCount MaximumSampleCount(VkSampleCountFlags supported)
+            {
+                if ((supported & VkSampleCountFlags.Count8) != 0)
+                    return MultisampleCount.X8;
+                if ((supported & VkSampleCountFlags.Count4) != 0)
+                    return MultisampleCount.X4;
+                if ((supported & VkSampleCountFlags.Count2) != 0)
+                    return MultisampleCount.X2;
+                return MultisampleCount.None;
+            }
+
+            // Reported per format, but computed from the intersection of the colour and depth masks
+            // rather than per format: telling colour and depth formats apart here would need a
+            // predicate Stride does not have on this side, and the two masks are identical on every
+            // driver worth supporting. Erring low only costs a sample count; erring high would hand
+            // out a count the device cannot attach.
+            var maximumMultisampleCount = MaximumSampleCount(colorSampleCounts & depthSampleCounts);
+
             for (int i = 0; i < mapFeaturesPerFormat.Length; i++)
-                mapFeaturesPerFormat[i] = new FeaturesPerFormat((PixelFormat) i, MultisampleCount.None, ComputeShaderFormatSupport.None, FormatSupport.None);
+                mapFeaturesPerFormat[i] = new FeaturesPerFormat((PixelFormat) i, maximumMultisampleCount, ComputeShaderFormatSupport.None, FormatSupport.None);
             //// Check features for each DXGI.Format
             //foreach (var format in Enum.GetValues(typeof(SharpDX.DXGI.Format)))
             //{
--- a/sources/engine/Stride.Graphics/Vulkan/GraphicsDeviceFeatures.Vulkan.cs
+++ b/sources/engine/Stride.Graphics/Vulkan/GraphicsDeviceFeatures.Vulkan.cs
@@ -42,31 +42,15 @@ namespace Stride.Graphics
 
             HasResourceRenaming = false;
 
-            // Only the framebuffer sample count limits are read, not per-format image properties:
-            // this runs for every PixelFormat at device creation, and the framebuffer limits are what
-            // bound a render target's sample count.
             var physicalDevice = deviceRoot.NativePhysicalDevice;
-            deviceRoot.NativeInstanceApi.vkGetPhysicalDeviceProperties(physicalDevice, out var physicalDeviceProperties);
-            var colorSampleCounts = physicalDeviceProperties.limits.framebufferColorSampleCounts;
-            var depthSampleCounts = physicalDeviceProperties.limits.framebufferDepthSampleCounts;
-
-            static MultisampleCount MaximumSampleCount(VkSampleCountFlags supported)
-            {
-                if ((supported & VkSampleCountFlags.Count8) != 0)
-                    return MultisampleCount.X8;
-                if ((supported & VkSampleCountFlags.Count4) != 0)
-                    return MultisampleCount.X4;
-                if ((supported & VkSampleCountFlags.Count2) != 0)
-                    return MultisampleCount.X2;
-                return MultisampleCount.None;
-            }
-
-            // Use the intersection of the color and depth masks for every format: there is no
-            // color/depth predicate on PixelFormat here, and erring low only costs a sample count.
-            var maximumMultisampleCount = MaximumSampleCount(colorSampleCounts & depthSampleCounts);
+            var instanceApi = deviceRoot.NativeInstanceApi;
 
             for (int i = 0; i < mapFeaturesPerFormat.Length; i++)
-                mapFeaturesPerFormat[i] = new FeaturesPerFormat((PixelFormat) i, maximumMultisampleCount, ComputeShaderFormatSupport.None, FormatSupport.None);
+            {
+                var pixelFormat = (PixelFormat) i;
+                var maximumMultisampleCount = GetMaximumMultisampleCount(instanceApi, physicalDevice, pixelFormat);
+                mapFeaturesPerFormat[i] = new FeaturesPerFormat(pixelFormat, maximumMultisampleCount, ComputeShaderFormatSupport.None, FormatSupport.None);
+            }
             //// Check features for each DXGI.Format
             //foreach (var format in Enum.GetValues(typeof(SharpDX.DXGI.Format)))
             //{
@@ -87,6 +71,29 @@ namespace Stride.Graphics
             //    //mapFeaturesPerFormat[(int)dxgiFormat] = new FeaturesPerFormat((PixelFormat)dxgiFormat, maximumMultisampleCount, computeShaderFormatSupport, formatSupport);
             //    mapFeaturesPerFormat[(int)dxgiFormat] = new FeaturesPerFormat((PixelFormat)dxgiFormat, maximumMultisampleCount, formatSupport);
             //}
+        }
+
+        private static MultisampleCount GetMaximumMultisampleCount(VkInstanceApi instanceApi, VkPhysicalDevice physicalDevice, PixelFormat pixelFormat)
+        {
+            if (!VulkanConvertExtensions.TryConvertPixelFormat(pixelFormat, out var format, out _, out _))
+                return MultisampleCount.None;
+
+            // Same usage as Texture.CreateImage for a render target or depth stencil of that format
+            var usage = VkImageUsageFlags.TransferSrc | VkImageUsageFlags.TransferDst;
+            usage |= Texture.IsDepthFormat(pixelFormat) ? VkImageUsageFlags.DepthStencilAttachment : VkImageUsageFlags.ColorAttachment;
+
+            var result = instanceApi.vkGetPhysicalDeviceImageFormatProperties(physicalDevice, format, VkImageType.Image2D, VkImageTiling.Optimal, usage, VkImageCreateFlags.None, out var imageFormatProperties);
+            if (result != VkResult.Success)
+                return MultisampleCount.None;
+
+            var sampleCounts = imageFormatProperties.sampleCounts;
+            if ((sampleCounts & VkSampleCountFlags.Count8) != 0)
+                return MultisampleCount.X8;
+            if ((sampleCounts & VkSampleCountFlags.Count4) != 0)
+                return MultisampleCount.X4;
+            if ((sampleCounts & VkSampleCountFlags.Count2) != 0)
+                return MultisampleCount.X2;
+            return MultisampleCount.None;
         }
     }
 }
--- a/sources/engine/Stride.Graphics/Vulkan/GraphicsDeviceFeatures.Vulkan.cs
+++ b/sources/engine/Stride.Graphics/Vulkan/GraphicsDeviceFeatures.Vulkan.cs
@@ -42,14 +42,9 @@ namespace Stride.Graphics
 
             HasResourceRenaming = false;
 
-            // Multisample support was never queried here: every format was reported as
-            // MultisampleCount.None, so Texture.New threw "the maximum supported level is None" for
-            // any multisampled render target - MSAA was simply unusable on Vulkan. Stride.Voxels'
-            // dominant-axis voxelization allocates an 8x target and died on it.
-            //
-            // Only the framebuffer masks are read, not per-format image properties: this runs for
-            // all 256 PixelFormat values at device creation, and the framebuffer limits are what
-            // actually bound a render target's sample count.
+            // Only the framebuffer sample count limits are read, not per-format image properties:
+            // this runs for every PixelFormat at device creation, and the framebuffer limits are what
+            // bound a render target's sample count.
             var physicalDevice = deviceRoot.NativePhysicalDevice;
             deviceRoot.NativeInstanceApi.vkGetPhysicalDeviceProperties(physicalDevice, out var physicalDeviceProperties);
             var colorSampleCounts = physicalDeviceProperties.limits.framebufferColorSampleCounts;
@@ -66,11 +61,8 @@ namespace Stride.Graphics
                 return MultisampleCount.None;
             }
 
-            // Reported per format, but computed from the intersection of the colour and depth masks
-            // rather than per format: telling colour and depth formats apart here would need a
-            // predicate Stride does not have on this side, and the two masks are identical on every
-            // driver worth supporting. Erring low only costs a sample count; erring high would hand
-            // out a count the device cannot attach.
+            // Use the intersection of the color and depth masks for every format: there is no
+            // color/depth predicate on PixelFormat here, and erring low only costs a sample count.
             var maximumMultisampleCount = MaximumSampleCount(colorSampleCounts & depthSampleCounts);
 
             for (int i = 0; i < mapFeaturesPerFormat.Length; i++)
--- a/sources/engine/Stride.Graphics/Vulkan/GraphicsDeviceFeatures.Vulkan.cs
+++ b/sources/engine/Stride.Graphics/Vulkan/GraphicsDeviceFeatures.Vulkan.cs
@@ -48,7 +48,7 @@ namespace Stride.Graphics
             for (int i = 0; i < mapFeaturesPerFormat.Length; i++)
             {
                 var pixelFormat = (PixelFormat) i;
-                var maximumMultisampleCount = GetMaximumMultisampleCount(instanceApi, physicalDevice, pixelFormat);
+                var maximumMultisampleCount = GetMaximumMultisampleCount(deviceRoot, instanceApi, physicalDevice, pixelFormat);
                 mapFeaturesPerFormat[i] = new FeaturesPerFormat(pixelFormat, maximumMultisampleCount, ComputeShaderFormatSupport.None, FormatSupport.None);
             }
             //// Check features for each DXGI.Format
@@ -73,14 +73,23 @@ namespace Stride.Graphics
             //}
         }
 
-        private static MultisampleCount GetMaximumMultisampleCount(VkInstanceApi instanceApi, VkPhysicalDevice physicalDevice, PixelFormat pixelFormat)
+        private static MultisampleCount GetMaximumMultisampleCount(GraphicsDevice deviceRoot, VkInstanceApi instanceApi, VkPhysicalDevice physicalDevice, PixelFormat pixelFormat)
         {
             if (!VulkanConvertExtensions.TryConvertPixelFormat(pixelFormat, out var format, out _, out _))
                 return MultisampleCount.None;
 
             // Same usage as Texture.CreateImage for a render target or depth stencil of that format
             var usage = VkImageUsageFlags.TransferSrc | VkImageUsageFlags.TransferDst;
-            usage |= Texture.IsDepthFormat(pixelFormat) ? VkImageUsageFlags.DepthStencilAttachment : VkImageUsageFlags.ColorAttachment;
+            if (Texture.IsDepthFormat(pixelFormat))
+            {
+                usage |= VkImageUsageFlags.DepthStencilAttachment;
+                // CreateImage substitutes a supported depth-stencil format, so query the one it would really create
+                format = Texture.GetFallbackDepthStencilFormat(deviceRoot, format);
+            }
+            else
+            {
+                usage |= VkImageUsageFlags.ColorAttachment;
+            }
 
             var result = instanceApi.vkGetPhysicalDeviceImageFormatProperties(physicalDevice, format, VkImageType.Image2D, VkImageTiling.Optimal, usage, VkImageCreateFlags.None, out var imageFormatProperties);
             if (result != VkResult.Success)
--- a/sources/engine/Stride.Voxels/Voxels/Voxelization/VoxelStorage/Processing/ClearBuffer.sdsl
+++ b/sources/engine/Stride.Voxels/Voxels/Voxelization/VoxelStorage/Processing/ClearBuffer.sdsl
@@ -6,9 +6,20 @@ namespace Stride.Rendering.Voxels
     {
         stage RWBuffer<uint> buffer;
         int offset;
+
+        // Direct3D11 allows at most 65535 thread groups per dispatch dimension, so a purely
+        // one-dimensional dispatch cannot address more than 65535 * 1024 elements. A 256^3
+        // anisotropic clipmap needs three times that, so the groups are spread over X and Y and
+        // the linear index is recomposed here. rowLength is how many elements one row of groups
+        // covers, and count bounds the last row, which generally overshoots.
+        int rowLength;
+        int count;
+
         override void Compute()
         {
-            buffer[streams.DispatchThreadId.x + offset] = 0;
+            uint index = streams.DispatchThreadId.x + streams.DispatchThreadId.y * (uint)rowLength;
+            if (index < (uint)count)
+                buffer[index + offset] = 0;
         }
     };
 }
--- a/sources/engine/Stride.Voxels/Voxels/Voxelization/VoxelStorage/Processing/ClearBuffer.sdsl
+++ b/sources/engine/Stride.Voxels/Voxels/Voxelization/VoxelStorage/Processing/ClearBuffer.sdsl
@@ -7,11 +7,12 @@ namespace Stride.Rendering.Voxels
         stage RWBuffer<uint> buffer;
         int offset;
 
-        // Direct3D11 allows at most 65535 thread groups per dispatch dimension, so a purely
-        // one-dimensional dispatch cannot address more than 65535 * 1024 elements. A 256^3
-        // anisotropic clipmap needs three times that, so the groups are spread over X and Y and
-        // the linear index is recomposed here. rowLength is how many elements one row of groups
-        // covers, and count bounds the last row, which generally overshoots.
+        // A dispatch may have at most 65535 thread groups per dimension, on every API alike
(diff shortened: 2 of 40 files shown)


sources/engine/Stride.Voxels/Voxels/Voxelization/VoxelStorage/Processing/ClearBuffer.sdsl
```diff
@@ -6,9 +6,20 @@ namespace Stride.Rendering.Voxels
     {
         stage RWBuffer<uint> buffer;
         int offset;
+
+        // Direct3D11 allows at most 65535 thread groups per dispatch dimension, so a purely
+        // one-dimensional dispatch cannot address more than 65535 * 1024 elements. A 256^3
+        // anisotropic clipmap needs three times that, so the groups are spread over X and Y and
+        // the linear index is recomposed here. rowLength is how many elements one row of groups
+        // covers, and count bounds the last row, which generally overshoots.
```
Ethereal77: As this is a limitation specific to Direct3D 11, maybe it would be better to enclose this in a
```csharp
#if STRIDE_GRAPHICS_API_DIRECT3D11
...
#endif
```
so as to not damage the performance of other APIs with extra data transfer or unnecessary computation.

Shaders also get those defines set when compiling. [See here](https://github.com/stride3d/stride/blob/44cc820f44742837dd5e3fb7d89a1d6117e2cfc8/sources/shaders/Stride.Shaders.Compilers/EffectCompiler.cs#L130).

sources/engine/Stride.Voxels/Voxels/Voxelization/VoxelStorage/VoxelStorageClipmaps.cs
```diff
@@ -246,6 +246,30 @@ public void CollectVoxelizationPasses(ProcessedVoxelVolume data, VoxelStorageCon
...
+        /// <summary>
+        /// Thread group counts for clearing <paramref name="elementCount"/> buffer elements, spread
+        /// over two dimensions.
+        /// </summary>
+        /// <remarks>
+        /// Direct3D11 rejects a dispatch with more than 65535 groups in any one dimension, which
+        /// caps a one-dimensional clear at about 67 million elements. A 256^3 clipmap storing six
+        /// directions needs 201 million, so it failed outright. ClearBuffer recomposes the linear
+        /// index from X and Y, using <paramref name="rowLength"/> as the width of one row of groups.
+        /// </remarks>
```
Ethereal77: This is a personal preference of mine, but I usually see XMLdocs as what I want to see in Intellisense or when hovering over some symbol. So, I'd keep just the bare minimum description about what the method does or what it is for, but those comments about D3D11-specific problems I'd move inside the method as regular comments. It does not add much value to a caller of the method that the implementation of a specific API has some drawbacks, imho.

sources/shaders/Stride.Shaders.Compilers/SDSL/SDSLC.cs
```diff
@@ -122,6 +122,13 @@ public readonly bool Compile(string? filename, string code, ObjectId hash, ReadO
             {
                 // Ignore (using C# codegen for now)
             }
+            else if (declaration is UsingShaderNamespace)
+            {
+                // Ignore: shader classes are resolved by name across every loaded namespace, so a
+                // `using` at the top of an .sdsl file carries no information the compiler needs.
+                // Erroring on it took down every shader that has one — Stride.Voxels'
+                // LightVoxelShader among them, which made voxel GI unusable.
+            }
```
Ethereal77, line 129: Excuse me if I'm misunderstanding this one.

`using` and `namespace`s don't work? Is this a regression or a problem with the new compiler?

The point of having namespaces in the SDSL language is the same as in C#: they serve as a means to avoid name clashes.

What's the point of using namespaces if they are not respected by the compiler?

sources/shaders/Stride.Shaders.Compilers/SDSL/ShaderMixer.Reflection.cs
```diff
@@ -211,6 +211,51 @@ private void RenameVariables(MixinGlobalContext globalContext, SpirvContext cont
...
+    /// start past its render targets - FXC otherwise rejects the shader with X4509. The engine
+    /// already assumes this: CommandList.OMSetSingleUnorderedAccessView binds with
+    /// <c>UAVStartSlot: currentRenderTargetViewsActiveCount</c> and indexes
+    /// <c>slot - currentRenderTargetViewsActiveCount</c>, which is negative if a UAV took u0.
+    /// </para>
+    /// <para>
+    /// Counted as the highest output Location plus one, so multiple render targets are handled,
+    /// and left at 0 when the module has no pixel shader - a compute shader keeps u0.
+    /// </para>
+    /// </summary>
```
Ethereal77, line 217: Similarly to my other comment, I think this `<summary>` contains too much too specific to D3D11 information.

I'd prefer to describe what the method does and a brief explanation of why in the remarks, and explain the D3D11-specific shenanigans in a comment inside for mantainers.

Oh, and as it is a method, I tend to prefer the verb-form both for the method name and the description (so `GetFirst...` and `Retrieves the...`. This is a personal take, however.

sources/shaders/Stride.Shaders.Parsers/Core/SymbolTypes.cs
```diff
@@ -69,19 +69,34 @@ public static bool TryGetBufferType(string name, TypeName? templateTypeName, [Ma
...
 
+        // An image - texture or typed buffer alike - cannot deliver 16-bit values: Vulkan requires
+        // its sampled type to be a 32-bit int, 64-bit int or 32-bit float
+        // (VUID-StandaloneSpirv-OpTypeImage-04656), and there is no extension that lifts it.
+        //
+        // Nothing is lost. The 16 bits live in the resource's pixel format, not in the shader's
+        // declaration: the texture unit decodes the stored halfs and delivers 32-bit floats to the
+        // registers, for free. `half` is already an alias of `float` in shader model 5 anyway, and
+        // code that wants native 16-bit registers converts after the read.
+        static SymbolType WidenImageElementType(SymbolType elementType) => elementType switch
```
Ethereal77: Does this not have any drawback? I mean, the old way was compiled succesfully by the old Stride / FXC shader compiler. Did it also widen under the hood?

I mean, even if when fetching a `half` it gets auto-converted to `float` for free by the HW, modern GPUs still have `half`-precission ALUs with better issue rate, if I'm not mistaken. I've seen shaders specifically optimized to use `half`s because of this. Is this not valid anymore?

Note also that Stride SPIR-V is a Stride-specific dialect not specific to Vulkan, but cross-compiled to be used by DX APIs also.

sources/shaders/Stride.Shaders.Parsers/Parsing/SDSL/AST/Expression.cs
```diff
@@ -364,14 +369,36 @@ protected void ProcessInputArguments(SymbolTable table, CompilerUnit compiler, F
...
+    /// <item>Opaque resources (see <see cref="SymbolTypeExtensions.IsOpaqueResource"/>): Vulkan
+    /// forbids OpStore to them, so they cannot live in a Function variable.</item>
+    /// <item>Geometry streams: appending goes through OpEmitVertexSDSL and the stage's output
+    /// variables, so the object holds nothing to copy - and the copy outlived the parameter, which
+    /// the interface processor removes from the signature, leaving SPIR-V reading an id that no
+    /// longer exists.</item>
+    /// </list>
+    /// The input and output sides both consult this: copying a result back out of something that
+    /// was never copied in would write through a pointer the callee already holds.
+    /// </summary>
```
Ethereal77: Same thing I wrote above about comments.

This is mixing the description of the method with a "commit-like" explanation.

sources/shaders/Stride.Shaders.Parsers/Parsing/SDSL/AST/Expression.cs
```diff
@@ -670,6 +697,53 @@ public partial class AccessorChainExpression(Expression source, TextLocation inf
...
+    /// one texel - rather than the image read indexing normally produces. Returns false when the
+    /// chain is not an atomic-capable buffer index, leaving the caller to compile it normally.
+    /// <para>
+    /// A typed buffer is not memory the shader can point into: it is a storage image, so SDSL
+    /// compiles <c>buffer[i]</c> to an OpImageRead and <c>buffer[i] = x</c> to an OpImageWrite,
+    /// both of which deal in values. An atomic needs the memory itself, and OpImageTexelPointer is
+    /// the only instruction that hands it over. Its result may only be consumed by atomics, which
+    /// is why this is offered to the `ref` argument path instead of being how every index compiles.
+    /// </para>
+    /// </summary>
```
Ethereal77: Again, too dense of a summary.

Better structure as a quick description in `<summary>`, a more thorough explanation in `<remarks>`, and what it returns in `<returns>`, instead of shoving all in a giant summary.

sources/shaders/Stride.Shaders.Parsers/Spirv/Processing/Interfaces/InterfaceProcessor.cs
```diff
@@ -55,21 +55,68 @@ internal EntryPointInfo(string name, int id, ExecutionModel model, List<int> int
...
+        /// through their composition variable and so can never be the shader's entry point.
+        /// </summary>
+        /// <remarks>
+        /// A composition that happens to inherit the same base as the shader it is composed into
+        /// inherits that base's entry point too, and lands in the same method group. Picking the
+        /// group's last member then picks the composition's copy: Stride.Voxels' Voxel2x2x2Mipmap
+        /// composes a Voxel2x2x2Mipmapper and both derive from ComputeShaderBase, so CSMain
+        /// resolved to the composition's, which calls the empty base Compute(). The real body -
+        /// and, once dead code was removed, the mipmap textures with it - disappeared, and voxel
+        /// GI silently contributed nothing.
```
Ethereal77: Better explain what it does and why it is needed, but without pointing specifically at `Voxel2x2x2Mipmap` (that is an implementation detail). That part is better inside as a comment for mantainers explaining why this was added, not as the thing that shows when hovering over the symbol.

sources/shaders/Stride.Shaders.Parsers/Spirv/Processing/Interfaces/InterfaceProcessor.cs
```diff
@@ -293,6 +344,91 @@ public Result Process(SymbolTable table, SpirvBuffer buffer, SpirvContext contex
...
+        /// OutputLineStrip / OutputTriangleStrip execution mode. So it survives until here, where
+        /// the entry point has already been stripped of it and everything else still has to be.
+        /// </para>
+        /// <para>
+        /// The function type is rewritten through GetOrRegister rather than in place: a method and
+        /// the entry point calling it share one OpTypeFunction when their signatures match, and
+        /// mutating it for one silently rewrites the other - which is how the entry point's own
+        /// removal left such a method with more parameters than its type declared.
+        /// </para>
+        /// </summary>
```
Ethereal77: Same as the last ones. Giant `<summary>`, too much too specific detail in there.

sources/shaders/Stride.Shaders.Tests/RenderingTests.cs
```diff
@@ -91,7 +91,7 @@ public void ComputeTest1(string shaderName, RendererBackend backend)
         File.WriteAllText($"{outputName}.spvdis", Spv.Dis(SpirvBytecode.CreateFromSpan(bytecode), DisassemblerFlags.Name | DisassemblerFlags.Id | DisassemblerFlags.InstructionIndex, true));
 
         // Validate SPIR-V
-        var validationResult = Spv.ValidateFile($"{outputName}.spv");
+        var validationResult = Spv.ValidateFile($"{outputName}.spv", targetVulkan: true);
```
Ethereal77: Even though the new shader backend uses SPIR-V as an intermediary format, it is a Stride-specific dialect used for cross-compilation to different backends.

Is this fixed `targetVulkan: true` desirable?

Ethereal77 (commented):
Lets wait for xen, the new SPIR-V shader compiler is his child.

In the meantime, I've left some comments. The ones about docs are my personal opinion, to aim for a better, more clear and readable and less intimidating documentation for end users vs. information specific for mantainers.

Also, I like to reinforce my point that (from what I saw when it was being implemented) Stride SPIR-V is a Stride-specific dialect not specific to Vulkan, but cross-compiled to be used by DX APIs also. Do not assume when the word SPIR-V appears it belongs automatically to the Vulkan API.

sources/engine/Stride.Voxels/Voxels/Voxelization/VoxelStorage/Processing/ClearBuffer.sdsl
Nicogo1705: The 65535 groups-per-dimension cap is not specific to Direct3D 11: D3D12's `Dispatch` documents the same `D3D11_CS_DISPATCH_MAX_THREAD_GROUPS_PER_DIMENSION (65535)` limit, and Vulkan's required minimum for `maxComputeWorkGroupCount` is 65535 per dimension too. So the 2D dispatch is needed on every backend, and it costs one multiply-add per thread. I've reworded the comment to say so rather than suggest a D3D11 branch.

Refs: https://learn.microsoft.com/en-us/windows/win32/api/d3d12/nf-d3d12-id3d12graphicscommandlist-dispatch and https://docs.vulkan.org/refpages/latest/refpages/source/Required_Limits.html

sources/engine/Stride.Voxels/Voxels/Voxelization/VoxelStorage/VoxelStorageClipmaps.cs
Nicogo1705: Done: one-line summary with `param`/`returns`, and the specifics moved inside the method as a comment.

sources/shaders/Stride.Shaders.Compilers/SDSL/SDSLC.cs
Nicogo1705: I don't know whether it is a regression - I haven't tried this on 4.3. What I can say is the current behaviour, which I've pinned down in tests (`NamespaceTests`): a shader class is resolved by its name alone (`ShaderSourceManager.LoadShaderSource(type)` loads `{type}.sdsl`), so a `using` takes no part in resolution; a `using` of a namespace no shader lives in is not an error; and two classes of one name in two namespaces of one file are not told apart by a `using` - the last declared one is mixed in. Rejecting the directive failed every shader that has one (Stride.Voxels' `LightVoxelShader`), so it is accepted and skipped. If namespaces are meant to take part in resolution, that is a separate change, and the tests will have to be updated knowingly.

sources/shaders/Stride.Shaders.Compilers/SDSL/ShaderMixer.Reflection.cs
Nicogo1705: Done: renamed to `GetFirstUnorderedAccessSlot`, short summary with `remarks` and `returns`, and the D3D11 register-space detail moved inside as a comment.

sources/shaders/Stride.Shaders.Parsers/Core/SymbolTypes.cs
Nicogo1705: No drawback, and it is what FXC did: per the HLSL docs `half` is "provided only for language compatibility" and "Direct3D 10 shader targets map all half data types to float data types", so `Texture3D<half4>` already declared a 32-bit register under the old compiler. The 16 bits live in the resource's pixel format (`R16G16B16A16_Float`), not in the shader declaration - the texture unit decodes them and hands floats to the ALU. Native 16-bit arithmetic is `min16float` (or `float16_t` with 16-bit types enabled), a separate type that is unaffected. On the SPIR-V side, `OpTypeImage`'s sampled type must be 32-bit for the module to validate, whichever backend it is cross-compiled to. The comment now states both.

Ref: https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-scalar

sources/shaders/Stride.Shaders.Parsers/Parsing/SDSL/AST/Expression.cs
Nicogo1705: Done: summary, `remarks` and `returns` split.

sources/shaders/Stride.Shaders.Parsers/Parsing/SDSL/AST/Expression.cs
Nicogo1705: Done: short summary, the explanation in `remarks`, and a `returns`.

sources/shaders/Stride.Shaders.Parsers/Spirv/Processing/Interfaces/InterfaceProcessor.cs
Nicogo1705: Done: the summary describes the mechanism; the `Voxel2x2x2Mipmap` case is now an inline note for maintainers.

sources/shaders/Stride.Shaders.Parsers/Spirv/Processing/Interfaces/InterfaceProcessor.cs
Nicogo1705: Done: summary and `remarks` split, the shared `OpTypeFunction` detail moved inside as a comment.

sources/shaders/Stride.Shaders.Tests/RenderingTests.cs
Nicogo1705: Agreed, it is a dialect. The tests now validate against Vulkan's rules only when Vulkan is the backend under test, which is what `EffectCompiler` does (`targetVulkan: effectParameters.Platform is GraphicsPlatform.Vulkan`).

Nicogo1705:
Thanks for the review. Docs reworked throughout as you describe: short summaries for users, the specifics as inline comments for maintainers. Point taken on the dialect: the "Vulkan requires" phrasing is gone where the constraint is SPIR-V's, and the rendering tests validate against Vulkan's rules only for the Vulkan backend. On namespaces I've added tests that pin down the current behaviour rather than claim what 4.3 did. Happy to wait for xen on the compiler side.

Nicogo1705:
Rebased on master and comments trimmed as on the other PRs (1 to 2 lines, no history). No code change since the last review round. Note: this shares three files with #3389; whichever merges second will need a rebase.

sources/engine/Stride.Graphics/Vulkan/GraphicsDeviceFeatures.Vulkan.cs
```diff
@@ -40,9 +42,31 @@ internal GraphicsDeviceFeatures(GraphicsDevice deviceRoot)
...
+                return MultisampleCount.None;
+            }
+
+            // Use the intersection of the color and depth masks for every format: there is no
+            // color/depth predicate on PixelFormat here, and erring low only costs a sample count.
+            var maximumMultisampleCount = MaximumSampleCount(colorSampleCounts & depthSampleCounts);
+
             for (int i = 0; i < mapFeaturesPerFormat.Length; i++)
-                mapFeaturesPerFormat[i] = new FeaturesPerFormat((PixelFormat) i, MultisampleCount.None, ComputeShaderFormatSupport.None, FormatSupport.None);
+                mapFeaturesPerFormat[i] = new FeaturesPerFormat((PixelFormat) i, maximumMultisampleCount, ComputeShaderFormatSupport.None, FormatSupport.None);
```
xen2, line 69: Doesn't this setup MSAA on every PixelFormat? (incl BC ones, etc.).
I think this won't work in `vkCreateImage` later.
Look into `vkGetPhysicalDeviceImageFormatProperties` (return sample count per format)

sources/shaders/Stride.Shaders.Parsers/Spirv/Processing/Interfaces/InterfaceProcessor.cs
```diff
@@ -293,6 +343,84 @@ public Result Process(SymbolTable table, SpirvBuffer buffer, SpirvContext contex
...
+        /// <remarks>
+        /// The parameter carries no data (appending goes through OpEmitVertexSDSL) but must survive
+        /// until EntryPointWrapperGenerator has read the output topology off it.
+        /// </remarks>
+        private static void RemoveGeometryStreamParameters(SpirvBuffer buffer, SpirvContext context)
+        {
+            // Function types are rewritten through GetOrRegister, not in place: functions with the same
+            // signature share one OpTypeFunction, and mutating it for one would rewrite the others.
+            // Which parameter index each function loses.
+            var removedParameters = new Dictionary<int, int>();
```
xen2, line 359: Probably rare, but this won't work for a method which has 2 stream, i.e. `void Emit(inout PointStream<Output> a, inout PointStream<Output> b)`

sources/shaders/Stride.Shaders.Compilers/SDSL/ShaderMixer.ShaderSourceEvaluator.cs
```diff
@@ -138,6 +138,29 @@ private void ProcessClasses(IExternalShaderLoader shaderLoader, SpirvContext con
             };
         }
 
+        // --- Step 2b: Build the promote-composition-to-parent callback ---
+        // Same as above, for the value of a `stage compose` rather than the shader declaring it.
+        // The root's callback writes into its compositions; nested levels pass the parent's along.
```
xen2, line 143: I need to dig a bit more about that one, I will review that part later.

xen2: I think we need to change a bit what is allowed in ShaderMixinSource to completely avoid this kind of issue.
I didn't understand the `ReferenceEquals` at first because it would almost always trigger due to how we generate those arrays, then I tried to review the use cases.
Also, I couldn't figure out if we needed to fail, or concat them, etc.

So, what about those new rules:
- a `stage compose` can be declared anywhere (as today)
- stage compose variables can only be declared on the root (where the compose end up being)

If we do that, we would need to do this:

Keep from the PR:
- eafee2e: the named error replacing the raw mixinSource.Compositions[variable.Key] indexer in MergeMixinNode. Valuable on its own.

Drop from the PR:
- `promoteCompositionToParent` and its closure
- StageCompositionPaths on ShaderMixinInstantiation
- the compositionPath / Nest() threading through EvaluateInheritanceAndCompositions → ProcessClasses → ProcessCompositions
- the basePath restore in MergeMixinNode
- the duplicate-value throw

Add in `ProcessCompositions`, check would be:

```
if (isSupplied && (variable.Flags & VariableFlagsMask.Stage) != 0 && !isRoot)
    log.Error($"'{variableName}' is a `stage compose` declared by '{shaderName}', so it is one slot "
            + $"for the whole effect and must be supplied at the root. It was supplied at '{compositionPath}'.");
```

Then some existing effects need adjustments, like Voxel GI can drop the `stage` from `AttributeSamplers` and adjust C# code that set this collection. Need to check other usage (Light shafts, etc.)

**Maybe the best approach here is to drop those specific controversial commits from PR (4df2f3d  and 6a90900).
That way we can merge the current PR more easily, and then we can deal with that in another discussion/PR?**

sources/engine/Stride.Graphics/Vulkan/GraphicsDeviceFeatures.Vulkan.cs
Nicogo1705: Right, it did. `f741c7a`: the count is now queried per format with `vkGetPhysicalDeviceImageFormatProperties`, using the same image type, tiling and usage `CreateImage` gives an attachment (`DepthStencilAttachment` for the depth formats, `ColorAttachment` otherwise). Formats without a Vulkan equivalent, or that the device refuses for that usage (BC and friends), report `MultisampleCount.None`. That needed a `TryConvertPixelFormat` next to `ConvertPixelFormat`, since about half of the `PixelFormat` values have no mapping.

Note that `CreateImage` still hardcodes `samples = Count1`, so Vulkan never actually creates a multisampled image today; I left that for a separate fix.

sources/shaders/Stride.Shaders.Parsers/Spirv/Processing/Interfaces/InterfaceProcessor.cs
Nicogo1705: Fixed in `978bfdd`: the processor kept one removed index per function, so the second stream parameter overwrote the first and shifted the index in the function type. It now keeps every index and rebuilds the call without all of them. Test `GeometryStreamsMethodWithTwoStreamParameters` (a helper taking two `TriangleStream<Output>`), failing before, passing after.

sources/shaders/Stride.Shaders.Compilers/SDSL/ShaderMixer.ShaderSourceEvaluator.cs
Nicogo1705: Agreed, dropped from this PR: `4df2f3d` and `6a90900` are gone, with their tests; `eafee2e` stays. The rule is worth its own thread since `Stride.Voxels` supplies `AttributeSamplers` from a light, which is always nested under `environmentLights[i]`, so "supply at the root" is not something a light renderer can do. I opened #3423 with your proposed rules and that constraint. Until it is settled this PR does not get `Stride.Voxels` through the mixer on its own; the remaining commits are all general fixes.

sources/shaders/Stride.Shaders.Compilers/SDSL/ShaderMixer.Reflection.cs
xen2: Line 269 still has comment "See FirstUnorderedAccessSlot." (old method name)

sources/shaders/Stride.Shaders.Compilers/SDSL/SDSLC.cs
xen2: Yes, namespace exist in the language but are ignored in practice by the compiler/resolver.
(I didn't touch this part when reimplementing new SDSL, I wanted to separate this as a follow up to reduce scope)

sources/shaders/Stride.Shaders.Compilers/SDSL/ShaderMixer.Reflection.cs
```diff
@@ -222,6 +264,12 @@ private unsafe void ProcessReflection(MixinGlobalContext globalContext, SpirvCon
         ref var cbufferSlot = ref slotCounts[options.ResourcesRegisterSeparate ? 2 : 0];
         ref var uavSlot = ref slotCounts[options.ResourcesRegisterSeparate ? 3 : 0];
 
+        // Only in the separate (Direct3D11) register space: a pixel shader's UAVs start past its
+        // render targets. Vulkan and D3D12 share one counter and one descriptor set, where the
+        // rule does not exist. See FirstUnorderedAccessSlot.
```
xen2:
```suggestion
        // rule does not exist. See GetFirstUnorderedAccessSlot.
```

sources/shaders/Stride.Shaders.Compilers/SDSL/ShaderMixer.Reflection.cs
xen2: Note: did it myself with a suggestion and applied it

PR #3383, opened 2026-08-31
feat: Enhance Stride.Launcher

Jklawreszuk:
# PR Details

This PR includes a list of visual improvements related to Stride.Launcher, designed to improve its appearance and ensure it blends in better with the Fluent2 design, from which Avalonia draws its inspiration: 
Source: https://fluent2.microsoft.design/

image

## Features

- Implement/Port custom TitleBar from WPF Launcher
- Add icons for "Getting Started" and News Tab 
- Add icon for News cards 
- Add styling for "primary" button
- Unify colors and some transparency for buttons 
- Small margin/padding tweaks, few refactoring changes for styles
- Set minimum size for Window
- Stretch community buttons
- Rounding corners for cards
- Add some text when synchronizing

## Types of changes

## Checklist


sources/launcher/Stride.Launcher/Views/MainView.axaml
```diff
@@ -13,34 +13,12 @@
...
-        <ResourceDictionary x:Key="Light">
-          <!-- FIXME xplat-editor find a background image working better with the light variant -->
-          <ImageBrush x:Key="ImageBackground"
-                      Source="/Assets/Images/robot.png" Stretch="UniformToFill" Opacity="0.4"/>
-        </ResourceDictionary>
-      </ResourceDictionary.ThemeDictionaries>
-    </ResourceDictionary>
-  </UserControl.Resources>
-  <UserControl.Background>
-    <DynamicResource ResourceKey="ImageBackground"/>
```
Kryptos-FR, line 35: Why remove the background image?

Jklawreszuk: It's just temporary, I'm experimenting with making it more flat 😅

Kryptos-FR: ok. I'll let you continue without interrupting. Let me know if you have something you want me to review.

Jklawreszuk:
Well, I'm not sure if the code is the most elegant, but I did the best I could. The changes are ready for review. @Kryptos-FR

sources/Directory.Packages.props
```diff
@@ -10,6 +10,7 @@
   <ItemGroup>
     <PackageVersion Include="BepuPhysics" Version="2.5.0-beta.28" />
     <PackageVersion Include="CommunityToolkit.HighPerformance" Version="8.4.0" />
+    <PackageVersion Include="CommunityToolkit.Mvvm" Version="8.4.2" />
```
Kryptos-FR, line 13: We don't want that dependency. We are avoiding it on purpose everywhere else.

You should use our own CommandBase implementation.

sources/presentation/Stride.Core.Presentation.Avalonia/Stride.Core.Presentation.Avalonia.csproj
```diff
@@ -19,6 +19,7 @@
     <!--Condition below is needed to remove AvaloniaUI.DiagnosticsSupport package from build output in Release configuration.-->
     <PackageReference Condition="'$(Configuration)' == 'Debug'" Include="AvaloniaUI.DiagnosticsSupport" />
     <PackageReference Include="MarkView.Avalonia" />
+    <PackageReference Include="CommunityToolkit.Mvvm" />
```
Kryptos-FR, line 22: Remove that dependency (see related comment).

sources/presentation/Stride.Core.Presentation.Avalonia/Commands/WindowingCommands.cs
```diff
@@ -0,0 +1,61 @@
...
+using Avalonia.Threading;
+using Avalonia.VisualTree;
+using CommunityToolkit.Mvvm.Input;
+using System;
+
+namespace Stride.Core.Presentation.Avalonia.Commands;
+
+public static class WindowingCommands
+{
+    public static RelayCommand<object> CloseWindowCommand { get; } = new(Close);
```
Kryptos-FR, line 23: Use ICommandBase/AnonymousCommand, or even better bind directly to the method since that's supported by Avalonia.

sources/presentation/Stride.Core.Presentation.Avalonia/Windows/TitleBar.axaml
Kryptos-FR, line 1: Why do we need a custom TitleBar? Avalonia already supports one on a Window.

sources/launcher/Stride.Launcher/App.axaml.cs
```diff
@@ -41,14 +41,6 @@ public override void OnFrameworkInitializationCompleted()
                 DataContext = InitializeMainViewModel()
             };
         }
-        else if (ApplicationLifetime is ISingleViewApplicationLifetime singleViewPlatform)
-        {
-            // don't remove; also used by visual designer.
```
Kryptos-FR, line 46: I think that comment was self-explanatory.

sources/launcher/Stride.Launcher/Views/MainView.axaml.cs
Kryptos-FR, line 1: The split MainView/MainWindow was done on purpose. Why this change?

Kryptos-FR (changes requested):
Don't really agree with the changes. I will need more explanations for them.

sources/Directory.Packages.props
Jklawreszuk: Hm...Interesting, can you explain why is that the case?

Kryptos-FR: One less dependency is one less dependency.

Another reason is that we have our own `DispatcherViewModel` and `CommandBase` implementation that properly deals with Dispatcher issue, i.e. when a property/command is updated outside the UI thread. It also exposes an `IsEnabled` property which allows to disable a command through binding.

sources/presentation/Stride.Core.Presentation.Avalonia/Windows/TitleBar.axaml
Kryptos-FR: It might cause publishing issue on macOS since it won't follow the native window style.

sources/Directory.Packages.props
Jklawreszuk: > One less dependency is one less dependency.

Well... it sounds convincing to me 😅

sources/presentation/Stride.Core.Presentation.Avalonia/Commands/WindowingCommands.cs
Jklawreszuk: Sadly, I've already tried the AnonymousCommand, before adding CommunityToolkit package as a last resort. The thing was if you replace it with the AnonymousCommand the whole app wont work anymore. I dont know why

sources/launcher/Stride.Launcher/Views/MainView.axaml.cs
Jklawreszuk: To be honest, I did it in a rush, mainly focusing on designing and had no idea how to match the titlebar to the background, but hey - I just managed to fix it, so there you go 😅

Jklawreszuk: Changes reverted

sources/launcher/Stride.Launcher/App.axaml.cs
Jklawreszuk: done but tbf the designer in rider was working fine

sources/presentation/Stride.Core.Presentation.Avalonia/Windows/TitleBar.axaml
Jklawreszuk: @Kryptos-FR Do you mean a custom titlebar in general, or the way I implemented it? I know there is WindowDrawnDecorations object but I had no clue how to operate on it because It doesnt do anything when I tried experiment on it. I will definitely look for other projects to find out more.

Jklawreszuk: Also...I think you're kinda right with the macos. I didn't take this into account because currently the launcher doesn't target this platform: only Linux and Windows
 ```xml
<RuntimeIdentifiers>linux-x64;win-x64</RuntimeIdentifiers>
```
But, Yes It doesnt mean  It wont change in the future. So would be nice to think about it right now.

Comment (Kryptos-FR) [reply sources/presentation/Stride.Core.Presentation.Avalonia/Windows/TitleBar.axaml:1]: I had some trouble making WindowDrawnDecorations work. The documentation is lacking but I did manage it somehow. I will push a commit in a derived branch if you want to have a look at it.
Apparently, they are planning to add platform-specific way to customize what the call the "chrome".

Comment (Kryptos-FR) [reply sources/presentation/Stride.Core.Presentation.Avalonia/Windows/TitleBar.axaml:1]: You can find my attempt in branch [`launcher_titlebar`](https://github.com/Jklawreszuk/stride/tree/launcher_titlebar) on your repo. I think its better to use this mechanism. And the day we support macOs, we can rely on [`OnPlatform`](https://docs.avaloniaui.net/docs/platform-specific-guides/xaml) to define a different template or even disable it and use the system one.

Bonus point, it bind automatically the button to the approrpate action so no need form command binding and `RelayCommand`. I'll let you look at it and deal with the "restore button" case: it needs a style switch to change the vertex icon.

File: sources/presentation/Stride.Core.Presentation.Avalonia/Commands/WindowingCommands.cs (hunk 4 above)
Comment (Kryptos-FR) [reply sources/presentation/Stride.Core.Presentation.Avalonia/Commands/WindowingCommands.cs]: Not needed anymore (see my other comment 😉 )

File: sources/presentation/Stride.Core.Presentation.Avalonia/Windows/TitleBar.axaml
Comment (Jklawreszuk) [reply sources/presentation/Stride.Core.Presentation.Avalonia/Windows/TitleBar.axaml:1]: Thanks @Kryptos-FR for the help, overall it simplified the logic. After some search, I was also able to adjust theme little bit (i.e. you can drag a window yey) and move it into separate file so that message dialogs can benefit from it as well.

File: sources/presentation/Stride.Core.Presentation.Avalonia/Windows/MessageBox.axaml [hunk 6]
```diff
@@ -11,17 +11,19 @@
         SizeToContent="WidthAndHeight" CanResize="False"
         WindowStartupLocation="CenterOwner"
         Padding="8,0" HorizontalContentAlignment="Stretch" VerticalContentAlignment="Center"
-        x:Class="Stride.Core.Presentation.Avalonia.Windows.MessageBox">
+        x:Class="Stride.Core.Presentation.Avalonia.Windows.MessageBox"
+        WindowDecorations="BorderOnly" ExtendClientAreaToDecorationsHint="True">
```
Kryptos-FR: Let's not impact anything else beside the Launcher, which is the scope of this PR. Especially, I don't expect other windows to be impacted. It should only be done on the main window of the launcher.

sources/presentation/Stride.Core.Presentation.Avalonia/Windows/TitleBar.axaml
Kryptos-FR: This should be limited to the Launcher and not shared in this library that is also used by the xunit.runner, and soon the new version of the GameStudio. Let's move this template to the main window in the launcher.

sources/presentation/Stride.Core.Presentation.Avalonia/Windows/CheckedMessageBox.axaml
```diff
@@ -11,17 +11,19 @@
         SizeToContent="WidthAndHeight" CanResize="False"
         WindowStartupLocation="CenterOwner"
         Padding="8,0" HorizontalContentAlignment="Stretch" VerticalContentAlignment="Center"
-        x:Class="Stride.Core.Presentation.Avalonia.Windows.CheckedMessageBox">
+        x:Class="Stride.Core.Presentation.Avalonia.Windows.CheckedMessageBox"
+        WindowDecorations="BorderOnly" ExtendClientAreaToDecorationsHint="True">
```
Kryptos-FR: Let's not impact anything else beside the Launcher, which is the scope of this PR. Especially, I don't expect other windows to be impacted. It should only be done on the main window of the launcher.

sources/presentation/Stride.Core.Presentation.Avalonia/Windows/MultiCheckedMessageBox.axaml
```diff
@@ -11,17 +11,19 @@
         SizeToContent="WidthAndHeight" CanResize="False"
         WindowStartupLocation="CenterOwner"
         Padding="8,0" HorizontalContentAlignment="Stretch" VerticalContentAlignment="Center"
-        x:Class="Stride.Core.Presentation.Avalonia.Windows.MultiCheckedMessageBox">
+        x:Class="Stride.Core.Presentation.Avalonia.Windows.MultiCheckedMessageBox"
+        WindowDecorations="BorderOnly" ExtendClientAreaToDecorationsHint="True">
```
Kryptos-FR: Let's not impact anything else beside the Launcher, which is the scope of this PR. Especially, I don't expect other windows to be impacted. It should only be done on the main window of the launcher.

sources/presentation/Stride.Core.Presentation.Avalonia/Windows/TitleBar.axaml
```diff
@@ -0,0 +1,73 @@
...
+            </WindowDrawnDecorationsContent.Overlay>
+          </WindowDrawnDecorationsContent>
+        </WindowDrawnDecorationsTemplate>
+      </Setter>
+      <!-- Pseudoclass styling -->
+      <Style Selector="^:maximized /template/ Border">
+        <Setter Property="BorderThickness" Value="0"/>
+      </Style>
+      <Style Selector="^:maximized /template/ Button#PART_MaximizeButton">
+        <Setter Property="Content" Value="{DynamicResource VectorRestoreWindow}" />
```
Kryptos-FR: 👍

Kryptos-FR:
I made commits in your branch to add a new background image for the light theme. But currently the colors for the light theme are broken, especially buttons and some background that lost their transparency. Can you have a look? (it was broken before your changes already, but in a different way 😅).

Also we shouldn't force the dark theme in users. One good feature from Avalonia is that it adapts to the user preferences. I want to keep it that way.

sources/launcher/Stride.Launcher/App.axaml
```diff
@@ -45,6 +57,38 @@
     </FluentTheme>
     <StyleInclude Source="avares://MarkView.Avalonia/Themes/MarkdownTheme.axaml" />
 
+    <Style Selector="Button">
+      <Setter Property="Background" Value="{DynamicResource NeutralBackgroundColor}" />
```
Kryptos-FR: I think this causes the most issues. Removing it looks better in light theme.

sources/presentation/Stride.Core.Presentation.Avalonia/Windows/GeometryResources.axaml
```diff
@@ -6,4 +6,39 @@
...
+
+  <!-- Window Icons -->
+  <PathIcon x:Key="VectorCloseWindow">
+      <PathIcon.Data>
+          <Geometry>F0 M12,12z M0,0z M1.1,2.8L5.3,7 1.1,11.2 1.81,11.91 6.01,7.71 10.21,11.91 10.92,11.2 6.72,7 10.92,2.8 10.21,2.09 6.01,6.29 1.81,2.09z</Geometry>
```
Kryptos-FR, line 13: What's the source of those icons? We might need to add a license or attribution.
Also it is better to just register the geometry here (hence the file name) so that they can be used in places other than PathIcon.

Jklawreszuk:
To be honest, I had no vision with light theme, so It made me think, "Hey, since the buttons are transparent now, we can consider this both a light and a dark theme isn't it?" Also It was something off with the previous light theme, but I guess It was simply background which was too much dark indeed. So I will try to think liltle bit longer about light theme.

sources/presentation/Stride.Core.Presentation.Avalonia/Windows/GeometryResources.axaml
Jklawreszuk: I didn't want to reinvent the wheel, so I took a look at the sources we currently use 😅 which was ModernUIIcons (https://github.com/Templarian/WindowsIcons)

Kryptos-FR:
I'm replacing some of the images with vector images. This should work better with dark/light theme.

Kryptos-FR:
I think I came up with a good compromise for the light theme. It's definitely better than before. There are still a few icons to replace with vertex geometry.

Dark:
image

Light:
image

Jklawreszuk:
I am not sure what to think about the button background change, but we can always reiterate by making another PR 😅. Overall I would say its 100% better with these new vector icons. The only this is missing icons for "Projects" and "Switch/update version". Other than that, I have no comments.

xen2:
@Kryptos-FR I think we'll want a new launcher release soon with all the changes (.NET 11, hiding 4th digit in version numbers for 4.4+, and I think there were a few other important bits to support 4.4 well)

Should we merge this in before final testing?

Maybe need to fix #3438 as well?
CrashReporter better stay in-process, right? (it's not like we load native libraries so it should be only .NET unhandled exception, easy to handle)

Kryptos-FR:
@xen2 I have 4 more images to replace with vector graphics. I already added a fix for #3438 in this branch (with a `MaxWidth`). For wrapping I need to check the behavior in `MarkView.Avaloni`a since we use a markdown viewer for the content.

We can keep the crash reporter in process for now. I'll need more time to find a proper solution (if any is needed).

Should we retarget this PR to `master-4.4`?

VaclavElias:
Light theme looks also nice.

Kryptos-FR:
Remaining icons converted. @xen2 ready to be tested and merged if no issues.

edit: I tried to pack the Launcher and add it to the local NugetDev. The old launcher picked up the update and then nothing. I'm not sure the current publish profile is working as expected.

edit2: my bad. I did not run `-t:BuildLauncher` before `-t:PackageLauncher` so my initial build wasn't single file.

Kryptos-FR (approved):

xen2:
> We can keep the crash reporter in process for now. I'll need more time to find a proper solution (if any is needed).

Yes, I think it works fine for our purpose as it is (pending final testing on real package).
No need to make it much more complex for cases we don't need (native crash, etc.).

I'd rather keep time/energy to focus on the launcher itself, Avalonia port, 4.4 release, 4.5, etc.

ferafiks:
Light mode is less readable than dark. It could be improved if the background image had boosted dark levels and a blur, similarly to the dark version. Here's a rough version of that, I don't have much time right now to build and test it:

image

Kryptos-FR:
@ferafiks I have actually "lighten" the original image.

PR #3387, opened 2026-09-06
[Rendering] feat: a per-view flag tells a material it is drawn for a shadow map

Nicogo1705:
# PR Details

A material whose surface is not its mesh (a ray-marched volume, a parallax search) needs to know when it is drawn for a shadow map, to settle for a depth good enough for a shadow and skip the rest. Nothing told it.

- `ShadowMapCasterPassInfo` carries `ShadowMapViewFlag` in a `PerView` constant buffer: 1 in a shadow map view, 0 in every other one. A material mixes it in and reads the flag in the pixel stage.
- The value is the view's, not a constant: the caster effect also runs the camera's depth-only prepass, where such a material must not write its shadow depth. `ShadowCasterRenderFeature` writes the flag for every view that has the group, and skips a layout the view draws nothing through, which has no resources prepared for it.

Reworked after review: no stream, no marker shader, no effect mixins, one loop over the view layouts. The bright pass clamp that was here moved to #3429.

Verified on Direct3D 11 with the voxel GI demo: the grid's shadow material casts in the shadow views, draws nothing in the camera prepass and the voxelization views.

## Types of changes

## Checklist

---


sources/engine/Stride.Rendering/Rendering/Images/BrightFilter/BrightFilterShader.sdsl
```diff
@@ -18,10 +18,22 @@ namespace Stride.Rendering.Images
...
         {
             float3 color = Texture0.Sample(PointSampler, streams.TexCoord).rgb;
 
+            // What leaves here is downsampled and blurred through a pyramid of half-float targets,
+            // where a value over 65504 becomes an infinity and the first blur weight of zero turns
+            // that into a NaN - which then spreads over the kernel and comes back up as a black
+            // block the size of the mip it happened on. Luma floors negatives; nothing capped the
+            // other end. Written as a comparison rather than isfinite() because a NaN fails every
+            // comparison, so this catches it, an infinity and a merely absurd value alike.
+            color = color > 0.0 ? min(color, MaxBrightPassValue) : float3(0, 0, 0);
```
Ethereal77: Possibly not important (I haven't benchmarked it), but maybe better as `color = min(max(color, 0.0), MaxBrightPassValue)` (less warp divergence, more SIMT/SIMD friendly than a branch).

sources/engine/Stride.Rendering/Rendering/Images/RangeConversion/RangeCompressorShader.sdsl
```diff
@@ -9,6 +9,15 @@ namespace Stride.Rendering.Images
...
             float3 color = Texture0.Sample(PointSampler, streams.TexCoord).rgb;
 
+            // Neither a negative colour nor an infinite one survives what follows: the square root
+            // below is a NaN for the first, and the division a NaN for the second, and that NaN is
+            // handed to FXAA as the pixel's luma - which then compares it against its neighbours
+            // and spreads it along whatever edge it finds.
+            // A NaN fails every comparison, so this keeps sane values, caps an infinity and floors
+            // a negative one - any of which would otherwise become the NaN that FXAA reads as this
+            // pixel's luma and then spreads along whatever edge it finds.
+            color = color > 0.0 ? min(color, 32000.0) : float3(0, 0, 0);
```
Ethereal77: Same as above.

Also, better introduce here also the `static const float MaxBrightPassValue = 32000.0` instead of hardcoding its value.

Even better would be if the const was defined just once and shared.

sources/engine/Stride.Rendering/Rendering/Shadows/ShadowMapCasterPassInfo.sdsl
```diff
@@ -0,0 +1,38 @@
...
+    /// afterwards then failed the depth test wherever the two disagreed.
+    /// </remarks>
+    shader ShadowMapCasterPassInfo
+    {
+        cbuffer PerView.ShadowMapView
+        {
+            /// 1 in a shadow map view, 0 in every other view - the camera's own depth-only
+            /// prepass included, which runs the very same caster effect. Written for every
+            /// view by ShadowCasterRenderFeature, so it is never stale.
+            stage float4 ShadowMapViewFlag;
```
Ethereal77: Why `float4` if only `.x` is ever read?

Ethereal77 (commented):
Maybe better as two separate PRs, but anyway:

1. The optimization I wrote about the clamping I think should be better **in theory**. Better to test its correctness and performance, of course. `min` + `max` should be branchless while also not falling into the previous `infinite`/`NaN` problem.

2. Nice idea the shadow caster pass identification!
   I would have preferred however if this were called something like `DepthOnlyPass` or something alike. It's weird (even if logically correct) that the depth-prepass invokes a shadow-caster pass. But maybe renaming those are a bigger change not worth the effort.

xen2:
@Ethereal77 
Sorry for the grouped PR, @Nicogo1705 asked me before submitting and I approved his proposed grouping.
But in retrospective, I agree it might have been better to split them further, for all the independent parts.
I think there is a large pass to do on the comments for all his PRs (too long/detailed, cf https://github.com/stride3d/stride/pull/3388#issuecomment-5579583029), then maybe @Nicogo1705 can reopen them with a bit more fine-grained PRs?

sources/engine/Stride.Rendering/Rendering/Images/BrightFilter/BrightFilterShader.sdsl
Nicogo1705: Done in both shaders. I compared the fxc output (ps_5_0, /O3) of the two forms on a minimal shader: neither branches, the ternary compiles to a compare plus a mask, so `min(max(...))` saves one of three instructions. Both replace a NaN by 0 (`lt` is false for a NaN, `min`/`max` return the non-NaN operand).

```
// color = color > 0.0 ? min(color, 32000.0) : 0
lt  r1.xyz, l(0, 0, 0, 0), r0.xyzx
min r0.xyz, r0.xyzx, l(32000, 32000, 32000, 0)
and o0.xyz, r0.xyzx, r1.xyzx

// color = min(max(color, 0.0), 32000.0)
max r0.xyz, r0.xyzx, l(0, 0, 0, 0)
min o0.xyz, r0.xyzx, l(32000, 32000, 32000, 0)
```

Verified on the VoxelGI demo with bloom on: same image, no NaN blocks.

sources/engine/Stride.Rendering/Rendering/Images/RangeConversion/RangeCompressorShader.sdsl
Nicogo1705: Done. Both shaders now call `HalfPrecisionUtils.ClampToSafeRange(color)`, a small shared utility shader in `Rendering/Images` on the `LuminanceUtils` pattern, which owns the 32000 constant.

sources/engine/Stride.Rendering/Rendering/Shadows/ShadowMapCasterPassInfo.sdsl
Nicogo1705: No reason, leftover from a first version. It is a `stage float` now, and `ShadowCasterRenderFeature` sets a plain float.

Nicogo1705:
Thanks for the review!
1. Branchless clamp applied in both shaders, see above.
2. On the name: I agree it reads oddly that the depth prepass runs a "shadow caster" effect. But the stream answers "am I in a shadow map view" and is 0 in the camera's depth prepass, so `IsShadowMapCasterPass` is the accurate name for what it carries.
3. Fair point on splitting, I will go one fix per PR from now on.

Nicogo1705:
Comments trimmed on this PR and on #3388 / #3389: 1 to 2 lines, no more references to the previous behaviour. If you would rather have these two fixes as separate PRs, I can reopen them that way, just say so.

Ethereal77:
Hey, no need to apologize. I'm not opposed to grouping if the changes are small. I understand it's more convenient sometimes to dump several changes into a single PR than open several. I just like that PRs sometimes serve as a kind of auto-documentation on how the software evolved and why some changes were needed, and combining several changes into one makes each change less discoverable and sometimes less understandable (context switching overhead 😄). Just my opinion, of course

Ethereal77:
> On the name: I agree it reads oddly that the depth prepass runs a "shadow caster" effect. But the stream answers "am I in a shadow map view" and is 0 in the camera's depth prepass, so `IsShadowMapCasterPass` is the accurate name for what it carries.

Oh, now I see what you mean. I have misread the PR and thought you have also rewritten the depth-prepass to use this and this just served to compute the depth-only output (hence my comment). But I see now that the depth-prepass was already using the shadow caster effect. Sorry for the confusion

sources/engine/Stride.Rendering/Rendering/Images/HalfPrecisionUtils.sdsl
```diff
@@ -0,0 +1,24 @@
...
+        /// Half the half-float maximum (65504), leaving headroom for the blurs and sums that follow.
+        /// </summary>
+        static const float MaxSafeValue = 32000.0;
+
+        /// <summary>
+        /// Clamps a color to [0, MaxSafeValue]. min/max return the other operand for a NaN, so a NaN is replaced too.
+        /// </summary>
+        static float3 ClampToSafeRange(float3 color)
+        {
+            return min(max(color, 0.0), MaxSafeValue);
```
xen2, line 21: Unfortunately, it works only on D3D.
Vulkan FMax/FMin result is undefined on NaN.
Comparison (as in your previous commit) should work everywhere:
```
static float3 ClampToSafeRange(float3 color)
{
    // A NaN fails every comparison, so it lands on MaxSafeValue like an infinity does.
    return color < MaxSafeValue ? max(color, 0.0) : MaxSafeValue;
}
```

sources/engine/Stride.Rendering/Rendering/Shadows/ShadowCasterRenderFeature.cs
```diff
@@ -27,6 +38,18 @@ public override void Prepare(RenderDrawContext context)
                 
                 // Process only shadow views
                 var shadowMapRenderView = view as ShadowMapRenderView;
+
+                // ShadowMapCasterPassInfo flag, written for every view (not only shadow ones) so the group is never left uninitialized.
+                foreach (var viewLayout in viewFeature.Layouts)
```
xen2, line 43: `viewFeature.Layouts` is looped twice (and they have different loop guards, yours is checking `resourceGroup` but not the other one; which is right?)

sources/engine/Stride.Rendering/Rendering/Shadows/ShadowMapCasterPassInfo.sdsl
```diff
@@ -0,0 +1,26 @@
...
+    shader ShadowMapCasterPassInfo
+    {
+        cbuffer PerView.ShadowMapView
+        {
+            /// 1 in a shadow map view, 0 in every other view. The caster effect is also used by the
+            /// camera's depth-only prepass, so this cannot be a constant.
+            stage float ShadowMapViewFlag;
+        }
+
+        stage stream float IsShadowMapCasterPass;
```
xen2, line 24: Question: why does user/material need to write 0?
Can't we do like in other shaders:
`override void ResetStream() { base.ResetStream(); streams.IsShadowMapCasterPass= 0.0f; }`

xen2: Never mind, ResetStream won't be called in this case.
But then another question: since this is computed in pixel shader, why using a stream at all? can't we use `ShadowMapViewFlag` directly?

sources/engine/Stride.Rendering/Rendering/Images/HalfPrecisionUtils.sdsl
Ethereal77: There is no way to make this branchless in Vulkan?

Nicogo1705: Right, min/max on a NaN is undefined in SPIR-V. Back to the comparison, kept in the shared helper: `color > 0.0 ? min(color, MaxSafeValue) : 0.0`. A NaN fails the comparison and lands on 0 on every backend; an infinity goes through `min` with a finite operand, which is defined.

@Ethereal77 it is still branchless: a ternary on a `float3` is a per-component select, which is what the fxc listing above showed, a compare and a mask, no jump.

The clamp now lives in its own PR, #3429, as suggested.

sources/engine/Stride.Rendering/Rendering/Shadows/ShadowCasterRenderFeature.cs
Nicogo1705: One loop now. The guard is the right one for both groups: a view can carry a layout it draws nothing through, and no resources are prepared for it then. The old loop never met that case because it only ran on shadow views.

sources/engine/Stride.Rendering/Rendering/Shadows/ShadowMapCasterPassInfo.sdsl
Nicogo1705: Agreed, the stream adds nothing. The flag is already written per view for every view that has the group, so a material reads `ShadowMapViewFlag` directly in the pixel stage. The marker, the three effect mixins and the `= 0` in the vertex stage are gone; `ShadowMapCasterPassInfo` is the constant buffer alone. It also removes an interpolant a geometry shader between the stages could drop.

Nicogo1705:
Both fixes were reworked after review, so this is the moment to split as suggested: this PR keeps the shadow map flag, the bright pass clamp moved to #3429. Rebased on master, one commit each. Verified on the voxel GI demo on Direct3D 11: shadows, camera prepass and voxelization views.

sources/engine/Stride.Rendering/Rendering/Shadows/ShadowCasterRenderFeature.cs
```diff
@@ -25,23 +36,28 @@ public override void Prepare(RenderDrawContext context)
...
-
-                        var shadowMapTexture = shadowMapRenderView.ShadowMapTexture;
-                        renderer.ApplyViewParameters(context, shadowMapRenderView.ViewParameters, shadowMapTexture);
-                        
-                        var resourceGroup = viewLayout.Entries[view.Index].Resources;
-                        resourceGroup.UpdateLogicalGroup(ref shadowCaster, shadowMapRenderView.ViewParameters);
-                    }
+                    // A layout the view draws nothing through has no resources prepared for it.
+                    var resourceGroup = viewLayout.Entries[view.Index].Resources;
+                    if (resourceGroup == null || resourceGroup.ConstantBuffer.Data == System.IntPtr.Zero)
```
xen2, line 44: "A layout the view draws nothing through has no resources prepared for it."
Not sure I get that.

`resourceGroup` should not be null (always created in RootEffectRenderFeatures).
I think `ConstantBuffer.Data`  should always exists.
`Data == Zero` only happens for a layout without a PerView cbuffer. Such a layout has neither logical group, so without the guard both `Hash == ObjectId.Empty` checks would skip it before anything reads `ConstantBuffer.Data`.

Did you hit a crash without it? If yes, can you share the repro? That would be a different bug. If not, let's drop the guard.

sources/engine/Stride.Rendering/Rendering/Shadows/ShadowMapCasterPassInfo.sdsl
```diff
@@ -0,0 +1,21 @@
...
+    /// A material whose surface is computed in the pixel shader (a ray-marched volume, a parallax
+    /// search) can settle for a depth good enough for a shadow and skip the rest.
+    /// <see cref="ShadowCasterRenderFeature"/> writes the flag for every view.
+    /// </remarks>
+    shader ShadowMapCasterPassInfo
+    {
+        cbuffer PerView.ShadowMapView
+        {
+            /// 1 in a shadow map view, 0 in every other view, the camera's depth-only prepass included.
+            stage float ShadowMapViewFlag;
```
xen2, line 18: I was wondering, if no `ShadowCasterRenderFeature`, this viewflag is basically undefined memory?
Should we accept that and add to doc, or do something for this case (i.e. always default to 0 in `MaterialRenderFeature`)?

sources/engine/Stride.Rendering/Rendering/Shadows/ShadowCasterRenderFeature.cs
Nicogo1705: No crash, I misread this: `viewFeature.Layouts` only holds layouts that passed `MarkAsUsed` and `PrepareResourceGroup` this frame, so a layout the view draws nothing through is never in the list. Guard dropped in dba0f2bc7.

sources/engine/Stride.Rendering/Rendering/Shadows/ShadowMapCasterPassInfo.sdsl
Nicogo1705: Moved the write to `MaterialRenderFeature` in a7c2a4d1c: 1 in a `ShadowMapRenderView`, 0 in every other view. `ShadowCasterRenderFeature` is back to master, so the flag is defined whenever materials render, with or without shadows in the compositor.

PR #3388, opened 2026-09-06
[Games] fix: stop skipping the discrete GPU of a hybrid laptop

Nicogo1705:
# PR Details

Stride games landed on the integrated GPU of a hybrid laptop. `FindBestDevices` skipped every adapter without a display output, meant to keep WARP and the Basic Render Driver out; on a hybrid laptop every display is wired to the integrated GPU, so the discrete one, which DXGI already sorts first, was dropped.

- `GraphicsAdapter.IsSoftwareAdapter` says what the skip meant: the DXGI software flag or the Microsoft vendor id on Direct3D, the `Cpu` device type on Vulkan. `FindBestDevices` skips those instead of output-less adapters.
- When nothing survives the skip (a VM, a CI runner with only a software adapter), the search runs again without it, so software rendering still works where it is all there is.
- Vulkan enumerates physical devices with the discrete GPU first, as DXGI's high-performance order already does on Direct3D. Not tested on a Vulkan machine with two GPUs.

Verified on a hybrid laptop (Intel iGPU with the display, NVIDIA dGPU without) on Direct3D 11: the game now creates its device on the NVIDIA. The software-only fallback and the Vulkan order compile but are not exercised here; a CI runner with only WARP or llvmpipe is the test for the first.

The GPU preference API that was part of this PR moved to its own issue, per review: #3428.

## Related Issue

#3428

## Types of changes

## Checklist

---


sources/engine/Stride.Graphics/Direct3D/GraphicsAdapter.Direct3D.cs
```diff
@@ -153,6 +159,9 @@ internal GraphicsAdapter(ComPtr<IDXGIAdapter1> adapter, uint adapterOrdinal)
             AdapterUid = dxgiAdapterDesc.AdapterLuid.BitCast<Luid, long>();
 
             VendorId = (int) dxgiAdapterDesc.VendorId;
+            // DXGI_ADAPTER_FLAG_SOFTWARE (2) marks WARP; the Basic Render Driver predates the
+            // flag on some Windows builds, so its Microsoft vendor id is the fallback test.
+            IsSoftwareAdapter = (dxgiAdapterDesc.Flags & 0x2) != 0 || dxgiAdapterDesc.VendorId == 0x1414;
```
Ethereal77: Instead of hardcoding flags, prefer to use the Silk-provided enum: `Silk.NET.DXGI.AdapterFlag.Software`

sources/engine/Stride.Graphics/Direct3D/GraphicsAdapterFactory.Direct3D.cs
```diff
@@ -178,17 +191,21 @@ static List<GraphicsAdapter> EnumerateAdapters()
             //
             // Enumerates all the Graphics Adapters in the system, using GPU preference (DXGI 1.6+).
             //
-            static List<GraphicsAdapter> EnumerateAdaptersPrefer(GpuPreference gpuPreference)
+            static List<GraphicsAdapter> EnumerateAdaptersPrefer(GpuPreference preference)
```
Ethereal77: Err... why rename this? 😀

Even though the type is different (it is now in Stride), the old name still made sense

sources/engine/Stride.Graphics/GpuPreference.cs
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) .NET Foundation and Contributors (https://dotnetfoundation.org/ & https://stride3d.net) and Silicon Studio Corp. (https://www.siliconstudio.co.jp)
```
Ethereal77: No need for Silicon Studio copyright in headers of new files

Ethereal77 (commented):
A nice addition. This lacked configurability.

I must say however that I put the `HighPerformance` as default for the exact case you mention. I was developing this at the tame on an old laptop with integrated Intel Graphics + a dedicated nVidia GPU, and this was enough for Stride to always choose the nVidia adapter. Did not work for you?

And sorry if I'm being annoying 🙏 but... the profiler change maybe should have been a separate PR, even if small. Having several unrelated changes in a single PR makes it difficult in the future to locate them or understand them at a glance. Just my honest opinion, I like to have things tidy 😁

xen2:
General remark (apply to your other PRs): comments are overly long and verbose (some of those 9 lines!), giving way too much details, relevant only to previous code or your specific context/discussion.
I think most of the 5+ lines comments could written in a much shorter and simpler form (3~4 lines max if really necessary, but most can be 1/2 lines).
Also they shouldn't mention about the previous issues which are now gone (hard to understand).

sources/engine/Stride.Graphics/Direct3D/GraphicsAdapter.Direct3D.cs
Nicogo1705: Done: `(uint) AdapterFlag.Software`, and the Microsoft vendor id moved to a named constant.

sources/engine/Stride.Graphics/Direct3D/GraphicsAdapterFactory.Direct3D.cs
Nicogo1705: Reverted, the parameter keeps its original name.

sources/engine/Stride.Graphics/GpuPreference.cs
Nicogo1705: Done, header trimmed to .NET Foundation and Contributors.

Nicogo1705:
Thanks! `HighPerformance` does put the NVIDIA first in the enumeration, that part worked. What broke it is `FindBestDevices` right after, which skipped every adapter without a display output. On my laptop all outputs are wired to the Intel GPU, so the NVIDIA has none and was dropped, and every game landed on the Intel. The first commit fixes that by testing "software adapter" instead of "no output"; the enum only makes your default settable.

On the profiler: no problem, done, it moved to its own PR: #3405.

Nicogo1705:
All trimmed or removed: the 9-line comment in `GamePlatform` is one line now, and none mention the previous behaviour any more. Same pass done on #3387 and #3389.

Ethereal77:
> What broke it is `FindBestDevices` right after, which skipped every adapter without a display output. On my laptop all outputs are wired to the Intel GPU, so the NVIDIA has none and was dropped, and every game landed on the Intel.

That's weird. My laptop's nVidia dGPU has no display attached either, yet using `HighPerformance` worked and Stride selected it instead of the Intel iGPU. At that's what happened back then when I was working on that part. I don't know if anything changed since.

Anyway, if your change is a more general solution, it's a nice addition.

sources/engine/Stride.Games/GamePlatform.cs
```diff
@@ -360,10 +360,8 @@ public virtual List<GraphicsDeviceInformation> FindBestDevices(GameGraphicsParam
                 if (!string.IsNullOrEmpty(preferredParameters.RequiredAdapterUid) && adapterUid != preferredParameters.RequiredAdapterUid)
                     continue;
 
-                // Skip adapters that don't have graphics output
-                // but only if no RequiredAdapterUid is provided (OculusVR at init time might be in a device with no outputs)
-                // Software rendering adapters (e.g. WARP) have no outputs either, so allow them through
-                if (graphicsAdapter.Outputs.Length == 0
+                // Skip software rasterizers (WARP, Basic Render Driver) unless explicitly requested.
+                if (graphicsAdapter.IsSoftwareAdapter
```
xen2, line 364: On Vulkan, there's always a fake output `graphicsOutputs = [ new GraphicsOutput() ];`
So previous code with `Outputs.Length == 0` never removed anything but the new one does.
Esp. software-only renderer like CI will only have IsSoftwareAdapter devices, which will be removed, so nothing left.

On D3D: same, since `VendorId == 0x1414` (Microsoft) means any machine with only Microsoft Basic device and/or VM with virtual GPU will go through that path.

Maybe your fix should only be applied if something survived.
If `graphicsDeviceInfos.Count == 0`, run it again without software filtering.
As a result, it can still allow software rendering if nothing else.

sources/engine/Stride.Graphics/Direct3D/GraphicsAdapterFactory.Direct3D.cs
```diff
@@ -183,12 +191,16 @@ static List<GraphicsAdapter> EnumerateAdaptersPrefer(GpuPreference gpuPreference
                 Debug.Assert(dxgiFactoryVersion >= 6);
                 var dxgiFactory6 = (IDXGIFactory6*) dxgiFactory;
 
+                var dxgiPreference = gpuPreference == GpuPreference.MinimumPower
```
xen2, line 194: What about Vulkan? GPUPreference is ignored?

sources/engine/Stride.Games/GamePlatform.cs
xen2, line 402: An adapter with no outputs still produces no fullscreen candidate, so a game that starts fullscreen keeps landing on the iGPU.

On a muxless laptop (panel wired to the Intel, no MUX to switch it):
GPU 0: NVIDIA GPU (0 output)
GPU 1: Intel UHD Graphics (1 output)

`GraphicsDeviceManager.IsFullScreen = true` -> the NVIDIA adds nothing here, only the Intel does, so `FindBestDevice` returns the Intel and the device is created on it. Windowed gets the NVIDIA, fullscreen gets the Intel, on the same machine.

However, exclusive fullscreen on an output-less adapter is not really possible in DXGI...
So not sure what the expected behavior is? Is it to do borderless fullscreen rather than an output lookup?

sources/engine/Stride.Graphics/GraphicsAdapterFactory.cs
```diff
@@ -18,6 +18,30 @@ public static partial class GraphicsAdapterFactory
...
         private static GraphicsAdapter defaultAdapter;
+        private static GpuPreference gpuPreference = GpuPreference.HighPerformance;
+
+        /// <summary>
+        ///   Gets or sets which GPU is enumerated first on machines that have several, and therefore
+        ///   which one <see cref="DefaultAdapter"/> is.
+        /// </summary>
+        /// <remarks>
+        ///   Set it before the first use of <see cref="Adapters"/> or <see cref="DefaultAdapter"/>, or call
+        ///   <see cref="Reset"/> afterwards. The <c>STRIDE_GPU_PREFERENCE</c> environment variable
```
xen2, line 29: Have to be careful with `Reset()` recommendation: it dispose adapters, which could still be used by an alive `GraphicsDevice`.

I wonder if we should change our API so that `GraphicsAdapterFactory` list doesn't magically change depending on some state like `GpuPreference` being setup early enough.

It could simply report what the system has, untouched. With proper fact like `IsSoftwareAdapter`, and discrete/integrated hint (Vulkan has this info). Also possibly `IsDefaultAdapter` (item 0, could be Intel UHD) or `PerformanceRank` (from DXGI `EnumAdapterByGpuPreference`).

GPU preference can be stored like other preference on the appropriate object, like in GameGraphicsParameters then GraphicsDeviceManager (i.e. PreferredGpu).

As a result, creation won't use indirectly this mutable static anymore.

Example API for GraphicsAdapter:
```
public enum GraphicsAdapterType { Unknown, Integrated, Discrete, Virtual, Software }

public GraphicsAdapterType Type { get; private init; }        // set by each backend ctor
public bool IsSoftwareAdapter => Type == GraphicsAdapterType.Software; // maybe not needed, we can use Type

// Index of this adapter in the system enumeration order. 0 is the system default (use if GPU preference is unspecified -- on Windows it should return the OS pref depending on power saving settings)
public int AdapterOrder { get; private init; }
public bool IsDefaultAdapter => AdapterOrder == 0;

// Position in the platform's fastest-first order. 0 is the fastest
public int PerformanceRank { get; private init; }

// Position in the platform's most-efficient-first order. 0 is the most power efficient
public int PowerEfficiencyRank { get; private init; }
```

D3D can fill using DXGI ordering method like `EnumerateLuidsByPreference`
Vulkan can use stuff like Discrete > Integrated > Virtual > Cpu for perf (and swap Discrete/Integrated for efficiency)

xen2 (commented):
~~Vulkan: nothing happens currently~~
sorry, text was incomplete

sources/engine/Stride.Games/GamePlatform.cs
Nicogo1705: Done. `FindBestDevices` runs the search with the software skip first, and again without it when nothing survived, so a machine with only WARP, the Basic Render Driver or llvmpipe still gets a device. Not exercised here; a software-only CI runner is the test.

sources/engine/Stride.Graphics/Direct3D/GraphicsAdapterFactory.Direct3D.cs
Nicogo1705: The preference API is out of this PR (see #3428), so there is no preference left to honour. What Vulkan gets instead is the same default as Direct3D: `GraphicsAdapterFactory` sorts the physical devices discrete > integrated > virtual > CPU before building the adapters, so the ordinal and `IsDefaultAdapter` follow the new order. Compiles, not tested on a Vulkan machine with two GPUs.

sources/engine/Stride.Graphics/GraphicsAdapterFactory.cs
Nicogo1705: Agreed, that is the wrong shape. The enum, `Reset()` and the environment variable are gone from this PR, which keeps only the bug fix. Your API sketch is in #3428 as the follow-up, with the ranks filled from DXGI on Direct3D and from the device type on Vulkan.

sources/engine/Stride.Games/GamePlatform.cs
Nicogo1705: Out of this PR: it is the same on master, and it needs a decision on what full screen means on an output-less adapter (borderless on the chosen adapter, most likely). Noted in #3428 with the preference API.

sources/engine/Stride.Graphics/Vulkan/GraphicsAdapterFactory.Vulkan.cs
```diff
@@ -62,7 +62,17 @@ internal static unsafe void InitializeInternal()
...
+            }
+
+            // The fastest kind first, as DXGI's high-performance order does on Direct3D: a hybrid laptop enumerates its integrated GPU first.
+            physicalDevices.Sort((a, b) => PerformanceRank(a.Properties.deviceType).CompareTo(PerformanceRank(b.Properties.deviceType)));
+
+            var adapterList = new List<GraphicsAdapter>();
+            for (int index = 0; index < physicalDevices.Count; index++)
+            {
+                var (device, properties, driverProps) = physicalDevices[index];
+                var adapter = new GraphicsAdapter(device, properties, driverProps, index);
```
xen2: Ordinal is now the sorted index, but `GetPhysicalDevice(enableValidation: true)` enumerates again unsorted and uses `nativePhysicalDevices[adapterOrdinal]`.
With the Debug flag (always on in Debug engine builds), the device can be created on a different GPU than the selected adapter.

sources/engine/Stride.Graphics/Vulkan/GraphicsAdapterFactory.Vulkan.cs
```diff
@@ -62,7 +62,17 @@ internal static unsafe void InitializeInternal()
                 if (properties.apiVersion >= VkVersion.Version_1_2)
                     defaultInstance.NativeInstanceApi.vkGetPhysicalDeviceProperties2(nativePhysicalDevices[index], &properties2);
 
-                var adapter = new GraphicsAdapter(nativePhysicalDevices[index], properties, driverProps, index);
+                physicalDevices.Add((nativePhysicalDevices[index], properties, driverProps));
+            }
+
+            // The fastest kind first, as DXGI's high-performance order does on Direct3D: a hybrid laptop enumerates its integrated GPU first.
+            physicalDevices.Sort((a, b) => PerformanceRank(a.Properties.deviceType).CompareTo(PerformanceRank(b.Properties.deviceType)));
```
xen2: I'd drop this sort from this PR: Vulkan never had the output skip bug (fake output), and reordering the factory list is what #3428 wants to avoid.
Discrete-first can go in future #3428 PR as a ranking in selection.

sources/engine/Stride.Graphics/Vulkan/GraphicsAdapterFactory.Vulkan.cs
xen2, line 75: Ordinal is now the sorted index, but `GetPhysicalDevice(enableValidation: true)` enumerates again unsorted and uses `nativePhysicalDevices[adapterOrdinal]`. With the Debug flag (always set in Debug engine builds) the device can end up on another GPU than the selected adapter.
Keep the original enumeration index for that lookup.

sources/engine/Stride.Games/GamePlatform.cs
```diff
@@ -349,6 +349,17 @@ protected void AddDevice(DisplayMode mode,  GraphicsDeviceInformation deviceBase
         }
 
         public virtual List<GraphicsDeviceInformation> FindBestDevices(GameGraphicsParameters preferredParameters)
+        {
+            // Software rasterizers (WARP, Basic Render Driver, llvmpipe) only when asked for, or when nothing else is there.
+            var skipSoftwareAdapters = string.IsNullOrEmpty(preferredParameters.RequiredAdapterUid)
+                && Environment.GetEnvironmentVariable("STRIDE_GRAPHICS_SOFTWARE_RENDERING") != "1";
+            var graphicsDeviceInfos = FindBestDevices(preferredParameters, skipSoftwareAdapters);
+            if (graphicsDeviceInfos.Count == 0 && skipSoftwareAdapters)
```
xen2, line 357: Suggestion: `skipSoftwareAdapters = RequiredAdapterUid is empty && any adapter is not software`, single pass.
- No env var check needed: with `STRIDE_GRAPHICS_SOFTWARE_RENDERING=1` D3D only enumerates WARP anyway, and CI then covers this path on every lane.
- A GPU without the requested profile keeps the "profile not supported" error instead of silently using WARP.

xen2 (commented):
The fix itself (skip software adapters instead of output-less ones) looks right.

Two things before approving:
1. The Vulkan sort: my earlier question was about the preference API, which is gone now, so there is nothing to honour on Vulkan anymore. The sort is fine to keep if you want Vulkan hybrid laptops fixed here, but it has a bug (inline). If you'd rather keep this PR small, dropping it is fine too and it can come with the follow-up of #3428.
2. The fallback: I'd rather skip software adapters only when a hardware adapter exists (inline), which removes the second pass and the env var check.

Note: I did not try to run it on a two-GPU machine

sources/engine/Stride.Graphics/Vulkan/GraphicsAdapterFactory.Vulkan.cs
Nicogo1705: Done in 70be09651: the adapter keeps its enumeration index next to the ordinal, and the validation lookup uses that one. `IsDefaultAdapter` stays on the sorted ordinal.

sources/engine/Stride.Games/GamePlatform.cs
Nicogo1705: Done in 70be09651: one pass, software adapters skipped only when the uid is unset and a hardware adapter exists. Env var check and second pass removed.

Nicogo1705:
Both points addressed in 70be09651 (see inline). Builds checked for Direct3D11 and Vulkan.

PR #3399, opened 2026-09-08
Fix Slider value coercion when Maximum changes

phcodesage:
## Summary

Re-coerce Slider.Value after Maximum changes so the value remains within the updated bounds.

## Changes

- Re-coerce Value in Slider maximum coercion.
- Add regression coverage for bounds changes, ValueChanged event counts, and tick snapping.

## Testing

- Stride.UI.Tests: 209 passed, 5 skipped, 0 failed.
- Full Stride.slnx run completed on macOS arm64 but reported unrelated platform and sample failures: asset tests, audio/sample screenshot tests, missing x64 .NET host, and an editor GameForm compile issue.


phcodesage:
@dotnet-policy-service agree

PR #3404, opened 2026-09-08
fix: priority queue losing order on item removal or reversed inserting

### LINKED ISSUE #3398: Priority queues return out-of-order values after removing an internal item
**Release Type**: GitHub source

**Version**: `3d034317171905c1889fd6b3820d648b2af73312`

**Platform(s)**: Reproduced on Linux using .NET SDK 10.0.100. This is managed collection code; no graphics or editor execution is required for the reproduction.

**Describe the bug**

`PriorityQueue<T>.Remove` and `PriorityNodeQueue<T>.Remove` can break priority order when removing an internal item. They replace the removed slot with the last element and only repair downward. When that replacement is smaller than its new parent, it must move upward instead.

**To Reproduce**

Run this against the `Stride.Core` queue implementation at the commit above:

```csharp
var queue = new Stride.Core.Collections.PriorityQueue<int>();
foreach (var value in new[] { 0, 1, 4, 2, 5, 6, 7, 3 })
    queue.Enqueue(value);

queue.Remove(5);

while (!queue.Empty)
    Console.WriteLine(queue.Dequeue());
```

Actual output: `0, 1, 2, 4, 3, 6, 7`.

**Expected behavior**

Output remains in priority order: `0, 1, 2, 3, 4, 6, 7`.

The same input reproduces with `PriorityNodeQueue<int>` when removing the node returned by `Enqueue(5)`. Both variants also fail under a descending custom comparer.

```csharp
var queue = new Stride.Core.Collections.PriorityNodeQueue<int>();
Stride.Core.Collections.PriorityQueueNode<int>? nodeToRemove = null;
foreach (var value in new[] { 0, 1, 4, 2, 5, 6, 7, 3 })
{
    var node = queue.Enqueue(value);
    if (value == 5)
        nodeToRemove = node;
}

queue.Remove(nodeToRemove!);

while (!queue.Empty)
    Console.WriteLine(queue.Dequeue());
```

**Screenshots**

Not applicable; the reproduction is a console collection test.

**Log and callstacks**

The regression assertion fails at the fourth dequeued value:

```text
Assert.Equal() Failure: Values differ
Expected: 3
Actual:   4

Original production sources: Total tests: 14; Passed: 8; Failed: 6
Patched production sources:  Total tests: 14; Passed: 14; Failed: 0
```

**Additional context**

These are active engine paths: `Scheduler.Unschedule` removes empty priority buckets, `AnimationChannel.Fitting` removes queued error nodes, and `AssetBuilderService.RemoveBuildUnit` removes queue nodes. The collection-level ordering failure is verified; I have not run the editor or claimed a particular visible scene failure.

A [prepared patch with regression tests](https://github.com/falseee1/stride/commit/2f63ef755325b88a886d7e15c8d3e70714ab89dc) repairs upward when needed while preserving the existing heap representation and maintaining both node indices on swaps. It includes six new regression cases: the eight-item example under both comparer directions, plus deterministic shuffled insertion/removal coverage for both queue variants. Compiling the three real production files and the two existing test files in an isolated xUnit project produced **8 passed / 6 failed before**, then **14 passed / 0 failed after**. No collection implementation was mocked or rewritten in the harness. The ordinary repository test invocation did not produce a final result in the captured log, so I am not claiming a passing engine build or full repository suite.

AI disclosure: OpenAI Codex prepared this investigation, patch, and executable regression tests for `falseee1`.

IXLLEGACYIXL:
# PR Details

PriorityQueue loses order on Remove/Reverse Inserting

## Related Issue

fix #3398

## Types of changes

internal bug fix

## Checklist

i was able to run around 4445 tests but it was stuck on bepu so few are missing, but the list queue tests pass now, also the old ones

### DIFF
--- a/sources/core/Stride.Core.Tests/TestPriorityQueue.cs
+++ b/sources/core/Stride.Core.Tests/TestPriorityQueue.cs
@@ -85,4 +85,28 @@ public class TestPriorityQueue
             lastItem = value;
         }
     }
+
+    [Fact]
+    public void TestRemovalSingleItem()
+    {
+        var queue = new PriorityQueue<int>();
+        foreach (var value in new[] { 0, 1, 4, 2, 5, 6, 7, 3 })
+            queue.Enqueue(value);
+
+        queue.Remove(5);
+
+        CheckPriorityQueue(queue);
+    }
+
+    [Fact]
+    public void TestRemovalSingleItemAtTheEnd()
+    {
+        var queue = new PriorityQueue<int>();
+        foreach (var value in new[] { 0, 1, 4, 2, 5, 6, 7, 3 })
+            queue.Enqueue(value);
+
+        queue.Remove(7);
+
+        CheckPriorityQueue(queue);
+    }
 }
--- a/sources/core/Stride.Core/Collections/PriorityQueue.cs
+++ b/sources/core/Stride.Core/Collections/PriorityQueue.cs
@@ -52,7 +52,7 @@ public class PriorityQueue<T>
         var parentIndex = index;
         while (true)
         {
-            var childIndex = parentIndex * 2;
+            var childIndex = (parentIndex * 2) + 1;
 
             // Check if there is any child, otherwise we're done
             if (childIndex >= maxCount)
@@ -87,7 +87,7 @@ public class PriorityQueue<T>
         // Bubble elements so that order is respected again
         while (childIndex != 0)
         {
-            var parentIndex = childIndex / 2;
+            var parentIndex = (childIndex - 1) / 2;
 
             // Should we swap with parent?
             if (comparer.Compare(items[childIndex], items[parentIndex]) >= 0)
@@ -133,7 +133,7 @@ public class PriorityQueue<T>
         while (true)
         {
             // Check if there is any child, otherwise we're done
-            var childIndex = parentIndex * 2;
+            var childIndex = (parentIndex * 2) + 1;
             if (childIndex >= maxCount)
                 break;
 


xen2:
Doesn't work in all cases:

```
var q = new PriorityQueue<int>();
foreach (var v in new[] { 26, 38, 33, 54, 69, 64, 21 }) q.Enqueue(v);
q.Remove(54);
// drains: 21, 26, 38, 33, 64, 69   <-- 33 comes after 38
```

Issue #3398 explains the real problem. When `Remove` deletes an item, it takes the last item in the queue and puts it in the empty slot. Then it only moves that item down the tree.
But the item can also be smaller than its new parent. In that case it has to move up instead.

IXLLEGACYIXL:
> Doesn't work in all cases:
> 
> ```
> var q = new PriorityQueue<int>();
> foreach (var v in new[] { 26, 38, 33, 54, 69, 64, 21 }) q.Enqueue(v);
> q.Remove(54);
> // drains: 21, 26, 38, 33, 64, 69   <-- 33 comes after 38
> ```
> 
> Issue #3398 explains the real problem. When `Remove` deletes an item, it takes the last item in the queue and puts it in the empty slot. Then it only moves that item down the tree. But the item can also be smaller than its new parent. In that case it has to move up instead.

ok so the current improvements fixes the regression that the old tests pass

i add this case as well

IXLLEGACYIXL:
ive added now your test and it passes to decide if order is broken up or downwards on removal

sources/core/Stride.Core.Tests/TestPriorityQueue.cs
```diff
@@ -85,4 +85,38 @@ private static void CheckPriorityQueue(PriorityQueue<int> priorityQueue)
...
+        foreach (var value in new[] { 0, 1, 4, 2, 5, 6, 7, 3 })
+            queue.Enqueue(value);
+
+        queue.Remove(7);
+
+        CheckPriorityQueue(queue);
+    }
+
+    [Fact]
+    public void TestRemovalOnBubbleDownPreservesOrder()
```
xen2, line 114: Nit: this test RemovalOnBubbleUp not Down

sources/core/Stride.Core.Tests/TestPriorityQueue.cs
xen2, line 78: CheckPriorityQueue only checks order. If Remove deleted the wrong item, the tests would still pass.
Could the new tests compare the drained values with the expected sorted list?

xen2 (commented):
`PriorityNodeQueue<T>.Remove` has the same bug.
It needs the same "move up if smaller than parent" step.
Please add tests in TestPriorityLinkedQueue too.

`Remove` and `Dequeue` now have the same sift-down loop.
Private SiftUp/SiftDown helpers would remove the copy (and maybe make the PriorityNodeQueue fix simple).

xen2:
Superseeded by #3456 (I took the liberty to restart from your commit and complete the PR).
Thanks again for the initial implementation!

Note: if you give me permissions on your org/fork, I could have pushed directly on this PR.

IXLLEGACYIXL:
thanks for completing

PR #3405, opened 2026-09-08
[Engineering] fix: profiler result page never clamped below 1

Nicogo1705:
# PR Details

Split out of #3388 as requested in review.

`GameProfilingSystem` clamps `CurrentResultPage` against the page count. On the first frames after profiling is enabled there are no results yet, so the page count is 0 and the page is clamped to 0, where it stays: once results arrive the profiler shows "PAGE 0 OF N" over an empty list. The clamp now never goes below page 1.

Verified with the in-game profiler (F2 / P in the VoxelGI demo): the first page shows its results right after enabling.

## Related Issue

None filed; small enough to be its own fix.

## Types of changes

## Checklist

---

### DIFF
--- a/sources/engine/Stride.Engine/Profiling/GameProfilingSystem.cs
+++ b/sources/engine/Stride.Engine/Profiling/GameProfilingSystem.cs
@@ -149,7 +149,7 @@ namespace Stride.Profiling
             var availableDisplayHeight = viewportHeight - 2 * TextRowHeight - 3 * TopRowHeight;
             var elementsPerPage = (int)Math.Floor(availableDisplayHeight / TextRowHeight);
             numberOfPages = (uint)Math.Ceiling(profilingResults.Count / (float)elementsPerPage);
-            CurrentResultPage = Math.Min(CurrentResultPage, numberOfPages);
+            CurrentResultPage = Math.Max(1, Math.Min(CurrentResultPage, numberOfPages));
 
             char sortByTimeIndicator = SortingMode == GameProfilingSorting.ByTime ? 'v' : ' ';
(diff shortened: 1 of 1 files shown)


sources/engine/Stride.Engine/Profiling/GameProfilingSystem.cs
```diff
@@ -149,7 +149,7 @@ private void UpdateProfilingStrings(bool containsMarks)
             var availableDisplayHeight = viewportHeight - 2 * TextRowHeight - 3 * TopRowHeight;
             var elementsPerPage = (int)Math.Floor(availableDisplayHeight / TextRowHeight);
             numberOfPages = (uint)Math.Ceiling(profilingResults.Count / (float)elementsPerPage);
-            CurrentResultPage = Math.Min(CurrentResultPage, numberOfPages);
+            CurrentResultPage = Math.Max(1, Math.Min(CurrentResultPage, numberOfPages));
```
Ethereal77, line 152: Better to [use this](https://learn.microsoft.com/en-us/dotnet/api/system.math.clamp?view=net-10.0#system-math-clamp(system-int32-system-int32-system-int32)).
Or, [even better this](https://learn.microsoft.com/en-us/dotnet/api/system.int32.clamp?view=net-10.0#system-int32-clamp(system-int32-system-int32-system-int32)), which is the Microsoft's recommended way now.
```csharp
int.Clamp(value, 1, numberOfPages)
```

Nicogo1705: Done: `uint.Clamp(CurrentResultPage, 1, Math.Max(1, numberOfPages))`. The upper bound is kept at 1 or more because `numberOfPages` is 0 until the first results arrive, and `Clamp` throws when max < min.

PR #3407, opened 2026-09-08
chore: small cleanup in shader api

Eideren:
# PR Details
In ShaderLoaderBase, redundant catch, finally already runs when an exception occurs.
In ShaderClassCode, empty array doesn't necessitate a stringbuilder, if that ever could be the case. Interpolation is more succinct than an sb and does create similar codegen.

## Related Issue
None

## Types of changes

## Checklist


sources/shaders/Stride.Shaders/ShaderClassCode.cs
xen2: I think this can be removed now

Eideren: done

PR #3409, opened 2026-09-09
Fix matrix truncating cast reading rows and columns swapped

### LINKED ISSUE #3323: Light Probes rendering bug
**Version**: 4.4.0 dev2
﻿
**Platform(s)**: Windows?
﻿
Image

**Describe the bug**
**There is an issue with Light Probes in the Stride 4.4.0 dev source code: when I place more than four Light Probes in a scene, the scene continuously renders and calculates Building effects (1 in queue), which keeps rendering indefinitely. After launching the game, an error occurs..** 

[stride440.txt](https://github.com/user-attachments/files/30729283/stride440.txt)

Image

xen2:
# PR Details

Fixes #3323: placing enough light probes to form a tetrahedron made the forward shading effect fail SPIR-V validation, so the scene never finished rendering.

`SpirvBuilder.Convert` read `MatrixType.Rows` and `MatrixType.Columns` the wrong way round when truncating a matrix. The type stores its dimensions HLSL-style, so `Columns` is the `OpTypeMatrix` column count and `Rows` is the size of each column vector.
The swap cancels out for square sources, so only `LightProbeShader` was affected: it is the only engine shader casting a `float3x4`.

The same routine also sized its value buffer by `Rows` and converted more entries than the truncating step filled, so `(float2x2)0` threw at compile time.

Tests: a render test covering row and column truncation, checking output pixels on Direct3D11 and Vulkan, plus a test compiling the real `LightProbeShader` in the `environmentLights` slot. Both fail with the fix reverted.

## Types of changes

## Checklist

### DIFF
--- a/sources/shaders/Stride.Shaders.Parsers/Spirv/Building/Builder.Expressions.cs
+++ b/sources/shaders/Stride.Shaders.Parsers/Spirv/Building/Builder.Expressions.cs
@@ -522,7 +522,14 @@ public partial class SpirvBuilder
             || (valueType is not ScalarType && valueType is not VectorType && valueType is not MatrixType))
             throw new NotImplementedException($"Cast only work between numeral types (cast from {valueType} to {castType})");
 
-        Span<int> values = stackalloc int[castType is MatrixType m ? m.Rows : 1];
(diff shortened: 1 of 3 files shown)


xen2:
Note: this doesn't fully fix the lightprobes, I am investigating independently and likely a follow-up PR.

PR #3411, opened 2026-09-09
Shader: Keep logical groups contiguous

xen2:
# PR Details

A logical group is assumed to be one contiguous run: `CreateLogicalGroup` scans until it hits a member that is not in the group, and stops there. But a group can be declared by several shaders (`cbuffer PerView.Lighting` comes from `LightClustered`, `DirectLightGroupPerView`, the `ShadowMap` shaders, `LightSkyboxShader` and more), and the parts were only grouped by cbuffer name, so their order followed the mixin order.

Light probes interleave: `PerView.LightProbes` lands between two `PerView.Lighting` declarations and splits the run. Everything after the split is then outside the group and never filled in.

* **Constant buffer**: only the cluster values were written. `LightCount`, `Lights`, the shadow cascade arrays, `AmbientLight`, `SkyMatrix`, `Intensity` and `SphericalColors` kept whatever the recycled buffer held, so the scene flickered a different colour every frame.
* **Descriptors**: the shadow map texture and the skybox cubemap were never bound and kept stale descriptors, so shadows and skybox were wrong.

Fixed by ordering the parts by logical group, so each group ends up in one run. The resource ordering is stable because consumers index a group positionally in declaration order.

Three related changes come with it:

* **A missing barrier**: the light probe tetrahedron ids texture was sampled by the opaque pass while still in the render target layout.
* **Constant buffer pools now stay mapped for the whole frame**, so the draw phase patches done by light probes and VR actually land. They were unmapped at the end of `Prepare`, which made those writes undefined on Vulkan and Direct3D12, while Direct3D11 always worked. A pool can now still be mapped at device teardown, after its buffer is gone, so `BufferPool` skips the unmap then (a null dereference on Direct3D12, a crash on MoltenVK).
* **Shader caches are now versioned**, otherwise cached bytecode would silently bring the old layout back. `ShaderCompilerVersion.Shader` covers the SPIR-V compiled from each shader class and invalidates everything (this fix bumps it); `ShaderCompilerVersion.Effect` covers what is derived from the SPIR-V and invalidates the effect caches only. An earlier version bumped `EffectBytecode.MagicHeader` instead, which rejects the 21 precompiled bytecodes checked in under `Stride.Graphics`; the header now stays for layout changes only.

A light probe render test, and two fixes found while writing it, follow in a PR stacked on this one.

## Types of changes

## Checklist

### DIFF
--- a/sources/engine/Stride.Assets/Effect/EffectCompileCommand.cs
+++ b/sources/engine/Stride.Assets/Effect/EffectCompileCommand.cs
@@ -59,6 +59,8 @@ namespace Stride.Assets.Effect
         {
             writer.Write(DataSerializer.BinaryFormatVersion);
             writer.Write(EffectBytecode.MagicHeader);
+            writer.Write(ShaderCompilerVersion.Shader);
+            writer.Write(ShaderCompilerVersion.Effect);
         }
 
         protected override Task<ResultStatus> DoCommandOverride(ICommandContext commandContext)
--- a/sources/engine/Stride.Engine/Rendering/Compositing/ForwardRenderer.LightProbes.cs
+++ b/sources/engine/Stride.Engine/Rendering/Compositing/ForwardRenderer.LightProbes.cs
@@ -351,6 +351,9 @@ namespace Stride.Rendering.Compositing
                 // context.CommandList.Draw...
             }
 
+            // The tetrahedron ids were just rendered into, and the opaque pass samples them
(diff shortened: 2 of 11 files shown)


xen2:
Before merge, I am checking if I could add a proper lightprobe test with gold so that this doesn't break again.
(it also extensively use various feature of the shader pipeline so that's more coverage)

PR #3412, opened 2026-09-09
Fix SDSL parser error on vector constructor argument containing a binary expression

azeno:
# PR Details

Fix SDSL parser error on vector constructor argument containing a binary expression

## Types of changes

## Checklist

### DIFF
--- a/sources/shaders/Stride.Shaders.Parsers/Parsing/SDSL/Parsers/LiteralParsers/LiteralParsers.cs
+++ b/sources/shaders/Stride.Shaders.Parsers/Parsing/SDSL/Parsers/LiteralParsers/LiteralParsers.cs
@@ -186,9 +186,7 @@ public record struct LiteralsParser : IParser<Literal>
                     while (!scanner.IsEof)
                     {
                         Parsers.Spaces0(ref scanner, result, out _);
-                        if (Vector(ref scanner, result, out var vec))
-                            p.Values.Add(vec);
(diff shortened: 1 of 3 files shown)


sources/shaders/Stride.Shaders.Parsers/Parsing/SDSL/Parsers/LiteralParsers/LiteralParsers.cs
```diff
@@ -186,9 +186,7 @@ public static bool Vector<TScanner>(ref TScanner scanner, ParseResult result, [M
                     while (!scanner.IsEof)
                     {
                         Parsers.Spaces0(ref scanner, result, out _);
-                        if (Vector(ref scanner, result, out var vec))
-                            p.Values.Add(vec);
-                        else if (ExpressionParser.Expression(ref scanner, result, out var exp))
+                        if (ExpressionParser.Expression(ref scanner, result, out var exp))
```
xen2, line 189: Good catch.
It seems it can now parse `float4(float2(1,2),float2(3,4))` (which is good)
But it can't parse `float2x2(float2(1,2), float2(1,2) * w)`
I think you need to do the same deletion in `MatrixParser`.

azeno: Thanks, fixed by next commit.

PR #3420, opened 2026-09-11
ci: build and ship spirv-cross for win-arm64

delustra:
ci: build and ship spirv-cross for win-arm64

# PR Details

`Stride.Dependencies.SpirvCross` ships `win-x64` only, so on `win-arm64` there is no
`spirv-cross.dll` to load and every shader-bearing asset fails to compile:

```
DllNotFoundException: Could not locate or load native library spirv-cross
```

`Silk.NET.SPIRV.Cross.Native` does have a `win-arm64` build, but this package's targets
strip Silk's native assets wholesale to keep conflict resolution deterministic — and that
strip is RID-agnostic, so on `win-arm64` it removes the only available DLL and puts nothing
back.

So, this PR adds an ARM64 build job (MSVC cross-compile on the same runner, mirroring
`dep-spirv-tools.yml`) and ships the result as `runtimes/win-arm64/native/`.

Windows-on-ARM natives landed in #2644; `spirv-cross` arrived a year later with the SDSL
rewrite and was never given the same treatment.

### Verification

Ran the modified workflow on a fork: both build jobs pass and the nupkg
contains both RIDs. Consuming that package, a `win-arm64` game build gets `spirv-cross.dll`
deployed on its own and runs natively on ARM64 (Direct3D11), and asset compilation works
under an arm64 `dotnet` host.

## Related Issue

No existing issue — regression found while building for `win-arm64`. Context: #2644.

## Types of changes

## Checklist

CI/packaging change, nothing to unit-test. Verified via workflow run, a game build, and
the editor; did not run the test suite.

Just in case, ai was used in process

### DIFF
--- a/.github/workflows/dep-spirv-cross.yml
+++ b/.github/workflows/dep-spirv-cross.yml
@@ -41,9 +41,39 @@ jobs:
           name: spirv-cross-win-x64
           path: spirv-cross-build/Release/spirv-cross-c-shared.dll
 
+  # Windows ARM64 — same host, MSVC ARM64 cross-compile (as in dep-spirv-tools.yml).
+  build-windows-arm64:
+    name: Build SPIRV-Cross (Windows ARM64)
+    runs-on: windows-2025-vs2026
+    steps:
+      - name: Checkout Stride
+        uses: actions/checkout@v7
+
+      - name: Checkout SPIRV-Cross
+        uses: actions/checkout@v7
+        with:
+          repository: stride3d/SPIRV-Cross
+          ref: ${{ github.event.inputs.branch || 'hlsl_tessel_shader' }}
+          path: spirv-cross-src
+
+      - name: Build
+        shell: pwsh
+        run: |
+          cmake -S spirv-cross-src -B spirv-cross-build -A ARM64 -Thost=x64 `
+            -DSPIRV_CROSS_SHARED=ON `
+            -DSPIRV_CROSS_CLI=OFF `
+            -DSPIRV_CROSS_ENABLE_TESTS=OFF
+          cmake --build spirv-cross-build --config Release --target spirv-cross-c-shared
+
+      - name: Upload build output
+        uses: actions/upload-artifact@v7
+        with:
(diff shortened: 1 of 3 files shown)


delustra:
@dotnet-policy-service agree

Kryptos-FR (approved):
LGTM. Waiting for someone more familiar with that tech to confirm.

Kryptos-FR:
@xen2 looks like changing files from `deps/` doesn't trigger any build. Maybe `Windows-Full` at the very least should run?

xen2:
Those specific `deps` workflow file are not invoked during any standard build.
Deps actual rebuilding is manual for several reasons:
- usually you still need to choose a version to compile against, i.e. spirv-cross `master` or a specific version/commits
   note: we do generate a [VERSION.txt](https://github.com/stride3d/stride/blob/5a712396750133b7d57f00dcb7368098e49f2473/deps/freetype/VERSION.txt) that is committed alongside .dll or inside nupkg on repo for tracking/repro
- build are not always 100% deterministic
- you typically want to verify/test the binary (it usually takes several round to get them right when adding a new platform)
- results comes later (either as binary or nupkg) so it needs another commit anyway

Currently, those workflow file mainly exist for easy reproducibility and targeting all platforms.
Before, we were just building based on some random outdated txt instructions or batch script, but it would never work well and people rarely had a mac or linux on hand.

In practice, the flow is:
- commit the dep file
- run it
  - if it generates a nupkg file, it is deployed on nuget.org => bump PackageVersion in Directory.Package.props, test & commit
  - if it generates binaries, test and commit them in deps/ folder

xen2:
> Ran the modified workflow on a fork: both build jobs pass and the nupkg
> contains both RIDs. Consuming that package, a `win-arm64` game build gets `spirv-cross.dll`
> deployed on its own and runs natively on ARM64 (Direct3D11), and asset compilation works
> under an arm64 `dotnet` host.

Thanks for doing the actual testing!
I will regenerate new package and do the version bump soon!

PR #3425, opened 2026-09-12
feat: improve crash report perf

Kryptos-FR:
# PR Details

Use `[LibraryImport]` for native calls and source generation for `System.Text.Json` (to avoid reflection).
Also formatted the files (e.g. remove unnecessary directives).

This prepares the work of eventually trimming the libraries or even using AOT.

## Related Issue

Will be needed for #3414 (if I find a good way of doing it).

## Types of changes

- [x] **Refactoring**

## Checklist

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -142,6 +142,7 @@ csharp_new_line_between_query_expression_clauses = true
 
 # Indentation preferences
 csharp_indent_case_contents = true
+csharp_indent_case_contents_when_block = false
 csharp_indent_switch_labels = true
 csharp_indent_labels = flush_left
 
--- a/sources/crashreport/Stride.CrashReport/CrashReportJsonContext.cs
+++ b/sources/crashreport/Stride.CrashReport/CrashReportJsonContext.cs
@@ -1,0 +1,10 @@
+using System.Text.Json.Serialization;
+
+namespace Stride.CrashReport;
+
+[JsonSourceGenerationOptions(WriteIndented = true)]
+[JsonSerializable(typeof(StoredCrash))]
+[JsonSerializable(typeof(HashSet<string>))]
+internal sealed partial class CrashReportJsonContext : JsonSerializerContext
+{
+}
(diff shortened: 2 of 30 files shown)


sources/crashreport/Stride.CrashReport/CrashReportJsonContext.cs
```diff
@@ -0,0 +1,10 @@
+using System.Text.Json.Serialization;
```
xen2, line 4: License header missing

sources/launcher/Stride.Cli/CliJsonContext.cs
```diff
@@ -0,0 +1,9 @@
+using System.Text.Json.Serialization;
```
xen2, line 4: License header missing

sources/crashreport/Stride.CrashReport/Stride.CrashReport.csproj
```diff
@@ -3,6 +3,8 @@
   <PropertyGroup>
     <!-- UI-agnostic and cross-platform: shared by the WPF window, the Avalonia launcher, and headless tools -->
     <TargetFramework>$(StrideXplatEditorTargetFramework)</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <LangVersion>latest</LangVersion>
```
xen2: I thought `latest` is automatic?

.editorconfig
```diff
@@ -142,6 +142,7 @@ csharp_new_line_between_query_expression_clauses = true
 
 # Indentation preferences
 csharp_indent_case_contents = true
+csharp_indent_case_contents_when_block = false
```
xen2, line 145: Agree with this, better than default behavior!

sources/crashreport/Stride.CrashReport/Stride.CrashReport.csproj
Kryptos-FR: By default, it is the latest of the current SDK used. So for example, if you build with a more recent SDK (let's say .NET 11), you can get some of the benefits (as long as it's only syntax difference with no required runtime support).
On the other hand, we do specify the .NET SDK we use in global.json so it likely has no impact in our case.

PR #3431, opened 2026-09-13
fix: keep one execution mode per entry point when a method overrides it

### LINKED ISSUE #3430: Overriding an entry point keeps the base method's execution modes
Found in #3304 review (https://github.com/stride3d/stride/pull/3304#discussion_r3964687578).

When a shader overrides `CSMain` and both methods carry `[numthreads]`, the base method's `OpExecutionMode LocalSize` stays in the merged module on a function that is no longer an entry point. `MergeSDSL` returns true, but validation fails:

```
OpExecutionMode Entry Point <id> '3[%CSNumThreadsOverrideBase_CSMain]' is not the Entry Point operand of an OpEntryPoint.
  OpExecutionMode %CSNumThreadsOverrideBase_CSMain LocalSize 8 8 1
```

Repro:

```
shader CSNumThreadsOverrideBase
{
    [numthreads(8, 8, 1)]
    void CSMain() { Compute(); }
    void Compute() {}
}

shader CSNumThreadsOverride : CSNumThreadsOverrideBase
{
    [numthreads(16, 16, 1)]
    override void CSMain() { base.CSMain(); }
}
```

Cause: `EntryPoint` is taken from the method name (`ShaderElements.MethodOrMember.cs:378`), so the base and the override both compile as the compute entry point and both emit `LocalSize`. `InterfaceProcessor.GenerateStreamWrapper` retargets execution modes only for the resolved entry point, so the base one is left behind. `ResolveHullPatchConstantEntryPoint` matches the `PatchConstantFuncSDSL` decoration the same way, so an `HSMain` override has the same problem.

This predates #3304. The same path runs on master.

To decide first (from @xen2): when an entry point is overridden, should its attributes transfer to the override, merging if the override repeats them, or drop unless restated?

Proposal: inherit, most-derived wins per mode. An override that restates an attribute replaces the base value; one that does not keeps it. `ComputeShaderBase.CSMain` carries `[numthreads(ThreadNumberX, ...)]`, so an override that only adds setup and calls `base.CSMain()` should not lose its `LocalSize`. The stricter option is the same code without the inheritance walk, so it is cheap to switch.

The fix belongs where the wrapper is generated: walk the override chain (the `FunctionGroupType` members with a matching signature), keep one instruction per mode, retarget it to the wrapper, drop the rest. Same walk for the hull patch constant decoration. I will open a PR for this.

sasvdw:
## PR Details

Each override of an entry point compiles with its own execution modes on its own function id. `InterfaceProcessor.GenerateStreamWrapper` moved only the modes of the most derived method to the wrapper. The base modes stayed on a function that is no longer an entry point, and `spirv-val` rejected the module while `MergeSDSL` returned true. A hull override without attributes failed earlier, because `FindOutputPatchSize` looked at the most derived method only.

**Policy.** The most derived declaration of each mode wins. An override that restates an attribute replaces the base value. An override that omits it keeps the base value. `ComputeShaderBase.CSMain` carries `[numthreads(ThreadNumberX, ...)]`, so an override that only adds setup and calls `base.CSMain()` keeps its `LocalSize`. The stricter option, drop unless restated, is the same code without the walk to the base, so it is cheap to switch if preferred.

**Change.** `ResolveOverrideChain` returns the function ids of the entry point and the methods it overrides, base first. The method group also holds overloads and composition functions, so the chain keeps only members with the entry point's signature and skips composition functions, as `ResolveEntryPoint` does. `GenerateStreamWrapper` walks the chain from the most derived method down, moves the first declaration of each mode to the wrapper, and drops repeats. `FindOutputPatchSize` and `ResolveHullPatchConstantEntryPoint` use the same walk.

**Tests.** New cases in `StrideShaderTests`, each against a corpus shader:
- `CSNumThreadsOverride`: the override restates `[numthreads]`. One `LocalSize`, the derived value.
- `CSNumThreadsInherit`: the override omits it. One `LocalSize`, the base value.
- `HSOverride`: the override declares no hull attributes. `outputcontrolpoints` and `patchconstantfunc` come from the base.
- `ComposeNumThreadsRoot` and `ComposeNumThreadsInheritRoot`: the composed helper derives from the same base and carries its own `[numthreads]`. It stays out of the chain, so the root's value, or the base's, wins.

The override cases fail on master, two with the validator error and one in `FindOutputPatchSize`. The composition cases fail without the composition filter.

## Related Issue

Fixes #3430

## Types of changes

## Checklist

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01VHYKv129fEUp6sxMbybZWf


xen2:
I think it needs a rebase (conflicts)

sources/shaders/Stride.Shaders.Parsers/Spirv/Processing/Interfaces/InterfaceProcessor.cs
```diff
@@ -55,67 +55,44 @@ internal EntryPointInfo(string name, int id, ExecutionModel model, List<int> int
...
-                    case Op.OpCompositionSDSL:
-                        depth++;
-                        break;
-                    case Op.OpCompositionEndSDSL:
-                        depth--;
-                        break;
-                    case Op.OpFunction when depth > 0:
-                        result.Add(((OpFunction)i).ResultId);
-                        break;
+                    if ((FunctionType)member.Type == signature)
```
xen2: After the rebase, this should skip composition functions, like `ResolveEntryPoint` now does on master (#3382).

A composition that also derives from `ComputeShaderBase` has its own `CSMain` in the same group. Right now it would end up in the chain, and its `[numthreads]` could win over the real entry point's.

```csharp
if ((FunctionType)member.Type == signature && !compositionFunctions.Contains(member.IdRef))
```

sasvdw: Good catch, missed that. Done.

sasvdw:
> I think it needs a rebase (conflicts)

rebased and ready for reveiw!

PR #3436, opened 2026-09-14
feat: BinaryOperator.AddMath

Eideren:
# PR Details
Material system's Multiply, Subtract, Divide map to the corresponding component-wise operation, Add doesn't. This PR adds AddMath to cover that discrepancy.

## Related Issue
None.

## Types of changes

## Checklist


sources/engine/Stride.Rendering/Rendering/Materials/ComputeColors/BinaryOperator.cs
```diff
@@ -12,10 +12,15 @@ namespace Stride.Rendering.Materials.ComputeColors
...
         /// <summary>
-        /// Add of the two textures.
+        /// Add '3ds' of the two textures.
         /// </summary>
         Add,
 
+        /// <summary>
+        /// Add of the two textures.
+        /// </summary>
+        AddMath,
```
Kryptos-FR: Adding an enum value in the middle is a binary breaking change. I someone somehow updates the version of stride while using a library that was compiled against a previous version (the scenario might be rare), it will use the wrong value.

Since we are adding a breaking change, I think we should explicitly assign values to any enum. That way, we can in the future add new values, keep them sorted (for documentation purpose) but with increasing backing value.

PR #3445, opened 2026-09-17
Fix types not being removed from AssetRegistry when assembly unregistered

### LINKED ISSUE #3444: Error while reloading a library with custom assets

**Version**: 4.4 beta 7

**Platform(s)**: Windows

**Describe the bug**

An exception is thrown when I try to reload scripts in Stride.

**To Reproduce**
1. Create a library project with custom asset classes and reference this library from the primary game project.
2. Ensure that the library has a module initializer as described by https://github.com/stride3d/stride/issues/2646.
3. Make some trivial change in the library C# files.
4. Accept when Stride asks to reload the assemblies.

**Expected behavior**

It should reload without giving an error.

**Log and callstacks**

```
Info: Patch for assembly [MyGame.Prototyping, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null]
Info: MyGame.Prototyping -> N:\StrideProjects\MyGame\MyGame.Prototyping\bin\Debug\net10.0\DreamingTogether.Prototyping.dll
Info: Unloading assembly MyGame.Prototyping, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null
Info: Loading assembly 
Error: Error loading assembly MyGame.Prototyping: 
ArgumentNullException: Value cannot be null. (Parameter 'assembly')
   at System.ArgumentNullException.Throw(String paramName)
   at System.ArgumentNullException.ThrowIfNull(Object argument, String paramName)
   at Stride.Core.Reflection.AssemblyRegistry.Register(Assembly assembly, IEnumerable`1 categories) in /_/sources/core/Stride.Core/Reflection/AssemblyRegistry.cs:line 180
   at Stride.Core.Reflection.AssemblyRegistry.Register(Assembly assembly, String[] categories) in /_/sources/core/Stride.Core/Reflection/AssemblyRegistry.cs:line 238
   at Stride.Assets.Presentation.AssemblyReloading.ReloadAssembliesOperation.LoadAssemblies(ILogger log, AssemblyContainer assemblyContainer, List`1 loadedAssemblies, Boolean newVersion, Boolean firstTime) in /_/sources/editor/Stride.Assets.Presentation/AssemblyReloading/ReloadAssembliesOperation.cs:line 93
```

ds5678:
# PR Details

This pull request ensures that AssetRegistry removes its references to an assembly's types when that assembly is unregistered. Before this pull request, the relevant collections had no calls to `Remove` in the codebase.

## Related Issue

Resolves #3444

## Types of changes

## Checklist

## Note

I do not know why `AlwaysMarkAsRootAssetTypes` needs a lock, but all other accesses to `AlwaysMarkAsRootAssetTypes` had one, so I put one here, too.


sources/assets/Stride.Core.Assets/AssetRegistry.cs
```diff
@@ -826,11 +826,34 @@ private static void UnregisterAssetAssembly(Assembly assembly)
                 RegisteredAssetFileExtensions.Remove(extensionToRemove);
             }
 
+            lock (AlwaysMarkAsRootAssetTypes)
+            {
+                foreach (var typeToRemove in AlwaysMarkAsRootAssetTypes.Where(type => type.Assembly == assembly).ToList())
+                {
+                    AlwaysMarkAsRootAssetTypes.Remove(typeToRemove);
```
Eideren: `AlwaysMarkAsRootAssetTypes.RemoveWhere` is more succinct and faster

sources/assets/Stride.Core.Assets/AssetRegistry.cs
```diff
@@ -826,11 +826,34 @@ private static void UnregisterAssetAssembly(Assembly assembly)
...
+                }
+            }
+
+            foreach (var typeToRemove in ContentToAssetTypes.Keys.Where(type => type.Assembly == assembly).ToList())
+            {
+                ContentToAssetTypes.Remove(typeToRemove);
+            }
```
Eideren: You can safely remove while enumerating a dictionary, no need to collect into a list first

sources/assets/Stride.Core.Assets/AssetRegistry.cs
```diff
@@ -826,11 +826,34 @@ private static void UnregisterAssetAssembly(Assembly assembly)
...
+
+            foreach (var typeToRemove in AssetToContentTypes.Keys.Where(type => type.Assembly == assembly).ToList())
+            {
+                AssetToContentTypes.Remove(typeToRemove);
```
Eideren, line 839: Same here

sources/assets/Stride.Core.Assets/AssetRegistry.cs
```diff
@@ -826,11 +826,34 @@ private static void UnregisterAssetAssembly(Assembly assembly)
...
+            }
+
             foreach (var upgraderToRemove in RegisteredPackageUpgraders.Where(keyValue => keyValue.Value.GetType().Assembly == assembly).Select(keyValue => keyValue.Key).ToList())
             {
                 RegisteredPackageUpgraders.Remove(upgraderToRemove);
             }
 
+            foreach (var pair in RegisteredAssetFactories.Where(pair => pair.Value.GetType().Assembly == assembly).ToList())
+            {
+                RegisteredAssetFactories.Remove(pair.Key);
```
Eideren, line 849: Yep

Eideren (commented):
Thanks! Small nitpick

sources/assets/Stride.Core.Assets/AssetRegistry.cs
ds5678: `ToList()` was already being called on other dictionaries in this method. Should I change those removals too?

Eideren: As you want, you can leave them as is given that you're not touching those lines, or you can change them if it bothers you

Kryptos-FR: > You can safely remove while enumerating a dictionary

I actually didn't know that. Apparently it is the case since .NET Core 3.0.

Reference: https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.remove?view=net-10.0#remarks

Eideren: Yes, hashset seem to provide the same behavior as you would expect, but I have not found documentation guaranteeing that, which is why I have not suggested doing so for the other cases here

xen2:
Pushed two small commits on top:
- `ContentToAssetTypes` also drops the unloaded assembly's asset types from the lists of content types owned by other assemblies (an asset targeting an engine content type
- the reload now passes its log to `LoadAssemblyFromPath` so the module initializer exception behind #3444 shows up instead of only the null assembly that follows.

Thanks for the fix!

PR #3449, opened 2026-09-18
[Shaders] fix: a stage composition is supplied at the root, and the voxel marchers own their samplers slot

### LINKED ISSUE #3423: [Shaders] `stage compose` supplied from a nested composition: which rule for the mixer?

The new mixer promotes any shader declaring stage members to the root, `stage compose` included. The value supplied for that composition does not follow: it stays in the `Compositions` dictionary of the node that supplied it. Supplied at the root it works (`CompositionArray1`). Supplied by a nested effect, the merge fails with `No composition was supplied for 'AttributeSamplers' ... while merging the mixin node '<root>'` (the named error added in #3382).

Case in the engine: `Stride.Voxels`. `MarchAttributes` declares `stage compose IVoxelSampler AttributeSamplers[]`. `LightVoxelEffect` supplies it, and is itself composed by the forward effect at `environmentLights[i]`. The marchers that sample through it inherit `MarchAttributes` two levels further down (`Marcher.diffuseMarcher.environmentLights[i]`, `specularMarcher.environmentLights[i]`). A light renderer never controls the root effect, so "supply it at the root" is not something a light can do.

#3382 carried two commits that made this work (4df2f3d: hoist the value with the declaring shader; 6a90900: keep the supply path so resource keys get the names the engine composes). They were removed from the PR at @xen2's request so the rest can merge; this issue is the follow-up.

Rules proposed by @xen2 in the #3382 review:

- a `stage compose` can be declared anywhere (as today);
- its value can only be supplied at the root. Supplying it from a nested node is an error, in `ProcessCompositions`:

```csharp
if (isSupplied && (variable.Flags & VariableFlagsMask.Stage) != 0 && !isRoot)
    log.Error($"'{variableName}' is a `stage compose` declared by '{shaderName}', so it is one slot "
            + $"for the whole effect and must be supplied at the root. It was supplied at '{compositionPath}'.");
```

Consequences:

- `Stride.Voxels` drops `stage` on `AttributeSamplers` and supplies the sampler to every marcher composition that uses it (each gets its own resource bindings, so the C# applies the sampler parameters once per marcher path).
- Other `stage compose` users to check: `LightShaftsShader.lightGroup`, `DirectLightGroupArray.directLightGroups[]`, `EnvironmentLightArray.environmentLights[]`. Those are supplied by the root effect today.

Hoist the value with the declaring shader (the two dropped commits). It needs a rule for two different values supplied for one stage slot (error, or first wins), and the supply path has to travel with the value so the resource keys keep the names the engine composes (`...AttributeSamplers[0].environmentLights[2]`, not `...AttributeSamplers[0]`). It is more state in the mixer, and the open points in review were the `ReferenceEquals` check and whether to fail or concatenate.

**Additional context**

- Review thread: #3382, `ShaderMixer.ShaderSourceEvaluator.cs`.
- Tests that covered the hoist and were removed with the commits: `CompositionArrayStageFromNested`, `StageCompositionSuppliedFromNestedKeepsItsSupplyPath` (both in the PR history).
- Question for the rule: is a shared composition below a light, which is always nested, meant to be impossible with `stage compose`? If so, what is the intended way for a light's nested compositions to share one sampler?

Nicogo1705:
# PR Details

A shader declaring a stage member is promoted to the root of the effect. When such a shader declares a `stage compose`, a value supplied for it at a nested composition path had nothing to attach to, and the merge failed later with a missing composition (#3374 on beta7).

- **Mixer**: the rule agreed in #3423. A `stage compose` can be declared anywhere, but its value can only be supplied at the root. Supplying it from a nested node is reported with the slot, the declaring shader and the path it was supplied at. The composition evaluation now runs under the same error handling as the merge, so it lands in the log like the other mixer errors.
- **Stride.Voxels**: `AttributeSamplers` was supplied by `LightVoxelEffect`, itself composed under `environmentLights[i]`. The slot moves to the marchers, the only shaders that read it: `MarchAttributes` loses `stage`, `LightVoxelShader` and `VoxelVisualizationViewShader` no longer inherit it, and `IVoxelMarchMethod` / `IVoxelMarchSet.GetMarchingShader` take the samplers to compose. `LightVoxelRenderer` and `VoxelVisualizationView` bind the attribute once per marcher path. `MarchAttributesEffect` had no user and goes with the key it served.
- Test: `StageCompositionSuppliedFromNestedIsReported`.

Verified on Direct3D 11: the shader test suite (846), the voxel GI demo (Cornell box, bounce and cones fine), and the MyGame2 project attached to #3374, which gets past the mixer.

Note: MyGame2 then hits a separate compiler gap, `SV_Coverage` as a fragment input is not mapped in `BuiltinProcessor` (used by `VoxelAttributeDirectionalCoverageShader`). That one is not in this PR; it will come as its own fix.

## Related Issue

Fixes #3423. Part of #3374.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change): `GetMarchingShader` gains a parameter on the two marcher interfaces.

## Checklist

---


xen2:
Thanks, the Voxels rewrite LGTM. I pushed two commits on top, please pull before continuing.

- **Mixer:** a `stage compose` declared only by a nested shader could not be supplied at the root either. Now a stage slot always takes its value from the root. Two tests added.
- **Voxels:** marchers bind their own samplers (`ApplyAttributeSamplers`), so `LightVoxelRenderer` no longer hard-codes the `Marcher` path.

Shader tests pass (849). I did not run the voxel GI demo, could you check it once?

Nicogo1705:
Pulled your two commits, thanks.

Tested with the VoxelGI Demo and the MyGame2 repro from #3374 (with #3450 on top, its compositor uses SV_Coverage) => seems good.

PR #3450, opened 2026-09-18
[Shaders] fix: SV_Coverage maps to SampleMask in the fragment stage

Nicogo1705:
# PR Details

A pixel shader declaring `stream uint coverage : SV_Coverage` failed to compile with `System-value Semantic not implemented: SV_COVERAGE for stage Fragment as Input`. In the engine, `VoxelAttributeDirectionalCoverageShader` uses it, so any voxel volume with the anti-aliasing modifier hit it (#3374 after #3449).

- `BuiltinProcessor` maps `SV_Coverage` in the fragment stage, input and output, to `BuiltIn.SampleMask`, with the one-element array SPIR-V expects; `ConvertInterfaceVariable` already handles the array/scalar conversion both ways.
- `EntryPointWrapperGenerator` loads a builtin input with its interface type before converting it, which is what the output side already did. Until now no fragment input had a type override, so this path was never exercised.
- Test: `CoverageIsSampleMaskInFragmentStage`, which validates the SPIR-V and checks the HLSL translation carries `SV_Coverage`.

Verified on Direct3D 11 with the MyGame2 project attached to #3374, together with #3449: it renders, shadows and voxel bounce included.

## Related Issue

Part of #3374.

## Types of changes

## Checklist

---


sources/shaders/Stride.Shaders.Parsers/Spirv/Processing/Interfaces/Generation/EntryPointWrapperGenerator.cs
```diff
@@ -393,7 +393,7 @@ void ProcessTessellationArguments(Symbol function, Span<int> arguments)
             foreach (var stream in streamLayout.InputStreams)
             {
                 var streamPointer = buffer.Add(new OpAccessChain(context.GetOrRegister(new PointerType(stream.Info.Type, Specification.StorageClass.Private)), context.Bound++, streamLayout.StreamsVariableId, [context.CompileConstant(stream.Info.StreamStructFieldIndex).Id])).ResultId;
-                var inputResult = buffer.Add(new OpLoad(context.Types[stream.Info.Type], context.Bound++, stream.Id, null, [])).ResultId;
+                var inputResult = buffer.Add(new OpLoad(context.GetOrRegister(stream.InterfaceType), context.Bound++, stream.Id, null, [])).ResultId;
```
xen2, line 396: Good fix. Note that the same load pattern exists in two other places in this file: line 130 (GS and tessellation array inputs) and line 224 (patch inputs). Both still load with `stream.Info.Type` instead of `stream.InterfaceType`.

No builtin with a type override reaches them today, but it would be good to fix them the same way you did for this one.

xen2 (commented):
I think the output side doesn't work: `InterfaceProcessor` only marks a pixel shader stream as an output for SV_Target* and SV_Depth, so the write to SV_Coverage in the test shader is dropped silently (the generated HLSL has no SV_Coverage output).

Please add SV_COVERAGE to that check and assert the output in the test (or limit the PR to input only).

```diff
-if ((semantic.ToUpperInvariant().StartsWith("SV_TARGET") || semantic.ToUpperInvariant() == "SV_DEPTH") && stream.Value.Write)
+if ((semantic.ToUpperInvariant().StartsWith("SV_TARGET")
+        || semantic.ToUpperInvariant() == "SV_DEPTH"
+        || semantic.ToUpperInvariant() == "SV_COVERAGE")
+    && stream.Value.Write)
     stream.Value.Output = true;
```
(`InterfaceProcessor.cs`, line 180)

Nicogo1705:
Both done in 7df901ccd: SV_COVERAGE added to the output check, the test asserts the SV_Coverage output and compiles the HLSL with fxc (ps_5_0), and the two other loads (GS/tessellation array inputs, patch inputs) use the interface type.

